PR #10, opened 2016-01-10
Add mkdocs.

hayd:
This also adds two static files to the build:
- mathjaxhelper.js (determine what mathjax sees as latex)
- Lapidary.css (for the moment this fixes a small annoyance which I have with mathjax rendering of LHS nav, but potentially could do some other julia-specific tweaks).

### DIFF
--- a/docs/build/index.md
+++ b/docs/build/index.md
@@ -1,0 +1,141 @@
+<a id='Main.Lapidary-1'></a>
+
+<a href='#Main.Lapidary-1'> # </a>**Constant**
+
+# Lapidary
+
+A documentation generator for Julia.
+
+[![Build Status](https://travis-ci.org/MichaelHatherly/Lapidary.jl.svg?branch=master)](https://travis-ci.org/MichaelHatherly/Lapidary.jl) [![codecov.io](http://codecov.io/github/MichaelHatherly/Lapidary.jl/coverage.svg?branch=master)](http://codecov.io/github/MichaelHatherly/Lapidary.jl?branch=master)
+
+## Brief Overview
+
+**Doctests**
+
+Any code block with `.language` set to `"julia"` containing either REPL prompts, `julia>`, or an `# output:` comment. For example:
+
+```
+    ```julia
+    julia> a = 1
+    1
+
+    julia> b = 2;
+
+    julia> a + b
+    3
+    ```
+```
+
+or
+
+```
+    ```julia
+    a = 1
+    b = 2
+    a + b
+
+    # output:
+
+    3
+    ```
+```
+
+Errors can be checked for using the `{throws ErrorName}` syntax, i.e.
+
+```
+    ```julia
+    julia> div(1, 0)
+
+        {throws DivideError}
+
+    ```
+```
+
+**Docstring Splicing**
+
+Docstrings for objects documented using Julia's docsystem can be spliced into the markdown files using code blocks containing `{docs}` as their first line:
+
+```
+```
+{docs}
+foo
+bar
+baz
+```
+```
+
+Operators must be enclosed, i.e. `(+)`.
+
+**Metadata**
+
+Metadata, such as the current module, can be set for a page using the `{meta}` code block:
+
+```
+```
+{meta}
+CurrentModule = Base
+```
+```
+
+Changing the module allows one to avoid having to fully qualify every object spliced into a `{docs}` block.
+
+**Contents & Index**
+
+Table of contents and docstring indexes can be automatically generated using either `{contents}` or `{index}` code blocks. These code blocks shouldn't contain any other text.
+
+```
+```
+{contents}
+```
+```
+
+expands to a nested list of all headers in all files found in the `src` directory. The depth of headers to be displayed can be set with
+
+```
+```
+{meta}
+ContentsDepth = 2
+```
+```
+
+prior to the `{contents}` block.
+
+```
+```
+{index}
+```
+```
+
+expands into a list of all documented objects spliced into the files found in the `src` directory. The module's to be included can be limited using
+
+```
+```
+{meta}
+IndexModules = [Foo, Bar]
+```
+```
+
+prior to the `{index}` block.
+
+**Auto Links**
+
+Automatic cross-referencing uses markdown link syntax with the `url` set to `{ref}`, i.e
+
+```
+This is the [Introduction]({ref}).
+```
+
+which links to the first header called `"Introduction"` in any file found in `src`.
+
+Spliced docstrings can be linked to using inline code syntax for the text of a link, i.e
+
+```
+This is the [`foo`]({ref}) function.
+```
+
+which will link to where `foo` was spliced into a `{docs}` block.
+
+**Building Docs**
+
+See the `src/build.jl` file for the standard build file used by Lapidary.
+<hr></hr>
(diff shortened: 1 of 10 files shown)


docs/build/index.md
```diff
@@ -0,0 +1,8 @@
+<a id='Main.Lapidary-1'></a>
+
+<a href='#Main.Lapidary-1'> # </a>**Constant**
+
+A documentation generator for Julia.
```
hayd: The alternative is to stick a copy of the README in here (Note: the `index.md` is required on mkdocs). I definitely like how that's generated in the rewrite branch!

docs/build/index.md
```diff
@@ -0,0 +1,8 @@
+<a id='Main.Lapidary-1'></a>
+
+<a href='#Main.Lapidary-1'> # </a>**Constant**
```
MichaelHatherly: This should be `**Module**`, needs an additional def in https://github.com/MichaelHatherly/Lapidary.jl/blob/7cb5b2ee8b3fd23b91c47b2acbc7dee471355e61/src/utilities.jl#L110-L114. (You can add it if you want, or just leave it for another commit later.)

src/passes.jl
```diff
@@ -306,6 +306,7 @@ end
 function process!(::SetupBuildDirectory, page::Page, content, block)
     dir = dirname(page.build)
     isdir(dir) || mkpath(dir)
+    addstaticfiles!(dir)
```
MichaelHatherly: I think `SetUpBuildDirectory` would run on every `page`, so if there's any `page` in a subdirectory then this will add the static files to that dir as well, i.e.

```
build/
    Lapidary.css
    manual/
        Lapidary.css
    ...
```

Is that then intended behaviour?

Would just need to check that the page's parent directory is the build dir itself to fix this I believe.

MichaelHatherly:
Thanks! Looks like I've been using 0.5-only features so travis is going to keep failing. For the moment probably best to just focus on getting this all working on 0.5 and do some compat fixes later on.

src/passes.jl
hayd: No! Ok, I thought the false at the end of it was something to do with that.

I had this somewhere else in a first pass but suspected you would feel strongly where it should go.... It feels kinda strange to have it as a pass (for every page for once time things like this).

The other option was to append! staticfiles to document.pages, that's probably the most elegant (but I couldn't get it working).

MichaelHatherly: The `false` tells the calling method, https://github.com/MichaelHatherly/Lapidary.jl/blob/7cb5b2ee8b3fd23b91c47b2acbc7dee471355e61/src/passes.jl#L31, whether or not to try the next action on the block or to move on to the next block. I think I might change the `Bool` to a custom `Enum` to make that move clear and also add some additional choices of what to do after an action.

Once we start trying to build larger docs, such as the Julia manual, I suspect that we'll need to optimise the number of times some things get run. For now though just checking the page's parent is `build/` should be sufficient.

hayd: Any thoughts on:

> append! staticfiles to document.pages

I think this would be my preferred refactor, if it were possible.

MichaelHatherly: > append! staticfiles to document.pages

Sorry, missed that part. Adding them as `Page`s would probably been best. Something like

```diff
diff --git a/src/types.jl b/src/types.jl
index 6533bd6..e9d5244 100644
--- a/src/types.jl
+++ b/src/types.jl
@@ -41,6 +41,13 @@ immutable Document
                 doc.pagemap[obj] = page
             end
         end
+        assetdir = normpath(joinpath(dirname(@__FILE__), "..", "assets"))
+        for asset in readdir(assetdir)
+            obj, out = paths(assetdir, asset, assetdir, build)
+            page = Page(doc, obj, out, format)
+            push!(doc.pages, page)
+            doc.pagemap[obj] = page
+        end
         doc
     end
 end
```

and store each static file in a `Lapidary/assets` dir as an actual file instead of the `Dict` in `staticfiles.jl`. That way they're easier to edit in their actual format rather than a Julia string.

(Any project-specific asset files can just be put into the user's `docs/src` dir and they will get copied over.)

docs/build/index.md
hayd: I forgot that this fallsback the the README. Which I think is going to be a bit circular here (if we generate the README!)...

hayd:
I've updated the branch with your change. It's much cleaner (which is great)!

We should move discussion of readme generation being circular to #12.

*This works with mkdocs, but you have to turn it on at readthedocs (as I'm sure you know)!*

http://lapidaryjl.readthedocs.org/en/latest/

MichaelHatherly:
Thanks!

MichaelHatherly:
RTDs appears to be a bit confused https://lapidaryjl.readthedocs.org/en/latest/. Or maybe I am... is this perhaps because the docs are in `docs/build` rather than `docs`?

MichaelHatherly:
Ah, looks like it's in https://lapidaryjl.readthedocs.org/en/latest/build/index.html. I guess that's not too bad.

hayd:
I think doc is the more usual naming convention. Perhaps we need to move to that.

Dang it looks terrible!! Is that with mkdocs checked?

hayd:
Eurgh, why can't everyone follow the CommonMark spec for fenced blocks. Sigh.

hayd:
Crap. That's probably my fault lol! Ok!!

It get's made plain to:

````
```
```julia
julia> a = 1
1

julia> b = 2;

julia> a + b
3
```
```
````

hayd:
The solution is for plain to look in code block for the max \`\`\` sequence and do one more that that. Since:

    ````
    ```julia
    julia> a = 1
    1

    julia> b = 2;

    julia> a + b
    3
    ```
    ````

is valid (and parsed correctly by mkdocs!!).

MichaelHatherly:
> I think doc is the more usual naming convention. Perhaps we need to move to that.

Fine with me. The main Julia repo names it `doc` so matching that would probably be for the best.

> Dang it looks terrible!! Is that with mkdocs checked?


There wasn't an option to check for mkdocs that I saw. I think it gets detected automatically, maybe?

> The solution is for plain to look in code block for the max ``` sequence and do one more that that.

If that looks alright everywhere then it's fine by me.

hayd:
Posting a fix on julia shortly.

The option for mkdocs is in the admin/settings, it may be selected by default nowadays. You can also set the language (though I'm unsure if that does anything).

hayd:
the "build" in https://lapidaryjl.readthedocs.org/en/latest/build/index.html I think means RTD is not reading the mkdocs file correctly... the LHS menu is not populated (it is with `mkdocs serve`), so I think there is something left to twiddle on their site.

hayd:
In fact at the bottom of the page it says:

> Built with Sphinx using a theme provided by Read the Docs.

when it should say MkDocs. http://lightgraphsjl.readthedocs.org/en/latest/

MichaelHatherly:
Yes, `mkdocs serve` works for me too locally. Perhaps a RTDs/mkdocs combo problem.

hayd:
you have to select mkdocs in the settings...

MichaelHatherly:
I was sure that was selected when I created the site. Wasn't when I check just now. Looks much better now though, thanks.

PR #32, opened 2016-03-21
fix REPL test when there is a semicolon before a comment

bicycle1885:
This fixes REPL test when there is a semicolon before a comment. In Julia's REPL, output is suppressed when there is a semicolon before a comment like below:

```julia
julia> [1,2,3];  # comment

```

This behavior was not reflected in Lapidary.jl, so I fixed it.


MichaelHatherly:
Thanks, lgtm.

PR #34, opened 2016-03-22
Added comment about leading white-space in api key

mfalt:
This problem took me several hours to solve. Github automatically includes a leading whitespace when displaying the key on the website. Travis will not be able to store the key, and will throw an error (small enough to quickly get lost in the log), but still exposing the key in the log (see issue https://github.com/travis-ci/travis-ci/issues/4139).
And Lapidary will simply conclude "skipping docs deployment." since apparently `github_api_key == ""`.


MichaelHatherly:
That's a nasty one! Thanks for working it out and documenting it.

PR #36, opened 2016-04-12
Extend mathjaxhelper.js functionality

lbenet:
See:
http://stackoverflow.com/questions/27882261/mkdocs-and-mathjax
https://github.com/mitya57/python-markdown-math

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -86,13 +86,18 @@ Linux. This can be changed using the `julia` and `osname` keywords as follows:
 
 ```julia
 deploydocs(
+    deps   = Deps.pip("python-markdown-math"),
     repo   = "github.com/USER_NAME/PACKAGE_NAME.jl.git",
     julia  = "0.4",
     osname = "osx"
 )
 ```
 
-This will deploy the docs from the OSX Julia 0.4 Travis build bot.
+This will deploy the docs from the OSX Julia 0.4 Travis build bot. The keyword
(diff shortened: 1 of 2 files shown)


docs/src/man/hosting.md
```diff
@@ -86,13 +86,18 @@ Linux. This can be changed using the `julia` and `osname` keywords as follows:
 
 ```julia
 deploydocs(
+    deps   = Deps.pip("python-markdown-math"),
```
MichaelHatherly: This will only install `python-markdown-math` and not `mkdocs`, which means it'll fail to deploy and could be confusing to readers. Probably best to describe the `deps` options and `mdx_math` in a separate example after this one. Perhaps directly after the

```
This will deploy the docs from the OSX Julia 0.4 Travis build bot.
``` 

line?

MichaelHatherly:
Thanks for the PR @lbenet, I've left one comment inline.

lbenet:
I will make the changes latter. I simply thought of a way to emphasize the installation of `python-markdown-math`, but you are right, one should do the same with mkdocs.

lbenet:
I have pushed a new commit following your suggestions. Please let me know if something else is needed.

MichaelHatherly:
Thanks.

lbenet:
Thank you. Documenter.jl is a great package!

MichaelHatherly:
Thanks for the kind words.

PR #47, opened 2016-05-11
remove horizontal rule after each doc

bicycle1885:
This pull request removes horizontal rules after docstrings. I think this is semantically wrong and visually misleading.

Semantically, the horizontal rules should not be used to separate header and description because:
> In HTML5, the `<hr>` tag defines a thematic break.

<http://www.w3schools.com/tags/tag_hr.asp>

Visually, it makes hard to find a description when a series of docstrings are rendered:
screen shot 2016-05-11 at 17 16 44


MichaelHatherly:
Seems reasonable to me. Thanks.

KristofferC:
I also thought this was not so good visually. Thanks for fixing.

PR #51, opened 2016-05-15
Deprecate {...} syntax.

MichaelHatherly:
---

**Todo:**

- [x] Deprecate old `{...}` syntax in favour of `@...`.
- [x] Update documentation to reflect syntax changes.
- [ ] Prepare PRs for known packages currently using Documenter.

---

Switches in-block `{...}` syntax to use the `.language` field instead as described in #5.

This is a first pass and syntax is up for discussion at this point. Docs changes will be made once a final decision is agreed upon.

**Syntax changes summary:**

        {docs}
        ...

becomes

    ```@docs
    ...
    ```

same goes for `{meta}`, `{autodocs}`, `{eval}`, `{index}`, `{contents}`, `{example}`, and `{repl}`.

`{repl <name>}` and `{example <name>}` become

    ```@repl <name>
    ...
    ```

and

    ```@example <name>
    ...
    ```

`{ref}` and `{ref#<id>}` become `@ref` and `@ref <id>`.

Named headers change from `[...]({ref#<id>})` to `[...](@id <id>)`.

This does not completely remove the old syntax, which will raise a deprecation warning suggesting using the newer syntax. I'll try go through known repos on github to switch the syntax over myself when this is ready to merge.

cc: @hayd, and anyone else that wants to do some syntax bikeshedding.


hayd:
This looks great, IMO the `@` is perfect for indicating it's going to be expanded.

For some bikeshedding, I think the `:` should rather be a space, since that's julia's macro syntax. (Assuming the single argument is the rest of the line stripped?)

MichaelHatherly:
> since that's julia's macro syntax

Perhaps it should be parsed *using* `Base.parse`... that way positional args and kwargs would be baked in if we needed them at some later stage.

hayd:
Updating every single user of Documenter for this breaking change is probably the best service I've ever seen from a library maintainer. Glorious. 👍

bicycle1885:
Really. Thank you @MichaelHatherly!

MichaelHatherly:
Thanks for the kind words.

PR #52, opened 2016-05-16
Update syntax.md

KristofferC:
Am I correct in that the nested block thing is not intended?


KristofferC:
Nvm I was wrong.

MichaelHatherly:
Yes, was intentional, though the syntax will change slightly once #51 is merged.

PR #56, opened 2016-05-17
Add Example

ChrisRackauckas:


MichaelHatherly:
Thanks @ChrisRackauckas, could you add it after `ControlSystems.jl` to keep the list ordered alphabetically.

ChrisRackauckas:
Fixed

MichaelHatherly:
Thanks.

PR #60, opened 2016-05-22
Implement SSH deploy keys

### LINKED ISSUE #43: Security of tokens
It's probably not good to be giving Travis personal access tokens, as this gives access to a users whole account (including private repositories, ability to change/delete lots of things).

A better approach would be to use [GitHub deploy keys](https://developer.github.com/guides/managing-deploy-keys/). [Here is some code](https://gist.github.com/simonbyrne/c22830777a15957e153c220904248f40) I wrote a while back when thinking about how to do this (this uses a slightly different approach of using Travis's encrypt functionality, which requires that the Travis command line tools be installed, but it is probably easier to use your approach of pasting it in to the Travis repository settings).

MichaelHatherly:
Adds support for deploying from travis to github using ssh deploy keys in addition to the current token approach. Using ssh keys does not grant the "deployer" access to all repos in a user account,
but rather just a single repo.

Support for both deployment methods will likely remain for the moment, but using ssh keys will be encouraged over use of tokens.

Also adds a `Travis.genkeys` function to partially automate the creation of the deploy keys and upload them to travis. The user must manually add the public key to github though.

Fixes #43.


hayd:
To ask the security question: Can someone can make a pull request (that builds on the travis server), having changed the `.travis.yml`, and either expose the key or be able to use the key on that build?

MichaelHatherly:
If I understand correctly how travis env variables work then I believe this is reasonably safe, though I'm no security expert.

From https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings it mentions that

> These variables are not automatically available to forks.

So any PR made from a fork of the repo will not have access to either the `GITHUB_API_TOKEN` or the encrypted key/iv pair that this PR adds support for.

Without either of those there shouldn't be a way to gain write access (or sensitive information) to the repo through travis.

This doesn't cover malicious actions from someone with read/write access to the repo who can open a PR from a branch on the main repo and expose the keys. I think this is probably unavoidable. Presumably anyone with that kind of access will be trusted not to though. Enabling 2FA (and always using signed commits) would probably help stop them from being compromised.

Other things that should probably be done to secure a repo that uses `deploydocs` would be to make `master` branch protected so that no one can force push to it and also require CI checks to pass before any branch can be merged with it. I've enable both of these for this repo. When all PRs are reviewed before merging I think the risks are probably quite low.

Also, if someone were to gain access to a repo and do something it can hopefully just be reverted using a trusted local clone. The tokens and ssh keys can also be revoked if problems arise.

Of course using travis to deploy documentation isn't a requirement, and it can all be done locally to completely avoid these risks. That's best left up to individual repo owners though.

---

If you, or anyone, can find holes in that that would be great. Feel free to open PRs that try to gain write access if you'd like :)

hayd:
From that doc page:

> Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.

which alleviates my concern. :)

Issue #57, opened 2016-05-18
Add a 'Pkg.generate'-like function for docs


MichaelHatherly:
Currently setting up Documenter for a package is a manual process. Would be nice to have a `setup`/`init`/`generate` function that did some of this automatically.

hayd:
(In the longer term /docs/src /docs/make.jl could be generated/stubbed in as part of [PkgDev](https://github.com/JuliaLang/PkgDev.jl) generate.)

mortenpi:
I suppose it would make sense to have `Documenter.generate()` available here and then `PkgDev` would just call that?

MichaelHatherly:
Yeah, that's probably the best approach, though integrating with `PkgDev` should wait a while I think, at least until `Documenter` is registered and tagged for a bit.

MichaelHatherly:
Should probably also have a second argument specifying the format. Just `"mkdocs"` as an option for the moment.

MichaelHatherly:
Closed by #66.

PR #66, opened 2016-05-26
Generate stub documentation

mortenpi:
Addresses #57.

Needs more polish before it can be merged, but thoughts on the exact API would be welcome. Right now I propose a very simple `Documenter.generate(pkgname)`.

Also, no format argument at the moment, just defaulting to mkdocs.

### DIFF
--- a/docs/src/lib/internals.md
+++ b/docs/src/lib/internals.md
@@ -88,6 +88,17 @@ Formats.Format
 Formats.mimetype
 ```
 
+## Generator
+
+```@docs
+Generator
+Generator.savefile
+Generator.make
+Generator.gitignore
+Generator.mkdocs
+Generator.index
+```
+
 ## Walkers
 
 ```@docs
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -8,6 +8,9 @@ Two functions are exported from this module for public use:
 - [`makedocs`](@ref). Generates documentation from docstrings and templated markdown files.
 - [`deploydocs`](@ref). Deploys generated documentation from *Travis-CI* to *GitHub Pages*.
 
+Additionally it provides the unexported [`Documenter.generate`](@ref), which can be used to
+generate documentation stubs for new packages.
+
 """
 module Documenter
 
@@ -29,6 +32,7 @@ submodule(:CrossReferences)
 submodule(:DocChecks)
 submodule(:Writers)
 submodule(:Deps)
+submodule(:Generator)
 
 # User Interface.
 # ---------------
@@ -439,6 +443,94 @@ function genkeys(package)
     end
 end
 
+end
+
+"""
+    generate(
+        pkgname;
+        dir = "<package directory>/docs"
+    )
+
+Creates a documentation stub for a package called `pkgname`. The location of
+the documentation is assumed to be `<package directory>/docs`, but this can
+be overriden with keyword arguments.
+
+It creates the following files
+
+    docs/
+        .gitignore
+        src/index.md
+        make.jl
+        mkdocs.yml
+
+# Positionals
+
+**`pkgname`** is the name of the package (without `.jl`). It is used to
+determine the location of the documentation if `dir` is not provided.
+
+# Keywords
+
+**`dir`** defines the directory where the documentation will be generated.
+It defaults to `<package directory>/docs`. The directory must not exist.
+
+# Examples
+
+    julia> using Documenter
+
+    julia> Documenter.generate("MyPackageName")
+    [ ... output ... ]
+
+"""
+function generate(pkgname::AbstractString; dir=nothing)
+    # TODO:
+    #   - set up deployment to `gh-pages`
+    #   - fetch url and username automatically (e.g from git remote.origin.url)
+
+    # Check the validity of the package name
+    if length(pkgname) == 0
+        error("Package name can not be an empty string.")
+    end
+    # Determine the root directory where we wish to generate the docs and
+    # check that it is a valid directory.
+    docroot = if dir === nothing
+        pkgdir = Pkg.dir(pkgname)
+        if !isdir(pkgdir)
+            error("Unable to find package $(pkgname).jl at $(pkgdir).")
+        end
+        joinpath(pkgdir, "docs")
+    else
+        dir
+    end
+
+    if ispath(docroot)
+        error("Directory $(docroot) already exists.")
+    end
+
+    # deploy the stub
+    try
+        info("Deploying documentation to $(docroot)")
+        mkdir(docroot)
+
+        # create the root doc files
+        Generator.savefile(docroot, ".gitignore") do io
+            write(io, Generator.gitignore())
+        end
+        Generator.savefile(docroot, "make.jl") do io
+            write(io, Generator.make(pkgname))
+        end
+        Generator.savefile(docroot, "mkdocs.yml") do io
+            write(io, Generator.mkdocs(pkgname))
+        end
+
+        # Create the default documentation source files
+        Generator.savefile(docroot, "src/index.md") do io
+            write(io, Generator.index(pkgname))
+        end
+    catch
+        rm(docroot, recursive=true)
+        rethrow()
+    end
+    nothing
 end
 
 end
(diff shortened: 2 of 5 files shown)


MichaelHatherly:
> Also, no format argument at the moment, just defaulting to mkdocs.

Sounds good for the moment. We can add to it later once #3 and #4 get implemented.

> but thoughts on the exact API would be welcome.

Looks good to me as is really.

Will need some kind of tests if possible, probably create a temp dir, run `generate` on it and check the structure looks alright.

src/Documenter.jl
```diff
@@ -8,6 +8,9 @@ Two functions are exported from this module for public use:
 - [`makedocs`](@ref). Generates documentation from docstrings and templated markdown files.
 - [`deploydocs`](@ref). Deploys generated documentation from *Travis-CI* to *GitHub Pages*.
 
+Additionally it provides the unexported [`generate`](@ref), which can be used to
```
MichaelHatherly: Perhaps ``[`Documenter.generate`](@ref)`` to make it more obvious that it's unexported?

src/Documenter.jl
```diff
@@ -441,4 +445,93 @@ end
...
+    julia> using Documenter
+
+    julia> Documenter.generate("MyPackageName")
+    [ ... output ... ]
+
+"""
+function generate(pkgname::AbstractString; dir=nothing)
+    # TODO:
+    #   - `format` argument to determine the deployment (e.g. mkdocs)
+    #   - set up deployment to `gh-pages`?
```
MichaelHatherly: Sounds reasonable, perhaps as a subsequent PR, since that would need some changes to the hosting docs as well.

src/Documenter.jl
```diff
@@ -441,4 +445,93 @@ end
...
+    end
+    Generator.savefile(docroot, "make.jl") do io
+        write(io, Generator.make(pkgname))
+    end
+    Generator.savefile(docroot, "mkdocs.yml") do io
+        write(io, Generator.mkdocs(pkgname))
+    end
+
+    # Create the default documentation source files
+    mkdir(joinpath(docroot,"src"))
```
MichaelHatherly: Should `savefile` handle creation of subfolders such as `docs/src`?

src/modules/Generator.jl
```diff
@@ -0,0 +1,86 @@
+"""
+Provides the functions related to generating documentation stubs.
+"""
+module Generator
+
+"""
+    savefile(f, root, filename)
+
+Attempts to save a file at `\$(root)/\$(filename)`. `f` will be called with file
+stream (see [`open`](http://docs.julialang.org/en/latest/stdlib/io-network/#Base.open)).
```
MichaelHatherly: Not totally relevant here at the moment of course, but would be cool to be able to use `@ref` across different packages, not sure the nicest way to do that though.

src/modules/Generator.jl
```diff
@@ -0,0 +1,86 @@
...
+
+"""
+Contents of the default `mkdocs.yml` file.
+"""
+function mkdocs(pkgname)
+    """
+    site_name:        $(pkgname).jl
+    #repo_url:         https://github.com/USER_NAME/PACKAGE_NAME.jl
+    #site_description: Description...
+    #site_author:      USER_NAME
```
MichaelHatherly: Could we automatically populate `USER_NAME` and `PACKAGE_NAME`? (I'm not too bothered if not though.)

MichaelHatherly:
Appveyor errors is unrelated.

src/Documenter.jl
mortenpi: It would make the code a bit nicer I think. One option would be to call as `savefile(root,"src/index.md")` and then have `savefile` split the name on `/`s and create the directories, if necessary.

src/modules/Generator.jl
mortenpi: Yeah, would be nice, but I am not sure where to get the information from (automatically). Of course, `generate` could take keywords, but I think it is easier for the user to change this stuff manually later.

I was thinking that we could check the repo's `origin` ref and take the URL from there, but I didn't look into that yet.

The `mkdocs()` could nevertheless take these things as arguments though, even if they do not get used right now..

MichaelHatherly: See [`getremote`](https://github.com/MichaelHatherly/Documenter.jl/blob/3b76ac806c42a5580354a0c4d42ac31d8cfc020a/src/modules/Utilities.jl#L355-L369) for how to get at the `origin` in `0.5`, it's also possible to get in `0.4`, but since `Base.LibGit2` doesn't exist there you need to shell out to `git` directly.

For the moment we can just got with the placeholder text since users will need to make some changes to the file any way.

src/Documenter.jl
MichaelHatherly: Yes, that sounds nicer.

src/Documenter.jl
```diff
@@ -441,4 +445,93 @@ end
 
 end
 
+"""
+    generate(
+        pkgname;
+        dir = "<package directory>"
```
hayd: Should read `"<package directory>/docs"` ?

docs/src/lib/internals.md
```diff
@@ -88,6 +88,16 @@ Formats.Format
 Formats.mimetype
 ```
 
+## Generator
```
MichaelHatherly: Blank line after headers, just to be consistent with the rest of the package.

src/modules/Generator.jl
```diff
@@ -0,0 +1,114 @@
...
+        modules = [$(pkgname)]
+    )
+
+    # Documenter can also automatically deploy documentation to gh-pages.
+    # See "Hosting Documentation" and deploydocs() in the Documenter manual
+    # for more information.
+    #=deploydocs(
+        repo = "<repository url>"
+    )=#
+    """
```
MichaelHatherly: Preferably just normal comments rather than a multiline one. Good idea with adding some explanation here though. +1

test/runtests.jl
```diff
@@ -216,6 +216,46 @@ end
 
 rm(joinpath(Documenter_root, "build"); recursive = true)
 
+# Try generating documentation with generate()
+
+pkg_generate = if VERSION >= v"0.5-"
+    Pkg.add("PkgDev")
```
MichaelHatherly: Can this be conditional on something like `isdir(Pkg.dir("PkgDev")) || Pkg.add("PkgDev")`? Otherwise the tests take quite a while to run for me whether `PkgDev` is installed or not.

MichaelHatherly:
Other than a couple of tiny nitpicks, LGTM.

src/modules/Generator.jl
```diff
@@ -0,0 +1,114 @@
...
+      - Home: index.md
+    """
+end
+
+"""
+Contents of the default `src/index.md` file.
+"""
+function index(pkgname)
+    """
+    # $(pkgname).jl
```
MichaelHatherly: As with the other comment, blank line after header here too.

src/modules/Generator.jl
mortenpi: I actually specifically went with the multilines so that it would be easier for the user to comment it in. I.e. this tries to be a commented out code block, not just a code example. For three lines it doesn't matter much of course, but would you still go with normal ones?

MichaelHatherly: OK, that makes sense, stick with a multiline one here then.

test/runtests.jl
mortenpi: I don't see that interestingly.. the "Nothing to be done" pops out quite quickly. But we can have that check just in case.

I was actually wondering -- is there a better way of handling deps for tests?

MichaelHatherly: It depends how many packages you've got, how recently `Pkg.update` was run, and the state of repos. Can take a while sometimes.

If `PkgDev` was installable in `0.4` then we could add it to a `test/REQUIRE` file, but as far as I know it's limited to `0.5` since it requires `Base.LibGit2`. The way you've done it would be how I would have do it as well.

mortenpi:
Fixed the nitpicks.

MichaelHatherly:
Great, I'll merge once appveyor finishes. Then later this evening I'll move the package over to [JuliaDocs](https://github.com/JuliaDocs) (issue #69) and add you as a co-owner so you can have a bit more freedom to implement what you've got planned. Probably best for neither of us to push direct to `master` from that point on of course.

PR #74, opened 2016-05-31
Refactor index and contents node code

MichaelHatherly:
Moves filter/sorter code out of the `Writers` modules and into the `Documents` module. Removes some duplicate code related to construction of `IndexNode`s and `ContentsNode`s.

These changes should help with simplifying support for alternate output formats such as HTML by removing non-output related code from the `Writers` module.

### DIFF
--- a/src/modules/Documents.jl
+++ b/src/modules/Documents.jl
@@ -17,6 +17,8 @@ import ..Documenter:
 
 using Compat
 
+import Compat: String
+
 # Pages.
 # ------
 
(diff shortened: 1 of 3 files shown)


MichaelHatherly:
Thanks for reviewing that @mortenpi.

mortenpi:
I am getting a lot of
```
WARNING: Base.String is deprecated, use AbstractString instead.
```
from `Documents.jl` (running on 0.4). Should we have a `using Compat: String` there?

I guess it would be nice to unify the different string types as well (right now, sometimes it's `AbstractString`, sometimes `String` and sometimes `Compat.String`).

MichaelHatherly:
Should be fixed now, thanks for catching that.

Re: string types. Agreed. If there are any fields with `AbstractString` then those should be changed to `String` and we should use `Compat.String` everywhere we need a concrete type. At least we're not needing to support 0.3 as well...

PR #76, opened 2016-06-01
Implement Selector interface

MichaelHatherly:
Currently the control flow in Documenter used to determine how to transform a `Document` is defined in a rather cryptic way.

The "selectors" added here should hopefully allow for a more straight forward approach to implementing the "pipeline" tuple and "expander" functions. Neither of those are done in this commit, but will be added in a subsequent one.


mortenpi:
How would you disable a particular match? E.g. in the case of the pipeline, say you don't want to run a particular action at this point. Right now it seems to me that you would have to define a completely new set of cases?

MichaelHatherly:
Hadn't thought of that particular case, probably just with overwriting the relevant definition:

```julia
Selectors.matcher(::Type{Foo}, x) = false
```

to disable matching `Foo`. That would print a warning though and wouldn't work once the `selector` method has been compiled.

Or perhaps provide a simple way to create a new selector without a particular `matcher`.

MichaelHatherly:
Or maybe allow `Selectors.order` to return a `(<order>, <precedence>)` and when only a scalar is returned default to `(<scalar>, 1.0)`. Then order the subtypes using `<order>` but when there are duplicates the the highest `<precedence>` "wins". Then you can override things using

```julia
Selectors.order(::Type{MyCase}) = 1.0

# ... later on

Selectors.order(::Type{MyReplacementCase}) = 1.0, 2.0
```

MichaelHatherly:
Or another option... a `Selectors.disable{T}(::Type{T}) = false` that can be overloaded to disable a particular `T` with

```julia
Selectors.disable(::Type{Foo}) = true
```

Then the generated function just needs to add a check for that. Would probably just get inlined all the time and compiled away to nothing which should be nice.

src/modules/Selectors.jl
```diff
@@ -0,0 +1,151 @@
+"""
+An extensible code selection interface.
+
```
mortenpi: I would have a quick overview of the interface here, I was initially somewhat confused with the example.

Perhaps something along the lines of:

> It makes writing code that has to dispatch based on a condition easy.
> The call to `selector(::T, x)` chooses the appropriate `runner` to call based on which of the `matcher`s
> returns `true`.

Perhaps it would be useful to say that `selector` effectively is a nice automatically generated if statement?
```julia
function selector(x)
    if matcher(::First, x)
        runner(::First, x)
    elseif matcher(::Second, x)
        runner(::Second, x)
    ...
```

src/modules/Selectors.jl
```diff
@@ -0,0 +1,151 @@
...
+strict{T <: AbstractSelector}(::Type{T}) = true
+
+"""
+Generated function that builds a specialised selector for each selector type provided, i.e.
+
+```julia
+Selectors.selector(MySelector, 1)
+```
+"""
+@generated function selector{T <: AbstractSelector}(::Type{T}, x...)
```
mortenpi: Not sure about this, but would it make sense to have the function name be a verb? `select`? Or maybe `dispatch`?

src/modules/Selectors.jl
MichaelHatherly: Yes, an example of the generated code would be good, as well as the better description.

src/modules/Selectors.jl
MichaelHatherly: Yeah, a verb would be better. Since `Base.select` is exported and does something completely different `dispatch` seems good to me.

mortenpi:
Out of the three, the first one actually seems the nicest, because of its simplicity. The `disable` one would mean that you would have to define it for each of them.. well or I guess you can define it once for the supertype.

Should there also be a macro to generate the boilerplate?
```julia
@selector EvenOdd Even Odd
```
It could generate the types, possibly the `disable` and a default ordering.

Can't write docstrings for the different cases though.

MichaelHatherly:
> The `disable` one would mean that you would have to define it for each of them

Only a single

```julia
disable{T <: AbstractSelector}(::Type{T}) = false
```

needs to be defined in the `Selectors` module, and a call to it in `selector` too. Seems to be the clearest approach since a user would only ever need to add definitions for ones they disable, can even do several at once

```julia
disable{T <: Union{A, B, C}}(::Type{T}) = true
```
Not sure how common disabling things would be really.

> Should there also be a macro to generate the boilerplate?

Perhaps. I won't be completely against adding one, probably another separate PR later on though.

> Can't write docstrings for the different cases though.

Yeah, would just have to be attached afterwards with

```julia
"..."
@selector EvenOdd Even Odd

"..."
Even

"..."
Odd
```

MichaelHatherly:
Option 1's warning isn't great and probably shouldn't be something that's ever visible to a user. Having tried out the `disable` function I think that's probably the best option. I'll update the PR later with that addition.

Issue #13, opened 2016-01-10
Documentation ordering


hayd:
At the moment, all the md files are built and dumped into build.

If using mkdocs (#10) these have to be specified in the yaml. One could do the same in a top level latex file to input these and add style. This seems like a duplication of effort (specifying the ordering).

I would propose, either:

1. mkdocs.yml pages-section is required, use this to decide the ordering and naming of sections. Warn/error if sections are missing in both src or yaml. Use this for generating single-page html/latex.
2. makedocs function allows passing of this ordering (as title => filename pairs). Warn/error if sections missing or do not exist in both src or yaml.

My preference is the mkdocs.yml as this already exists, and allows nested files etc.

This is crucial in compiling to a single document (e.g. #4).

MichaelHatherly:
I don't mine either option really. Go with the `mkdocs.yml` and we can see how it works out in practise.

MichaelHatherly:
Specifying page order was added in https://github.com/JuliaDocs/Documenter.jl/commit/e7de65974efcbeadbadc16e40feec2f1d8aa14ad (currently internal and undocumented since markdown output doesn't really need it). Closing this issue since https://github.com/JuliaDocs/Documenter.jl/issues/100 supersedes this I think.

PR #82, opened 2016-06-03
Minor `Writers` refactoring

MichaelHatherly:
Moves some non-rendering code into more suitable modules.

Only check for the git remote url once on construction of the `Document` object and then reuse that when needed.

Add `Utilties.unwrap` to call a function on the value wrapped by a `Nullable` if it exists.

cc @mortenpi, it's not a complete refactor of course, but let me know if there's other bits in particular you'd like to see fixed up.

### DIFF
--- a/src/modules/CrossReferences.jl
+++ b/src/modules/CrossReferences.jl
@@ -69,7 +69,7 @@ function basicxref(link::Markdown.Link, meta, page, doc)
         # No `name` was provided, since given a `@ref`, so slugify the `.text` instead.
         text = strip(sprint(Markdown.plain, Markdown.Paragraph(link.text)))
         if ismatch(r"#[0-9]+", text)
-            issue_xref(link, text, meta, page, doc)
+            issue_xref(link, lstrip(text, '#'), meta, page, doc)
         else
             name = Utilities.slugify(text)
             namedxref(link, name, meta, page, doc)
@@ -134,18 +134,8 @@ end
 # Issues/PRs cross referencing.
 # -----------------------------
 
-if isdefined(Base, :LibGit2)
-    function issue_xref(link::Markdown.Link, num, meta, page, doc)
-        remote =
-            LibGit2.with(LibGit2.GitRepoExt(dirname(doc.user.root))) do repo
-                LibGit2.with(LibGit2.GitConfig(repo)) do cfg
-                    Utilities.getremote(cfg)
-                end
-            end
-        isnull(remote) || (link.url = "https://github.com/$(get(remote))/issues/$(num[2:end])")
-    end
-else
-    issue_xref(link::Markdown.Link, num, meta, page, doc) = nothing
+function issue_xref(link::Markdown.Link, num, meta, page, doc)
+    link.url = "https://github.com/$(doc.internal.remote)/issues/$num"
 end
 
 end
--- a/src/modules/CrossReferences.jl
+++ b/src/modules/CrossReferences.jl
@@ -100,6 +100,7 @@ function namedxref(link::Markdown.Link, slug, meta, page, doc)
             # Replace the `@ref` url with a path to the referenced header.
             anchor   = get(Anchors.anchor(headers, slug))
             path     = relpath(anchor.file, dirname(page.build))
+            path     = Formats.extension(doc.user.format, path)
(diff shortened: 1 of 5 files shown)


MichaelHatherly:
Second commit adds the only (hopefully) required changes to make adding HTML rendering relatively straightforward to implement.

All links are automatically rewritten to have the correct extension depending on the value of the `document.user.format` field.

Some stub methods for the HTML and LaTeX `render` methods have been added as well.

MichaelHatherly:
So ideally the plan for implementing HTML (and LaTeX some time later) rendering is as follows:

- replicate the markdown `render` methods, but with `text/html` instead `text/plain` for their mimetype argument
- adjust the markdown specific output for each of those new methods
- add some structural HTML `div` tags (or whatever's needed), such as around individual `DocsNode`s, that can be used to attach some nice css styling to elements
- wrap each page in whatever header and footer html is needed.

I'm guessing the later steps will probably be more time-consuming to get something that's on the same level as `mkdocs` output.

Other things that aren't too important for a first pass at this:

- site navigation, currently we don't have a way to specify what order the individual pages should be arranged in some kind of side/top bar nav (related #13)
- support for user-defined themes like sphinx and mkdocs, not too important right now, but might need to keep that kind of stuff in mind while designing things
- like above, user-defined javascript files.

MichaelHatherly:
Also: regarding rendering of several formats at once. I think we should just get single rendering "passes" working well for all the formats first and then look into ways to make things for fancy.

src/modules/CrossReferences.jl
```diff
@@ -134,18 +136,8 @@ end
...
-                LibGit2.with(LibGit2.GitConfig(repo)) do cfg
-                    Utilities.getremote(cfg)
-                end
-            end
-        isnull(remote) || (link.url = "https://github.com/$(get(remote))/issues/$(num[2:end])")
-    end
-else
-    issue_xref(link::Markdown.Link, num, meta, page, doc) = nothing
+function issue_xref(link::Markdown.Link, num, meta, page, doc)
+    link.url = "https://github.com/$(doc.internal.remote)/issues/$num"
```
mortenpi: Not very relevant for this PR, but should we have a `github_url` or something argument for mkdocs to determine this? The remote may not be 100% reliable, e.g. if you're building on a cloned fork. We should default to origin remote though, I think.

src/modules/Writers.jl
```diff
@@ -157,4 +154,21 @@ dropheaders(h::Markdown.Header) = Markdown.Paragraph(Markdown.Bold(h.text))
 dropheaders(v::Vector) = map(dropheaders, v)
 dropheaders(other) = other
 
+
+# LaTeX Output.
+# -------------
+
+# TODO
+
+function render(io::IO, ::MIME"text/latex", node, page, doc)
```
mortenpi: Hopefully not very relevant for long, but perhaps they should `error` at the moment?

mortenpi:
I would add "document `Document`" to the todo list for supporting other formats, so that it would be clear what information a node contains.

rendering multiple: I think it would be a good thing to keep in mind though.. it would force us to compartmentalize the code more, which is good. But yeah, having a fancy multi-format build-deploy workflow going probably shouldn't be the highest priority.

src/modules/CrossReferences.jl
MichaelHatherly: Yes, probably not ideal, though for the moment I think it's reasonable to assume it exists since `Pkg/PkgDev` rely on existence of remote `origin` if I remember correctly.

src/modules/Writers.jl
MichaelHatherly: Sure, seems reasonable to me.

MichaelHatherly:
I'll add some docs for the things in the `Documents` module in a separate PR a little later today probably.

For rendering multiple formats, it'll probably only need to following changes:

- your `wd/` idea for each format
- tracking of `Markdown.Link` objects that need to have their `.url` fields changed depending on format
- running the rendering stage of the pipeline however many times is necessary.

As far as I can tell that can be largely added independently of additional renderers, and would probably be best to do once we have more than just markdown to try it out on, instead of adding additional code now without being able to test it properly.

PR #86, opened 2016-06-07
Add Currencies.jl to examples

TotalVerb:
I converted Currencies.jl from Sphinx to Documenter recently. It was very easy to use; great work!


MichaelHatherly:
Thanks!

PR #87, opened 2016-06-07
Fix a typo in Utilities.jl

mortenpi:
It didn't come up in tests. Should we add some?


MichaelHatherly:
Thanks for catching that, not sure if that code path can ever actually be hit.

Yeah, we should probably start adding some unit tests at some point soonish, but doesn't have to be right away.

mortenpi:
It can definitely be called, that's how I happened to stumble upon it.

MichaelHatherly:
Oops, yeah, I wasn't looking far enough up the diff...

PR #90, opened 2016-06-09
Implement method lists

mortenpi:
Things left to do:
 - [x] Make it work with macros (fixed by getting rid of an `eval`)
 - [x] Fix 0.4 (-> don't support at the moment)
 - [x] Typevars
 - [x] The last Type method (`(::Type{T})`) (-> filtered out)
 - [x] Fix docstring source links (too much whitespace) (-> won't fix)
 - [x] Filtering by module.
 - [x] Documentation.

Decisions:
 - [x] Disabling the printing of method tables? (-> possibly implement this in the future)
 - [x] Don't print for abstract types? (-> implemented)

---

A preliminary version of method lists for the markdown output. It puts the methods from the `methods(obj)` function as an ordered list to the end to the docstring for every name that has methods.

Basically looks like this:
typesigs

Currently macros are not working since an `eval` fails. Also, due to differences between 0.4 and 0.5, 0.4 is currently not working. One option would be to disable this feature for 0.4 completely, but in principle it should be possible to work around the current problems.

### DIFF
--- a/assets/Documenter.css
+++ b/assets/Documenter.css
@@ -5,3 +5,14 @@ div.wy-menu-vertical ul.current li.toctree-l3 a {
 a.documenter-source {
   float: right;
 }
+
+.documenter-methodtable pre {
+    margin-left: 0px;
+    margin-right: 0px;
+    margin-top:  0px;
+    padding:     0px;
+}
+
+.documenter-methodtable pre.documenter-inline {
+    display: inline;
+}
--- a/src/modules/Expanders.jl
+++ b/src/modules/Expanders.jl
@@ -261,16 +261,88 @@ end
 # @docs
 # -----
 
+immutable MethodNode
+    method  :: Method
+    visible :: Bool
+end
+
 immutable DocsNode
-    docstr :: Any
-    anchor :: Anchors.Anchor
-    object :: Utilities.Object
-    page   :: Documents.Page
+    docstr  :: Any
+    anchor  :: Anchors.Anchor
+    object  :: Utilities.Object
+    page    :: Documents.Page
+    """
+    Vector of methods associated with this `DocsNode`. Being nulled means that
+    conceptually the `DocsNode` has no table of method (as opposed to having
+    an empty table).
+    """
+    methods :: Nullable{Vector{MethodNode}}
 end
 
 immutable DocsNodes
     nodes :: Vector{DocsNode}
 end
+
+"""
+Returns a `Nullable{Vector{MethodNode}}` with the methods associated with the `object`.
+
+Null is returned if the object conceptually does not have a table of methods (e.g.
+modules). If instead the object should have a table of methods which just happens
+to be empty then it returns an empty vector.
+
+The methods are also filtered, with `MethodNode.visible` set to false for those
+methods not defined in this package. This allows the writers to choose how they
+want to treat trivial constructors, functions imported from `Base` and such.
+
+Due to differences between 0.4 and 0.5, only 0.5 is supported currently. On 0.4
+we always drop the methods table.
+"""
+function docsnode_methodlist end
+
+if VERSION > v"0.5-"
+    function docsnode_methodlist(object::Utilities.Object, page, doc)
+        if !isdefined(object.binding.mod, object.binding.var)
+            info("Unable to find $(object)")
+            return Nullable{Vector{MethodNode}}()
+        end
+
+        # Fetch the actual object and its category
+        object_deref = getfield(object.binding.mod, object.binding.var)
+        doccat = Utilities.doccat(object)
+
+        # We do not attach a method table to a `DocsNode` of an abstract type,
+        # except when it has 2 or more methods. While abstract types can not be
+        # constructed, they do have the default (::Type{T})(args..) method attached
+        # to them, but this is not very interesting to the user. However, it is
+        # possible to add additional methods to abstract types and if that is the
+        # case then we should display the method table, hence the "less than two"
+        # check.
+        if (doccat == "Type") && !isleaftype(object_deref) && length(methods(object_deref)) < 2
+            return Nullable{Vector{MethodNode}}()
+        end
+
+        # We only generate the method table for a Function, Type and Macro.
+        # The only other one with methods should be Method itself (e.g. defined
+        # with `foo(x,y)` in the @docs block), but in that case the user has
+        # already picked a method and we shouldn't generate a single element
+        # table.
+        if doccat == "Function" || doccat == "Type" || doccat == "Macro"
+            ms = map(methods(object_deref)) do m
+                # We filter out the methods not defined in this package.
+                # That is we only show the method if it is defined in any of the
+                # modules in `doc.user.modules`, which is derived from the `modules`
+                # setting in `makedocs`.
+                MethodNode(m, m.module in doc.user.modules)
+            end
+            return Nullable(ms)
+        else
+            return Nullable{Vector{MethodNode}}()
+        end
+    end
+else
+    docsnode_methodlist(args...) = Nullable{Vector{MethodNode}}()
+end
+
 
 function Selectors.runner(::Type{DocsBlocks}, x, page, doc)
     failed = false
@@ -303,7 +375,8 @@ function Selectors.runner(::Type{DocsBlocks}, x, page, doc)
         # Update `doc` with new object and anchor.
         docstr   = get(filtered)
         anchor   = Anchors.add!(doc.internal.docs, object, slug, page.build)
-        docsnode = DocsNode(docstr, anchor, object, page)
+        ms = docsnode_methodlist(object, page, doc)
+        docsnode = DocsNode(docstr, anchor, object, page, ms)
         doc.internal.objects[object] = docsnode
         push!(nodes, docsnode)
     end
--- a/src/modules/Writers.jl
+++ b/src/modules/Writers.jl
@@ -59,6 +59,31 @@ function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNodes, page,
     end
 end
 
+"""Wrap a string `str` in a `<span class="<cls>">`."""
+span(cls,str) = "<span class=\"$(cls)\">$(str)</span>"
+
+"""
+Converts the function argument tuple `(name, type)` into a string.
+
+The tuple comes from the second return element of the `Base.arg_decl_parts(::Method)`
+and it seems they are always both `::String` (`::ASCIIString` in 0.4).
+It also appears that if the type is not declared for the method, `arg_decl_parts`
+returns an empty string.
+
+The returned string is `name::type` or just `name`, if the type is not declared.
+
+If the keyword argument `html` is true (default), then it also puts `<span>`s
+around the characters for code highlighting.
+"""
+function join_decl(decl; html::Bool=true)
+    n, t = decl
+    if html
+        isempty(t) ? span(:n,n) : span(:n,n) * span(:p,"::") * span(:n,t)
+    else
+        isempty(t) ? n : "$(n)::$(t)"
+    end
+end
+
 function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNode, page, doc)
     # Docstring header based on the name of the binding and it's category.
     anchor = "<a id='$(node.anchor.id)' href='#$(node.anchor.id)'>#</a>"
@@ -66,6 +91,90 @@ function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNode, page, 
     println(io, anchor, "\n", header, "\n\n")
     # Body. May contain several concatenated docstrings.
     renderdoc(io, mime, node.docstr, page, doc)
+
+    # Table of methods.
+    # If DocsNode.methods is nulled, then we assume that we should not render a
+    # table. However, if the list of methods is there but has no elements then
+    # we output an appropriate note saying that the name has no methods associated
+    # with it.
+    Utilities.unwrap(node.methods) do methodnodes
+        name = node.object.binding.var # name of the method without the modules
+
+        # We filter out the methods that are marked `visible`
+        ms = [m.method for m in filter(m -> m.visible, methodnodes)]
+
+        println(io, "<strong>Methods</strong>\n")
+
+        # We print a small notice of the methods table is completely empty,
+        # and an unordered list of methods if there are some to display.
+        if isempty(methodnodes)
+            println(io, "This function has no methods.\n")
+        elseif isempty(ms)
+            println(io, "This function has no methods to display.\n")
+        else
+            # A regexp to match filenames with an absolute path
+            r = Regex("$(Pkg.dir())/([A-Za-z0-9]+)/(.*)")
+
+            print(io, """
+            <ul class="documenter-methodtable">
+            """)
+
+            for m in ms
+                tv, decls, file, line = Base.arg_decl_parts(m)
+                decls = decls[2:end]
+                file = string(file)
+                url = get(Utilities.url(doc.internal.remote, m.module, file, line), "")
+                file_match = match(r, file)
+                if file_match !== nothing
+                    file = file_match.captures[2]
+                end
+                # We'll generate the HTML now.
+                #
+                # We also apply code highlighting that tries to be consistent with
+                # how the code blocks are highlighted in mkdocs. In a nutshell, all
+                # characters have to be wrapped in <span>s with specific classes.
+                # The classes seem to have the following semantics:
+                #
+                #   - p   punctuation, e.g. {} () :: ,.
+                #   - k   keyword, e.g. type, return
+                #   - nf  function name, in a function definition
+                #   - n   name (generally, as used in code)
+                #
+                # TODO: the type expressions (in typevars or after ::) are colored
+                #       as normal names, but it would be nice to have them properly
+                #       highlighted (e.g. if there's a string in the type name,
+                #       like in MIME"text/html")
+                #
+                tvars = isempty(tv) ? "" :
+                    span(:p,"{") * join([span(:n,t) for t in tv], span(:p,", ")) * span(:p,"}")
+                # If the list of arguments is too long (which can happen quite easily
+                # due to long type names), we will display them in a multiline block
+                # instead.
+                args_raw = join([join_decl(d, html=false) for d in decls], span(:p,", "))
+                args,preclass = if length(args_raw) <= 50
+                    join([join_decl(d) for d in decls], span(:p,", ")),
+                    " class=\"documenter-inline\""
+                else
+                    "\n" * join([(" "^4)*join_decl(d) for d in decls], span(:p,",\n")) * "\n",
+                    ""
+                end
+                print(io, """
+                <li>
+                    <pre$(preclass)>$(span(:nf,name))$(tvars)$(span(:p,"("))$(args)$(span(:p,")"))</pre>
+                    defined at
+                    <a target="_blank" href="$(url)">$(file):$(line)</a>
+                </li>
+                """)
+            end
+            print(io, "</ul>\n\n")
+        end
+
+        # we print a small notice if we are not displaying all the methods
+        nh = length(methodnodes)-length(ms) # number of hidden methods
+        if nh > 0
+            println(io, "_Hiding $(nh) method$(nh==1?"":"s") defined outside of this package._\n")
+        end
+    end
 end
 
 function renderdoc(io::IO, mime::MIME"text/plain", md::Markdown.MD, page, doc)
--- a/docs/src/lib/internals.md
+++ b/docs/src/lib/internals.md
@@ -82,6 +82,7 @@ Expanders.IndexBlocks
 Expanders.ContentsBlocks
 Expanders.ExampleBlocks
 Expanders.REPLBlocks
+Expanders.docsnode_methodlist
 ```
 
 ## Formats
@@ -128,6 +129,8 @@ Walkers.walk
 ```@docs
 Writers
 Writers.render
+Writers.join_decl
+Writers.span
 ```
 
 ## Utilities
@@ -149,4 +152,5 @@ Utilities.object
 Utilities.docs
 Utilities.doccat
 Utilities.nodocs
+Utilities.issubmodule
 ```
--- a/src/modules/Utilities.jl
+++ b/src/modules/Utilities.jl
@@ -448,4 +448,20 @@ function withoutput(func, stream::CombinedStream, buffer::IOBuffer)
     end
 end
 
+"""
+    issubmodule(sub, mod)
+
+Checks whether `sub` is a submodule of `mod`. A module is also considered to be
+its own submodule.
+
+E.g. `A.B.C` is a submodule of `A`, `A.B` and `A.B.C`, but it is not a submodule
+of `D`, `A.D` nor `A.B.C.D`.
+"""
+function issubmodule(sub, mod)
+    if (sub === Main) && (mod !== Main)
+        return false
+    end
+    (sub === mod) || issubmodule(module_parent(sub), mod)
+end
+
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -87,6 +87,8 @@ using Compat
 
 module UnitTests
 
+module SubModule end
+
 type T end
 
 "Documenter unit tests."
@@ -108,6 +110,14 @@ let doc = @doc(length)
     @test !contains(stringmime("text/plain", get(c)), "Documenter unit tests.")
     @test isnull(d)
 end
+
+# Documenter.Utilities.issubmodule
+@test Documenter.Utilities.issubmodule(Main, Main) === true
+@test Documenter.Utilities.issubmodule(UnitTests, UnitTests) === true
+@test Documenter.Utilities.issubmodule(UnitTests.SubModule, Main) === true
+@test Documenter.Utilities.issubmodule(UnitTests.SubModule, UnitTests) === true
+@test Documenter.Utilities.issubmodule(UnitTests.SubModule, Base) === false
+@test Documenter.Utilities.issubmodule(UnitTests, UnitTests.SubModule) === false
 
 
 # Integration tests for module api.


assets/Documenter.css
```diff
@@ -5,3 +5,16 @@ div.wy-menu-vertical ul.current li.toctree-l3 a {
 a.documenter-source {
   float: right;
 }
+
+/* .methodtable */
+.methodtable pre {
```
MichaelHatherly: Should these classes be prefixed with a `documenter-` just to make sure they stay separate from user-defined things?

MichaelHatherly: Prefix class with `documenter-`? That's what I've done with `source` above, but if you don't feel it'll be needed feel free to change that one as well.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+
+The tuple comes from the second return element of the `Base.arg_decl_parts(::Method)`
+and it seems they are always both `::String` (`::ASCIIString` in 0.4).
+It also appears that if the type is not declared for the method, `arg_decl_parts`
+returns an empty string.
+
+The returned string is `name::type` or just `name`, if the type is not declared.
+"""
+function _function_arg(decl)
+    n,t = decl
```
MichaelHatherly: Space after `,`.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+    if length(t)==0
```
MichaelHatherly: Can be a `?:` since it's pretty short. Also preferable to use `isempty(t)` rather than a `length` check.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+
+"""
+Returns a markdown paragraph (`Markdown.Paragraph`) with a list of methods.
+
+The return value will be wrapped in a `Nullable`, and a `null` is returned
+if the object in `ex` has no methods.
+"""
+function methodtable_md(mod::Module, ex, page, doc)
+    # TODO:
+    #   - fails on 0.4
```
MichaelHatherly: For the moment lets just limit this feature to `0.5` and work on `0.4` support later.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+    #   - doesn't work with macros (problem with eval)
+    obj = try
+        mod.eval(ex)
```
MichaelHatherly: Should be possible to just use the `object` defined in the `DocsBlock` `runner` method which contains the type signature and `Binding` to the actual object rather than re-evaluate it again here.

Will also need to limit the methods displayed by using `methods(object, signature)` when `signature !== Union{}` otherwise we'll end up with duplicate lists when someone includes something like

    ```@docs
    func(x)
    func(x, y)
    ```

in their docs I think.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+    catch e
+        warn("Unable to eval `$(ex)` in module $(mod).")
+        info("Error: "*string(e))
+        nothing
+    end
+
+    # If we were not able to eval the expression or the object doesn't have any
+    # methods (such as modules), we just return a `null`.
+    if obj === nothing || length(methods(obj)) == 0
```
MichaelHatherly: `isempty(methods(obj))` here.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+      return Nullable{Markdown.Paragraph}()
+    end
+
+    pkgdir = Pkg.dir()
+    s = """
+    <p><strong>Methods</strong></p>
+    <ol class="methodtable">
+    """
```
MichaelHatherly: Better to use an `IOBuffer` and `print` to it instead of using string concatenation. Then just do `Markdown.Paragraph(takebuf_string(buffer))` at the end.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+
+    # Determine the name of the method.
+    # In general the name is Module.Submodule.object, and so we drop all
+    # the modules and just pick the name.
+    name = split(string(obj), ".")[end]
```
MichaelHatherly: By using the `object` mentioned above you'll be able to get the `name :: Symbol` from `object.binding.var`.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,102 @@ immutable DocsNodes
...
+
+    # A regexp to match filenames with an absolute path
+    r = Regex("$(Pkg.dir())/([A-Za-z0-9]+)/(.*)")
+
+    for m=methods(obj)
```
MichaelHatherly: `in` rather than `=` for anything other than a literal `Range` object. Same with the comprehensions below. (I generally only use `=` when it's a literal `Range` object.)

MichaelHatherly:
The output looks good so far, nice! :+1:

The last method in the list shown is the default `(::Type{T}){T}(arg)` constructor. That one should be filtered out I think.

assets/Documenter.css
MichaelHatherly: (Hmm, so GitHub did post the first comment I made... weird.)

src/modules/Expanders.jl
mortenpi: Hmm, I didn't know that was supported. An option would be to actually drop the methods list in that case since the user is specifying the method anyway.

MichaelHatherly: That's a good solution. So only display method/constructor lists when it's a `Function` or `DataType`? (Slightly unrelated, might be nice to add an export list to `Module` docs automatically, similar to this, at some point.)

> Hmm, I didn't know that was supported.

I'm not sure how common it's use is currently, but it can come in handy sometimes I think.

MichaelHatherly:
Currently this is doing the "rendering" at expansion time, which means that only markdown/HTML output would work, not LaTeX or any other potential formats. Can this be split into two stages: collect all the needed info and store it in a new field in each `DocsNode` object in `runner`, and then do the format-specific rendering in `Writers`?

I'm also wondering whether we should have a way to option in/out of autogenerated content such as this on a per-`@docs` block basis? Something like

    ```@docs
    ShowMethodLists = false
    foo
    bar
    baz
    ```

or something like that.

mortenpi:
Yup, I agree, it should be divided up, I just wanted to start with the simplest approach. However, I do feel that this would be a part of a slightly bigger refactoring, because I feel that relying on the `Base.Markdown` bits to represent the docstrings etc. is also not that LaTex/etc-friendly. Strictly speaking, what I push to the docstring is valid markdown.

About turning it on/off, I think that would be a good feature, to cater to different use cases.  I would, however, have it as an argument to `@docs`, e.g.

    ```@docs methods=off
    foo
    bar
    baz
    ```

Not 100% sure how it should look though, especially to keep it consistent with the current style.

mortenpi:
> The last method in the list shown is the default (::Type{T}){T}(arg) constructor. That one should be filtered out I think.

It is a valid method though and you sometimes end up calling it.. so even though is always exists, in the interest of accuracy, I would keep it in the list actually.

mortenpi:
> Yup, I agree, it should be divided up, I just wanted to start with the simplest approach. However, I do feel that this would be a part of a slightly bigger refactoring, because I feel that relying on the Base.Markdown bits to represent the docstrings etc. is also not that LaTex/etc-friendly. Strictly speaking, what I push to the docstring is valid markdown.

Although, this does not block doing this properly now. So I am actually thinking about having `MethodsTableNode` and `MethodNode` (are they "nodes" though?), which would store all the information.

MichaelHatherly:
> However, I do feel that this would be a part of a slightly bigger refactoring

Whether there's other things to refactor shouldn't mean that additional things get added that will also need refactoring if we can avoid it.

> Although, this does not block doing this properly now. So I am actually thinking about having `MethodsTableNode` and `MethodNode` (are they "nodes" though?), which would store all the information.

It's part of the `DocsNode` rather than a totally separate bit of data, so just needs a new `Vector{Method}` field in the `DocsNode` type. The valid methods can be filtered during expansion and then formatted separately during the rendering phase.

> It is a valid method though and you sometimes end up calling it.. so even though is always exists, in the interest of accuracy, I would keep it in the list actually.

Not really a question of whether it always exists, but rather whether providing a link to it is helpful to the user. I don't feel that it adds much additional information, though if you do decide to keep it in it needs to be displayed with the correct `(::Type{T})` syntax if we're going for accuracy.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,104 @@ immutable DocsNodes
...
+"""
+function methodtable_md(mod::Module, object::Utilities.Object, page, doc)
+    # In case the object is a method (e.g. given as `foo(x,y)` in the @docs
+    # block) we do not generate a methods table.
+    if Utilities.doccat(object) == "Method"
+        return Nullable{Markdown.Paragraph}()
+    end
+
+    ms = try
+        methods(eval(object.binding.mod, object.binding.var))
```
MichaelHatherly:
```
isdefined(object.binding.mod, object.binding.var)
```

and

```
getfield(object.binding.mod, object.binding.var)`
```

rather than `try` and `eval` here.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,104 @@ immutable DocsNodes
...
+    catch e
+        warn("Unable to eval $(object)")
+        info("Error: $e")
```
MichaelHatherly: Can silently discard undefined bindings rather than printing warnings since it's autogenerated content.

src/modules/Expanders.jl
```diff
@@ -272,6 +272,104 @@ immutable DocsNodes
...
+        nothing
+    end
+
+    # If we were not able to eval the expression or the object doesn't have any
+    # methods (such as modules), we just return a `null`.
+    if ms === nothing || isempty(ms)
+      return Nullable{Markdown.Paragraph}()
+    end
+
+    pkgdir = Pkg.dir()
```
MichaelHatherly: Not used elsewhere?

src/modules/Expanders.jl
```diff
@@ -272,6 +272,104 @@ immutable DocsNodes
...
+    s = IOBuffer()
+    print(s, """
+    <p><strong>Methods</strong></p>
+    <ol class="documenter-methodtable">
+    """)
+
+    # Determine the name of the method.
+    # In general the name is Module.Submodule.object, and so we drop all
+    # the modules and just pick the name.
+    name = string(object.binding.var)
```
MichaelHatherly: `string` not really needed since it's only used in the two string interpolations below.

src/modules/Expanders.jl
```diff
@@ -302,6 +400,18 @@ function Selectors.runner(::Type{DocsBlocks}, x, page, doc)
 
         # Update `doc` with new object and anchor.
         docstr   = get(filtered)
+
+        # If there are any methods then we add a segment to the docstring.
+        # The markdown content of the docstring is actually the first element
+        # in the docstring content, and we push to that vector.
+        # Also, since there are differences between 0.4 and 0.5, we only support
+        # 0.5 at this point. There is not table generated for 0.4.
+        if VERSION > v"0.5-"
```
MichaelHatherly: I prefer to keep `VERSION` checks at toplevel so they're easier to find when removing them later. So rather add a mock `methodtable_md` definition when on version `0.4`:

```julia
if VERSION < v"0.5.0-dev"
    function methodtable_md(...)
        # the real def
    end
else
    methodtable_md(...) = Nullable{Markdown.Paragraph}()
end
```

mortenpi:
> It's part of the `DocsNode` rather than a totally separate bit of data, so just needs a new `Vector{Method}` field in the `DocsNode` type. The valid methods can be filtered during expansion and then formatted separately during the rendering phase.

Yeah, the table would be completely unnecessary. I would, however, like to wrap the `Method`, because it's not really documented and seems to change between versions. Then at least we'd have our own version which is stable and if something changes, you don't have to fix all the output formats.

> Not really a question of whether it always exists, but rather whether providing a link to it is helpful to the user. I don't feel that it adds much additional information, though if you do decide to keep it in it needs to be displayed with the correct `(::Type{T})` syntax if we're going for accuracy.

It's one of the constructors, so it should be there, in my opinion. Having the proper syntax I can get behind, although at this point I am still slightly confused about what it does exactly.

MichaelHatherly:
> because it's not really documented and seems to change between versions.

Do you mean the `Method` object? At least since `0.3` `collect(methods(func))` is just a vector of methods of `func`; hasn't changed much externally.

> at this point I am still slightly confused about what it does exactly.

Hence why I don't feel it should be included. It's pretty much just a fallback to `convert` for any type `T`'s constructor. (We should filter by module if `doc.internal.modules` is not empty, which would pretty much always filter it out anyway.)

mortenpi:
> Do you mean the `Method` object? At least since `0.3` `collect(methods(func))` is just a vector of methods of `func`; hasn't changed much externally.

Yeah, I mean the fields of `Method` and such, i.e. actually getting useful info (e.g. line no) out of it programmatically. The fields have changed quite a bit between 0.4 and 0.5.

mortenpi:
> For the moment we can just have it work on 0.5, and add compat fixes later on.

Well, there's more than just compat. Getting argument names is slightly non-trivial (needs `Base.arg_decl_params`). Also, possibly, URL or link construction. I feel in the rendering part all the necessary information should be accessible as simply and obviously as possible.

mortenpi:
> > For the moment we can just have it work on 0.5, and add compat fixes later on.

> Well, there's more than just compat. Getting argument names is slightly non-trivial (needs `Base.arg_decl_params`). Also, possibly, URL or link construction. I feel in the rendering part all the necessary information should be accessible as simply and obviously as possible.

Alternatively, might make more sense to have a bunch of methods available, something along the lines of `lineno(::Method)`, to easily access all this information. Although, then I think you have to pass multiple arguments to some.. so maybe a mix of the two somehow.

MichaelHatherly:
We need:

- `module` via `object.binding.mod`
- `name` via `object.binding.var`
- `tvars` via `Base.arg_decl_parts`
- `args` via `Base.arg_decl_parts`
- `file` via `Base.arg_decl_parts`
- `line` via `Base.arg_decl_parts`

Any other data that we need? That seems straightforward, am I missing any others?

> Alternatively, might make more sense to have a bunch of methods available

That's fine with me if necessary.

src/modules/Expanders.jl
```diff
@@ -262,16 +262,59 @@ end
...
+
+        # We only generate the method table for a Function and a Type.
+        # The only other one with methods should be the Method itself (e.g.
+        # defined as `foo(x,y)` in the @docs block), in that case the user has
+        # already picked a method, so we shouldn't generate a single element
+        # table.
+        doccat = Utilities.doccat(object)
+        if doccat == "Function" || doccat == "Type"
+            ms = methods(getfield(object.binding.mod, object.binding.var))
+            return Nullable([m for m in ms])
```
MichaelHatherly: Would `Nullable(collect(ms))` work here instead?

src/modules/Writers.jl
```diff
@@ -59,13 +59,91 @@ function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNodes, page,
...
     println(io, anchor, "\n", header, "\n\n")
     # Body. May contain several concatenated docstrings.
     renderdoc(io, mime, node.docstr, page, doc)
+
+    # Table of methods.
+    # If DocsNode.methods is nulled, then we assume that we should not render a
+    # table. However, if the list of methods is there but has no elements then
+    # we output an appropriate note saying that the name has no methods associated
+    # with it.
+    Utilities.unwrap(node.methods) do ms
```
MichaelHatherly: Might be worth factoring this out since it's quite big. Then just calling something like

```julia
Utilities.unwrap(ms -> rendermethods(io, node.object, ms), node.methods)
```

if that's possible. Would also help to avoid vars leaking into the closure unintentionally.

src/modules/Writers.jl
```diff
@@ -59,13 +59,91 @@ function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNodes, page,
...
+        println(io, "<p><strong>Methods</strong></p>")
+
+        if isempty(ms)
+            println(io, "<p><emph>This name has no methods.</emph></p>")
```
MichaelHatherly: Should `name` be `function` / `type` instead?

src/modules/Writers.jl
```diff
@@ -59,13 +59,91 @@ function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNodes, page,
...
+                </li>
+                """)
+            else
+                args = join([(" "^4)*join_decl(d) for d in decls], ",\n")
+                print(io, """
+                <li>
+                    <pre><code><span class="n f">$(name)$(tvars)</span>(
+                $(args)
+                )</code></pre>
+                    defined at
```
MichaelHatherly: Should we just follow the wording used in the REPL and print `at` rather than `defined at`?

MichaelHatherly:
The updates are coming along great I think! :100:

A couple of notes after building the docs locally with this branch:

- I'm thinking we should have this on by default, and no option for disabling it for the moment. If we get sufficient interest in adding a `ShowMethodList = false` (or some other name) option then we can add one later.
- `abstract` types shouldn't list their constructors (only the `(::Type{T})` is shown anyway) since they can't be constructed.
- Maybe use the word `Constructors` for when it's a `DataType` and `Methods` when it's a `Function`?
- Filtering by module using `doc.user.modules` will be needed otherwise adding a method for `length` to a module will display 66 methods, which I think wouldn't be what's expected.
- Is it possible for the multiline signatures not to have a grey background? To match the single line ones.
- Perhaps use an unordered list, since if there's just one method it looks a little out-of-place.

(Those are just small nitpicks/suggestions though, probably close to being complete I'd think.)

src/modules/Writers.jl
mortenpi: I was thinking that the same code can most likely be used for HTML rendering as well, and so should be factored out, but thought to do it when we come to that.

src/modules/Writers.jl
mortenpi: Consistency would be nice, but I actually like the space the word "defined" puts between the signature and the link. Just "at" seems crowded to me, especially if it is small/zoomed out.

outs

src/modules/Writers.jl
mortenpi: This is a good question about terminology actually, also ties into "constructor" vs "method" discussion a little bit.

My own impression based on what is see in the julia documentation is that you can refer to every item in a module collectively as a "name" (from [`methods()`](http://docs.julialang.org/en/latest/stdlib/base/?highlight=names#Base.names)). Also, [constructors are technically functions](http://docs.julialang.org/en/latest/manual/constructors/), which can have several methods associated with them. So, does the type have methods or the function associated with the type? Or are methods associated with the names?

Regarding constructor vs method, again the docs refer to different constructors as methods. And you could (although probably shouldn't) write a "constructor" that returns an object of a different type, so I would stick to calling them methods.

Basically, as a philosophy, I would stick with the terms that are technically correct, rather than making assumptions about the user's intentions (e.g. not assume that all type methods are constructors), even if that goes against widespread practices. But there's still the matter of figuring out what is technically correct.

mortenpi:
> - I'm thinking we should have this on by default, and no option for disabling it for the moment. If we get sufficient interest in adding a `ShowMethodList = false` (or some other name) option then we can add one later.

What about a global option though?

> - `abstract` types shouldn't list their constructors (only the `(::Type{T})` is shown anyway) since they can't be constructed.

Darn, I'm torn. Something like
```
abstract Foo
type Bar <: Foo end
Foo(Bar())
```
is actually valid code and works, so the function does exist. But, of course, it doesn't do anything useful, so it's a fight between accuracy and trying to avoid noise. You can define methods for abstract types as well though, so I don't think we should completely throw away methods for abstract types.

There could be a compromise though -- with styles we could maybe grey the default/automatic methods out to make it clear that they're not very interesting.

> - Filtering by module using `doc.user.modules` will be needed otherwise adding a method for `length` to a module will display 66 methods, which I think wouldn't be what's expected.

Yup, something needs to be done for this. Also, as you pointed out before, this might solve the default constructor issue.

> - Is it possible for the multiline signatures not to have a grey background? To match the single line ones.

Do you mean the slightly different shade of gray? Or the big box? I would keep the box, because it is kind-of a code block actually.

But I have to think about style a bit more actually. Also, I have to see what happens if you build with a different template. Currently I am using a simple `<code>` or `<pre><code>` block without any customization (except margins to get lists working in material).

> - Perhaps use an unordered list, since if there's just one method it looks a little out-of-place.

Yeah, I implicitly assumed that order probably matters, but actually I guess that's not the case.. ambiguities seem to be explicit errors, not determined by the order in which things were defined (or something).

src/modules/Writers.jl
MichaelHatherly: Fair enough.

src/modules/Writers.jl
MichaelHatherly: Since we've got access here to whether it's a function or a type we should include the relevant one. Though it will only ever be empty for a `Function` object defined with

```julia
function func end
```

so probably just put `This function has no methods.` and call it done I think.

Regarding constructor/method, lets just leave as you've written it. If anyone comes along to complain about it we can deal with it then.

src/modules/Writers.jl
MichaelHatherly: Fair enough, not like we're limited by a character count here I guess.

MichaelHatherly:
> What about a global option though?

Do you mean a setting in `makedocs` or one per-page in a `@meta` block?

> actually valid code and works

Yes, and it returns a `Bar` object from the `::Type{T}` constructor:

```
julia> @which Foo(Bar())
(::Type{T}){T}(arg) at sysimg.jl:50
```

Let's sort out the module filtering first and then come back to this discussion if it's still relevant, though I doubt it will be then :)

> fight between accuracy and trying to avoid noise

My view is that avoiding noise is what good documentation should be *all* about. There's always the source itself if one needs to see exactly what's happening rather than getting the broad idea of it.

> Do you mean the slightly different shade of gray? Or the big box? I would keep the box, because it is kind-of a code block actually.

The grey behind the small signatures wasn't showing very clearly for me (glare on my screen at the time), but I can see now they both have. So that can definitely just stay as is.

> build with a different template.

Yes, would be good to build with a few other templates to be sure.

> not determined by the order

Yeah, ordering of definitions shouldn't have an effect on dispatch as far as I'm aware, so best to avoid implying some kind of order here.

mortenpi:
> > What about a global option though?
>
> Do you mean a setting in `makedocs` or one per-page in a `@meta` block?

I meant `makedocs`, just to give the users the opportunity to opt out. Although.. we could again maybe just wait until someone asks for it.

MichaelHatherly:
> maybe just wait until someone asks for it.

I'd prefer that, otherwise `makedocs` will end up accumulating a lot of options eventually &mdash; potentially unnecessarily.

src/modules/Writers.jl
```diff
@@ -59,13 +59,91 @@ function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNodes, page,
...
+                    <a href="$(url)">$(file):$(line)</a>
```
MichaelHatherly: I find it useful to have these kind of URLs open in separate tabs, rather than the same one. Just needs a `target='_blank'` like `source` links for docstrings do.

MichaelHatherly:
Worth while getting this into an initial `0.1` release I think. No rush though, I'm going to going over docs for the next few days to see that everything is in order before tagging and registering.

mortenpi:
As sort of a compromise I went with filtering by `CurrentModule` and then adding a small note that some methods are being omitted (if that indeed is the case):
out

In the future HTML output I would like to have those methods present as well though (in the one-per-page function docs). I imagine having different blocks, with sections like "Methods defined in `Base`" etc., which should probably be collapsed initially. But in the Markdown output I now actually think it makes sense to omit to indeed have less noise.

MichaelHatherly:
> 66 methods are not being displayed.

I'd prefer the wording as `66 methods from other packages not displayed.` or something similar if we got with this approach, since it provides a reason for why things have been left out.

For the `issubmodule` method the following should be much quicker and not allocate:

```julia
is_submodule(sub, mod) =
    sub === mod ? true :
    sub === Main ? false :
    is_submodule(module_parent(sub), mod)
```

which I believe covers all the cases...

src/modules/Writers.jl
```diff
@@ -142,7 +144,15 @@ function render(io::IO, mime::MIME"text/plain", node::Expanders.DocsNode, page,
...
-        print(io, "</ul>")
+        print(io, "</ul>\n\n")
+
+        # we print a small notice if we are not displaying all the methods
+        nhidden = length(methodnodes)-length(ms)
+        if nhidden == 1
+            print(io, "_1 method is not being displayed._\n\n")
+        elseif nhidden > 1
+            print(io, "_$(nhidden) methods are not being displayed._\n\n")
+        end
```
MichaelHatherly: Can be

```julia
println(io, "$n method$(n == 1 ? "" : "s") not being displayed.\n")
```

to avoid the duplicate text. That's what I've done elsewhere in the package, though up to you if you prefer to keep the current style.

src/modules/Expanders.jl
```diff
@@ -261,6 +261,11 @@ end
 # @docs
 # -----
 
+immutable MethodsNode
```
MichaelHatherly: Should this be `MethodNode` rather, since it's only storing a single method?

MichaelHatherly:
Regarding use of `CurrentModule` for filtering:

- `doc.user.modules` better matches the use case here, since it's already being used to filter concatenated docstrings. `CurrentModule` is used to evaluate expressions in a particular module.
- Since Documenter does not allow for duplicate `Object`s within a `Document` if `CurrentModule` is set to some submodule of the package root module then it's possible that methods may be filtered out of the final list that are part of the module, but not the current subtree. As we can't then add the same `Object` somewhere else in the documentation with a different `CurrentModule` those methods that are filtered out can't ever be added to the document.

So that leads me to think using `doc.user.modules` is the way this should be filtered for now.

src/modules/Writers.jl
mortenpi: :+1: in general, but it did have is/are as well and then it would become a bit hard to read (not an issue after I rephrased it). I would drop the whitespace though, to make it more compact.

src/modules/Expanders.jl
```diff
@@ -261,17 +261,75 @@ end
...
+        # with `foo(x,y)` in the @docs block), but in that case the user has
+        # already picked a method and we shouldn't generate a single element
+        # table.
+        doccat = Utilities.doccat(object)
+        if doccat == "Function" || doccat == "Type" || doccat == "Macro"
+            ms = map(methods(getfield(object.binding.mod, object.binding.var))) do m
+                # We filter out the methods not defined in this package.
+                # That is we show the method if it is defined in any of the
+                # `doc.user.modules` modules defined in `makedocs`.
+                visible = any(map(mod->Utilities.issubmodule(m.module, mod), doc.user.modules))
```
MichaelHatherly: Since `doc.user.modules` is a `Set{Module}` can this just be

```julia
visible  = m.module in doc.user.modules
```

since we store all the submodules there already.

MichaelHatherly:
Can you rebase rather than doing merge commits. And also squash out WIP intermediate commits once this is ready to go. Thanks.

mortenpi:
So, this is the look now (not that many visible differences actually):
out-a
out-b
I am still a bit annoyed by the space the source link leaves.. but with markdown there isn't much we can do I think. We could play with styles, but that wouldn't be very friendly to the different styles (probably would work well only with one style or something).

mortenpi:
> `abstract` types shouldn't have anything printed I think.

I'm fine with that. However, should we have an "Abstract Type" `doccat` in princple?

Beyond that, @MichaelHatherly if it looks good to you know then I think it's time to merge now. Once you green light the last changes I'll squash it and it will be ready to go.

MichaelHatherly:
Yes, squash and merge now, looks good to me. We can add an `Abstract Type` in another PR, since it's not particularly related to the methodlists.

Issue #44, opened 2016-04-29
Issues with multiple named `{repl}`


lbenet:
Thanks (again) for implementing https://github.com/MichaelHatherly/Documenter.jl/commit/71685248e1066e2e0489a85e00cde922eeb35556. I really like its functionality. Here, I'd like to point out few issues I have encountered while using it:

- It seems to me that `{repl}` is not redirecting correctly `STDOUT` and `STDERR`. Actually, when an event related to `STDERR` is encountered, the whole `{repl}` block is simply parsed as a code block. Warnings and info messages are not displayed (but do not cause a code block as the error).

- The trick (see your [comment](https://github.com/MichaelHatherly/Documenter.jl/issues/37#issuecomment-208989100)) of using `nothing # hide` to suppress output (`@time do_somethin()`) does not working in `{repl}`.

- I'm getting a weird error (on Julia v0.4) which leads to hanging in "Documenter: skipping docs deployment" (in Julia 0.5 it simply hangs there). The context is two files which use named `{repl}` blocks which are independent. Taking away either file from the `src/` directoryyields no error, but merging the files into one does.) It is worth saying that the files created in `build/` appear correctly. The error is
```
ERROR (unhandled task failure): readcb: bad file descriptor (EBADF)
 in yieldto at /usr/local/julia/julia-d829429d34/lib/julia/sys.dylib
 in wait at /usr/local/julia/julia-d829429d34/lib/julia/sys.dylib (repeats 2 times)
 in wait_readnb at ./stream.jl:374
 in eof at ./stream.jl:96
 in watchbuffer at /Users/benet/.julia/v0.4/Documenter/src/modules/DocChecks.jl:343
 in anonymous at ./task.jl:447
```

MichaelHatherly:
> It seems to me that `{repl}` is not redirecting correctly `STDOUT` and `STDERR`.

Currently, thrown errors will abort execution of a `{repl}` block and just splice in the original code block. So this one is expected at the moment, but can be changed relatively easily to rather print the error and continue. Will fix this one.

> Warnings and info messages are not displayed

Does it not output something like the following:

```
...
Documenter: expanding markdown templates.
 !! failed to run code block.

DivideError() [src/tests.md]
...
```

> using `nothing # hide` to suppress output

Should just support using `;` for this. Can also be fixed quite easily.

> The context is two files which use named `{repl}` blocks which are independent.

[...]

Issue #45, opened 2016-05-05
Travis CI build non responsive for 10m


Ward9250:
Hi,

On Bio.jl travis builds for julia v0.4 on OSX we've started to experience build failures even though all tests pass, the log indicating something freezes or becomes non-responsive after docs deployment. In the travis log it will say "Skipping docs deployment" for a merge into the master branch, and then will be non responsive for 10m. So I'm just investigating to see if this is Documenter.jl related.

This is the current log for Bio.jl master: https://travis-ci.org/BioJulia/Bio.jl/jobs/128102434

Many thanks,
Ben.

MichaelHatherly:
Sorry for not being very responsive here. I'll try take a look at this in the next few days.

Is this just happening on travis OSX? Or have you been able to reproduce locally too?

bicycle1885:
This is not reproducible on my OS X. So, I'm not sure whether this is due to a bug of Documenter.jl.

MichaelHatherly:
The OSX travis build for Documenter itself completes fine, so might have something to do with how long tests are taking for Bio.jl. Might be worth seeing what happens if doc gen is moved prior to other parts of the build such as code coverage.

There's also another possible OSX issue, https://github.com/MichaelHatherly/Documenter.jl/issues/44 (`bad file descriptor`), that *might* be related.

Ward9250:
I can confirm as @bicycle1885 says, my OSX does not do this either. I can try and swap the order and see what happens.

Ward9250:
Hi @MichaelHatherly,

I swapped the order of Documenter with Coveralls steps as you suggested (So the order was Tests -> Documenter -> Coveralls). The build only got as far as "Skipping Docs Deployment", and did not go on to the coveralls step. Compared to before (Tests -> Coveralls -> Documenter), which would complete the coveralls step, and the Documenter step, and then hang. The build is here: https://travis-ci.org/BioJulia/Bio.jl/jobs/129402418.  

Best,
@Ward9250

MichaelHatherly:
Thanks for checking. Seems like it is more than likely a Doucmenter issue then. I'll see if I can replicate it on travis myself.

MichaelHatherly:
Turning off doctesting with `makedocs(doctest = false)` allowed the build to finish, so this is most probably a problem with the output redirection code in Documenter. (Until I've resolved it you could conditionally disable doctests on the OSX 0.4 build to allow it to finish properly.)

[...]

PR #93, opened 2016-06-10
Fixup redirected output

MichaelHatherly:
Don't create new tasks for each redirect and only redirect the `STDOUT` and `STDERR` streams once during `Document` construction.

Might help solve the OSX issues related to hanging and too many open files.

@lbenet, if you have the chance to test this out that would be great, might potentially help with #44.

### DIFF
--- a/src/modules/Builder.jl
+++ b/src/modules/Builder.jl
@@ -138,6 +138,7 @@ function Selectors.runner(::Type{CheckDocument}, doc::Documents.Document)
 end
 
 function Selectors.runner(::Type{RenderDocument}, doc::Documents.Document)
+    Utilities.redirect_stream(doc.internal.stream)
     Utilities.log("rendering document.")
     Documenter.Writers.render(doc)
 end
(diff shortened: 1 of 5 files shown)


MichaelHatherly:
Success perhaps?

- Without this branch: https://travis-ci.org/MichaelHatherly/TaylorSeries.jl/jobs/136787308
- With this branch: https://travis-ci.org/MichaelHatherly/TaylorSeries.jl/jobs/136788666

Was the problem intermittent or happening every time without fail?

MichaelHatherly:
I'll rebuild that job a couple of times to check for sure.

MichaelHatherly:
Issue #45 for `Bio.jl` also appears to be fixed by this:

- Without this branch: https://travis-ci.org/MichaelHatherly/Bio.jl/jobs/136891138
- With this branch: https://travis-ci.org/MichaelHatherly/Bio.jl/jobs/136893501

cc: @Ward9250, @bicycle1885 if you get a chance to test this out locally as well that would be great.

lbenet:
I just checked it in my mac, and it the problem is gone, including julia 0.4 ! Thank you very much!

MichaelHatherly:
Excellent! Thanks for checking @lbenet.

PR #96, opened 2016-06-14
Cross-version compatible docsystem interface

MichaelHatherly:
Adds a `DocSystem` module that provides a programmatic interface to the Julia docsystem rather than having to rely on `at-doc` and concatenated `MD` objects.

Main interface is `DocSystem.getdocs` which takes a `Binding` and optional type signature. Keyword arguments for the signature comparison and which modules to search are also provided. `getdocs` returns a `Vector{DocStr}` containing matching docstrings.

The differences in `0.4`'s internal representation is handled automatically and shouldn't be visible to the caller. Some metadata is only available in `0.5` since it wasn't captured in `0.4`.

### DIFF
--- a/src/modules/DocSystem.jl
+++ b/src/modules/DocSystem.jl
@@ -1,0 +1,318 @@
+"""
+Provides a consistent interface to retreiving `DocStr` objects from the Julia
+docsystem in both `0.4` and `0.5`.
+"""
+module DocSystem
+
+using Compat
+
+## Bindings ##
+
+"""
+Converts an object to a `Base.Docs.Binding` object.
+
+    binding(object) :: Binding
+
+Supported inputs are:
+
+- `Binding`
+- `DataType`
+- `Function`
+- `Module`
+- `Symbol`
+
+Note that unsupported objects will throw an `ArgumentError`.
+"""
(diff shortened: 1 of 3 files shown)


src/modules/DocSystem.jl
```diff
@@ -0,0 +1,262 @@
...
+        CACHED[meta] = docs
+    end
+    CACHED[meta]::ObjectIdDict
+end
+const CACHED = ObjectIdDict()
+
+#
+# Is the documented object `x` valid in the docsystem in `0.5`?
+#
+validdoc(x) = !isa(x, Union{Number, AbstractString, ObjectIdDict})
```
mortenpi: Any reason not to have this as a proper docstring (and some other similar ones below)? If there's already some documentation present anyway maybe have it be available through the docsystem as well, even if the docstring is not super-thorough?

MichaelHatherly: I'll probably just inline the `validdoc` definition into the method above, since it's not useful elsewhere. So won't need a docstring after that. Re: the other comment blocks, probably the second `getdocs` can be a docstring I think.

(If you've got any requests for functionality that would be useful to have in this module do let me know.)

MichaelHatherly:
I'll merge this shortly as is. It's not currently being used as a `@doc` replacement, but is tested. We can switch over later on.

PR #102, opened 2016-06-18
Update Currencies.jl link

TotalVerb:
Currencies.jl has moved to JuliaFinance.


MichaelHatherly:
Thanks for the update.

PR #103, opened 2016-06-18
Add a simple HTML DSL

MichaelHatherly:
A different approach than that taken by Hiccup.jl.

All nodes are the same type (`Node`) and no symbolic `tag` type
parameter. Each node's children are all also `Node`s and attributes are
a `Vector` rather than a `Dict` with concrete type parameters (`Symbol`
and `String`).

Also has a bit of syntax sugar:

```julia
div[".class#id"](
    ul(map(li, split("foo bar baz"))),
    img[:src => "path/to/file.jpg"]
)
```

No tags are exported at all. ~~They can just be defined when needed using~~

```julia
const div, ul, li, img = map(Tag, (:div, :ul, :li, :img))
```

~~We can add a macro to do that as well if we do use this.~~

Now with a `@tags` macro so the above is just

```julia
@tags div ul li img
```

### DIFF
--- a/docs/src/lib/internals.md
+++ b/docs/src/lib/internals.md
@@ -165,3 +165,15 @@ Utilities.doccat
 Utilities.nodocs
 Utilities.issubmodule
 ```
+
+### DOM
+
+```@docs
+Utilities.DOM
+Utilities.DOM.@tags
+Utilities.DOM.Tag
+Utilities.DOM.Node
+Utilities.DOM.escapehtml
+Utilities.DOM.flatten!
+```
+
--- a/src/Utilities/DOM.jl
+++ b/src/Utilities/DOM.jl
@@ -1,0 +1,302 @@
+"""
+Provides a domain specific language for representing HTML documents.
+
+# Examples
+
+```julia
+using Documenter.Utilities.DOM
+
+# `DOM` does not export any HTML tags. Define the ones we actually need.
+@tags div p em strong ul li
+
+div(
+    p("This ", em("is"), " a ", strong("paragraph."),
+    p("And this is ", strong("another"), " one"),
+    ul(
+        li("and"),
+        li("an"),
+        li("unordered"),
+        li("list")
+    )
+)
+```
+
+*Notes*
+
+All the arguments passed to a node are flattened into a single vector rather
+than preserving any nested structure. This means that passing two vectors of
+nodes to a `div` will result in a `div` node with a single vector of children
+(the concatenation of the two vectors) rather than two vector children. The
+only arguments that are not flattened are nested nodes.
+
+String arguments are automatically converted into text nodes. Text nodes do not
+have any children or attributes and when displayed the string is escaped using
+[`escapehtml`](@ref).
+
+# Attributes
+
+As well as plain nodes shown in the previous example, nodes can have attributes
+added to them using the following syntax.
+
+```julia
+div[".my-class"](
+    img[:src => "foo.jpg"],
+    input[\"#my-id\", :disabled]
+)
+```
+
+In the above example we add a `class = "my-class"` attribute to the `div` node,
+a `src = "foo.jpg"` to the `img`, and `id = "my-id" disabled` attributes to the
+`input` node.
+
+The following syntax is supported within `[...]`:
+
+```julia
+tag[\"#id\"]
+tag[".class"]
+tag[\".class#id\"]
+tag[:disabled]
+tag[:src => "foo.jpg"]
+# ... or any combination of the above arguments.
+```
+
+# Internal Representation
+
+The [`@tags`](@ref) macro defines named [`Tag`](@ref) objects as follows
+
+```julia
+@tags div p em strong
+```
+
+expands to
+
+```julia
+const div, p, em, strong = Tag(:div), Tag(:p), Tag(:em), Tag(:strong)
+```
+
+These [`Tag`](@ref) objects are lightweight representations of empty HTML
+elements without any attributes and cannot be used to represent a complete
+document. To create an actual tree of HTML elements that can be rendered we
+need to add some attributes and/or child elements using `getindex` or `call`
+syntax. Applying either to a [`Tag`](@ref) object will construct a new
+[`Node`](@ref) object.
+
+```julia
+tag(...)      # No attributes.
+tag[...]      # No children.
+tag[...](...) # Has both attributes and children.
+```
+
+All three of the above syntaxes return a new [`Node`](@ref) object. Printing of
+`Node` objects is defined using the standard Julia display functions, so only
+needs a call to `print` to print out a valid HTML document with all nessesary
+text escaped.
+"""
+module DOM
+
+using Compat
+
+if VERSION < v"0.5.0-dev"
+    typealias String UTF8String
+    tostr(p::Pair) = p[1] => utf8(p[2])
+else
+    tostr(p::Pair) = p
+end
+
+export @tags
+
+#
+# The following sets are based on:
+#
+# - https://developer.mozilla.org/en/docs/Web/HTML/Block-level_elements
+# - https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements
+# - https://developer.mozilla.org/en-US/docs/Glossary/empty_element
+#
+const BLOCK_ELEMENTS = Set([
+    :address, :article, :aside, :blockquote, :canvas, :dd, :div, :dl,
+    :fieldset, :figcaption, :figure, :footer, :form, :h1, :h2, :h3, :h4, :h5,
+    :h6, :header, :hgroup, :hr, :li, :main, :nav, :noscript, :ol, :output, :p,
+    :pre, :section, :table, :tfoot, :ul, :video,
+])
+const INLINE_ELEMENTS = Set([
+    :a, :abbr, :acronym, :b, :bdo, :big, :br, :button, :cite, :code, :dfn, :em,
+    :i, :img, :input, :kbd, :label, :map, :object, :q, :samp, :script, :select,
+    :small, :span, :strong, :sub, :sup, :textarea, :time, :tt, :var,
+])
+const VOID_ELEMENTS = Set([
+    :area, :base, :br, :col, :command, :embed, :hr, :img, :input, :keygen,
+    :link, :meta, :param, :source, :track, :wbr,
+])
+const ALL_ELEMENTS = union(BLOCK_ELEMENTS, INLINE_ELEMENTS, VOID_ELEMENTS)
+
+#
+# Empty string as a constant to make equality checks slightly cheaper.
+#
+const EMPTY_STRING = ""
+const TEXT = Symbol(EMPTY_STRING)
+
+"""
+Represents a empty and attribute-less HTML element.
+
+Use [`@tags`](@ref) to define instances of this type rather than manually
+creating them via `Tag(:tagname)`.
+"""
+immutable Tag
+    name :: Symbol
+end
+
+Base.show(io::IO, t::Tag) = print(io, "<", t.name, ">")
+
+"""
+Define a collection of [`Tag`](@ref) objects and bind them to constants
+with the same names.
+
+# Examples
+
+Defined globally within a module:
+
+```julia
+@tags div ul li
+```
+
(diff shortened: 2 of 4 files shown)


MichaelHatherly:
@mortenpi, this started out as just some tinkering around this afternoon, so we *can* still just go with Hiccup, though I have started to kind of like this approach. Figured I shouldn't just throw it away. :laughing:

MichaelHatherly:
Benchmarking against Hiccup suggests that this is a bit faster in all the cases that I've tried:

```
julia> @benchmark div[".foo#main"](img[".image-main", :src => "foo.png"])
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     9
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  3.25 kb
  allocs estimate:  65
  minimum time:     2.95 μs (0.00% GC)
  median time:      3.04 μs (0.00% GC)
  mean time:        3.57 μs (12.10% GC)
  maximum time:     376.84 μs (97.77% GC)

julia> @benchmark Hiccup.div(".foo#main", Hiccup.img(".image-main", Dict(:src => "foo.png")))
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  4.27 kb
  allocs estimate:  70
  minimum time:     9.02 μs (0.00% GC)
  median time:      10.76 μs (0.00% GC)
  mean time:        11.36 μs (2.56% GC)
  maximum time:     2.97 ms (97.84% GC)

julia> f(b, s) =  print(b, Hiccup.div(Hiccup.ul(map(Hiccup.li, s))));

julia> g(b, s) = print(b, div(ul(map(li, s))));

julia> s = [randstring(100) for n = 1:1000];

julia> @benchmark g(IOBuffer(), s)
BenchmarkTools.Trial: 
  samples:          5737
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  610.14 kb
  allocs estimate:  7037
  minimum time:     783.25 μs (0.00% GC)
  median time:      805.14 μs (0.00% GC)
  mean time:        871.13 μs (5.58% GC)
  maximum time:     3.13 ms (67.14% GC)

julia> @benchmark f(IOBuffer(), s)
BenchmarkTools.Trial: 
  samples:          573
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  3.24 mb
  allocs estimate:  39057
  minimum time:     7.97 ms (0.00% GC)
  median time:      8.24 ms (0.00% GC)
  mean time:        8.73 ms (4.32% GC)
  maximum time:     12.58 ms (22.05% GC)
```

I'll probably add some docs for this in the next few days, and merge. Seems to be a win regarding performance as far as I can tell.

src/Utilities/Tags.jl
```diff
@@ -0,0 +1,120 @@
+module Tags
```
mortenpi: Could be called `HTML` perhaps, so it would more clear what it is.

src/Utilities/Tags.jl
```diff
@@ -0,0 +1,120 @@
...
+                for each in n.data
+                    show(io, each)
+                end
+            end
+            print(io, "</", n.name, '>')
+        end
+    end
+end
+
+isvoid(n::Node) = n.name in VOID_ELEMENTS
```
mortenpi: Very minor, but perhaps have them defined before the previous `show`, since they're used in there.

src/Utilities/Tags.jl
```diff
@@ -0,0 +1,120 @@
...
+immutable Tag
+    name :: Symbol
+end
+
+Base.show(io::IO, t::Tag) = print(io, "<", t.name, ">")
+
+macro tags(args...) esc(tags(args)) end
+tags(s) = :(const ($(s...),) = $(map(Tag, s)))
+
+typealias Attributes Vector{Pair{Symbol, String}}
```
mortenpi: Why `Vector` not a simple `Dict`? Performance?

src/Utilities/Tags.jl
```diff
@@ -0,0 +1,120 @@
...
+
+immutable Node
+    name :: Symbol
+    text :: String
+    attr :: Attributes
+    data :: Vector{Node}
```
mortenpi: I think `.elements` would the correct term.

src/Utilities/Tags.jl
```diff
@@ -0,0 +1,120 @@
...
+
+    Node(name::Symbol, attr::Attributes, data::Vector{Node}) = new(name, EMPTY_STRING, attr, data)
+    Node(name::Symbol, text::AbstractString) = (@assert(name === TEXT); new(name, text))
```
mortenpi: An option here would also be `Node(text::AbstractString) = new(TEXT,text)` to easily create text nodes. It shouldn't clash, but perhaps it is better to be explicit.

mortenpi:
I am not entirely sold on using the array access syntax for constructing the elements. For me it would feel a bit more natural to have them work as constructors (`div(".class#id", ...)`), which is what `Hiccup` does.

Regarding performance -- is `Hiccup` somehow fundamentally slower or could perhaps the code just be improved?

Final thought -- rolling our own thing. This would be (afaict) a third way of doing DOM/HTML Julia (besides Hiccup there's also [Patchwork](https://github.com/shashi/Patchwork.jl)). It's of course quite simple, so maintaining it wouldn't actually be a problem, but I feel that there should be a compelling argument to ditch the existing things.

I'll think a bit more on this.

src/Utilities/Tags.jl
MichaelHatherly: `DOM` sounds good to me.

src/Utilities/Tags.jl
MichaelHatherly: I typically put helper functions after the main function. Though in this case they're only ever used in the show method, so maybe just fold them into that.

src/Utilities/Tags.jl
MichaelHatherly: The attributes are never going to contain a terribly large number of elements, and we don't need to have O(1) element retrieval. It just needs to store and then print out the contents in order. So a `Dict` would be wasteful here.

src/Utilities/Tags.jl
MichaelHatherly: Yes, I'll change those to `attributes` and `nodes`. (Just had to have them line up nicely :laughing:)

src/Utilities/Tags.jl
MichaelHatherly: Yes, that's better.

MichaelHatherly:
> I am not entirely sold on using the array access syntax for constructing the elements.

I wasn't at first either, but when writing a large document it makes it quite clear where the attributes are, since the brackets kind of jump out at you. Just having the selector string as the first argument to a `Node` constructor runs into the edge case where you have something like `p(".class", ...)`. Is the `".class"` a fragment of text or an attribute? I prefer to keep them separated by some syntax. It's also just an internal DSL, so I don't feel to bad about appropriating some syntax here :smile:

 > Regarding performance -- is `Hiccup` somehow fundamentally slower or could perhaps the code just be improved?

Non-concrete fields and parameter types. Parametric types for differentiating tags. Using a `Dict` for attributes. Changing those would probably break peoples' code. Patchwork seems to have similar internals as far as I can tell and writes out a JSON diff rather than plain HTML I think. (Both have quite large dependency trees too.) Benchmarking Patchwork DOM creation gives the following for a simple case:

```
julia> @benchmark div[".foo#main"](img[".image-main", :src => "foo.png"])
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     8
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  3.25 kb
  allocs estimate:  65
  minimum time:     3.06 μs (0.00% GC)
  median time:      3.19 μs (0.00% GC)
  mean time:        4.00 μs (16.34% GC)
  maximum time:     648.74 μs (98.52% GC)

julia> @benchmark Patchwork.Elem(:div, Patchwork.Elem(:img, class = "image-main", src =  "foo.png"),  id = "main", class = "foo")
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  46.86 kb
  allocs estimate:  325
  minimum time:     444.05 μs (0.00% GC)
  median time:      450.28 μs (0.00% GC)
  mean time:        458.67 μs (1.06% GC)
  maximum time:     3.53 ms (80.58% GC)
```

> third way of doing DOM/HTML

It won't be a package, so shouldn't lead to any confusion in the package ecosystem I'd imagine and, as you say, it's pretty simple so won't be a burden really.

MichaelHatherly:
TODO: `lib/internals.md` is becoming much to large and needs some subdividing at some point.

src/Utilities/Tags.jl
mortenpi: Ah, I would have them before normally, so that if you read linearly you would know the definitions of everything. But I don't know which is better and consistency is the key, so if it is after, it is after.

src/Utilities/Tags.jl
mortenpi: I am not sure, but would it matter that much? I.e. I would go with the code-wise simpler/more obvious option unless the performance is critical, and I feel that a `Dict` would the more idiomatic one here. But I'm ok with both.

src/Utilities/Tags.jl
mortenpi: One more: should there also be a `convert(::Type{Node}, s::String) = Node(s)` ?

mortenpi:
Aight, I'm sold now. Rolling our own also means that we have more freedom and do not have to worry about upstream etc., which will probably be good. And yes, separating is good, so for our own DSL I'm now completely OK with that syntax.

In other words: :100:.

src/Utilities/Tags.jl
MichaelHatherly: I'd say that a `Vector` is simpler that a `Dict`. Changing between `Dict` and `Vector` only requires the following change, everything else is generic enough to handle either:

```julia
# typealias Attributes Vector{Pair{Symbol, String}}
typealias Attributes Dict{Symbol, String}
```

A `Dict` has three different `Vector` fields, so it's definitely going to use a bit more space. In this case we don't need to do any attribute access at all, so using a `Dict` really wouldn't be worthwhile. We just need to append results and then print them out in order later.

Switching to a `Dict` gives me the following results:

```
julia> @benchmark div(ul(map(li, s))) # Vector
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  337.30 kb
  allocs estimate:  6020
  minimum time:     140.30 μs (0.00% GC)
  median time:      148.29 μs (0.00% GC)
  mean time:        181.16 μs (16.02% GC)
  maximum time:     2.49 ms (90.70% GC)

julia> @benchmark div(ul(map(li, s))) # Dict
BenchmarkTools.Trial: 
  samples:          7634
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  854.00 kb
  allocs estimate:  9027
  minimum time:     552.36 μs (0.00% GC)
  median time:      568.65 μs (0.00% GC)
  mean time:        654.44 μs (11.50% GC)
  maximum time:     4.84 ms (80.05% GC)
```

Not a *huge* difference, and it does come down slightly once you add attributes. That might have something to do with array resizing and copying though.

(Also, using a `Dict` could give us different orders of attributes in the final rendering on each call  to `julia make.jl` which would be pretty annoying if you want to check whether some code changes cause any subtle changes in the output via `git diff`.)

src/Utilities/Tags.jl
MichaelHatherly: I'm not too bothered which way they should be written (I've simply removed them in this particular case), but perhaps we should put together a short style guide for the package with these kind of things. There's also some other places that could do with some consistency fixes as well.

src/Utilities/Tags.jl
MichaelHatherly: Probably not needed I'd imagine, though we can definitely add one if the need arises.

MichaelHatherly:
Great, thanks for looking over this. Much appreciated.

PR #114, opened 2016-06-22
Add converter from Markdown to Node objects

MichaelHatherly:
Should be useful for generating HTML docs.

### DIFF
--- a/src/Utilities/DOM.jl
+++ b/src/Utilities/DOM.jl
@@ -269,6 +269,8 @@ function Base.show(io::IO, n::Node)
     end
 end
 
+@compat Base.show(io::IO, ::MIME"text/html", n::Node) = print(io, n)
+
 """
 Escape characters in the provided string. This converts the following characters:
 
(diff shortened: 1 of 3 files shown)


src/Utilities/DOM.jl
```diff
@@ -298,5 +300,81 @@ function escapehtml(text::AbstractString)
...
+
+"""
+Convert a markdown object to a `DOM.Node` object.
+
+The `parent` argument is passed to allow for context-dependant conversions.
+"""
+mdconvert(md) = mdconvert(md, md)
+
+
+mdconvert(text::AbstractString, parent) = text
```
mortenpi: It gets converted automatically, but should this be explicitly `= Node(text)`, so that `mdconvert` would always return a `Node`?

mortenpi:
It seems that HTML within Markdown gets escaped (`<` gets converted to `&lt;` etc.). Of course, converting all the HTML to DOM would, I think, require a whole HTML parser, but I am still wondering whether we should do something about it.

Otherwise :+1:!

MichaelHatherly:
> It seems that HTML within Markdown gets escaped

I think there's probably an open issue about that in `Base`. No one has had the opportunity to solve that, since, as you say, we need a full HTML parser there to correctly parse things.

src/Utilities/DOM.jl
MichaelHatherly: Yes, can do for consistency, (though it's not a type stability issue.)

mortenpi:
> > It seems that HTML within Markdown gets escaped

> I think there's probably an open issue about that in `Base`. No one has had the opportunity to solve that, since, as you say, we need a full HTML parser there to correctly parse things. 

The Markdown parser in `Base`, strictly speaking, doesn't have to parse the HTML since it is fine to store the HTML as is and later just print it out.

One option would be to handle the HTML properly somewhere down the road (and have an open issue for now?). But perhaps a better option would be to have a no-HTML policy? I.e. we wouldn't support full markdown in a way, but you really shouldn't be using HTML in the docstrings anyway, so it's OK. We should probably document that somewhere though.

MichaelHatherly:
> no-HTML policy

Agreed, having lots of bells and whistles to make use of in a docstring can lead to overly-complex docs as far as I'm concerned.

It might be worth documenting the exact syntax that `Markdown.parse` currently supports. Probably in the Julia manual. The current description hasn't been updated since it was first added and is a little vague http://docs.julialang.org/en/latest/manual/documentation/#markdown-syntax-notes.

PR #116, opened 2016-06-27
Split Writers.jl into multiple files

mortenpi:
In order accommodate multiple output formats the Writers module should
be in multiple files. Futhermore, in order for the internal code for
each format not to conflict with other code, they should each be
wrapped in submodules as well.

A couple of questionable decisions currently though:

  - Having the suffix `W`, even though it is not per say necessary.
  - While each format is a module, I like having just one line per format in `Writers.jl`. Hence the fact that the files are not named after the modules and that there is an `using` statement in them.

### DIFF
--- a/src/Writers/Writers.jl
+++ b/src/Writers/Writers.jl
@@ -1,0 +1,43 @@
+"""
+Provides a rendering function, [`render`](@ref), for writing each supported
+[`Formats.Format`](@ref) to file.
+
+Note that currently `Formats.Markdown` is the **only** supported format.
+
+"""
+module Writers
+
+import ..Documenter:
+    Anchors,
+    Builder,
+    Documents,
+    Expanders,
+    Formats,
+    Documenter,
+    Utilities
+
+using Compat
+
+# Driver method for document rendering.
+# -------------------------------------
+
+"""
+Writes a [`Documents.Document`](@ref) object to `build` directory in specified file format.
+"""
+function render(doc::Documents.Document)
+    mime = Formats.mimetype(doc.user.format)
+    for (src, page) in doc.internal.pages
+        open(Formats.extension(doc.user.format, page.build), "w") do io
+            for elem in page.elements
+                node = page.mapping[elem]
+                render(io, mime, node, page, doc)
+            end
+        end
+    end
+end
+
+include("MarkdownWriter.jl")
+include("HTMLWriter.jl")
+include("LaTeXWriter.jl")
+
+end
--- a/docs/src/lib/internals.md
+++ b/docs/src/lib/internals.md
@@ -140,8 +140,11 @@ Walkers.walk
 ```@docs
 Writers
 Writers.render
-Writers.join_decl
-Writers.span
+Writers.MarkdownWriter
+Writers.MarkdownWriter.join_decl
+Writers.MarkdownWriter.span
+Writers.HTMLWriter
+Writers.LaTeXWriter
 ```
 
 ## Utilities
@@ -178,4 +181,3 @@ Utilities.DOM.flatten!
 Utilities.DOM.MarkdownConverter
 Utilities.DOM.MarkdownConverter.mdconvert(md)
 ```
-
--- a/src/Builder.jl
+++ b/src/Builder.jl
@@ -8,7 +8,6 @@ doctests, etc.
 module Builder
 
 import ..Documenter:
-
     Anchors,
     Selectors,
     Documents,
--- a/src/CrossReferences.jl
+++ b/src/CrossReferences.jl
@@ -4,7 +4,6 @@ Provides the [`crossref`](@ref) function used to automatically calculate link UR
 module CrossReferences
 
 import ..Documenter:
-
     Anchors,
     Builder,
     Documents,
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -4,7 +4,6 @@ Provides two functions, [`missingdocs`](@ref) and [`doctest`](@ref), for checkin
 module DocChecks
 
 import ..Documenter:
-
     Builder,
     Documents,
     Expanders,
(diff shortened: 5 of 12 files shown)


src/Writers/Writers.jl
```diff
@@ -0,0 +1,51 @@
...
+The code for each format should generally be in a separate file (e.g. `markdown.jl`)
+and wrapped in a module. In the interest of code readability, the module names
+should be suffixed with a "W", to indicate that it is a "Writer" (e.g. `MarkdownW`),
+even if there is not direct conflict.
+
+"""
+module Writers
+
+import ..Documenter:
+
```
mortenpi: I would also actually like to remove this space, so it would be clear that the block of names is tied to the `import`.

MichaelHatherly: Fair enough, though all the other `import`s should be fixed as well at some point. Preferably in a separate PR later on.

MichaelHatherly:
Thanks for tackling this one, :+1:

> Having the suffix W, even though it is not per say necessary.

Preferably not. I doubt there would be any naming conflicts in the `Writers` module. If we do really need to suffix these then spelling it out as `MarkdownWriter` might be best.

> files are not named after the modules

Is that necessary? The rule I've been following is files with `module` are named the same as the module, and files without `module` are lowercase (though we don't have any right now). Would be best to just have a single naming scheme, whichever way we go on this.

 > there is an `using` statement in them.

Since the module's don't export anything is the `using` actually needed? I think it would be best to keep internals all as qualified uses to make it easier to track down where things are coming from.

src/Writers/html.jl
```diff
@@ -0,0 +1,27 @@
+"""
+Provides the `render()` methods to write the documentation as HTML files
+(`MIME"text/html`).
```
MichaelHatherly: Missing a closing `"`. Can you also escape the opening and closing `"` so that GitHub doesn't mangle the highlighting? With `MIME\"text/html\"`. Same with the others like this.

mortenpi: Thanks for catching the typo. However, I would prefer not to escape, since it's a GitHub bug, and not an issue in, say, Atom. And as long as we have the closing `"` present as well, the highlighting doesn't break for the whole file, so I think we could live with it.

I also reported the highlighting bug to GitHub just now.

src/Writers/Writers.jl
mortenpi: I'll only make the change in the `Writers` module right now as part of this PR.

mortenpi:
> > Having the suffix W, even though it is not per say necessary.

> Preferably not. I doubt there would be any naming conflicts in the `Writers` module. If we do really need to suffix these then spelling it out as `MarkdownWriter` might be best.

My thinking was along the lines of not confusing this `Documenter.Writers.Markdown` with `Base.Markdown`, just from the readability POV, when scanning the `import` statements and so on (I also don't foresee any actual naming conflicts within the submodule). But I now think I was probably overthinking it and the namespacing by the modules is enough. 

> > files are not named after the modules

> Is that necessary? The rule I've been following is files with `module` are named the same as the module, and files without `module` are lowercase (though we don't have any right now). Would be best to just have a single naming scheme, whichever way we go on this.

> > there is an `using` statement in them.

> Since the module's don't export anything is the `using` actually needed? I think it would be best to keep internals all as qualified uses to make it easier to track down where things are coming from.

When I was writing this, I got the impression that the `render` method's overloads from the submodules did not get loaded into the parent scope without the `using`, but that is wrong, so the `using`s are indeed useless. This also now means that the each file is a module and naming becomes a no-brainer.

mortenpi:
> I also don't foresee any actual naming conflicts within the submodule

Ah, actually there is a small conflict. `Base.Markdown` gets automatically loaded into the scope as `Markdown`, so having your own module `Markdown` overwrites that binding. A simple `import Base.Markdown: <stuff>` or explicit `Base.Markdown.<stuff>` will fix things (there is some `Markdown.MD` and `Markdown.Header` types used in the current code).

But, I am now thinking again that maybe a `MarkdownWriter` would be better (since we are not actually replacing the `Base.Markdown`). The namespacing `Writers.Markdown` makes it clear what our module is, but you can't make use of this within the `Writers` module.

MichaelHatherly:
> Ah, actually there is a small conflict. `Base.Markdown` gets automatically loaded into the scope as `Markdown`, so having your own module `Markdown` overwrites that binding.

Yeah, that's kind of annoying. I'd be in favour of qualifying `Base.Markdown.<name>` so that it's easier to `grep` for things when needed. Not terribly bothered though with which naming you'd like to go with though, so `MarkdownWriter` is fine with me as well.

mortenpi:
I feel that avoiding the conflict is worth it, so I would go with the `SomethingWriter` convention. Ready to be merged I think, once tests complete.

src/Writers/LatexWriter.jl
```diff
@@ -0,0 +1,24 @@
+"""
+Provides the [`render`](@ref) methods to write the documentation as Latex files
+(`MIME"text/latex"`).
+"""
+module LatexWriter
```
MichaelHatherly: Just noticed, this should be `LaTeXWriter` to follow official spelling for the formats, i.e. `HTMLWriter` not `HtmlWriter`. Other than that, looks ready to merge to me.

src/Writers/LaTexWriter.jl
```diff
@@ -0,0 +1,24 @@
+"""
+Provides the [`render`](@ref) methods to write the documentation as Latex files
+(`MIME"text/latex"`).
+"""
+module LaTexWriter
```
MichaelHatherly: The `x` should also be a capital here 😄. (Sorry for the nitpicking.)

mortenpi: Yes, it should be. My ability to sometimes systematically not notice such things never ceases to amaze me -.-

mortenpi:
Also, since there are no open PRs right now, I figured we could remove the the empty lines in `import`s, keep master consistent. I think I got them all.

MichaelHatherly:
LGTM, merge when ready.

PR #117, opened 2016-06-29
Move stream redirection later in pipeline

MichaelHatherly:
Should help make output redirection less visible and easier to debug.

@mortenpi this *might* solve your `julia -i file.jl` troubles.

### DIFF
--- a/docs/src/lib/internals.md
+++ b/docs/src/lib/internals.md
@@ -36,9 +36,11 @@ Builder
 Builder.DocumentPipeline
 Builder.SetupBuildDirectory
 Builder.CopyAssetsDirectory
+Builder.RedirectOutputStreams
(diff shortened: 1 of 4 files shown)


mortenpi:
Yup, it fixes it. It now nicely goes to REPL after it's done with the file.

MichaelHatherly:
Cheers, thanks for checking.

mortenpi:
Just accidentally ran across this -- I guess [Expanders.jl#L31](https://github.com/JuliaDocs/Documenter.jl/blob/b77da2625f5b9cb17db563166f9a4c119df05e34/src/Expanders.jl#L31) should have been changed as well.

MichaelHatherly:
Oops, thanks for catching that.

PR #121, opened 2016-07-01
User defined page order

MichaelHatherly:
To remain undocumented for the moment. Will be used later on for auto-generated tables of contents in LaTeX and HTML documents as well as the `mkdocs.yml` configuration file for markdown.

The pages are stored as a flat vector. Page "nesting" should be determined automatically based on the layout of the `src/` directory sub-folders rather being specified by the user.

### DIFF
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -68,6 +68,7 @@ immutable User
     clean   :: Bool           # Empty the `build` directory before starting a new build?
     doctest :: Bool           # Run doctests?
     modules :: Set{Module}    # Which modules to check for missing docs?
+    pages   :: Vector{Any}    # Ordering of document pages specified by the user.
 end
 
 """
@@ -102,6 +103,7 @@ function Document(;
         clean    :: Bool             = true,
         doctest  :: Bool             = true,
         modules  :: Utilities.ModVec = Module[],
+        pages    :: Vector           = Any[],
         others...
     )
     Utilities.check_kwargs(others)
@@ -113,7 +115,8 @@ function Document(;
         format,
         clean,
         doctest,
-        Utilities.submodules(modules)
+        Utilities.submodules(modules),
+        pages,
     )
     internal = Internal(
         Utilities.assetsdir(),


mortenpi:
That was roughly the next thing I intended to tackle, because you can't really do HTML without this.

But you really need hierarchy too, and pretty names. Basically, I think, we need to reproduce the structure currently provided by the `pages:` section of `mkdocs.yml`.

So I would say Dicts would be the way to go, and it actually looks pretty reasonable. This is what I was toying around with:
```julia
makedocs(
    modules = Documenter,
    clean   = false,
    pages = Dict(
        "Home" => "index.md",
        "Manual" = Dict(
            "Guide" => "man/guide.md",
            "Examples" => "man/examples.md",
            "Syntax" => "man/syntax.md",
            "Doctests" => "man/doctests.md",
            "Hosting Documentation" => "man/hosting.md",
            "LaTeX Syntax" => "man/latex.md",
            "Internals" => "man/internals.md",
        ),
        "Library" => Dict(
            "Public" => "lib/public.md",
            "Internals" => "lib/internals.md"
        )
    )
)
```

mortenpi:
I managed to ignore the second paragraph. I actually think making use of the information provided by the file system could be a good idea, but, how would you determine the "pretty names"? Especially for a subdirectory (e.g. `man/` -> Manual, `lib/` -> Library)?

MichaelHatherly:
Using a `Dict` doesn't preserve any ordering though, which is needed for LaTeX rendering at least.

For page "titles" I'm thinking those should be either determined via the first header on a page or, if it exists, via a `PageTitle = "..."` in a `@meta` block. And nesting can be determined via the folder structure as I mentioned in the commit message.

MichaelHatherly:
> Especially for a subdirectory (e.g. man/ -> Manual, lib/ -> Library)?

Maybe just by naming the folder `Manual` or `Library`? Though that's not too nice I think.

Perhaps a middle ground between the `Vector{String}` and `Dict`? A `Vector{Union{String, Pair{String, String}}}`, i.e

```julia
makedocs(
    modules = Documenter,
    clean   = false,
    pages = [
        "index.md",
        "man/" => "Manual",
        "man/guide.md",
        "man/examples.md",
        "man/syntax.md",
        "man/doctests.md",
        "man/hosting.md",
        "man/latex.md",
        "man/internals.md",
        "lib" => "Library",
        "lib/public.md",
        "lib/internals.md",
    ],
)
```

So either a `=>` is provided to give the pretty name, or it's found in a `@meta` block, or lastly guessed via the first header on the page.

MichaelHatherly:
Lets go with an explicitly untyped vector here for now, since we will definitely need it to be order-preserving. Once we start actually using it we'll know better what kind of data we're storing and can tighten the type a bit if needed.

mortenpi:
Yes, I forgot that a `Dict` loses the ordering information, so definitely a vector. But I think we can make the decision how the user should define this now.

Falling back to the headers in the files in most cases is a good idea. `mkdocs.yml` often contains redundant information in that sense. I also like the possibility to have the possibility of a "local override" (`PageTitle` in `@meta`). Keeping the headers in the `.md` file, in general, is good.

But I wonder if we can do better with the hierarchy and directory names. What would the following mean?

```julia
pages = [
    "index.md",
    "man/" => "Manual",
    "lib" => "Library",
    "page1.md",
    "lib/public.md",
    "page2.md",
    "lib/internals.md",
]
```
Could, of course, be a "compilation error", or we could just determine the order on a subdirectory basis, but would be nice if that wouldn't be possible.

MichaelHatherly:
I'd think it would be an error. I can't really come up with a situation where you'd want to have the pages of a subdirectory split up like that. But perhaps requiring nesting for subdirectories like so:

```julia
pages = [
    "index.md",
    "Manual" => [
        "man/guide.md",
        "man/examples.md",
        "man/syntax.md",
        "man/doctests.md",
        "man/hosting.md",
        "man/latex.md",
        "man/internals.md",
    ],
    "Library" => [
        "lib/public.md",
        "lib/internals.md",
    ],
]
```

would be clearest.

mortenpi:
> I can't really come up with a situation where you'd want to have the pages of a subdirectory split up like that.

I also think it's meaningless, so I would be nice to not allow that (instead of an error, if possible).

I like that one you proposed. I guess though we wouldn't enforce the subdirectories then? I.e. this would also be OK (although bad style):

```julia
pages = [
    "index.md",
    "Manual" => [
        "man/guide.md",
        "examples.md"
    ],
    "lib/internals.md"
]
```

and would yield a navigation bar

```julia
- Index
- Manual
  - Guide
  - Examples
- Internals
```

MichaelHatherly:
> I guess though we wouldn't enforce the subdirectories then?

Yeah, probably not for the moment. A `"..." => [...]` element would just be for
naming a subsection of the docs that doesn't actually have real page associated
with it as you've illustrated in that example. I'd also agree that that's probably bad
style though.

PR #122, opened 2016-07-01
Minor refactors for multi-format support

mortenpi:
Several small changes, not directly related to each other. Descriptions in the commit messages.

### DIFF
--- a/src/Builder.jl
+++ b/src/Builder.jl
@@ -23,7 +23,6 @@ using Compat
 The default document processing "pipeline", which consists of the following actions:
 
 - [`SetupBuildDirectory`](@ref)
-- [`CopyAssetsDirectory`](@ref)
 - [`RedirectOutputStreams`](@ref)
 - [`ExpandTemplates`](@ref)
 - [`CrossReferences`](@ref)
@@ -38,11 +37,6 @@ abstract DocumentPipeline <: Selectors.AbstractSelector
 Creates the correct directory layout within the `build` folder and parses markdown files.
 """
 abstract SetupBuildDirectory <: DocumentPipeline
-
-"""
-Copies the contents of the `assets` directory into the `build` folder.
-"""
-abstract CopyAssetsDirectory <: DocumentPipeline
 
 """
 Replace `STDOUT` and `STDERR` streams with a single stream to capture doctest output.
@@ -76,13 +70,12 @@ Writes the document tree to the `build` directory.
 abstract RenderDocument <: DocumentPipeline
 
 Selectors.order(::Type{SetupBuildDirectory})   = 1.0
-Selectors.order(::Type{CopyAssetsDirectory})   = 2.0
-Selectors.order(::Type{RedirectOutputStreams}) = 3.0
-Selectors.order(::Type{ExpandTemplates})       = 4.0
-Selectors.order(::Type{CrossReferences})       = 5.0
-Selectors.order(::Type{CheckDocument})         = 6.0
-Selectors.order(::Type{RestoreOutputStreams})  = 7.0
-Selectors.order(::Type{RenderDocument})        = 8.0
+Selectors.order(::Type{RedirectOutputStreams}) = 2.0
+Selectors.order(::Type{ExpandTemplates})       = 3.0
+Selectors.order(::Type{CrossReferences})       = 4.0
+Selectors.order(::Type{CheckDocument})         = 5.0
+Selectors.order(::Type{RestoreOutputStreams})  = 6.0
+Selectors.order(::Type{RenderDocument})        = 7.0
 
 Selectors.matcher{T <: DocumentPipeline}(::Type{T}, doc::Documents.Document) = true
 
@@ -114,23 +107,6 @@ function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
         end
     else
         error("source directory '$(abspath(source))' is missing.")
-    end
-end
-
-function Selectors.runner(::Type{CopyAssetsDirectory}, doc::Documents.Document)
-    Utilities.log("copying assets to build directory.")
-    assets = doc.internal.assets
-    if isdir(assets)
-        builddir = joinpath(doc.user.build, "assets")
-        isdir(builddir) || mkdir(builddir)
-        for each in readdir(assets)
-            src = joinpath(assets, each)
-            dst = joinpath(builddir, each)
-            ispath(dst) && Utilities.warn("Overwriting '$dst'.")
-            cp(src, dst; remove_destination = true)
-        end
-    else
-        error("assets directory '$(abspath(assets))' is missing.")
     end
 end
 
--- a/src/Writers/Writers.jl
+++ b/src/Writers/Writers.jl
@@ -18,23 +18,26 @@ import ..Documenter:
 
 using Compat
 
-# Driver method for document rendering.
-# -------------------------------------
+"""
+A parametric type that allows us to use multiple dispatch to pick the appropriate
+writer for each output format.
+
+The parameter `f` should be an instance of the [`Formats.Format`](@ref) enumeration.
+"""
+immutable Writer{f} end
+Writer(f::Formats.Format) = Writer{f}()
 
 """
-Writes a [`Documents.Document`](@ref) object to `build` directory in specified file format.
+Writes a [`Documents.Document`](@ref) object to `.user.build` directory in
+the format specified in `.user.format`.
+
+The method should be overloaded in each writer as
+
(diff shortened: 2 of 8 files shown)


MichaelHatherly:
Thanks!

MichaelHatherly:
Yeah, other than the one comment, LGTM.

mortenpi:
> Might it be better to pack all the logic to determine `name` into `addpage!`?

I had to think for a sec, but: my argument would be that when pushing a new page, its name should be specified explicitly, not determined magically by `addpage!`. So the semantics of `addpage!(doc, name, src)` would be to "add page `name` to `doc` from the file system path `src`". So, in principle, the path could be anything (e.g. `/tmp/something`).

Just to note, the idea is not to support some crazy hypothetical here, but for me this way the semantics of `addpage!` are conceptually simpler (instead of the `src` being some weird semi-relative path relative to something in `doc`. I could live with determining the name from a path relative to `.user.src`, but I think being explicit is even better).

Also, looking at it again now, I would completely remove the `dst` field and determine the writing path in the writer from `name`. This could be a separate PR however.

I'm in favor of simplifying the code however (didn't know about `splitext`). The `last` made me think though -- perhaps `first` would be better style than `name, _ = ...`?

MichaelHatherly:
> not determined magically by `addpage!`

The argument for moving that logic into `addpage!` is that computing `name` is
irrelevant to everything else that happens in the `runner` method.

I'd say the `addpage!` signature should probably just be reduced to

```julia
addpage!(doc::Document, sourcefile::AbstractString)
```

`doc` has access to `.user.src` etc. which should be enough to calculate a
`name`, `src`, and whatever else the `Page` needs.

It's not really anything magical happening as far as I can tell, just
separating out parts of a largish method that aren't as relevant to the main
goal of that method &mdash; collecting all the source files and doing something
(either copying or building a `Page`) to each one.

> perhaps `first` would be better style than `name, _ = ...`?

Can do, I don't particularly mind which.

mortenpi:
> aren't as relevant to the main goal of that method — collecting all the source files and doing something (either copying or building a `Page`) to each one.

In my head determining the name of the page is relevant to the `runner`, since the one who pushes pages should also say what the name of the page is (like adding an element to a dictionary). In the interest of modularity and clarity, I think you shouldn't deal with paths any more once you're physically pushing a page to a `Document`, even if you technically have access to that information (helps testing and hacking, among others).

Nevertheless, if you would like to see it in `addpage!`, I'll make the change.

Currently I made only a small change to this commit. I stuck with the path magic, instead of `split`, because for me that is more readable.

mortenpi:
Ok, just saw that removing `dst` (by which I really meant removing the whole `.build` from `Page`) is a bit more work than I thought due to the cross-references and such, so definitely not part of this PR.

MichaelHatherly:
> the one who pushes pages should also say what the name of the page is

Is the page name going to ever be anything other than the source file name without the file extension and the source dir? So `src/foo.md` becomes `foo`. If that's the case then the `name` is already implicit in the file name and we don't need to also specify it.

> I think you shouldn't deal with paths any more once you're physically pushing a page to a `Document`

Not sure I'm following you here. The paths are always going to be needed, especially for things like reading page content, calculating autolinks, switching to the correct directory for `@eval` blocks, and writing the final generated content to file. They're not just there for testing and hacking.

> is a bit more work than I thought due to the cross-references and such, so definitely not part of this PR.

The `dst` path can stay for the moment, there's no need to make that change unless it's actually causing problems for the HTML rendering work you're working on.

src/Builder.jl
```diff
@@ -104,36 +97,23 @@ function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
...
                 # Non-markdown files are simply copied over to `build`.
                 # Markdown files get added to the document tree as `Page` objects.
-                endswith(src, ".md") ?
-                    Documents.addpage!(doc, src, dst) :
+                if endswith(file, ".md")
+                    name = first(splitext(path))
+                    Documents.addpage!(doc, name, src, dst)
+                else
                     cp(src, dst; remove_destination = true)
+                end
```
MichaelHatherly: The more minimal change for this would just be

```diff
diff --git a/src/Documents.jl b/src/Documents.jl
index b26be4d..4ba9395 100644
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -132,7 +132,8 @@ end
 
 function addpage!(doc::Document, src::AbstractString, dst::AbstractString)
     page = Page(src, dst)
-    doc.internal.pages[src] = page
+    name, _ = splitext(relpath(src, doc.user.source))
+    doc.internal.pages[name] = page
 end
```

which I feel would be much more preferable.

src/Writers/Writers.jl
```diff
@@ -18,23 +18,26 @@ import ..Documenter:
...
+Writes a [`Documents.Document`](@ref) object to `.user.build` directory in
+the format specified in `.user.format`.
+
+The method should be overloaded in each writer as
+
+    render(::Writer{format}, doc)
+
+where `format` is one of the values of the [`Formats.Format`](@ref) enumeration.
+"""
+render(doc::Documents.Document) = render(Writer(doc.user.format), doc)
```
MichaelHatherly: If we can avoid this type instability here that would be great since `Writer(...)` will always be unstable. I doubt it would ever be a performance hotspot, but would be good to avoid anyway.

Here one possible approach:

Don't extend `Writers.render` in each submodule. Just define a new `MarkdownWriter.render` etc. I don't think we'd ever need to dispatch to methods between different modules so including `Writer{Format}` in the signatures doesn't seem worth it. Instead of the dynamic dispatch being used here, we just hardcode the required `if`s:

```julia
function render(doc::Documents.Document)
    doc.user.format === Formats.Markdown ? MarkdownWriter.render(doc) :
    doc.user.format === Formats.HTML     ?     HTMLWriter.render(doc) :
    doc.user.format === Formats.LaTeX    ?    LaTeXWriter.render(doc) : error("unhandled format.")
end
```

I don't see us allowing the output formats to be user-defined any time in the *near* future.

---

Or perhaps write a new `Selector` for this:

```julia
abstract WriterFormat <: Selectors.AbstractSelector

abstract MarkdownFormat <: WriterFormat
abstract HTMLFormat <: WriterFormat
abstract LaTeXFormat <: WriterFormat

Selectors.order(::Type{MarkdownFormat}) = 1.0
Selectors.order(::Type{HTMLFormat})     = 2.0
Selectors.order(::Type{LaTeXFormat})    = 3.0

Selectors.matcher(::Type{MarkdownFormat}, doc) = doc.user.format === Formats.Markdown
Selectors.matcher(::Type{HTMLFormat}, doc)     = doc.user.format === Formats.HTML
Selectors.matcher(::Type{LaTeXFormat}, doc)    = doc.user.format === Formats.LaTeX

Selectors.runner(::Type{MarkdownFormat}, doc) = MarkdownWriter.render(doc)
Selectors.runner(::Type{HTMLFormat}, doc)     = HTMLWriter.render(doc)
Selectors.runner(::Type{LaTeXFormat}, doc)    = LaTeXWriter.render(doc)

render(doc::Documents.Document) = Selectors.dispatch(WriterFormat, doc)
```

which would be user-extensible *if* we choose to export/document output formats at some point.

mortenpi: I intend to use user-defined formats to prototype the HTML code "out of tree", so to speak. And in any case, if we're going to have an `if`, might as well let the type system handle it. A `Selector` makes 3 lines of code 14, so I feel that the current one is preferred.

src/Builder.jl
mortenpi: I think I still disagree, but, after sleeping on it, I do see that `addpage!` can have alternative semantics, where this makes sense (and my original change seems to have been changing the semantics). Also, links are an issue I didn't really consider, so I need to put in a bit more thought (i.e. beyond this PR).

Changed to the code you proposed.

src/Writers/Writers.jl
MichaelHatherly: Sure, yes it's shorter. Making the type system do the work that conditionals should be doing isn't great if we can avoid it, but if it makes your work easier for the moment to work out-of-tree then we can go with the dynamic dispatch approach for now and revisit this later on.

MichaelHatherly:
Marked as "pending", since it seems ready to go as far as I can tell.

mortenpi:
If you're OK with the `Writer` stuff as is, then feel free to merge.

src/Writers/Writers.jl
mortenpi: Yup, the code is self-contained, so changing it in the future should be easy.

PR #125, opened 2016-07-05
Add definitions needed for markdown admonitions

### LINKED ISSUE #124: Custom div blocks / admonitions
The general idea is to have (extended Markdown) syntax available to wrap Markdown in custom `<div>`-type blocks. This is to support "Tip", "Note", "Warning" etc. type blocks which are a bit out of the normal flow of the document and require special formatting.

Borrowing from Python's Markdown docs, something like this
```markdown
!!! note
    You should note that the title will be automatically capitalized.
```
would get rendered as
```html
<div class="admonition note">
<p class="admonition-title">Note</p>
<p>You should note that the title will be automatically capitalized.</p>
</div>
```
and additional styling could then be defined in the CSS to make the section stand out.

There is precedent for this in Markdown in the form of [rST-style admonitions](https://pythonhosted.org/Markdown/extensions/admonition.html). The rST admonitions are also used in [in the base docs](http://docs.julialang.org/en/latest/manual/mathematical-operations/#updating-operators), meaning this should also be relevant for julialang/julia#12573.

Potential syntax could be either (a) Python's admonitions syntax, or (b) provide another ```` ```@something ```` block in Documenter. (b) would be relatively straightforward to implement in-house, but adopting existing syntax, I feel, would be preferred (but would require extending the Markdown parser in `Base`).

MichaelHatherly:
Will close #124.

### DIFF
--- a/src/Utilities/DOM.jl
+++ b/src/Utilities/DOM.jl
@@ -373,6 +373,15 @@ if isdefined(Base.Markdown, :Footnote)
     footnote(id, text,       parent) = span["#footnote-$(id)"](mdconvert(text, parent))
 end
 
+if isdefined(Base.Markdown, :Admonition)
+    import Base.Markdown: Admonition
+    mdconvert(a::Admonition, parent) =
+        div[".admonition.$(a.category)"](
+            p[".admonition-title"](a.title),
+            mdconvert(a.content, a),
+        )
+end
+
 end
 
 end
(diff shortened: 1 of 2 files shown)


vchuravy:
Is it possible to make admonitions working under v0.4?

MichaelHatherly:
Unfortunately not, this requires some changes to `Base.Markdown.parse` which can't be backported since they're new features rather than simply bugfixes.

PR #129, opened 2016-07-07
Move populate! calls to a separate Builder step

mortenpi:
Currently the writer populates the `IndexNode`s and `ContentsNode`s with links, but that should be a separate step. This separates the `populate!` calls into a separate `Builder.Populate` step, which runs just before the `Builder.RenderDocument`.

It fixes the issue of duplicate links in `Index`/`ContentsNode`s when `Builder.RenderDocument` gets called multiple times after the document has been constructed by the previous steps in the pipeline.

---

I took the simplest route of moving the `populate!` calls to a separate step, where I just loop through the elements of a `Page`. This makes the implicit assumption that all the `Contents`/`IndexNode`s are in the top-level of the Markdown files, which seems like a bad assumption (e.g. the same nodes in docstrings?).

Potential options could be:

- Make a full walk in the `Populate`
- Store a list of `Index`/`ContentsNode`s in the `Document` in one of the previous steps and then just loop over that.
- Perhaps the `populate!` calls can be merged into an existing walk in the pipeline?

but I wanted to RFC this first, before going for anything more complicated.

### DIFF
--- a/src/Builder.jl
+++ b/src/Builder.jl
@@ -28,6 +28,7 @@ The default document processing "pipeline", which consists of the following acti
 - [`CrossReferences`](@ref)
 - [`CheckDocument`](@ref)
 - [`RestoreOutputStreams`](@ref)
+- [`Populate`](@ref)
 - [`RenderDocument`](@ref)
 
 """
@@ -65,6 +66,11 @@ Switch back to the real `STDOUT` and `STDERR` that were changed in [`RedirectOut
 abstract RestoreOutputStreams <: DocumentPipeline
 
 """
+Populates the `ContentsNode`s and `IndexNode`s with links.
+"""
+abstract Populate <: DocumentPipeline
+
+"""
 Writes the document tree to the `build` directory.
 """
 abstract RenderDocument <: DocumentPipeline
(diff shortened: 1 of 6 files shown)


MichaelHatherly:
> in the top-level of the Markdown files, which seems like a bad assumption

That assumption should be fine. Non-toplevel code blocks can't be used as `@`-blocks. Docstrings should always be pure markdown, without any Documenter-specific blocks.

> Store a list of `Index`/`ContentsNodes` in the `Document` in one of the previous steps and then just loop over that.

Yes, that's best I'd think. Just add two fields to the `.internals` object and loop over each of those. Can be done in the `Expanders` module where the `Index`/`ContentsNodes` are created.

src/Builder.jl
```diff
@@ -136,6 +143,20 @@ function Selectors.runner(::Type{RestoreOutputStreams}, doc::Documents.Document)
     Utilities.restore_output_stream!(doc.internal.stream)
 end
 
+function Selectors.runner(::Type{Populate}, doc::Documents.Document)
+    Utilities.log("populating indices")
```
MichaelHatherly: `.` at end of message. Also needs the code block in https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/man/guide.md#building-an-empty-document that displays the usual output to be updated to reflect this change:

```
Documenter: redirecting output streams.
Documenter: restoring output streams.
Documenter: copying assets to build directory.
```

MichaelHatherly:
(Coverage stats don't make much sense. Seems like it's missing lines in random places. Appears to have been happening for a little while now.)

mortenpi:
Re-implemented it with two new fields in `.internal`. I had to shuffle stuff in `Documents.jl` a bit, so that types would be defined in the right order. Rebasing could be fun, so I'll wait with that until #134 is merged.

I did notice that, in principle, the `page.mapping` could be looped over as well, but I think using the new fields is cleaner actually.

MichaelHatherly:
LGTM, rebase and merge when ready.

PR #134, opened 2016-07-08
Link to source when repo is not hosted on GitHub

tribut:
You have nice links to the relevant part of the source in your docs which I wanted as well but my projects are hosted on an internal gitlab installation. This PR is my attempt at making that feature available to non-GitHub-hosted projects.
Let me know if you are unhappy with the name of the parameter or how it's implemented.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -60,6 +60,7 @@ export Deps, makedocs, deploydocs
         clean   = true,
         doctest = true,
         modules = Module[],
+        repo    = "",
     )
 
 Combines markdown files and inline docstrings into an interlinked document.
@@ -124,6 +125,22 @@ $(strip(readstring(joinpath(dirname(@__FILE__), "..", "docs", "make.jl"))))
 
 and so any docstring from the module `Documenter` that is not spliced into the generated
 documentation in `build` will raise a warning.
+
+**`repo`** specifies a template for the "link to source" feature. If you are
+using GitHub, this is automatically generated from the remote. If you are using
+a different host, you can use this option to tell Documenter how URLs should be
+generated. The following placeholders will be replaced with the respective
+value of the generated link:
+
+  - `{commit}` Git commit id
+  - `{path}` Path to the file in the repository
+  - `{line}` Line (or range of lines) in the source file
+
+For example if you are using GitLab.com, you could use
+
+```julia
+makedocs(repo = "https://gitlab.com/user/project/blob/{commit}{path}#L{line}")
+```
 
 # See Also
 
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -69,6 +69,7 @@ immutable User
     doctest :: Bool           # Run doctests?
     modules :: Set{Module}    # Which modules to check for missing docs?
     pages   :: Vector{Any}    # Ordering of document pages specified by the user.
+    repo    :: Compat.String  # Template for URL to source code repo
 end
 
 """
@@ -104,6 +105,7 @@ function Document(;
         doctest  :: Bool             = true,
         modules  :: Utilities.ModVec = Module[],
         pages    :: Vector           = Any[],
+        repo     :: AbstractString   = "",
         others...
     )
     Utilities.check_kwargs(others)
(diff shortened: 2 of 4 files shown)


MichaelHatherly:
> Let me know if you are unhappy with the name of the parameter

Not unhappy, but maybe something like `template_url` or `url_template`, since the argument acts as a template or sorts? @mortenpi any other naming suggestions?

Other than that the implementation seems fine to me. Thanks!

src/Documenter.jl
```diff
@@ -125,6 +126,22 @@ $(strip(readstring(joinpath(dirname(@__FILE__), "..", "docs", "make.jl"))))
 and so any docstring from the module `Documenter` that is not spliced into the generated
 documentation in `build` will raise a warning.
 
+**`sources_url`** specifies a template for the "link to source" feature. If you
+are using GitHub, this is automatically generated from the remote. If you are
+using a different hosting feature, you can use this option to tell Documenter
```
MichaelHatherly: `hosting feature` to `host` maybe?

MichaelHatherly:
(Just thinking out loud: might be nice to handle this automatically for other public hosts such as GitLab, BitBucket, etc. and not just GitHub. Private hosts would obviously still need to use the template though. Of course none of that needs to be implemented in this PR though.)

src/Documenter.jl
tribut: Sounds good, updated.

tribut:
Hm, `source_url_template` maybe?

mortenpi:
> > Let me know if you are unhappy with the name of the parameter

> Not unhappy, but maybe something like `template_url` or `url_template`, since the argument acts as a template or sorts? @mortenpi any other naming suggestions?

More options: `repository`, `repo`, `repo_url` or `repo_template`, with the "repo" part implying that it resolves to the remote repository. We use the word "source" to pretty systematically refer to the local source of the docs itself, so my two cents would be not to reuse it for this, if possible.

There's also the matter of issue references, which assume github right now (via [`issues_xref`](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/CrossReferences.jl#L138)).

MichaelHatherly:
Maybe just `repo` then, and we can reuse the data in other places rather than limiting it to source link templates. (`repo` would also match what we have in `deploydocs`, kind of.)

Regarding the `issues_xref`, the only reason it works for GitHub URLs is that `/issues/<number>` will redirect to `/pull/<number>` if `<number>` is a pull request and not an issue. I'm not sure whether other sites do the same thing. Might be best to leave that part as GitHub-specific for now.

tribut:
I'm not sure I understand. How would it be possible to re-use `repo` in other places? Add `repo` as `https://gitlab.com/user/project` and then also add `repo_tree_template` as just `{commit}{path}#L{line}`?

Regarding issues, GitLab at least does not use the same numbering scheme for merge requests and issues.

MichaelHatherly:
Just add `repo` with the templating included as

```julia
makedocs(repo = "https://gitlab.com/user/project/blob/{commit}{path}#L{line}")
```

I think. We can later extract other bits from `repo` if we need to. (Your implementation is fine as is, just needs `sources_url` to `repo` name change.

mortenpi:
(This is beyond this PR.) I am not sure how wise it is to make assumptions about the `repo`, beyond that it gives you the file. Do you have something specific in mind @MichaelHatherly where/how we could use it?

> Regarding issues, GitLab at least does not use the same numbering scheme for merge requests and issues.

Yup, it enumerates them separately. And instead of `pull/`, it's `merge_requests/`.

MichaelHatherly:
> Do you have something specific in mind @MichaelHatherly where/how we could use it?

Nothing in mind at the moment.

> Yup, it enumerates them separately. And instead of `pull/`, it's `merge_requests/`.

Ah, ok, lets leave that until someone *really* wants that then.

tribut:
I've updated the PR accordingly and rebased onto master.

MichaelHatherly:
Thanks for you contribution @tribut! Much appreciated.

MichaelHatherly:
Now available in the latest tagged release, `0.1.4`.

PR #139, opened 2016-07-13
Prepare for upcoming deletion of Pkg.Git module

tkelman:
get GITHUB_REGEX from Base.LibGit2 when it's defined


tkelman:
Lexicon might need a similar change, but is deprecated for 0.5 right?

MichaelHatherly:
Thanks. Lexicon and Docile are pretty much deprecated in 0.5 at the moment so no need to fix this there as well.

PR #145, opened 2016-07-20
Fix highlighting in docs

lopezm94:


MichaelHatherly:
Thanks @lopezm94. The highlighting was intentional in this case since `include` is the function used and file is "include-d", though maybe that's a bit subtle and slightly confusing. I'm happy to accept this PR if you find it less confusing with your wording.

lopezm94:
Didn't think about that, "include-d" is better. Unnecessary PR.

MichaelHatherly:
Ok, will close then. Thanks for taking the time to open a PR nonetheless!

PR #149, opened 2016-07-27
Move mdconvert to HTMLWriter

mortenpi:
The Markdown to DOM conversion is not always unique (e.g. element CSS classes) and is specific to the HTML writer. The `HTMLWriter` will also extend `mdconvert` with Documenter specific nodes and other processing, so it is preferred if all the code is in the same module.

Wrapping `mdconvert` into a submodule does not add much while in `HTMLWriter`, since other parts of the writer module will have to access the `DOM.*` and `Markdown.*` objects as well. However, in order not to pollute the `HTMLWriter` namespace, `@tags` are replaced by calls to `Tag` (for oneliners) or moved into function scopes (for multi line definitions). Similarly, all `Markdown.*` types are prepended with `Markdown.`, instead of importing them to the namespace.

---

This is in preparation for #3. I'd say that if it makes sense to split something generic into `Utilities`, we could just move it there again. Right now this would make merging the HTML output into Documenter a bit easier.


MichaelHatherly:
LGTM, merge when ready.

PR #150, opened 2016-07-27
Handle disabling stream redirection gracefully

mortenpi:
Currently, if `STDOUT` and `STDERR` have not been redirected by `redirect_output_stream!`, `restore_output_stream!` and `withoutput` calls throw `UndefRefError`s, since the `CombinedStream` fields are not initialized. With this change restoring or using non-redirected streams just prints an error.

---

Normally the streams should, of course, be redirected, but it's nice to disable redirection when debugging to reveal errors and stacktraces, which seem to disappear sometimes.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -417,6 +417,8 @@ type RedirectedStream
     from  :: IO
     to    :: IO
 
+    RedirectedStream(stream::IO) = new(stream, stream, DevNull, stream)
+
     function RedirectedStream(stream::IO, f)
         real = stream
         from, input = f()
@@ -431,7 +433,7 @@ type CombinedStream
     tout :: Task
     terr :: Task
 
-    CombinedStream() = new()
+    CombinedStream() = new(RedirectedStream(STDOUT), RedirectedStream(STDOUT), false)
 end
 
 function redirect_output_stream!(stream::CombinedStream)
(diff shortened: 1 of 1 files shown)


MichaelHatherly:
Could this be done with a dummy `RedirectedStream` for `.out` and `.err` that both just reference the normal `STDOUT` and `STDERR`? Where possible I don't think we should rely on `isdefined` to decide control flow.

Have you looked into using `Pipe` as suggested in your `start_reading` PR? That seems like it could be a nice approach possibly. If not, I may take a look into that tomorrow.

mortenpi:
Yes, a dummy would be much better and I've updated the PR accordingly. Since only `watchbuffer` would ever read from `.from`, it should be safe to have it be a `DevNull`, and none of this should affect the normal case when the streams are redirected anyway.

We could merge this as a placeholder until #151 or something similar gets tested (this PR needs cleanup first though).

MichaelHatherly:
> Yes, a dummy would be much better and I've updated the PR accordingly.

Yes, thanks, that looks like how I'd have done it too.

> We could merge this as a placeholder until #151 or something similar gets tested (this PR needs cleanup first though).

 I'll be merging #151 shortly in any case since it seems to work well, but thanks for putting the time into this one nonetheless.

mortenpi:
Superseded by #151.

PR #151, opened 2016-07-27
Use a `Pipe` for output redirection

MichaelHatherly:
(Sorry @mortenpi, I kind of just got carried away here... hope you hadn't already started on this as well.)

`Pipe`s *seem* to work fine, though this will need some extensive checks on travis with some of the packages that were giving problems before the change to `CombinedStream` etc., hence this is very WIP at the moment.

This returns to the original method of output redirection of only changing streams without the `withoutput` `do` block rather than globally in the `Document` object with some changes.

No `at-async` to combine the output streams. To deal with `STDOUT`/`STDERR` interleaving we use a `Pipe` and call `Base.link_pipe` (unexported) to redirect both `STDOUT` and `STDERR`.

Interestingly, the tests all passed without any problems: so either good, or very bad 😆. As a nice side-effect the single `info` calls that were previously getting lost are now captured correctly.

### DIFF
--- a/docs/src/lib/internals.md
+++ b/docs/src/lib/internals.md
@@ -35,11 +35,9 @@ Anchors.isunique
 Builder
 Builder.DocumentPipeline
 Builder.SetupBuildDirectory
-Builder.RedirectOutputStreams
 Builder.ExpandTemplates
 Builder.CrossReferences
 Builder.CheckDocument
-Builder.RestoreOutputStreams
 Builder.Populate
 Builder.RenderDocument
 ```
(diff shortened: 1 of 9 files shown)


MichaelHatherly:
(`0.4` errors are expected, need to correctly handle creating a `String` from `Vector{UInt8}`. I'll do that tomorrow though.)

mortenpi:
Nope, I didn't start on it, but the `Pipe` idea sounds promising and this looks great! I tested this with my code and it seems to work nicely, and stacktraces also survive now when there's an error.

MichaelHatherly:
This appears to work fine for the two problem packages, `Bio` and `TaylorSeries`, as far as I can tell, so I'll merge as is shortly.

PR #156, opened 2016-07-29
Use `==` for docstring lookup

MichaelHatherly:
And fallback on `<:` when no unique result is found.

This could potentially be worse from a usability point-of-view and cause havoc with packages using Documenter. Then again, it might be nicer since if you have two docstrings

```julia
"""
abstract
"""
f(x::Integer) = x

"""
concrete
"""
f(x::Int) = x
```

Then `getdocs` will return

```
julia> Documenter.DocSystem.getdocs(f, Tuple{Int})
1-element Array{Base.Docs.DocStr,1}: ...

julia> Documenter.DocSystem.getdocs(f, Tuple{Integer})
1-element Array{Base.Docs.DocStr,1}: ...
```

as opposed to

```
julia> Documenter.DocSystem.getdocs(f, Tuple{Int})
2-element Array{Documenter.DocSystem.DocStr,1}: ...

julia> Documenter.DocSystem.getdocs(f, Tuple{Integer})
1-element Array{Documenter.DocSystem.DocStr,1}: ...
```

Note the different length of results returned.

Not sure whether this is a better behaviour, but figured it might be worth testing out.

### DIFF
--- a/src/DocSystem.jl
+++ b/src/DocSystem.jl
@@ -277,7 +277,7 @@ Returns a `Vector{DocStr}` ordered by definition order in `0.5` and by
 function getdocs(
         binding::Docs.Binding,
         typesig::Type = Union{};
-        compare = (<:),
+        compare = (==),
         modules = Docs.modules,
         aliases = true,
     )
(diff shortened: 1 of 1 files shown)


MichaelHatherly:
Doesn't seem to cause any noticeable problems as far as I can tell.

PR #157, opened 2016-07-29
WIP: Catch parsing errors

mortenpi:
Currently various bad inputs in links and `@docs` blocks crash Documenter via errors thrown by `parse`, `eval` etc. With this we capture these errors, try to give the user a semi-helpful error message and allow Documenter to continue.

---

These types of errors came up when trying to run the converted juliadoc through Documenter. I think this is sort of an half-solution only -- there should be a better way of handling these errors in general, but in any case we should try/catchyify the `parse` etc. calls.

Also, I am not sure whether we should abort Documenter or not if one of these errors arises.


mortenpi:
(hmm, apparently in 0.4 you can't return from a catch block if the try is trying to assign)

MichaelHatherly:
Yes, you'll need to do something like

```julia
local x
try
    x = f(...)
catch err
   return
end
# use `x`
# ...
```

for `0.4` support. Bit ugly, but it does work.

Regarding throwing an error or continuing after exceptions are caught: so far it's mostly been a "carry on and log the error" approach. Problem with that is users can just ignore the warnings and end up with half-built docs, which won't reflect positively on Documenter. Might be worth testing out throwing an error instead.

test/errors/src/index.md
```diff
@@ -19,6 +19,20 @@ Modules = [NonExistantModule]
 NonExistantModule
 ```
 
+```@docs
+# comment in a @docs block
+```
+
+Linking to problematic items: [`ccall`](@ref), [`for`](@ref)
```
MichaelHatherly: The (undocumented) syntax for keyword refs is `` [`:ccall`](@ref) ``, though would be better to get it to just handle it properly I suppose.

test/errors/src/index.md
```diff
@@ -34,4 +48,3 @@ This is the footnote [^1]. And [^another] [^another].
 [^nested]:
 
     Duplicate [^1] nested footnote.
-
```
MichaelHatherly: Newline?

mortenpi: Yup, an extra one, which Atom removes automatically on save.

test/errors/src/index.md
mortenpi: I think that (in the long run) it would be nice to support keywords without extra syntax. In any case, currently they are creating an error, as they should, so it seems that they are appropriate in the errors section.

mortenpi:
> Might be worth testing out throwing an error instead.

Well, they currently do that, since they're not getting caught. In that case we would just get perhaps slightly more informative errors (currently it ends up being something like `ParseError: unexpected ,`, which is not that helpful). I guess the question is whether we should treat these errors (various sytanx errors basically) the same way as a missing reference and just warn and carry on, or at least fail the docs block.

By the way, there is no way of getting line numbers for the parsed Markdown, right? I think this has come up before as well. Would be great for the errors though :)

Also, one more thing that came up -- what about comments in the blocks? They end up as `nothing` expressions (at least lines) and thus error. Should we just ignore them?

MichaelHatherly:
> what about comments in the blocks

Comments are unsupported. There's no real need to allow them. The `build_sysimg` comment in the julia-docs conversion will be removed once I've sorted out including the file.

> By the way, there is no way of getting line numbers for the parsed Markdown, right? I think this has come up before as well. Would be great for the errors though :)

No, no line numbers.

test/errors/src/index.md
MichaelHatherly: > I think that (in the long run) it would be nice to support keywords without extra syntax

Done in #158.

test/errors/src/index.md
MichaelHatherly: Unrelated changes shouldn't be included. Especially whitespace ones. We can look into formatting everything in a consistent way in a separate PR if you'd like to (and have the time).

mortenpi:
Rebased, fixed v0.4 and split the whitespace change into a separate commit.

Having thought about it a bit, I think we should go with "warn the user and continue" for these errors. Primarily, I think Documenter should report all errors it finds before it quits. Otherwise you end up in this loop where you see an error, Documenter stops, you fix the error, run Documenter again, find the second errors, etc. I find this behaviour really annoying in general (e.g. in the deps check part of `./configure` scripts), and I think it would be better to allow Documenter to continue to the end so that you'd see all the errors.

There is the matter of reporting the errors. I think everything else is fine, but printing the stacktrace might be a bit noisy.. but it could be useful under certain circumstances. Should we maybe make that somehow optional? The `Utilities.__log__` flag or something?

Regarding whitespace: I agree it shouldn't be in the same commit as other changes, but I feel that it's fine to make related fixes in the same PR, if it came up due to these changes (in this case, editing one of the files). Generally it seems that we don't have empty lines in the end, so I removed the few that did (checked the `.md` files manually though, so there might be more; now I am going a bit beyond this PR though). I can also cherry pick the commit, or drop it completely if there's a reason for those lines.

MichaelHatherly:
> Having thought about it a bit, I think we should go with "warn the user and continue" for these errors. Primarily, I think Documenter should report all errors it finds before it quits. Otherwise you end up in this loop where you see an error, Documenter stops, you fix the error, run Documenter again, find the second errors, etc.

Yes, that sounds fine to me.

> but printing the stacktrace might be a bit noisy

Yeah, definitely too noisy most of the time. Perhaps trim it down to the top one/two frames if possible?

> The `Utilities.__log__` flag or something?

Could do, though less user options is better where possible. Not sure that's possible in this case.

Re: whitespace changes: I'm not particularly bothered which way we go on it, just needs to be in a separate commit as you've now done. So this is fine.

mortenpi:
I decided to remove the stacktrace altogether for now -- you can determine the location of the error from the message really and the expressions are simple, so I am not sure it would add much (trimmed or otherwise).

There was also some repeating code, so I broke it into a separate `warn` function. The warning are now printed in a style consistent with the other ones as well.

I think it should be ready to merge, if this behaviour seems fine.

src/CrossReferences.jl
```diff
@@ -116,11 +116,37 @@ function docsxref(link::Markdown.Link, meta, page, doc)
...
+    local ex
+    if haskey(Docs.keywords, keyword)
+        ex = QuoteNode(keyword)
+    else
+        try
+            ex = parse(code)
+        catch err
+            !isa(err, ParseError) && rethrow(err)
+            Utilities.warn(page.source, "Unable to parse the reference '[`$code`](@ref)'.")
+            return
```
MichaelHatherly: Codecov is saying this block isn't being tested. Can we add a malformed `code` to cover it?

src/CrossReferences.jl
```diff
@@ -116,11 +116,37 @@ function docsxref(link::Markdown.Link, meta, page, doc)
...
+
     # Find binding and type signature associated with the link.
-    local binding = Documenter.DocSystem.binding(mod, ex)
-    local typesig = eval(mod, Documenter.DocSystem.signature(ex, rstrip(code)))
+    local binding
+    try
+        binding = Documenter.DocSystem.binding(mod, ex)
+    catch err
+        Utilities.warn(page.source, "Unable to get the binding for '[`$code`](@ref)'.", err, ex, mod)
+        return
```
MichaelHatherly: Same as the one above. Needs something to trigger the `catch`. Perhaps `[`1`](@ref)` would work, or some other object that can't be converted to a `Binding`.

MichaelHatherly:
Other than the two comments this looks fine to merge.

mortenpi:
Added test cases for those. I would've expected symbols to fail as well (i.e. ``[`:sym`](@ref)``), but I don't think there's any reason to make the code more complicated just to fail these, and just let them resolve as they do.

But am I right in guessing that, other than the keyword thing which is out now, symbols wouldn't really make sense as references?

mortenpi:
(the previous test didn't actually test the parse error, fixed)

MichaelHatherly:
> I would've expected symbols to fail as well

Symbols, or actually `QuoteNode` is what they're parsed as, are needed so you can use `@doc` as follows:

```julia
"An undefined binding that gets defined later on, perhaps at runtime."
:STDOUT
```
Since `DocSystem.binding` just falls back to `Docs.Binding` eventually, there's no need to restrict it further.

PR #170, opened 2016-08-03
Parse doc.user.pages

mortenpi:
Proposing a way of parsing `doc.user.pages` and storing it into a convenient data structure. Necessary for #3 and #4, opening it up for discussion.

In a nutshell: `doc.internal.navtree` allows for walking across the tree for a tree-like navbar (like in HTML, but also probably in LaTeX), `doc.user.navlist` is useful for sequential things only on pages. `NavNode`'s `.next`, `.prev` etc. are useful for other navlinks.

~~**TODO:** It could probably use a few tests as well.~~

**`UTF8String`:** One thing I am doing is using `Compat.UTF8String` instead of `Compat.String` for type fields (which goes against the previous practice), but should actually be the correct one, if I interpret [Compat.jl's documentation](https://github.com/JuliaLang/Compat.jl#type-aliases) correctly. Specifically:

> Compat also provides an unexported `Compat.String` /--/ It should **not** be used as the default type for variables or fields holding strings, as it introduces type-instability in Julia 0.3 and 0.4: use Compat.UTF8String or Compat.ASCIIString instead.

In 0.4 `Compat.String = ByteString = Union{ASCIIString,UTF8String}`.

I am thinking that maybe we should have a `String = Compat.UTF8String` under every `using Compat` to get uniform behaviour and then use just `String` in fields. In this PR I could change back to `Compat.String` though, for consistency.

### DIFF
--- a/src/Builder.jl
+++ b/src/Builder.jl
@@ -112,7 +112,59 @@ function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
             end
         end
     end
+
+    # If the user hasn't specified the page list, then we'll just default to a
+    # flat list of all the markdown files we found, sorted by the filesystem
+    # path (it will group them by subdirectory, among others).
+    userpages = isempty(doc.user.pages) ? sort(mdpages) : doc.user.pages
+
+    # Populating the .navtree and .navlist.
+    # We need the for loop because we can't assign to the fields of the immutable
+    # doc.internal.
+    for navnode in walk_navpages(userpages, nothing, doc)
+        push!(doc.internal.navtree, navnode)
+    end
+
+    # Finally we populate the .next and .prev fields of the navnodes that point
+    # to actual pages.
+    local prev::Nullable{Documents.NavNode} = nothing
+    for navnode in doc.internal.navlist
+        navnode.prev = prev
+        Utilities.unwrap(prev) do prevnode
+            prevnode.next = navnode
+        end
+        prev = navnode
+    end
 end
+
+"""
+    walk_navpages(x, parent, doc)
+
+Recursively walks through the [`Documents.Document`](@ref)'s `.user.pages` field,
+generating [`Documents.NavNode`](@ref)s and related data structures in the
+process.
+
+This implementation is the de facto specification for the `.user.pages` field.
+"""
+walk_navpages(ps::Vector, parent, doc) = [walk_navpages(p, parent, doc)::Documents.NavNode for p in ps]
+walk_navpages(p::Pair, parent, doc) = walk_navpages(p.first, p.second, parent, doc)
+function walk_navpages(title::Compat.String, children::Vector, parent, doc)
+    nn = Documents.NavNode(nothing, title, parent)
+    nn.children = walk_navpages(children, nn, doc)
+    nn
+end
+function walk_navpages(title::Compat.String, page::Compat.String, parent, doc)
+    nn = walk_navpages(page, parent, doc)
+    nn.title_override = title
+    nn
+end
+function walk_navpages(src::Compat.String, parent, doc)
+    src in keys(doc.internal.pages) || error("'$src' is not an existing page!")
+    nn = Documents.NavNode(src, nothing, parent)
+    push!(doc.internal.navlist, nn)
+    nn
+end
+
 
 function Selectors.runner(::Type{ExpandTemplates}, doc::Documents.Document)
     Utilities.log(doc, "expanding markdown templates.")
--- a/src/Builder.jl
+++ b/src/Builder.jl
@@ -76,30 +76,41 @@ Selectors.strict{T <: DocumentPipeline}(::Type{T}) = false
 
 function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
     Utilities.log(doc, "setting up build directory.")
+
     # Frequently used fields.
     build  = doc.user.build
     source = doc.user.source
 
+    # The .user.source directory must exist.
+    isdir(source) || error("source directory '$(abspath(source))' is missing.")
+
+    # We create the .user.build directory.
+    # If .user.clean is set, we first clean the existing directory.
     doc.user.clean && isdir(build) && rm(build; recursive = true)
     isdir(build) || mkdir(build)
-    if isdir(source)
-        for (root, dirs, files) in walkdir(source)
-            for dir in dirs
-                d = normpath(joinpath(build, relpath(root, source), dir))
-                isdir(d) || mkdir(d)
-            end
-            for file in files
-                src = normpath(joinpath(root, file))
-                dst = normpath(joinpath(build, relpath(root, source), file))
-                # Non-markdown files are simply copied over to `build`.
-                # Markdown files get added to the document tree as `Page` objects.
-                endswith(src, ".md") ?
-                    Documents.addpage!(doc, src, dst) :
-                    cp(src, dst; remove_destination = true)
+
+    # We'll walk over all the files in the .user.source directory.
+    # The directory structure is copied over to .user.build. All files, with
+    # the exception of markdown files (identified by the extension) are copied
+    # over as well, since they're assumed to be images, data files etc.
+    # Markdown files, however, get added to the document and also stored into
+    # `mdpages`, to be used later.
+    mdpages = Compat.String[]
+    for (root, dirs, files) in walkdir(source)
+        for dir in dirs
+            d = normpath(joinpath(build, relpath(root, source), dir))
+            isdir(d) || mkdir(d)
+        end
+        for file in files
+            src = normpath(joinpath(root, file))
+            dst = normpath(joinpath(build, relpath(root, source), file))
+            if endswith(file, ".md")
+                push!(mdpages, Utilities.srcpath(source, root, file))
+                Documents.addpage!(doc, src, dst)
+            else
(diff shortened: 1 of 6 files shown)


src/Builder.jl
```diff
@@ -76,33 +76,101 @@ Selectors.strict{T <: DocumentPipeline}(::Type{T}) = false
...
         end
-    else
-        error("source directory '$(abspath(source))' is missing.")
+    end
+
+    # Load all the pages to the document.
+    for path in mdpages
+        src = joinpath(source, path)
+        dst = joinpath(build, path)
+        Documents.addpage!(doc, src, dst)
```
MichaelHatherly: Do we need to have this happen after the previous `for` loop? Can it not remain with line 106 where it was before?

src/Builder.jl
```diff
@@ -76,33 +76,101 @@ Selectors.strict{T <: DocumentPipeline}(::Type{T}) = false
...
+    end
+
+    # If the user hasn't specified the page list, then we'll just default to
+    # all the markdown files we found, listing them in a relatively arbitrary
+    # order, i.e. whatever came out of walkdir().
+    userpages = isempty(doc.user.pages) ? mdpages : doc.user.pages
```
MichaelHatherly: > relatively arbitrary order

Perhaps worth making the order stable then by calling `sort` perhaps?

src/Builder.jl
```diff
@@ -76,33 +76,101 @@ Selectors.strict{T <: DocumentPipeline}(::Type{T}) = false
...
+generating [`Documents.NavNode`](@ref)s and related data structures in the
+process.
+
+This implementation is the de facto specification for the `.user.pages` field.
+"""
+walk_navpages(ps::Vector, parent, doc) = map(p -> walk_navpages(p, parent, doc), ps)
+function walk_navpages{T}(p::Pair{String, Vector{T}}, parent, doc)
+    nn = Documents.NavNode(nothing, p.first, parent)
+    nn.children = walk_navpages(p.second, nn, doc)
+    nn
```
MichaelHatherly: Needs coverage, same with the one below.

src/Documents.jl
```diff
@@ -134,6 +134,31 @@ immutable EvalNode
...
 
+# Navigation
+# ----------------------
+
+"""
+Element in the navigation tree of a document, containing navigation references
+to other page, reference to the [`Page`](@ref) object etc.
+"""
+type NavNode
+    page           :: Nullable{Compat.UTF8String} # unless null, should point to a valid Page in the document
```
MichaelHatherly: Which are the cases where this could be an invalid page as opposed to a "valid Page"?

src/Documents.jl
```diff
@@ -134,6 +134,31 @@ immutable EvalNode
...
+    next           :: Nullable{NavNode}
+end
+NavNode(page, title_override, parent) = NavNode(page, title_override, parent, [], nothing, nothing)
+
+"""
+Constructs a list of the parents of the `navnode`, ordered such that the `navnode`
+itself is the last item.
+"""
+parents(navnode::NavNode) = isnull(navnode.parent) ? [navnode] :
+    push!(parents(get(navnode.parent)), navnode)
```
MichaelHatherly: > such that the `navnode` itself is the last item.

Calling this method `parents` doesn't seem to fit if it includes the node itself and parents of parents etc.

Perhaps `ancestors` or something similar?

MichaelHatherly:
> One thing I am doing is using Compat.UTF8String instead of Compat.String for type fields (which goes against the previous practice), but should actually be the correct one, if I interpret Compat.jl's documentation correctly. Specifically:

There's other places in the code based with type instability at the moment as well. If we actually run into bugs by using `Compat.String` then we can change it everywhere at once. Until then, just go with the current approach. (I'd like to remove `0.4` support *very* soon after `0.5` is properly released anyway, so it would be extra work for little noticeable gain I think.)

src/Builder.jl
mortenpi: Both would work, but I like that this breaks it into two separate steps -- walking the file system and actually parsing the files. The copying should be moved out by that argument too, true, but that should be moved into the writers anyway at some point (with the walk just creating a list of files).

src/Documents.jl
mortenpi: What it unclearly tries to say is that there should a `Page` object in `doc.internal.pages` that has this name. Needs rephrasing indeed.

src/Builder.jl
MichaelHatherly: The parsing is done is a separate function, it's separate enough like that. Doing it in a different loop afterwards isn't worth I think, the function is long enough as it is anyway.

src/Documents.jl
MichaelHatherly: Throw an error if that happens perhaps?

src/Documents.jl
mortenpi: I agree with the comment, but doesn't the word "ancestor" have the same issue though? "Parent" is good since I am calling it `.parent` in `NavNode`.

MichaelHatherly: Does `parents` return all the `NavNode`s all the way up until root or just the single `.parent` of a node?

MichaelHatherly: In other words: "ancestor" is the recursive version of "parent". If this does something recursive then it can't be called "parent".

src/Documents.jl
mortenpi: It will anyway, once the code tries to fetch the `Page` item from the `doc`. I think there's any point in checking for this in the constructor. The comment really is just a tip for whoever is trying to figure out what the fields contain exactly, with the "should" meaning "it will most likely do if the code works correctly".

src/Documents.jl
mortenpi: Ah, true enough! There is only one parent anyway. The plural was trying to imply that it goes all the way up, but parents indeed go in breadth, not depth, my bad. `ancestors` is the appropriate word.

One other option would be `navpath`, since effectively it gives the list of links one should click (sort of) from the root to this one, and "path" as in "path in a tree". Also avoids the is-end-included problem.

src/Builder.jl
```diff
@@ -76,32 +76,99 @@ Selectors.strict{T <: DocumentPipeline}(::Type{T}) = false
...
+    walk_navpages(x, parent, doc)
+
+Recursively walks through the [`Documents.Document`](@ref)'s `.user.pages` field,
+generating [`Documents.NavNode`](@ref)s and related data structures in the
+process.
+
+This implementation is the de facto specification for the `.user.pages` field.
+"""
+walk_navpages(ps::Vector, parent, doc) = [walk_navpages(p, parent, doc)::Documents.NavNode for p in ps]
+# NOTE (Compat): the `T <: Compat.String` is required for 0.4 compatibility,
```
mortenpi: I could definitely use a second opinion on this hack.

test/runtests.jl
```diff
@@ -143,6 +143,10 @@ end
 @test UnitTests.A.B.C in Documenter.Utilities.submodules(UnitTests.A)
 @test UnitTests.A.B.C.D in Documenter.Utilities.submodules(UnitTests.A)
 
+## NavNode tests
+
+include("navnode.jl")
```
mortenpi: Should we keep it in a separate file. If so, the location of the `include` might not be ideal.

MichaelHatherly: Yes, keeping it separate sounds fine to me, as does shuffling around `include`s in needed.

src/Builder.jl
MichaelHatherly: A `walk_navpages(p::Pair, parent, doc)` method that dispatches to a wrapper with `walk_navpages(p.first, p.second, parent, doc)` that has `walk_navpages(pfirst::String, psecond::Vector` and `walk_navpages(pfirst::AbstractString, psecond::AbstractString)` methods?

mortenpi: Yes, that sounds better. It gets rid of all the type parameters, including the one that we would otherwise have even in 0.5.

mortenpi:
Ok, I don't think I have anything to add to this now, except squashing and rebasing. Shall we wait with the merge until 0.2 gets tagged?

test/navnode.jl
```diff
@@ -29,7 +29,7 @@ pages = [
     "page6.md",
 ]
 doc = FakeDocument()
-doc.internal.pages = Dict(["page$i.md" => nothing for i=1:8])
+doc.internal.pages = Dict(["page$i.md" => nothing for i=1:8]) # NOTE (Compat): deprecated Dict syntax for 0.4 compatibility
```
MichaelHatherly: I think `Dict(map(i -> "page$i.md" => nothing, 1:8))` should work fine on all versions and it's hardly more verbose than the generator syntax anyway.

MichaelHatherly:
> Shall we wait with the merge until 0.2 gets tagged?

If this doesn't do anything noticeable to the user then it's fine to merge now and include in `0.2`, otherwise it should wait probably.

mortenpi:
> If this doesn't do anything noticeable

It doesn't, just populates the two new `doc.internal` fields that aren't used anywhere at the moment. But it does introduce a bit of new code (which, granted, seems to work without issues) that is not relevant right now, but still runs.

MichaelHatherly:
Fine to merge then. Just needs the intermediate commits squashed and then it good to go I think.

mortenpi:
Squashed and rebased.

Issue #154, opened 2016-07-28
Modernise tests


MichaelHatherly:
We should switch over to using `@testset` at some point.

mortenpi:
I was experimenting with `@testset` a bit and I came across a few small things:

* v0.4 doesn't have `@testset`s and [BaseTestNext.jl](https://github.com/JuliaCI/BaseTestNext.jl) should be used. Should we drop that into `test/REQUIRE`, even though it will then get installed on 0.5 as well (but not used because of in-code version checks)?

* Modules can't be defined within a `@testset` (not the top-level), but we use them quite extensively (testsets within a module are fine, of course). Basically, the question is how should we group our tests? I would seem convenient to have one master `@testset`, but that would be problematic because of that. Should we just have multiple "top-level" testsets, corresponding, I guess, to the top-level modules that we have?

Also, while we're at it, should we perhaps split `test/runtests.jl` into multiple files as well? The tests are a bit hard to read at the moment.

MichaelHatherly:
> Should we drop that into `test/REQUIRE`


Yes, no calls to `Pkg.add` within `runtests.jl`.

>  Basically, the question is how should we group our tests

Have `runtests.jl` wrap a bunch of `include`s that each contain the test modules and their inner `@testset`s? Not sure if that works though.

Just one toplevel testset if possible.

Yes, `runtests.jl` needs splitting. It's really just grown organically as needed until now. Definitely getting a bit big.

mortenpi:
Should this have a 0.4 milestone as well?

MichaelHatherly:
Yup, `0.4` sounds good to me.

PR #171, opened 2016-08-03
Implement HTML output

mortenpi:
Addresses #3. I am finally now opening the WIP PR to discuss this step by step.

Current issues / TODOs:
 
- [x] Duplication in the search index.
- [x] Internal links
- [x] Tests !
- [x] Docs !

Style issues:

- [x] Fonts
- [x] FontAwesome CDN
- [x] Custom admonitions style
- [x] Separate docstrings (`DocsNode`s) style-wise
- [x] Code block background (highlightjs conflict)
- [x] Search results style.

Separate PRs

- [x] Move HTMLWriter (#149)
- [x] `doc.user.page` parsing (#170)

--- 

Immediate question is the use of `@testset`. The whole HTML output will require quite a lot of new tests and since we have #154, I think it would make sense to start using `@testset`s for new tests. We might have to do some bikeshedding on exactly how we want to organize the tests though.


src/Utilities/MDFlatten.jl
```diff
@@ -0,0 +1,99 @@
+"""
+Proved the [`mdflatten`](@ref) function that can "flatten" Markdown objects into
```
MichaelHatherly: `Proved --> Provides`

src/Utilities/MDFlatten.jl
```diff
@@ -0,0 +1,99 @@
...
+"""
+function mdflatten(md)
+    io = IOBuffer()
+    mdflatten(io, md)
+    takebuf_string(io)
+end
+
+mdflatten(io, md) = mdflatten(io, md, md)
+mdflatten(io, md::MD, parent) = mdflatten(io, md.content, md)
+mdflatten(io, vec::Vector, parent) = map(x->mdflatten(io, x, parent), vec)
```
MichaelHatherly: Spaces before and after `->`.

src/Utilities/MDFlatten.jl
```diff
@@ -0,0 +1,99 @@
...
+    end
+end
+
+# Block level MD nodes
+mdflatten{N}(io, h::Header{N}, parent) = mdflatten(io, h.text, h)
+mdflatten(io, p::Paragraph, parent) = mdflatten(io, p.content, p)
+mdflatten(io, bq::BlockQuote, parent) = mdflatten(io, bq.content, bq)
+mdflatten(io, ::HorizontalRule, parent) = nothing
+function mdflatten(io, list::List, parent)
+    for (idx,li) in enumerate(list.items)
```
MichaelHatherly: Spaces after `,`. Same below as well.

src/Utilities/MDFlatten.jl
```diff
@@ -0,0 +1,99 @@
...
+        for (jdx,x) in enumerate(li)
+            mdflatten(io, x, list)
+            jdx != length(li) && print(io, '\n')
```
MichaelHatherly: Replace with `==` and `||` to follow other occurrences of this idiom in the code base. Same below as well.

Use `println(io)` instead of `print(io, '\n')`?

src/Utilities/MDFlatten.jl
```diff
@@ -0,0 +1,99 @@
...
+    end
+end
+
+# Inline nodes
+mdflatten(io, text::AbstractString, parent) = print(io, text)
+mdflatten(io, link::Link, parent) = mdflatten(io, link.text, link)
+mdflatten(io, b::Bold, parent) = mdflatten(io, b.text, b)
+mdflatten(io, i::Italic, parent) = mdflatten(io, i.text, i)
+mdflatten(io, i::Image, parent) = print(io, "[Image: $(i.alt)]")
+mdflatten(io, m::LaTeX, parent) = print(io, "``$(m.formula)``")
```
MichaelHatherly: If `mdflatten` is meant to strip formatting then the double backticks should probably not be printed, correct? Same with `Image` above and `Footnote` below, or is there a particular reason they need to formatting?

src/Utilities/MDFlatten.jl
```diff
@@ -0,0 +1,99 @@
...
+mdflatten(io, ::LineBreak, parent) = print(io, '\n')
+
+# Is both inline and block
+mdflatten(io, c::Code, parent) = print(io, c.code)
+
+# Special (inline) "node" -- due to JuliaMark's interpolations
+mdflatten(io, expr::Union{Symbol,Expr}, parent) = print(io, expr)
+
+
```
MichaelHatherly: One too many blank lines?

test/mdflatten.jl
```diff
@@ -0,0 +1,77 @@
...
+using Base.Markdown
+using Documenter.Utilities.MDFlatten
+
+@testset "MDFlatten" begin
+    @test mdflatten(Markdown.Paragraph("...")) == "..."
+    @test mdflatten(Markdown.Header{1}("...")) == "..."
+
+    # a simple test for blocks in top-level (each gets two newline appended to it)
+    @test mdflatten(Markdown.parse("# Test\nTest")) == "Test\n\nTest\n\n"
+    block_md = md"""
```
MichaelHatherly: Can `@md_str` be replaced with `Markdown.parse` in this and the cases below? I'd prefer only using `Markdown.parse` if possible.

test/mdflatten.jl
```diff
@@ -0,0 +1,77 @@
...
+
+    # test some inline blocks
+    @test mdflatten(md"`code` *em* normal **strong**") == "code em normal strong\n\n"
+    @test mdflatten(md"[link text *parsed*](link/itself/ignored)") == "link text parsed\n\n"
+    @test mdflatten(Markdown.parse("- a\n- b\n- c")) == "a\nb\nc\n\n"
+    @test mdflatten(Markdown.parse("A | B\n---|---\naa|bb\ncc | dd")) == "A B\naa bb\ncc dd\n\n"
+
+    # Math
+    @test mdflatten(md"$e=mc^2$") == "``e=mc^2``\n\n"
+    # TODO: backticks and blocks for math only in 0.5, i.e. these fail on 0.4
```
MichaelHatherly: `if VERSION < v"0.5.0-dev"`?

MichaelHatherly:
>  We might have to do some bikeshedding on exactly how we want to organize the tests though.

Just group them into logical units that try to test one particular method each if possible.

src/Utilities/MDFlatten.jl
mortenpi: Just to clarify -- the idiom basically is to minimize the number of `!`s?

Just asking because, my own go-to idiom would have been `&&` for normal control flow and `||` for errors (after having thought about it for a sec I think that's too strict in any case, although the "something or error()" part still seems appealing).

(Will make the change in any case, consistency is the key.)

As for `println`: I would maybe stick with `print`, with the reason being that I'm rather picky about what gets outputted when, how many newlines etc. Also with the change we'd have `print`s and `println`s mixed, and I think it would make it less readable. And `print("\n\n")` -- should that be `println('\n')` or `println(); println()`?

src/Utilities/MDFlatten.jl
mortenpi: This is a good one. In my head it strips most of the formatting, but leaves some semantic information, so that it would still be readable. The use case I have in mind is the `docs.julialang.org` search where would also display a small snippet on context, not only the page name (even though our search doesn't currently do that).

Hence, I would keep `(Image: <alt text>)`, so that if it would come up you could see that it's an image, etc. Similarly for footnotes, although maybe less necessary..

As for LaTeX -- that's the trickiest. TeX equation aren't really that searchable. Perhaps best to drop the equation and have replaced by `(Equation)`.

src/Utilities/MDFlatten.jl
MichaelHatherly: > Just to clarify -- the idiom basically is to minimize the number of `!`s?

Just to reduce the number of negations within an expression, i.e.

> `jdx` is equal to `length(li)` or `print(io, '\n')`

rather than

> `jdx` is not equal to `length(li)` and `print(io, '\n')`

---

> As for `println`

Ok, that's fine. Just so long as it's consistent.

src/Utilities/MDFlatten.jl
MichaelHatherly: Ok, maybe just remove equations for now. We can always add them back later.

mortenpi:
Ok, I've copied over and cleaned up the code, and it's rebase on the lastest master. There are still some details to work out, but generally I think it can be reviewed. There's also an example (under `test/html/`) that will build Documenter's docs with this.

The previous feedback on `MDFlatten` has been implemented.

A few things:

 - The `nightly` test fails because `BaseTestNext.jl` doesn't have a 0.6-installable version tagged (on master the version restriction was lifted, JuliaCI/BaseTestNext.jl#13). I am thinking that as a temporary workaround we could define our own dummy `@testset`, which just returns the `begin`/`end` block. 

 - I am now realizing that something like `Dict{Symbol,Any}` might be better instead of the `HTMLContext`. Not sure it's worth changing at the moment though.

 - Beyond the Julia code, css and js need some cleanup as well.

- Switched to `highlight.js` among others, so that we wouldn't have to deal with the Pygments dep.

- I'm bundling some JS code, I should check if there's a CDN option for them.

mortenpi:
Regarding the appveyor failure -- my guess is the use if `/`s in paths in the test file.

MichaelHatherly:
Done a quick scan over it, will do a more thorough review later tonight hopefully, but no major problems stood out to me, so :+1:.

> Not sure it's worth changing at the moment though.

Fine to leave as is for the moment.

> I am thinking that as a temporary workaround we could define our own dummy `@testset`, which just returns the begin/end block. 

Sounds reasonable to me.

> Switched to highlight.js among others, so that we wouldn't have to deal with the Pygments dep.

> I'm bundling some JS code, I should check if there's a CDN option for them.

👍, https://cdnjs.com/ probably has all the necessary libs I'd think.

MichaelHatherly:
> Regarding the appveyor failure -- my guess is the use if `/`s in paths in the test file.

Yeah, hardcoded paths on windows probably...

mortenpi:
cbd5fa7 should fix appveyor errors (basically `is_windows() && replace(/, \\)`). Should we do that for all input paths though? I would guess that if one does `source = "../docs"`, for example, it would also fail on windows.

mortenpi:
> fix appveyor errors (basically is_windows() && replace(/, \\))

Actually `normpath` should be a better solution (51cf717). And that's already called in other cases.

assets/html/normalize.css
```diff
@@ -0,0 +1,419 @@
...
+
+/**
+ * 1. Correct the inability to style clickable types in iOS and Safari.
+ * 2. Change font properties to `inherit` in Safari.
+ */
+
+::-webkit-file-upload-button {
+  -webkit-appearance: button; /* 1 */
+  font: inherit; /* 2 */
+}
```
MichaelHatherly: Does `normalize` need to be included, or can it be CDN'd?

src/Writers/HTMLWriter.jl
```diff
@@ -4,24 +4,472 @@ Provides the [`render`](@ref) methods to write the documentation as HTML files
...
+    )
+    open(Formats.extension(Formats.HTML, page.build), "w") do io
+        print(io, htmldoc)
+    end
+end
+
+function render_head(ctx, navnode)
+    @tags head meta link script title
+    src = get(navnode.page)
+    page_title = "$(mdflatten(pagetitle(ctx, navnode))) :: $(ctx.doc.user.sitename)"
```
MichaelHatherly: A `-` seems more common than `::` for the separator. I don't mind too much though and it might grown on me if you'd rather keep it as `::`.

src/Writers/HTMLWriter.jl
```diff
@@ -4,24 +4,472 @@ Provides the [`render`](@ref) methods to write the documentation as HTML files
...
+    end
+    domify(lb)
+end
+
+function domify(ctx, navnode, index::Documents.IndexNode)
+    @tags ul li a
+    lis = map(index.elements) do _
+        object, doc, page, mod, cat = _
+        url = string(page, "#", Utilities.slugify(object))
+        li(a[:href=>url]("$(object.binding)"))
```
MichaelHatherly: Wrap `"$(object.binding)"` in a `code` node?

MichaelHatherly:
Looks great so far! A few minor comments inline.

Also a couple of styling requests/suggestions (if you're looking for any of those 😆):

- Serif for the article body text (I'm a fan of Merriweather light) and a sans-serif for the headers and nav menu (perhaps Open Sans bold).
- I find Fantasque Sans Mono to be *really* easy to read and distinguish characters, or maybe Source Code Pro if that's not possible, or Ubuntu Mono.
- Can the main body be a bit wider. I've always found the Julia manual to be slightly too narrow. Also maybe it could be centered between the navmenu and the right hand side?
- The github icon (fontawesome icons?) isn't loading, just a blank box. (This is on Firefox.)
- The slight transparency of the search overlay makes it a little hard to read sometimes. Also, clicking on the links in the search results seems to require clicking exactly on the letters rather than just over the general area of the link. Perhaps that's related to the transparency?

Note, all these suggestions aren't *requirements*, just some things that would be good to try out and see how they look before committing to an official theme since this will (hopefully) become one of the most public things new users see when they start learning Julia.

MichaelHatherly:
Another one: should each `DocsNode` have some sort of "separator" or border to distinguish it from surrounding content and other docstrings?

src/Utilities/MDFlatten.jl
```diff
@@ -0,0 +1,98 @@
...
+    end
+end
+
+# Inline nodes
+mdflatten(io, text::AbstractString, parent) = print(io, text)
+mdflatten(io, link::Link, parent) = mdflatten(io, link.text, link)
+mdflatten(io, b::Bold, parent) = mdflatten(io, b.text, b)
+mdflatten(io, i::Italic, parent) = mdflatten(io, i.text, i)
+mdflatten(io, i::Image, parent) = print(io, "(Image: $(i.alt))")
+mdflatten(io, m::LaTeX, parent) = print(io, "(equation)")
```
MichaelHatherly: For page titles such as the "LaTeX syntax" one, it's printed as "(equation) syntax". Perhaps just strip punctuation characters, but leave the rest intact? Something like `replace(m.formula, r"[^\w\s]", "")` might be enough.

I don't think it's worth spending too much time on though, since page titles shouldn't really contain anything too complex and can be set explicitly in the `pages` keyword if needed.

src/Writers/HTMLWriter.jl
mortenpi: It seems to be way more common. I intended to change that anyway, but forgot. However, I now noticed that Github uses a  `·` (middle dot), which feels pretty nice as well to me, so I thought to give it a try.

MichaelHatherly: > so I thought to give it a try.

LGTM.

assets/html/style.css
```diff
@@ -0,0 +1,323 @@
...
+}
+
+code {
+    padding: 0.1em;
+}
+
+pre code {
+    background-color: initial;
+    padding: 0;
+}
```
MichaelHatherly: Code blocks that have highlight.js highlighting enabled seem to have two background colours, one in front of the other, with a small amount of padding between them.

2016-08-10-212035_4966x1080_scrot

Note the lighter grey border around the code block.

src/Utilities/MDFlatten.jl
mortenpi: Yes, that's better. I like it that it actually makes the equation somewhat searchable as well, which might be relevant for headers.

It might make sense retain a few math characters as well, so I would propose `r"[^()+\-*^=\w\s]"`, but nothing more complex.

I.e. `sin(x^2) = \\frac{n!}{m(n-m)!} \\LaTeX` -> `sin(x^2) = fracnm(n-m) LaTeX`.

mortenpi:
> if you're looking for any of those 😆

Absolutely! I am pretty much improvising, suggestions are very much welcome! And yes, I agree, it's worth spending some time and effort on the style, and try out different things.

> Can the main body be a bit wider. I've always found the Julia manual to be slightly too narrow. Also maybe it could be centered between the navmenu and the right hand side?

Yup, we can try slightly wider. The centering I was thinking about myself as well, I'll try to replicate what the current Julia manual does.

> The github icon (fontawesome icons?) isn't loading, just a blank box. (This is on Firefox.)

Good catch. I had it installed, so I didn't think about it. I CDNned it now, can you check if it works properly now? [Latest HTML build of Documenter here.](http://mortenpi.eu/Documenter.jl/html/)

Other fonts: I'll check out the suggestions. The current "web-safe" fonts could definitely use replacing. 

Regarding serif for the body -- it would be consistent with the current manual, yes, but it seems generally sans-serif fonts are considered more readable on screens and hence more appropriate for webpages. Someone also mentioned this on julia-users in relation to the manual and julialang.org (can't find it though).

Personally I don't actually have a preference, and Merriweather looks pretty nice. Current font is Georgia.

> The slight transparency of the search overlay makes it a little hard to read sometimes. Also, clicking on the links in the search results seems to require clicking exactly on the letters rather than just over the general area of the link. Perhaps that's related to the transparency?

The style of the search is still a prototype, will work on that. Now that you pointed it out I can also see the link clicking issue.. interesting.

I've also updated the OP checklists.

MichaelHatherly:
> but it seems generally sans-serif fonts are considered more readable on screens and hence more appropriate for webpages

Yeah, I'll concede that one, I think I'm probably just in a minority that doesn't have much of a problem with it :)

If we go with a sans-serif then Lato would be a decent choice perhaps. It's not too hard-edged like other sans fonts, I find. Making paragraph colour off-black would probably also help with readability. I believe the usual on most sites is `#333`.

For the monospace another choice would be DejaVu Sans Mono, which I believe is used in Juno (at least last I looked). That's probably the best choice, since it supports a wide range of Unicode compared to some others.

---

The logo is brilliant btw! I'd been wondering for a while what icons should be in the circles &ndash; those choices are great. Please do add it as the org logo if/when it's ready.

mortenpi:
I've added a small thingy to play with the fonts [of the HTML build](http://mortenpi.eu/Documenter.jl/html/), makes it easy to compare different options. Note that on Firefox it seems that you need to press enter for the change to register if you use the keyboard to select.

I must say, I quite like the feel of Lato for the text body - has this dense but clear feel to it.

> Making paragraph colour off-black

Agreed. Also links. Went with colors from the current manual.

MichaelHatherly:
> I've added a small thingy to play with the fonts

Neat!

> I must say, I quite like the feel of Lato for the text body - has this dense but clear feel to it.

Yes, I think we're in agreement about Lato. Reads really nicely.

It's surprising how similar Ubuntu Mono and Fantasque are. The `m`s in Ubuntu look slightly odd to me, but either of those two choices would be good with me. The others seem to be slightly too big for the surrounding text.

> Agreed. Also links. Went with colors from the current manual.

Link colour is good I think. Not too glaring, but still "link colour" so people aren't (hopefully) going to get confused.

Here's my vote:

- Lato and Fantasque (or Ubuntu);
- `#333` for font colour;
- `48em` width;
- normal font weight.

---

Any thoughts on the rounded corners of the search bar? It's the only significant element that's noticeably rounded. Does it look better with square corners? Not too bothered either way, just wanted to mention it.

Also, `<hr>`s have that little default "shadow". Should we follow the current style of the manual and have a `solid 1px` style for that?

src/Writers/HTMLWriter.jl
```diff
@@ -4,24 +4,471 @@ Provides the [`render`](@ref) methods to write the documentation as HTML files
...
+        print(io, htmldoc)
+    end
+end
+
+function render_head(ctx, navnode)
+    @tags head meta link script title
+    src = get(navnode.page)
+    page_title = "$(mdflatten(pagetitle(ctx, navnode))) · $(ctx.doc.user.sitename) documentation"
+    head(
+        meta[:charset=>"UTF-8"](),
```
MichaelHatherly: For `DOM` nodes with attributes that don't have children the `()` isn't needed since `[...]` will return a `Node` anyway.

src/Writers/HTMLWriter.jl
```diff
@@ -4,24 +4,471 @@ Provides the [`render`](@ref) methods to write the documentation as HTML files
...
+    scripts :: Vector{Compat.String}
+    documenter_js :: Compat.String
+    documenter_css :: Compat.String
+    search_index :: IOBuffer
+    search_index_js :: Compat.String
+end
+HTMLContext(doc) = HTMLContext(doc, "", [], "", "", IOBuffer(), "")
+
+"""
+Returns a page from the [`HTMLContext`](@ref).
```
MichaelHatherly: `Builder.CrossReferences` is complaining that `HTMLContext` isn't documented.

src/Writers/HTMLWriter.jl
```diff
@@ -4,24 +4,471 @@ Provides the [`render`](@ref) methods to write the documentation as HTML files
...
+# nothing to show for MetaNodes, so we just return an empty list
+domify(ctx, navnode, node::Documents.MetaNode) = Vector{DOM.Node}()
+
+
+# Utilities
+# ------------------------------------------------------------------------------
+
+"""
+Get the relative hyperlink between two [`Documents.NavNode`](@ref)s. Assumes that both
+[`Documents.NavNode`](@ref)s have an associated [`Page`](@ref) (i.e. `.page` is not null).
```
MichaelHatherly: `Page` needs to be `Documents.Page` I think since `Page` isn't defined within this module.

src/Writers/HTMLWriter.jl
```diff
@@ -4,24 +4,471 @@ Provides the [`render`](@ref) methods to write the documentation as HTML files
...
+
+    for navnode in doc.internal.navlist
+        render_page(ctx, navnode)
+    end
+
+    open(joinpath(doc.user.build, ctx.search_index_js), "w") do io
+        println(io, "var documenterSearchIndex = {\"docs\": [\n")
+        write(io, takebuf_string(ctx.search_index))
+        println(io, "]}")
+    end
```
mortenpi: Would be great but I think I ran into problems with the same-origin policy when I was trying to load the search index with ajax on a local build (some of the `.html` files are in a subdirectory). Hence the need to "load" it with a `<script>` tag.

src/Writers/HTMLWriter.jl
mortenpi: Good point. I do like the style aspect of explicitly constructing an empty `Node` instance though. In my head it looks uninitialized without the parens.

MichaelHatherly: We *could* add a 0-arg method that just passes the `Node` back to avoid additional allocations. I find the extra `()` redundant, but perhaps that just from knowing how it all works underneath.

MichaelHatherly:
> Now that you pointed it out I can also see the link clicking issue.. interesting.

Would using a separate `search.html` page instead of an overlay help to solve that? I guess we wouldn't get the realtime results in that case though.

MichaelHatherly:
Distinguishing docstrings from adjacent ones will probably need more than just the small header we currently use. Here's a possible suggestion using a thin border just on the left had side and also adding a bit of padding between them. Also with the `source` link moved to the left, since I think it's probably being missed by readers since its all the way over to the side.

2016-08-11-215729_4966x1080_scrot

Not *too* visually overpowering, but still distinctive enough, perhaps?

---

Might also be worth seeing whether making the entire `Binding` name the permalink rather than just the small `#` to it's left works. Would make it easier to click on probably.

mortenpi:
On docstrings: yup, the left border looks pretty good as well. My own proposal would've been to wrap it in a box:

screenshot from 2016-08-12 00-14-26

Links -- both good thoughts. Moving the source link to the left also "removes" the weird whitespace which you can see on this screencap.

MichaelHatherly:
> On docstrings: yup, the left border looks pretty good as well. My own proposal would've been to wrap it in a box:

How's the full box version look when there's a whole page of one-liner docstrings? Hopefully that won't be too common, but it might look quite busy with all the horizontal lines perhaps? Of course that would be incentive for people to write longer docstrings to make their docs look less cluttered.

mortenpi:
>  How's the full box version look when there's a whole page of one-liner docstrings?

cmp

[Also here](http://mortenpi.eu/documenter-html/dynamic/styles.html#Docstrings-1). The [Unit Testing](http://mortenpi.eu/documenter-html/stdlib/test.html) page might be a good example of the style in action as well.

To my eyes it doesn't look too busy (lines are quite faint), but still keeps things everything distinct. I feel that _maybe_ the left bar doesn't keep them distinct enough, but it's hard to say without actually using it for a while.

Also, one idea would be to keep some of these things we discuss undecided for a while, get some additional feedback from people (i.e. when we don't immediately feel what the right answer is).

---

Regarding `search.html`: I was thinking about doing a complete overlay somehow (perhaps mkdocs style). However, a separate page might have another benefit -- as you can perhaps see from the `julia-docs` build, a large manual is a bit slow due to the index construction. While it should be possible to improve on the current situation, there will always be quite a bit of computation on the index on every page load.

Hence it might make sense to only load the index when the user actually searches. It should be possible to do that on the same page as well, but the lazy approach here would be to only load `search_index.js` on `search.html`.

MichaelHatherly:
> To my eyes it doesn't look too busy (lines are quite faint), but still keeps things everything distinct.

Yes, the full boxes look fine to me as well. A single left bar definitely isn't distinct enough.

> Also, one idea would be to keep some of these things we discuss undecided for a while, get some additional feedback from people (i.e. when we don't immediately feel what the right answer is).

Quite right, I think the current styling that you've got at the moment is good for now. As you say, we can just take feed back over the next few weeks to tweak things slightly if needed, but not too much.

---

> However, a separate page might have another benefit -- as you can perhaps see from the `julia-docs` build, a large manual is a bit slow due to the index construction.

That's probably a good idea if it's not too much trouble to change over to.

---

My only other suggestions at this point are

- perhaps removing the `#` in the docstring headers now that the whole thing is a link;
- and turning all the headers on pages into permalinks as well, but without the blue link colour. Kind of like how Rust does.

Other than those (and some docs for the new features) this seems like it's in a good enough state now to prepare to merge soonish.

MichaelHatherly:
Just some other random thoughts worth noting:

On the question of customisation (like `mkdocs` allows with the `extra_css` and `extra_javascript`) I think we should probably, for the time being at least, not allow any form of customisation aside from a logo image.

This would force people who would like to make style changes to need to upstream their work and hopefully avoid fragmenting the official style that you've worked on.

Having the entire package ecosystem have a consistent style would be great I think, would make it much easier for readers to browse the docs for different packages without having to get used to totally different fonts/colours.

mortenpi:
Fixing the docstring and section headers -- good thoughts, done.

I've split up the search onto a separate page.. now the search js and index only load when the user goes there (still slow for the [Juliadocs build](http://mortenpi.eu/documenter-html/search.html?q=workflow), but at least the rest is usable :smile: ). Still needs some style fixes though.

---

Regarding customisation: I agree with the sentiment that we should keep the style uniform and changes be made upstream. However, I think having the ability to drop additional `.js` and `.css` files might still be convenient.. e.g. currently for dev purposes I am adding `debug.css` and `debug.js`.. would be nice if I wouldn't have to change the Julia code for that.

To do it without additional user options I was thinking about the following logic:

1. copy over `Documenter/assets/html/*` to `assets/`
2. copy over `docs/assets/*` to `assets/*
3. if there's a `assets/logo.{png,svg,jpg}`, use it 
4. if there's any `assets/*.js` (except `documenter.js`), compile them in as additional `<script>` tags

Note that in step 2 Documenter's default files could we overwritten, which is intentional.

As for the style fragmentation -- we could keep this feature undocumented for the time being (i.e. "this is a hack, not an official way of changing theme"). But in general I think people will be lazy, i.e. they will run with the default style anyway, especially since changing the theme here means writing your own CSS, not just swapping out the theme's name. Hence perhaps the fragmentation will not be such an issue, even if we allow it.

That being said, this could be a subsequent PR.

MichaelHatherly:
> I've split up the search onto a separate page

Much snappier now, thanks!

> Still needs some style fixes though.

Yes, could use some style tweaks. I've not got any suggestions at the moment.

---

> would be nice if I wouldn't have to change the Julia code for that.

True, that. The steps you outline sounds reasonable to me. Leaving undocumented
initially would be fine by me.

> That being said, this could be a subsequent PR.

Yes, definitely. Let's not go adding additional features to this PR.

MichaelHatherly:
Was testing this out on `DocStringExtensions` and noticed that with a very short navmenu the navmenu doesn't extend to the bottom of the page:

2016-08-12-231838_4966x1080_scrot

which looks slightly out of place I think. Is it possible to extend the right hand border and background colour of the navmenu all the way to the bottom of the screen even if there aren't enough items?

MichaelHatherly:
@mortenpi how's progress looking on the last few tasks, since the submission date is pretty much upon us? I think, for the moment, there's enough test coverage of the HTML rendering. All there really needs is a small section in the docs that mentions HTML rendering as an alternative to the standard markdown, but notes that it is not the default just yet (that will need to wait to 0.4 to give users a chance to prepare for it). Other than that I think we should merge this after it has had a rebase and perhaps a squash of some intermediate commits.

mortenpi:
I'll get on these things first thing tomorrow morning. Yes, I think after a few fixes to the search and navigation style (thanks for catching that, forgot to add a `min-height` after changing it) + squash & rebase, we could merge this.

I agree, it shouldn't be the default just yet. I would also like to write up a list of things that should still be done (and to get another PR in next week addressing these).

A note would probably do for now, I suppose? Any thoughts on where should we have it in the docs?

MichaelHatherly:
> Any thoughts on where should we have it in the docs?

At the end of the `man/guide` page perhaps? In an "alternative formats" section or something like that.

> I agree, it shouldn't be the default just yet. I would also like to write up a list of things that should still be done (and to get another PR in next week addressing these).

Sounds good to me. As far as I can tell from https://summerofcode.withgoogle.com/dashboard/timeline/ the "coding" part ends tomorrow. With "submissions" happening for the subsequent week.

As far as I'm concerned that'll just mean that the work you've done up to that point will be what gets evaluated and you're free to continue with anything extra after that (with the knowledge that it can't be part of the evaluation). And you're definitely more than welcome (and encouraged) to continue working on all this stuff.

I assume you've also had a read through https://developers.google.com/open-source/gsoc/help/work-product to check what it is you'll need to submit?

mortenpi:
I was always under the impression that it's fine to do final small changes on the final week as well, just that it should definitely be nearly finished and under review by then. In any case, not that imporant -- let's get this merged tomorrow and it's quite fine by me if the things after that are not formally part of GSoC anymore :smile: 

> I assume you've also had a read through https://developers.google.com/open-source/gsoc/help/work-product to check what it is you'll need to submit?

Yup, some time ago though.

MichaelHatherly:
> I was always under the impression that it's fine to do final small changes on the final week as well, just that it should definitely be nearly finished and under review by then.

:+1: fine by me.

mortenpi:
Addressed the things we discussed and squashed the old intermediate commits. I also put `BaseTestNext` back in since a 0.6-compatible version has been tagged.

One final issue cropped up though --  internal text links are not handled at the moment (i.e. `[...](page.md)` -type links don't have their extension changed). I think a simple solution would be to do some heuristics in the `mdconvert` -- if it doesn't have a protocol (i.e. `(.*)://(.*)` and ends with `.md`) then we'll pass it to `Formats.extension()`, otherwise we'll keep the link as is.

MichaelHatherly:
> internal text links are not handled at the moment

Should those kind just be `@ref` links to the first header on a page which should have the same kind of effect without needing to worry about paths, do we actually need to handle them differently? I'm not opposed to doing it, just wondering how necessary it is.

mortenpi:
Hmm, I had actually forgotten that `@ref` links work for headers as well. By quickly looking at the examples packages I saw that people use relative links (`page/`), absolute links and `@ref`s for inter-page links (one for each). However, it's the "MkDocs-way", so I feel that it would make sense to be compatible, for the time being. I do think that parsing these links should be done in an earlier step, but with #184 coming up, I didn't want to spend time on that (and they already work on MkDocs).

More generally though, if we'd decide to not support these links then we should probably print a warning to the user that there's an invalid relative link. But I must say that I feel that the `page.md#section` links seem more intuitive to me and they keep the source more explicit (by pointing to a concrete file). On the other hand, `@ref` links are very good when you want to convert text to a link directly, without specifying the link by hand, so they would definitely have a role as well.

MichaelHatherly:
You can also use the `@id <id>` and `@ref <id>` approach to get a more explicit autoref without needing to match the header text exactly. This also avoids needing to fix relative links if you happen to change the folder structure around.

My view is that we should just have one way to do this. `@ref` seems to cover everything I've needed to do. Also, if this isn't specific to the HTML rendering it should probably just be in another PR afterwards.

mortenpi:
Well, it's a HTML-specific issue in the sense that it would be a regression compared to the current MkDocs case. A more thorough solution would probably warrant a separate PR though.

While I do strongly agree that we shouldn't have multiple ways of doing the same thing, I also feel that the `[...](page.md)` has such an obvious meaning that it should "just work". As for breaking -- `@ref` links would break (especially for pages) when you decide to rephrase the heading (which to me seems roughly as likely as changing files or directories around).

I did swap to using `@ref` for the one link I introduced in the doc though.

MichaelHatherly:
> links would break (especially for pages) when you decide to rephrase the heading

Yes, `@ref` would break in that way, hence the `@id` feature.

I see what you mean about the HTML-specificness of it. If it's tested and documented then I guess it's fine to add now then rather than as a separate PR.

(Feel free to merge this when you're happy with it.)

mortenpi:
It's c998e2d. Not documented though, and I'm not sure whether it should be, if we're not sure if we wan't to endorse this style.

We could keep it as it is, so that the links like `[...](page.md)` are silently working as expected, and some time in the future we add some code to check the validity of the relative links (before the rendering phase). At that point we could then finally decide if we wish to support this linking syntax or thrown an error. Alternatively, I can also drop that commit and these links will just point to the raw URL, failing silently for now.

Since the current code is somewhat of a hack/workaround (i.e. it doesn't do existence checks etc), but should generally work (as long as anyone doesn't actually try to link a non-page `.md` file), I'm not actually sure which would be the right way to go.

MichaelHatherly:
That makes sense. Probably go with undocumented for now, and as you say we can work out whether we actually want it officially supported later.

> I'm not actually sure which would be the right way to go.

Looks alright for now, lets go with it as is.

mortenpi:
Will squash and merge then.

PR #180, opened 2016-08-06
Remove autogenerated methodlists

### LINKED ISSUE #163: Liberate autogenerated method lists
Currently the method lists that get generated don't appear within the actual docstrings, which seems a bit unfortunate. Would be good to be able to add them to any docstring independent of Documenter and have them show up within the REPL. Also, in some cases having the method list show up at all is overkill and would be good to be able to opt in/out on a per-docstring basis.

(@mortenpi, I'm busy beginning to working on a solution to this, and just wanted to give you a heads up on this, since I'm sure you'll want to have some input on this.)

Basic idea is to have:

```julia
"""
...

$__METHODS__

...
"""
func(x, y) = # ...
```

Expand `$__METHODS__` into

    ...

    # Methods

    ```
    func(x, y)
    ```
    defined at [<filename>](<giturl>).

    ...

Naming scheme of `__METHODS__`-type "abbreviations" still to be decided.

MichaelHatherly:
Closes #163. These are now available in the DocStringExtensions package.

---

Appologies for taking this out @mortenpi, I should have probably thought ahead when you were initially implementing it. Sorry about that.

I do think it's definitely a good idea to decouple this kind of autogenerated content from Documenter so that it can be viewed in the REPL and elsewhere, and not just trapped in Documenter's output. If there's any improvements you'd like to see with the methodlists in DocStringExtensions please do go ahead and add/suggest them if you'd like to.

### DIFF
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -117,12 +117,6 @@ immutable DocsNode
     anchor  :: Anchors.Anchor
     object  :: Utilities.Object
     page    :: Documents.Page
-    """
-    Vector of methods associated with this `DocsNode`. Being nulled means that
-    conceptually the `DocsNode` has no table of method (as opposed to having
-    an empty table).
-    """
-    methods :: Nullable{Vector{MethodNode}}
 end
 
 immutable DocsNodes
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -230,67 +230,6 @@ end
 # @docs
 # -----
 
-"""
-Returns a `Nullable{Vector{MethodNode}}` with the methods associated with the `object`.
-
-Null is returned if the object conceptually does not have a table of methods (e.g.
-modules). If instead the object should have a table of methods which just happens
-to be empty then it returns an empty vector.
-
-The methods are also filtered, with `MethodNode.visible` set to false for those
-methods not defined in this package. This allows the writers to choose how they
-want to treat trivial constructors, functions imported from `Base` and such.
-
-Due to differences between 0.4 and 0.5, only 0.5 is supported currently. On 0.4
-we always drop the methods table.
-"""
-function docsnode_methodlist end
-
-if VERSION > v"0.5-"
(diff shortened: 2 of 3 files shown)


mortenpi:
No problem! I agree that the way I did them (that is, effectively coupling them to the docstring) is a bit intrusive. I don't think you should have them in the docstring at all though -- a bit too noisy. Having the `METHODLIST` available though is fine, it puts the decision at the package author's discretion.

And I love the other features, such as auto-generating signatures into docstrings, which indeed should be separate from Documenter. :100:

I do think that in the long term we should automatically generate a separate page for every method, type etc. which we could then (over)load with information, including method lists. It becomes possible with our own HTML output. Maybe I should open an issue for this.

The role of docstring interpolation would then be for including docstrings into manual pages where they would add information. For example, I think [Unit test](http://docs.julialang.org/en/latest/stdlib/test/) section in Julia manual is really well done in this respect.

MichaelHatherly:
> such as auto-generating signatures into docstrings

Currently it's *slightly* limited since it won't display the argument types or default values. Whether it should actually do that or not is, I think, a rather difficult question to answer.

On the one hand it does mean that you can end up with signatures that are a bit cryptic since sometimes most of the information is encoded in the types rather than the argument names.

On the other hand, Julia's method signatures can sometimes be a bit overloaded with type information (such as `StridedMatrix` typealiases...) which isn't very useful to get a general overview of a method.

It's a fine balance to get the most useful signature possible, and will definitely need fine tuning.

> I do think that in the long term we should automatically generate a separate page for every method

We can definitely do that. Probably just by piecing together different `Abbreviations` into a custom `DocStr` object, parsing it, and then rendering it to a new page on it's own. At first perhaps we could just do that for objects that lack any handwritten docs, and later add it for all docs if that goes well.

> For example, I think Unit test section in Julia manual is really well done in this respect.

Very true, that's one of the nicest pages in the manual.

> Maybe I should open an issue for this.

Yes, please do.

Issue #168, opened 2016-08-02
Reduce number of different internal node types used


MichaelHatherly:
Speculative at the moment, but might be worth considering unifying all the different node types that we currently use to represent a document into a single (or just a couple) node &mdash; possibly based on the current `DOM` node, but probably with a little bit more metadata per node.

Also, instead of keeping `MD` objects hanging around all the way until rendering, might be good to convert them all to this unified node type as soon as possible. We would lose the built in markdown and html rendering the `MD` already has, but that's not much work to reimplement (better) here.

(Very low priority at the moment. Let's get the HTML stuff sorted first.)

mortenpi:
I have been thinking about something like this, that we'd have a clearer and easier to reason about structure for the document. It is currently a bit of a hotchpotch, so definitely sounds good.

I guess one question is the role of the `mkdocs` output in future. Going full-HTML internally and then back to markdown would be a bit of a headache.

Buy yes, I agree, let's get HTML in based on the current tree-ish thing first.

MichaelHatherly:
> tree-ish thing

🌳 

> Going full-HTML internally and then back to markdown would be a bit of a headache.

That did cross my mind. I think it would still be doable, though I'll need to actually try first.

>  clearer and easier to reason about structure for the document

Yes, that's the main motivation.

PR #184, opened 2016-08-08
Towards a unified doctree

MichaelHatherly:
Ref #168.

Definitely not in a mergable/usable state at the moment and we'll need to
wait until the dust has settled after the HTML work is merged and tagged
before actually considering this refactoring.

What it does:

  * A single "doctree" node type, `Node`. Mostly immutable apart from links
    to parents and children nodes. HTML tag attributes and additional metadata
    are stored in `ImmutableDict`s, much better for our use case than `Dict`.

  * Since `Node` is immutable to set the correct autolink for `at-ref` links
    a `Documenter.Renderers.attributes` function is provided that can be overloaded
    to map the correct URLs during the rendering stage based on data collected
    during earlier stages.

  * The current `Documents.Document` type will be stored instead as `.metadata` in
    the root `Node` of the doctree. Every `Node` will have access to it by traversing
    their `.parent` nodes. This doesn't appear to be much of a bottleneck, so adding
    a direct `.root` `Node` doesn't seem worth it.

  * This uses the `Selectors` module to avoid dynamic dispatch when rendering doctrees
    rather than the current approach where we're looping over `Vector{Any}`s. Everything
    important appears to be type-stable as far as I can tell so far.

  * Using a single unified doctree structure *should* make it possible to add new outputs
    by simply adding a new `Renderers/<format>` module that matches the layout of the
    others. Adding a new input format should just need a new function like `mdconvert`
    that builds a tree of `Node`s from files. (Not that there are any other parsers yet...)

  * Throughout the rest of the internal stages, i.e. not the input or output stages, we
    should just be able to work with `Node` (and `Tag`) objects. Extra formatting can just
    be stripped by `Renderers` that can't handle them. This should mean we can remove all
    the markdown-specific bits that are used internally and replace them with just the
    parts in `HTMLWriter` that are currently being worked on.

  * "navtrees" and next/prev pages should be pretty easy to construct directly from the
    doctree without needing too much work, though I've not yet looked too carefully into
    that.

@mortenpi, if you have any questions/suggestions/concerns then you're very welcome to raise them here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ os:
   - osx
 
 julia:
-  - 0.4
   - 0.5
   - nightly
 
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.4
+AbstractTrees
 Compat 0.8
 DocStringExtensions 0.1
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,5 @@
 environment:
   matrix:
-  - JULIAVERSION: "julialang/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
   - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
   - JULIAVERSION: "julianightlies/bin/winnt/x64/julia-latest-win64.exe"
 
--- a/src/DocTrees/DocTrees.jl
+++ b/src/DocTrees/DocTrees.jl
@@ -1,0 +1,202 @@
+module DocTrees
+
+import AbstractTrees
+
+#
+# Constants.
+#
+
+const NULL_TAG = Symbol("#NULL#")
+const TEXT_TAG = Symbol("#TEXT#")
+
+const EMPTY_TEXT = ""
+
+const HEADER_TAGS = (:h1, :h2, :h3, :h4, :h5, :h6)
+
+
+#
+# Metadata and Attributes.
+#
+
+immutable Metadata
+    dict::Base.ImmutableDict{Symbol, Any}
+
+    function Metadata(args...)
+        local dict = Base.ImmutableDict{Symbol, Any}()
+        for (k, v) in args
+            dict = Base.ImmutableDict{Symbol, Any}(dict, k, v)
+        end
+        return new(dict)
+    end
+end
+
+immutable Attributes
+    dict::Base.ImmutableDict{Symbol, String}
+
+    function Attributes(args...)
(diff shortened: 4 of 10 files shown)


mortenpi:
I didn't see this in the code, but the current `Page`s would also become `Node`s? So we'd have the root node, which would have a bunch of `Node(:PageTag)` (or something) as children?

Secondly, my impression is that currently we'd convert the document into basically the HTML DOM structure we have, including doing many of the rendering decisions? I am wondering if it would make more sense to keep the nodes more semantic and let the renderer decide more? I.e. in that sense I wouldn't deviate that much from what we have now (a `DocsNode` etc), just implement our own tags do that we'd have more control, could add metadata, parents etc (as opposed to mixing `Base.Markdown` objects with `Documenter.Documents` objects etc). Not sure what it would look like exactly, but basically I wouldn't add much additional formatting information (e.g. how exactly `DocsNode` divides up into a header `div` and a body `div`). Also not sure how well that would work with a unified `Node` type.

I am very much in favor of having access to the whole tree and the ability to store metadata in the tree (mostly in the root node, as you said). Would be so much nicer if you wouldn't have to pass stuff along all the time.

Just these quick questions for now, I should to give it some more thought at some point.

MichaelHatherly:
> I didn't see this in the code, but the current `Pages` would also become `Nodes`? So we'd have the root node, which would have a bunch of `Node(:PageTag)` (or something) as children?

Yes, they'd just be `Node`s. Probably named similar to the `#TEXT#` and `#NULL#` tags, probably just `#PAGE#`. We'd also need a `#SECTION#` tag as well that contains several `#PAGE#`s to handle the navmenu I'd think.

> Secondly, my impression is that currently we'd convert the document into basically the HTML DOM structure we have, including doing many of the rendering decisions?

Yes, pretty much that. Convert `MD` (and its friends) to `Node` objects as early as possible. Manipulate the resulting tree, i.e. expand all `@` nodes, do links, adding styling, and other things like that. Then pass the tree to the renderer and let it output as much of the tree as possible. For example: the markdown renderer would just ignore things like `div` tags. As far as I can tell we'd probably not loose any of the current markdown output.

> Not sure what it would look like exactly, but basically I wouldn't add much additional formatting information (e.g. how exactly DocsNode divides up into a header div and a body div). Also not sure how well that would work with a unified Node type.

I think it'll probably work fine with a single `Node` type, since something like the markdown parser just ignores things like classes and ids, and just prints out a best approximation to what the HTML one does.

> Would be so much nicer if you wouldn't have to pass stuff along all the time.

Definitely agree on that one.

Issue #191, opened 2016-08-15
Markdown.Image does not render properly


Hydrotoast:
Hi all,

I want to use `Markdown.Image` to render the results of `@eval` onto the page. I used [the example on the documentation](https://juliadocs.github.io/Documenter.jl/stable/lib/internals/expanders/#Documenter.Expanders.EvalBlocks). Unfortunately, the plots do not render properly.

My example project structure is as follows:

```
$ tree .
.
├── build
│   ├── assets
│   │   ├── Documenter.css
│   │   └── mathjaxhelper.js
│   ├── index.md
│   └── plot.svg
├── make.jl
├── mkdocs.yml
└── src
    └── index.md
```

The source for the example in `index.md`:

    ```@eval
    using PyPlot
    x = linspace(-π, π)
    y = sin(x)
    plot(x, y, color = "red")
    savefig("plot.svg")
    Markdown.Image("Plot", "plot.svg")
    ```

I compiled the document with `julia make.jl` and then built/served the documentation using

```
mkdocs build
mkdocs serve
```

Here is the rendering:

![documenter_markdown_image](https://cloud.githubusercontent.com/assets/246905/17653251/3314473a-6247-11e6-928e-7f45d32abf59.png)

MichaelHatherly:
Looks like `Markdown.Image` is an inline element rather than a block element to you'll need to wrap it in a `Markdown.Paragraph` to make it render correctly:

```@eval
using PyPlot
x = linspace(-π, π)
y = sin(x)
plot(x, y, color = "red")
savefig("plot.svg")
Markdown.Paragraph(Markdown.Image("Plot", "plot.svg"))
```

The docs should be updated, not sure why it didn't include `Paragraph`. Might actually be best to use `Markdown.parse("![Plot](plot.svg)")` since `Paragraph` and `Image` aren't actually exported from `Base`.

As the note in the [Syntax](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@eval-block-1) suggests it's usually best to try to use `@repl` and `@example`. Do they not fit with what you're trying to do?

The issue with `nothing` showing up when `@eval` block return `nothing` will be fixed shortly on `master` and I'll tag a new point release.

Hydrotoast:
Thanks @MichaelHatherly 

[...]

PR #193, opened 2016-08-15
Update Documentation referencing `Markdown.Image`

Hydrotoast:
Prefer to use `Markdown.Parse` instead of `Markdown.Image` in the documentation because `Markdown.Image` alone will not render images directly. See #191

### DIFF
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -423,8 +423,8 @@ evaluating each block the present working directory, `pwd`, is set to the direct
 `build` where the file will be written to.
 
 Also, instead of returning `nothing` in the example above we could have returned a new
-`Markdown.Image` object directly. This can be more appropriate when the filename is not
-known until evaluation of the block itself.
+`Markdown.MD` object through `Markdown.parse`. This can be more appropriate when the 
+filename is not known until evaluation of the block itself.
 
 !!! note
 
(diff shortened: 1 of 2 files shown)


MichaelHatherly:
Thanks! Should be `parse` rather than `Parse`, other than that looks good to me.

mortenpi:
Sorry for being a bit nitpicky, but saying "to return a `Markdown.Parse` object" is not entirely correct, since `Markdown.parse` is a function that returns a `Markdown.MD` object.

Perhaps we could phrase it "returned a `Markdown.parse` return value directly" or, based on [the comment](https://github.com/JuliaDocs/Documenter.jl/issues/191#issuecomment-239740735), "returned a Markdown block object (node?) or a `Markdown.MD` object directly"?

MichaelHatherly:
Good catch, thanks for spotting that. Your wording sounds good to me.

Hydrotoast:
Addressed both issues raised by @MichaelHatherly and @mortenpi. Thanks for reviewing the PR.

MichaelHatherly:
Thanks for the contribution @Hydrotoast.

MichaelHatherly:
@Hydrotoast could you squash the five commits down to a single one? Thanks.

Hydrotoast:
@MichaelHatherly squashed as requested

PR #195, opened 2016-08-15
Add `Deps.none`

MichaelHatherly:
This adds a dependency installer that doesn't install anything. Needed since the default behaviour of `deploydocs` is to install `mkdocs` and `pygments`.

@mortenpi any suggestions for a better name or alternative approach?


mortenpi:
One possibility would be to simply pass `deps = nothing` to `deploydocs`.

MichaelHatherly:
That sounds good. One less export is better probably.

PR #197, opened 2016-08-16
Allow `make` to be bypassed

MichaelHatherly:
Adds check for `make === nothing` to allow for extra build sets, i.e. `mkdocs`, to be skipped since HTML rendering doesn't require extra build steps.

Also removes `cleandir` call since `target` should either not exist or be the the directory where `makedocs` has output files to.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -305,9 +305,12 @@ function deploydocs(;
         # Change to the root directory and try to deploy the docs.
         cd(root) do
             Utilities.log("setting up target directory.")
-            Utilities.cleandir(target)
-            Utilities.log("building documentation.")
-            make()
+            isdir(target) || mkpath(target)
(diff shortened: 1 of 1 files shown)


src/Documenter.jl
```diff
@@ -305,9 +305,12 @@ function deploydocs(;
         # Change to the root directory and try to deploy the docs.
         cd(root) do
             Utilities.log("setting up target directory.")
-            Utilities.cleandir(target)
-            Utilities.log("building documentation.")
-            make()
+            isdir(target) || mkpath(target)
+            # Run extra build steps defined in `make` if required.
+            if make !== nothing
+                Utilities.log("building documentation.")
```
mortenpi: Maybe have the logging be "running extra build steps." or something along those lines? "building documentation." could sound a bit like it's running `makedocs`.

PR #204, opened 2016-08-17
Switch the docs to HTML

### LINKED ISSUE #194: Use HTML renderer for docs
Now that #171 is merged it would be good to start using it for our own docs. @mortenpi would you like to do the honours?

mortenpi:
Resolves #194.


docs/make.jl
```diff
@@ -3,9 +3,47 @@ using Documenter
 makedocs(
     modules = [Documenter],
     clean   = false,
+    format   = Documenter.Formats.HTML,
+    sitename = "Documenter.jl",
+    pages    = [( # Compat: parens needed for 0.4 syntax compat
```
MichaelHatherly: I think using `Any[` should work shouldn't it?

mortenpi:
Yup, `Any` looks better. I would still keep one of the `Compat` notes, since on 0.5+ only the `Any` syntax should probably be discouraged.

mortenpi:
As I was adding a test to build the `test/examples` with HTML, I discovered that I had missed the `EvalNode` in `HTMLWriter`. So added the necessary function and also the test.

MichaelHatherly:
LGTM, merge when ready.

> I discovered that I had missed the `EvalNode` in `HTMLWriter`. So added the necessary function and also the test.

:+1:

Issue #208, opened 2016-08-18
Overwriting Julia's internal docstring objects


mortenpi:
Documenter overwrites Julia's internal docstring objects when resolving `@ref` links for docstrings. This means that when when running `makedocs` for the second time the cross-references in docstrings are not resolved again.

Currently this breaks cross-references within docstrings in our docs, since we build the docs in `runtests.jl` and a Markdown build runs before, resolving the links to `.md` files.

I have a feeling that we probably shouldn't be modifying Julia's docstrings. A simple `deepcopy` around `parsedocs` didn't work though, since at some point the docstrings refer to a `Module`, which cannot be copied. #184 should fix this though, so I am not sure if there's any point in doing anything complicated for this at the moment.

As for fixing our docs: there are a few workarounds.

MichaelHatherly:
Yes, we shouldn't really be mutating anything, hence the push for making as much of #184 immutable. Since it's only a problem with running `makedocs` multiple times in the same Julia session we probably won't see users running into to just yet.

PR #209, opened 2016-08-18
Fix mdconvert's link parsing

mortenpi:
The link parsing wasn't working properly -- splitting was limited to zero splits, so it never actually split at a #, and it also hid the call to a non-existent `second` function.

This will also work around the broken links of #208, since the `mdconvert` gets called on the resolved docstring links, fixing the wrong extensions.


MichaelHatherly:
LGTM.

PR #213, opened 2016-08-19
Fix spelling typos

MichaelHatherly:


tkelman:
also L222 and the top of src/Deps.jl

PR #217, opened 2016-08-20
Update HTML docs

mortenpi:
Fix some mistakes and add a few details to the "Output formats" section. The style and code block changes are required so that the updated docs would render correctly.

The rendered version, including the style of `<h1>` in admonitions vs a top-level `<h2>`:

screenshot from 2016-08-20 09-58-31

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -590,6 +590,15 @@ end
 # mdconvert
 # ------------------------------------------------------------------------------
 
+md_block_nodes = [Markdown.MD, Markdown.BlockQuote, Markdown.List]
+if isdefined(Base.Markdown, :Admonition) push!(md_block_nodes, Markdown.Admonition) end
+
+"""
+[`MDBlockContext`](@ref) is a union of all the Markdown nodes whose children should
+be blocks. It can be used to dispatch on all the block-context nodes at once.
+"""
+typealias MDBlockContext Union{md_block_nodes...}
+
 """
 Convert a markdown object to a `DOM.Node` object.
 
@@ -607,7 +616,7 @@ mdconvert(b::Markdown.BlockQuote, parent) = Tag(:blockquote)(mdconvert(b.content
 
 mdconvert(b::Markdown.Bold, parent) = Tag(:strong)(mdconvert(b.text, parent))
 
-function mdconvert(c::Markdown.Code, parent::Markdown.MD)
+function mdconvert(c::Markdown.Code, parent::MDBlockContext)
     @tags pre code
     language = isempty(c.language) ? "none" : c.language
     pre(code[".language-$(language)"](c.code))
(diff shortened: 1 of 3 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -667,6 +669,7 @@ if isdefined(Base.Markdown, :Admonition)
             div[".admonition-text"](mdconvert(a.content, a))
         )
     end
+    mdconvert(c::Markdown.Code, parent::Markdown.Admonition) = mdconvert_codeblock(c, parent)
 end
```
MichaelHatherly: The other toplevel blocks should also have this change, not just `Admonition`s and `MD`s, correct? `BlockQuote`, `Footnote`, and `List` elements as well?

MichaelHatherly:
CSS changes look good to me.

MichaelHatherly:
Should be backported to `0.3.1`?

mortenpi:
> The other toplevel blocks should also have this change, not just `Admonition`s and `MD`s, correct? `BlockQuote`, `Footnote`, and `List` elements as well?

Yes, indeed, forgot those. `Footnote` is a bit tricky though.. I am wrapping it in a `<span>`, i.e. I am assuming that footnotes should only have inline content. Perhaps we should make that change in a separate PR.

Also, `LaTeX` has a context-dependent dispatch for the same thing as well. I decided to go with a `Union` type to dispatch on, instead of writing a separate signature for each block node.

I did notice that parsing lists can run into issues if the whitespace is inconsistent, i.e. in this

    - A list

       ```julia
       associated
       ```

      ```math
      \sin(x)
      ```

     - ```math
       \sin(x)
       ```
some of it gets wrapped in paragraphs, the math block doesn't get parsed etc. Not sure if they're bugs though actually -- with consistent whitespace everything seems to work.

> Should be backported to `0.3.1`?

Probably, I suppose. I guess we should do bugfixes for the HTML part on `0.3` branch as well, since (hopefully) some people will switch their docs over.

MichaelHatherly:
> I am assuming that footnotes should only have inline content

`Footnote` should be able to contain any nested block content, just like `Admonition`s.

> Not sure if they're bugs though actually -- with consistent whitespace everything seems to work.

Technically the commonmark "spec" is quite lenient with indentation, but I've gone with requiring consistent indentation so that all docs will be just a little bit more uniform (which I think is a good thing).

 > Probably, I suppose. I guess we should do bugfixes for the HTML part on 0.3 branch as well, since (hopefully) some people will switch their docs over.

Yes, wherever we can backport non-invasive things we probably should. It can wait 'til this is merged to `master` though.

mortenpi:
> > I am assuming that footnotes should only have inline content

> `Footnote` should be able to contain any nested block content, just like `Admonition`s.

Yup, I agree that we should change that. I'd do that in a separate PR though, since it also requires changes to the rendering code.

Shall we merge?

PR #218, opened 2016-08-20
HTML formatting: add max-width 100%

r9y9:
Without this I cannot show big images in a page. I believe this simple change should be harmless.


MichaelHatherly:
Seems reasonable to me. @mortenpi look fine to you?

mortenpi:
Yes, a good thought! Thanks for the contribution!

PR #219, opened 2016-08-20
Disable codecov inline comments

MichaelHatherly:


mortenpi:
These are the Codecov PR posts, right?

MichaelHatherly:
Yes, apparently that's the way to disable the inline comments. It should still show a status with the other CI things though.

MichaelHatherly:
Might as well also backport it.

PR #231, opened 2016-08-23
Enable AppVeyor on release-* branches

mortenpi:
I added a `^` as well, so that it wouldn't match things like `mp/release-appveyor`.

If it works it should be forward-ported to `master` as well.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ environment:
 branches:
   only:
     - master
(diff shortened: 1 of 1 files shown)


mortenpi:
> If it works

It seems it does, an AppVeyor build started after I had opened the PR.

MichaelHatherly:
LGTM.

PR #234, opened 2016-08-23
Fix index ordering when extension changes

### LINKED ISSUE #226: HTML contents not sorted correctly
I just tried out the HTML renderer on 0.3.0 and I found that the `@contents` blocks are not sorted as I wrote them, e.g.

    ```@contents
    Pages = [
        "algorithms.md",
        "graph-types.md",
        "graphs-builtin.md",
        "interface.md",
        ]
    Depth = 1
    ```

will show the links in the order provided with mkdocs, but it will show "graph-types", "graphs-builtin", "interface", "algorithms" with the HTML renderer.

BTW Is there any way to only show the contents from depth 2 for a page, i.e. excluding the page title as redundant?

mortenpi:
Currently the extensions are not changed for the pagesmap variables,
but are for the paths in `IndexNode`s and `ContentsNode`s. This
confuses the sort when the extension is not `.md` anymore. With this
patch we shift the changing of extensions to the writing step, making
the IndexNode and ContentsNode sort format-independent.

Fixes #226.

---

@carlobaldassi, would it be possible for you to build your docs with this branch and confirm that #226 is fixed?

### DIFF
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -286,7 +286,6 @@ function populate!(index::IndexNode, document::Document)
         # Include *all* signatures, whether they are `Union{}` or not.
         cat  = Symbol(lowercase(Utilities.doccat(object.binding, Union{})))
         if _isvalid(page, index.pages) && _isvalid(mod, index.modules) && _isvalid(cat, index.order)
-            page = Formats.extension(document.user.format, page)
             push!(index.elements, (object, doc, page, mod, cat))
         end
     end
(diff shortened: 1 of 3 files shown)


carlobaldassi:
Yes, it fixes the issue, thanks.

MichaelHatherly:
LGTM.

MichaelHatherly:
Seems like travis is being quite slow tonight. @mortenpi we should probably both stop creating our PR branches on `JuliaDocs/Documenter` and instead use our personal forks to avoid "double building" with both `travis-ci/pr` and `travis-ci/push`.

Issue #237, opened 2016-08-24
Travis can't push Documenter documentation to github


richardreeve:
I've tried to follow your [instructions](https://juliadocs.github.io/Documenter.jl/latest/man/hosting.html#SSH-Deploy-Keys-1) to get Travis to automatically update my documentation on github for my [Diversity](https://github.com/richardreeve/Diversity.jl) package. I can generate the docs on my laptop, I have an (old) [gh-pages](https://github.com/richardreeve/Diversity.jl/tree/gh-pages) branch on github, and I have the .documenter.enc file on github, and Travis has the appropriate keys on its site. However, when I hook Documenter into Travis via the [.travis.yml](https://github.com/richardreeve/Diversity.jl/blob/master/.travis.yml) file and commit something to master, the [report](https://travis-ci.org/richardreeve/Diversity.jl/jobs/154544145) says that it generates the docs, but then doesn't have permission to push the gh-pages branch:

```
INFO    -  Building documentation to directory: /home/travis/.julia/v0.5/Diversity/doc/site 
Documenter: pushing new documentation to remote: github.com/richardreeve/Diversity.jl.git:gh-pages.
Agent pid 3817
Identity added: /home/travis/.julia/v0.5/Diversity/doc/.documenter (/home/travis/.julia/v0.5/Diversity/doc/.documenter)
Initialized empty Git repository in /tmp/tmp6i91w7/.git/
Warning: Permanently added the RSA host key for IP address '192.30.253.113' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
```

Any suggestions would be welcome...

EDIT: I appreciate this is unlikely to be directly a Documenter issue, but I have tried to follow Documenter's own online docs, without any noticeable success...

MichaelHatherly:
Have you copied the private key displayed by `Travis.genkeys` to your GitHub repo settings page?

richardreeve:
No, I don't see that in the docs... is there any way of doing it retrospectively?

MichaelHatherly:
Have you still got the terminal open that you called `genkeys` in?

MichaelHatherly:
If not then you'll just need to redo the `genkeys` step again and replace the `.documenter.enc` file.

richardreeve:
[...]

PR #238, opened 2016-08-24
Clarifying instructions for Travis to automatically publish the documentation online

richardreeve:
As a follow-up to #237, here is some additional documentation along with REPL output explaining how to get Travis to talk to github. It also deletes the docs on using github security tokens.


MichaelHatherly:
Excellent, thanks!

> Feel free to ignore it if it's not what you're looking for.

No, this is pretty much what I had in mind :+1:. Only bit that could probably be clarified is the `[...]` that is being used to represent the ssh keys. Can you add a note after its first appearance just to explain that it isn't the actual output users should expect.

richardreeve:
Okay, I've removed one of the `[...]` sections and explained the second.

MichaelHatherly:
Thanks, LGTM.

PR #239, opened 2016-08-24
Add first-level headings to navmenu

### LINKED ISSUE #225: HTMLWriter: sidebar doesn't reflect contents?
I may be doing something wrong, but the sidebar contents is only picking up the level 2 (`##`) headings from `src/index.md`, whereas the @contents macro is doing the right thing. Presumably they're compiled differently somehow, but shouldn't they be more similar? Or have I omitted some stages (I've just got a single `index.md` file, for example.)

screen shot 2016-08-22 at 13 35 42

mortenpi:
Add first-level headings to the navigation menu. The only exception is
when the very first block of a page is a first-level heading -- in this
case that particular heading is ignored. This is consistent with how
the headings are usually used, with the first heading being the page
title.

The `<li>` tags of the first-level headers have the class `.toplevel`
attached to them, so that we can style them in the CSS.

Fixes #225.

---

@cormullion and @MichaelHatherly: What I am wondering is whether the behavior that a `<h1>` is a page title if and only if it is the very first block of a page is reasonable. Also, should we go with this, I guess we should also change the [`pagetitle` function](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Writers/HTMLWriter.jl#L560-L565), so that only the very first block would be used to automagically determine a page title for the nav.menu?

We could also maybe use a small doc. note on this.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -2,6 +2,17 @@
 Provides the [`render`](@ref) methods to write the documentation as HTML files
 (`MIME"text/html"`).
 
+# Page outline
+
+The [`HTMLWriter`](@ref) makes use of the page outline that is determined by the
+headings. It is assumed that if the very first block of a page is a level 1 heading,
+then it is intended as the page title. This has two consequences:
+
+1. It is then used to automatically determine the page title in the navigation menu
+   and in the `<title>` tag, unless specified in the `.pages` option.
+2. If the first heading is interpreted as being the page title, it is not displayed
+   in the navigation sidebar.
+
 # Default and custom assets
 
 Documenter copies all files under the source directory (e.g. `/docs/src/`) over
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -557,12 +557,17 @@ end
 Tries to guess the page title by looking at the `<h1>` headers and returns the
 header contents as a `Nullable` (nulled if the algorithm was unable to determine
 the header).
+
+It is assumed that the intended page title can only be guessed if the very first
+block of the page is `<h1>` heading. If there is something before the first heading
+or the first heading is a lower level heading, then the return value is nulled.
 """
 function pagetitle(page::Documents.Page)
-    for e in page.elements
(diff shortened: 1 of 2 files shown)


cormullion:
In documentation circles, a small amount of selective redundancy is considered acceptable (more so than in coding circles)... So it wouldn't be too bad if the first heading was also used elsewhere. But this solution is also fine -- particularly if added to the really fine manual.

src/Writers/HTMLWriter.jl
```diff
@@ -576,15 +577,22 @@ function pagetitle(ctx, navnode::Documents.NavNode)
...
+the page also happens to be a level 1 heading. In that case it is assumed to be
+the page title and is dropped from the list of subsections.
 """
 function collect_subsections(page::Documents.Page)
-    # TODO: Should probably be replaced by a proper outline algorithm.
-    #       Currently we ignore the case when there are multiple h1-s.
-    hs = filter(e -> isa(e, Base.Markdown.Header{2}), page.elements)
-    map(hs) do e
+    hs = filter(enumerate(page.elements)) do _
+        i, e = _
```
MichaelHatherly: Can `i` and `e` have more descriptive names?

MichaelHatherly:
> What I am wondering is whether the behavior that a `<h1>` is a page title if and only if it is the very first block of a page is reasonable.

Yes, that seems reasonable to me.

mortenpi:
I think we should treat the page outline consistently, both in the sidebar and when guessing a page title. So I've added a change that we'd only use the very first `<h1>` block to guess the page title (and also a small doc. note).

Assuming it looks good, I'd propose that we would only backport the first commit -- the second commit changes the behavior of Documenter.

MichaelHatherly:
> the second commit changes the behavior of Documenter.

It only changes the behaviour of the HTML output, correct? If so then we *can* get away with changing it in a point release since we've explicitly stated that it is experimental at this point. Probably better to change it sooner rather than later anyway, otherwise people will end up relying on it if we leave it too long.

MichaelHatherly:
(Also, LGTM, so merge whenever you're ready @mortenpi.)

PR #249, opened 2016-09-02
add a setup block, fixes #244

### LINKED ISSUE #244: Hide whole code block for being displayed?
I'm doing a lot of the following in Gadfly's docs:

    ```@example 1
    using Gadfly # hide
    using RDatasets # hide
    srand(1234) # hide
    ```

    ```@example 1
    df = dataset("ggplot2", "diamonds") # hide
    p = plot(df, # hide
             x = :Price, color = :Cut, # hide
	    	 Stat.histogram, # hide
		     Geom.bar) # hide
    p_rendered = deepcopy(p) # hide
    ```

Is it possible to hide the whole code block instead of having to add `#hide` everywhere?

tlnagy:
working towards dcjones/Gadfly.jl#879

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -517,21 +517,17 @@ function Selectors.runner(::Type{SetupBlocks}, x, page, doc)
     name = Utilities.getmatch(matched, 1)
     sym  = isempty(name) ? gensym("ex-") : Symbol("ex-", name)
     mod  = get!(page.globals.meta, sym, Module(sym))::Module
-    # Evaluate the code block. We redirect STDOUT/STDERR to `buffer`.
-    result, buffer = nothing, IOBuffer()
-    for (ex, str) in Utilities.parseblock(x.code, doc, page)
-        (value, success, backtrace, text) = Utilities.withoutput() do
-            cd(dirname(page.build)) do
-                eval(mod, :(ans = $(eval(mod, ex))))
-            end
+
+    # Evaluate whole @setup block at once instead of piecewise
+    page.mapping[x] =
+    try
+        cd(dirname(page.build)) do
+            eval(mod, :(include_string($(x.code))))
         end
-        result = value
-        print(buffer, text)
-        if !success
-            Utilities.warn(page.source, "failed to run code block.\n\n$(value)")
-            page.mapping[x] = x
-            return
-        end
+        Markdown.MD([])
+    catch err
+        Utilities.warn(page.source, "failed to run `@setup` block.\n\n$(err)")
+        x
     end
     # ... and finally map the original code block to the newly generated ones.
     page.mapping[x] = Markdown.MD([])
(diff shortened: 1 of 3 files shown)


MichaelHatherly:
Thanks, will review tomorrow morning.

src/Expanders.jl
```diff
@@ -499,6 +507,36 @@ function Selectors.runner(::Type{REPLBlocks}, x, page, doc)
...
+            end
+        end
+        result = value
+        print(buffer, text)
+        if !success
+            Utilities.warn(page.source, "failed to run code block.\n\n$(value)")
+            page.mapping[x] = x
+            return
+        end
+    end
```
MichaelHatherly: Since a `@setup` block doesn't need to do the parsing piece-by-piece it would probably be better to just do

```julia
eval(mod, :(include_string($(x.code))))
```

test/examples/src/lib/functions.md
```diff
@@ -116,6 +116,16 @@ d = 1
...
+Test setup function
+
+```@setup testsetup
+w = 5
+```
+
+```@example testsetup
+@assert w === 5
+```
+
```
MichaelHatherly: `@setup` should also work with `@repl` blocks, correct? Might be worth adding a `@repl` block test as well.

src/Expanders.jl
MichaelHatherly: (Will still need the `cd` and `try` blocks though.)

tlnagy: I'm not 100% sure what the code is doing here, to be honest.

MichaelHatherly: So the code in a `@setup` block doesn't need to be parsed expression by expression (using `Utilities.parseblock`) like it does for `@example` and `@repl` blocks since we don't need to evaluate the `ans` variable for each toplevel expression. We also don't need to redirect `STDOUT` and `STDERR` streams (using `Utilities.withoutput`) to `buffer`.

This all just means we can write something like

```julia
page.mapping[x] =
    try
        cd(dirname(page.build)) do
            eval(mod, :(include_string($(x.code))))
        end
        Markdown.MD([])
    catch err
        Utilities.warn(page.source, "failed to run `@setup` block.\n\n$(err)")
        x
    end
```

instead of evaluating the block piece-by-piece in a `for` loop like it is done for `@example` and `@repl`. (`page.mapping` is used to associate the original markdown AST node with the object that should actually be printed in the final documentation.)

tlnagy:
Thanks @MichaelHatherly for your feedback. I've clean up my PR with your suggestions.

MichaelHatherly:
Thanks, lgtm.

MichaelHatherly:
(Since this is a new feature it can't really be added to the `release-0.3` branch, so the `0.4` milestone has been pushed back to `0.5` and `@setup` blocks will be part of a new `0.4` release to be made shortly.)

tlnagy:
Sure, that works for me.

PR #258, opened 2016-09-11
Update examples.md, Luxor now registered

cormullion:
Managed to sneak Luxor past Tony "[Heimdall](https://en.wikipedia.org/wiki/Heimdall_(comics))" Kelman and got it registered... :)


MichaelHatherly:
Thanks.

PR #259, opened 2016-09-12
add OhMyREPL.jl to examples

KristofferC:


MichaelHatherly:
Thanks, #258 caused a conflict here. Could you do a quick rebase?

MichaelHatherly:
btw, OhMyREPL is brilliant!

KristofferC:
Thanks a lot. Comments like that gives me inspiration to work more on it :). Documenter is also really awesome and getting better and better all the time.

KristofferC:
I just made a new one, it was faster than having to fix my remote and rebase etc.

PR #271, opened 2016-09-15
Fixup navmenu headers

MichaelHatherly:
@mortenpi, look alright to you?

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -592,23 +592,23 @@ function pagetitle(ctx, navnode::Documents.NavNode)
 end
 
 """
-Returns an ordered list of tuples `(toplevel, anchor, text)`, corresponding to
-level 1 and 2 headings on the page. The only exception is if the first block on
-the page also happens to be a level 1 heading. In that case it is assumed to be
-the page title and is dropped from the list of subsections.
+Returns an ordered list of tuples, `(toplevel, anchor, text)`, corresponding to level 1 and 2
+headings on the `page`. Note that if the first header on the `page` is a level 1 header then
+it is not included -- it is assumed to be the page title and so does not need to be included
+in the navigation menu twice.
 """
 function collect_subsections(page::Documents.Page)
-    hs = filter(enumerate(page.elements)) do _
-        idx, element = _
-        isa(element, Base.Markdown.Header) || return false # ignore non-headers
-        (idx == 1) && (Utilities.header_level(element) == 1) && return false # if first elem. <h1> => ignored
-        Utilities.header_level(element) <= 2 # only let <h1> and <h2> through
+    local sections = []
+    for element in page.elements
+        if isa(element, Base.Markdown.Header) && Utilities.header_level(element) < 3
+            local toplevel = Utilities.header_level(element) === 1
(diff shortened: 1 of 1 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -598,17 +598,17 @@ the page also happens to be a level 1 heading. In that case it is assumed to be
...
-        Utilities.header_level(element) <= 2 # only let <h1> and <h2> through
-    end
-    map(hs) do _
-        idx, heading = _
-        anchor = page.mapping[heading]
-        (Utilities.header_level(heading) == 1), "#$(anchor.id)-$(anchor.nth)", heading.text
+    local sections = []
+    for element in page.elements
+        if isa(element, Base.Markdown.Header) && Utilities.header_level(element) < 3
+            local toplevel = Utilities.header_level(element) === 1
```
mortenpi: Just a small question -- wouldn't `==` be preferred when comparing numbers?

src/Writers/HTMLWriter.jl
```diff
@@ -598,17 +598,17 @@ the page also happens to be a level 1 heading. In that case it is assumed to be
 the page title and is dropped from the list of subsections.
 """
 function collect_subsections(page::Documents.Page)
```
mortenpi: Actually, the docstring should probably be updated, to make it clear that it's the first level 1 heading that gets dropped.

src/Writers/HTMLWriter.jl
MichaelHatherly: Shouldn't make a difference here. They should either both be `Int32` or `Int64`, not one of each (which would be a problem.)

src/Writers/HTMLWriter.jl
MichaelHatherly: +1

src/Writers/HTMLWriter.jl
mortenpi: Agreed that it should work in this instance, but I meant as a general style rule.

MichaelHatherly: I prefer more restrictive comparisons where possible, just a habit really.

MichaelHatherly: Saves some silly typos sometimes: `true == 1` vs `true === 1`.

mortenpi: Can break in other cases though (`BigInt(1) === BigInt(1)`). But I guess it comes down to a judgement call, specific to the situation. In this case, since we know we should be dealing with bitstypes, being more strict is probably better, as you pointed out.

mortenpi:
One small thing still is that if we have e.g. `<h2>, <h1>, ...`, then the `<h1>` still get's used as a page title, but is also present in the sidebar. Probably fine though -- I don't think we should drop the `<h1>` if there are `<h2>`s before it. But I am wondering whether we should use the first `<h2>` then as the page title or not.

MichaelHatherly:
> But I am wondering whether we should use the first `<h2>` then as the page title or not.

Let's just keep the rule as simple as possible: first h1 header is the title. I don't think we should worry too much about supporting pages where there are `h2`s in front of `h1`s, it's probably not a common case.

PR #283, opened 2016-09-22
Mh/fix path and list

MichaelHatherly:


mortenpi:
Whoops :smile:.

The `fieldtype` case could perhaps be `<: Integer`? Although I'm not sure that we'd win much.. and lose performance?

LGTM by the way.

MichaelHatherly:
Since it's toplevel code there shouldn't be any performance loss. `== Int` should be fine since the field itself is `Int`.

Thanks for reviewing, not sure what I was thinking with the original condition there.

mortenpi:
> I was thinking

I actually thought it was my code (hence the whoops), but I now saw that this line was actually added later.

MichaelHatherly:
Yeah, was my error, not yours :laughing:.

PR #286, opened 2016-09-23
Add linkchecking doccheck

### LINKED ISSUE #285: linkcheck
useful feature of sphinx, detects urls and checks if links are dead or redirects. This is run as part of the base Julia release-candidate target. Should be doable via Requests.

MichaelHatherly:
Checks whether external URLs found in documentation exist.

Fixes #285.

---

Interestingly, this check picked up an incorrect link in the examples page for `DifferentialEquations.jl`.

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -25,6 +25,7 @@ defined in the `modules` keyword passed to [`Documenter.makedocs`](@ref).
 Prints out the name of each object that has not had its docs spliced into the document.
 """
 function missingdocs(doc::Documents.Document)
+    println(" > checking for missing docstrings.")
     bindings = allbindings(doc.user.modules)
     for object in keys(doc.internal.objects)
         if haskey(bindings, object.binding)
@@ -100,6 +101,7 @@ the document generation when an error is thrown. Use `doctest = false` keyword i
 """
 function doctest(doc::Documents.Document)
     if doc.user.doctest
+        println(" > running doctests.")
         for (src, page) in doc.internal.pages
             empty!(page.globals.meta)
             for element in page.elements
@@ -353,6 +355,7 @@ end
 
 if isdefined(Base.Markdown, :Footnote)
     function footnotes(doc::Documents.Document)
+        println(" > checking footnote links.")
         # A mapping of footnote ids to a tuple counter of how many footnote references and
         # footnote bodies have been found.
         #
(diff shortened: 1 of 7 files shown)


tkelman:
That's shockingly little code. Nice!

src/DocChecks.jl
```diff
@@ -405,4 +408,38 @@ else
...
+end
+
+function linkcheck(link::Base.Markdown.Link, doc::Documents.Document)
+    if !(link in doc.internal.locallinks)
+        local good = try
+            Requests.statuscode(Requests.head(link.url)) == 200
+        catch err
+            false
+        end
+        print_with_color(good ? :green : :red, good ? "     " : "   X ", link.url, "\n")
```
mortenpi: Could be helpful to print the `statuscode` here as well.

mortenpi: Also, the perhaps the links could be more indented. My first thought was that the gray page names also were links, but internal ones and therefore got ignored or something:

```
 > checking external URLs:
   - man/contributing.md
   - lib/internals/writers.md
   - man/guide.md
     http://www.mkdocs.org/
     https://github.com/JuliaDocs/Documenter.jl/issues
   - lib/public.md
   - man/examples.md
     https://juliafinance.github.io/Currencies.jl/latest/
   X http://chrisrackauckas.github.io/DifferentialEquations.jl/latest/
```

This I think looks more clear:

```
 > checking external URLs:
   - man/contributing.md
   - lib/internals/writers.md
   - man/guide.md
       http://www.mkdocs.org/
       https://github.com/JuliaDocs/Documenter.jl/issues
   - lib/public.md
   - man/examples.md
       https://juliafinance.github.io/Currencies.jl/latest/
     X http://chrisrackauckas.github.io/DifferentialEquations.jl/latest/
```

MichaelHatherly: With latest change it now prints as

```
...
 > checking external URLs:
   - lib/internals/dom.md
   - lib/internals.md
   - lib/internals/builder.md
   - man/contributing.md
   - lib/internals/writers.md
   - man/guide.md
         http://www.mkdocs.org/
         https://github.com/JuliaDocs/Documenter.jl/issues
   - lib/public.md
   - lib/internals/mdflatten.md
   - man/examples.md
         http://juliacontrol.github.io/ControlSystems.jl/latest/
         https://juliafinance.github.io/Currencies.jl/latest/
     404 http://chrisrackauckas.github.io/DifferentialEquations.jl/latest/
         https://juliadocs.github.io/Documenter.jl/latest
...
```

mortenpi: Maybe one more character? In case there are a lot of status codes then it becomes a bit hard to read (e.g. if all links of a page fail etc.)
```
...
 > checking external URLs:
   - lib/internals/dom.md
   - lib/internals.md
   - lib/internals/builder.md
   - man/contributing.md
     200 http://juliacontrol.github.io/ControlSystems.jl/latest/
     200 https://juliafinance.github.io/Currencies.jl/latest/
   - lib/internals/writers.md
   - man/guide.md
      200 http://www.mkdocs.org/
      200 https://github.com/JuliaDocs/Documenter.jl/issues
   - lib/public.md
   - lib/internals/mdflatten.md
```

Also, I am wondering if, for consistency, we should also show the `200`s. Of course, in that case, if there's no color, there's nothing indicating whether the link was considered a success or not. OTOH, without it the user doesn't really know why the link was considered a success (unless they go and read the code and see that we're checking for the status code 200).

MichaelHatherly: I'll add an extra char.

> Also, I am wondering if, for consistency, we should also show the 200s.

The intention was to make it inconsistent so that it's easier to tell (without colour) whether the link was good.

> why the link was considered a success

If is succeeds then they shouldn't really need to be concerned about why (I'd assume), only when it fails.

mortenpi: > If is succeeds then they shouldn't really need to be concerned about why (I'd assume), only when it fails.

Yup, that's fair enough. LGTM then!

MichaelHatherly: Thanks for the review.

PR #288, opened 2016-09-26
LaTeX rendering

MichaelHatherly:


MichaelHatherly:
A link to the generated pdf for Documenter's own docs: https://dl.dropboxusercontent.com/u/14501513/document.pdf.

There's very little customisation at the moment, so it looks pretty bland at the moment. Bikeshedding welcome.

src/Writers/LaTeXWriter.jl
```diff
@@ -13,12 +13,415 @@ import ...Documenter:
...
+function latex(io::IO, contents::Documents.ContentsNode, page, doc)
+    local depth = 1
+    local needs_end = false
+    println(io, "\\begin{itemize}")
+    for (count, path, anchor) in contents.elements
+        local header = anchor.object
+        local level = Utilities.header_level(header)
+        local id = string(hash(string(anchor.id, "-", anchor.nth)))
+        level < depth && println(io, "\\end{itemize}")
+        level > depth && (println(io, "\\begin{itemize}"); needs_end = true)
```
hayd: I don't recall if julia exports it, but IIRC there are some neat withXXX latex methods, which wrap begins/ends in a slightly cleaner way than printing (though it requires more indenting).

hayd:
Woah! You're having travis build the pdf, that's brilliant!

So, to tweak this ourselves (to try and change the configs), we just run make.jl on this branch? (and edit assets/latex/documenter.sty ?)

src/Writers/LaTeXWriter.jl
MichaelHatherly: I'm using `with*` for some further down, but this particular case needed it done manually.

MichaelHatherly:
> Woah! You're having travis build the pdf, that's brilliant!

Trying to, though at the moment it looks as if installing all the needed texlive packages is going to take too long to do on every build.

MichaelHatherly:
> So, to tweak this ourselves (to try and change the configs), we just run make.jl on this branch? (and edit assets/latex/documenter.sty ?)

Yes, pretty much, though you'll need a *full* texlive install.

bicycle1885:
Great! I really wanted this kind of PDF manuals for a long time.

Some bikesheddings from my preferences:
* Narrower margins.
* Using pale colors instead of vivid ones.
* Sans-serif fonts.

These are the styles used in [vignettes of Bioconductor project](https://www.bioconductor.org/help/package-vignettes/). Here is an example: https://bioconductor.org/packages/release/bioc/vignettes/GenomicRanges/inst/doc/GenomicRangesHOWTOs.pdf.

Also, I think it would be good to use [minted](https://www.ctan.org/pkg/minted?lang=en) to highlight code, which internally calls [pygments](http://pygments.org/) and the next release of it will be shipped with my patches to enhance Julia's highlighter.

MichaelHatherly:
Yeah, narrower margins and sans-serif would probably look better. The lack of minted was due to wanting to avoid too many deps on travis, though it looks unlikely for the time being that we'll actually be able to build on travis in a timely manner, so I'll definitely switch over to minted.

Bioconductor docs look nice, thanks, I'll see whether I can match that style.

MichaelHatherly:
And a preliminary version of the main manual: https://dl.dropboxusercontent.com/u/14501513/julia.pdf. Still some missing chars here and there that need fixing, but seems to compile alright.

hayd:
A few comments:

- Chapter 1 seems superfluous (considering the contents page before).
- Chapter vs section labelling isn't quite right (all chapter titles are labelled `X.1`, rather than `X`) and there's no `X.1.1`, `X.2` etc.
- Are the figs missing?

This is going to be a huge upgrade from [RTD](https://media.readthedocs.org/pdf/julia/latest/julia.pdf)!

MichaelHatherly:
Yes, chapter/section structure is a bit odd at the moment. Probably needs to have `\part`s for each toplevel pagegroup in `pages` and bump all the `\*section`s up one level. That would probably better match the current manual layout I think.

> Are the figs missing?

Which figs?

hayd:
I may have been confused, perhaps it's just some strange whitespace:

http://docs.julialang.org/en/release-0.5/manual/mathematical-operations/?highlight=trigonometric#trigonometric-and-hyperbolic-functions

screen shot

MichaelHatherly:
Hmm, looks like the tables in that section are just rearranging themselves however they like. I think there's probably some incantation to force tables to go to the right spot...

MichaelHatherly:
For the moment this will be regarded as an experimental output type (no docs about it yet), and will be subject to change. It currently manages to build both Documenter's docs and the main Julia Manual, so I'm calling it good enough for now.

There won't be support for building PDFs on Travis for the moment, since the dep requirements are a bit much to impose on a free service like that. PDF generation should be done locally and manually uploaded by the package author.

hayd:
Final comments / for the future:
- `bclogo` is an option for those warn/note blocks.
- wrapping manual docstrings like is done in html (in a rectangle)
- the following _sparse_ table is slightly broken:

sparse table

MichaelHatherly:
> `bclogo` is an option for those warn/note blocks.

Those look good. Thanks for the link.

> wrapping manual docstrings like is done in html (in a rectangle)

Yes, we could do, though there's a few places with overly long lines which I felt look slightly better without a line cutting through them. Though the proper solution there is just to sort out the markdown source file rather.

> the following sparse table is slightly broken:

Those tables aren't great, though without manually setting column widths it's pretty hit-and-miss as to whether `tabulary` manages to get it right. If you've got any suggestions for other table packages that handle things better I'm definitely happy to switch over.

tkelman:
> There won't be support for building PDFs on Travis for the moment, since the dep requirements are a bit much to impose on a free service like that.

Other than a latex distribution, what's needed? That's pretty big but it's not that outrageous to do on Travis.

MichaelHatherly:
Current set of deps would be:

- full TexLive install, we're building with LuaLaTeX;
- minted (pygments and python);
- Lato Sans and Fantasque Sans Mono (non-default system fonts everywhere I think) to give us a consistent look between the HTML and PDF docs.

For the Manual, we'll also need some fallback fonts for some of the more obscure characters. I've been testing it out so far with the DejaVu Sans and Noto Sans CJK families, which appears to cover everything that's needed.

Probably wouldn't be too bad. The most amount of time spent is actually on highlighting code blocks with minted. The Manual has been taking about 10 min to compile because of that.

tkelman:
Interesting, I guess we'll eventually have to write a higher-performance replacement for pygments too then! Is that 10 minutes for the base manual, or Documenter's own docs? If it's the former, then that's probably a worst case.

MichaelHatherly:
> I guess we'll eventually have to write a higher-performance replacement for pygments too then!

Yes, definitely something worth writing.

> Is that 10 minutes for the base manual, or Documenter's own docs? If it's the former, then that's probably a worst case.

10 minutes for the Julia Manual, so, yes, most likely worst case (by quite a bit).

ChrisRackauckas:
Are there docs for how to do this?

MichaelHatherly:
Should just need to switch `format = :html` with `format = :latex`. You'll also need to make sure you've got all the TeX packages and other bits and pieces needed:

- `latexmk`;
- `lualatex`;
- `minted` (and `pygments` by extension);
- `Lato` and `Fantasque Sans Mono` fonts;
- a couple of other packages listed in [the style file](https://github.com/JuliaDocs/Documenter.jl/blob/master/assets/latex/documenter.sty).

PR #294, opened 2016-09-28
Use at-testset for tests

mortenpi:
A pretty big bunch of changes attempting to organize tests in a more readable manner. Addresses #154. I see that #288 doesn't have any tests yet, so I'm hopefully not conflicting with it. Motivation comes from #282 -- I wanted to make a few small changes to `examples/` for it.

#### A summary of the changes:
- Move to `@testset`.
- Allow building `examples/` separately (useful for testing changes to HTML writer etc.).
- No toplevel stuff (modules, types) within a `@testset` macro, so we need to define modules/types outside it. Hence `navnode.jl` depends on `faketypes.jl` and `examples/tests.jl` (and due to an `include`, `runtests.jl`) depends on `examples/make.jl`.
- Style change: indent module contents (except if it is the only thing in the file) for readability.
- Consider `pages/` to be part of the `examples/` (but also use them + the modules `Mod` and `AutoDocs`, defined now in `examples/make.jl`, in `runtests.jl` for unit tests).

Things are quite interlinked here and so this ended up being this one big commit at the moment. I am considering splitting it up into several commits to make it easier to review (most importantly to make sure that we're not losing or making useless any of the existing tests). However a discussion on what the final result should look like would be useful before I get to that.

#### TODO:
- [x] Fix 0.4 tests.
- [x] A couple of tests failed due to the module structure changes (marked with `TODO`s).
- [x] Add more explanatory comments.


MichaelHatherly:
Thanks for doing this! I'll take a closer look once I've gotten the LaTeX stuff out of the way.

docs/src/man/contributing.md
```diff
@@ -21,6 +21,7 @@ Follow the style of the surrounding text when making changes. When adding new fe
 ### Julia
 
   * 4-space indentation;
+  * modules spanning entire files should not be indented, but modules that have surrounding code should;
```
mortenpi: Added a note here about the modules. Thoughts on the phrasing very welcome.

test/docsystem.jl
```diff
@@ -0,0 +1,85 @@
...
+if VERSION >= v"0.5.0-dev+7720"
+    using Base.Test
+else
+    using BaseTestNext
+    const Test = BaseTestNext
+end
+
+import Documenter: Documenter, DocSystem
+
+const alias_of_getdocs = DocSystem.getdocs # NOTE: won't get docstrings if in a @testset
```
mortenpi: I had to move this here, since the test on line 66 started failing (`d_4` had 0 docstrings).

test/formats/latex.jl
```diff
@@ -0,0 +1,63 @@
...
+    const Test = BaseTestNext
+end
+
+using Documenter
+
+# Documenter package docs
+info("Building Documenter's docs with LaTeX.")
+const Documenter_root = normpath(joinpath(dirname(@__FILE__), "..", "..", "docs"))
+doc = makedocs(
+    debug = true,
```
mortenpi: Added a debug flag here.

test/utilities.jl
```diff
@@ -0,0 +1,60 @@
...
+    "Documenter unit tests."
+    Base.length(::T) = 1
+end
+
+@testset "Utilities" begin
+    let doc = @doc(length)
+        a = Documenter.Utilities.filterdocs(doc, Set{Module}())
+        b = Documenter.Utilities.filterdocs(doc, Set{Module}([UnitTests]))
+        c = Documenter.Utilities.filterdocs(doc, Set{Module}([Base]))
+        d = Documenter.Utilities.filterdocs(doc, Set{Module}([UtilitiesTests]))
```
mortenpi: Had to change the module name here.

mortenpi:
Appveyor failures due to path handling, I'll have to fix those. I was wondering why we have the `joinpath`s here and there :laughing:.

Best practice wise though -- should we construct all paths with `joinpath` or just when we actually use them to compare paths etc.? It doesn't seem necessary for `include`s, for example.

It is nice to see the testsets in action though:
```
Test Summary: | Pass  Fail  Error  Total
  Documenter  |  278     2      1    281
  Utilities   |   17                  17
  NavNode     |   19                  19
  DocSystem   |   34                  34
  DOM         |  132                 132
  MDFlatten   |   20                  20
  Examples    |   35     2      1     38
  Markdown    |   34     2      1     37
  HTML        |    1                   1
  Markdown    |   20                  20
  LaTeX       |    1                   1
```

docs/src/man/contributing.md
MichaelHatherly: Wording sounds good to me.

MichaelHatherly:
> Best practice wise though -- should we construct all paths with joinpath or just when we actually use them to compare paths etc.? It doesn't seem necessary for includes, for example.

I'd say just always use `joinpath` when creating paths.

> It is nice to see the testsets in action though:

Awesome!

mortenpi:
I fixed the paths in `examples/tests.jl` that were causing issues. So, other than the things below, it should be good to go.

However, I do feel that for `include`s (e.g. in `runtests.jl`) it should be sufficient to use the forward slash. Julia handles these paths correctly on Windows and it seems more readable:
`include("examples/make.jl")`
vs
`include(joinpath(dirname(@__FILE__), "examples", "make.jl"))`.

`runtests.jl` is slightly inconsistent at the moment though -- the old ones use `joinpath`, but the new ones I added don't. I should probably make it consistent though, so I would like to drop the `joinpath`s.

I also took the liberty of dropping branch restrictions from `appveyor.yml` -- I couldn't build custom branches on my fork and I don't see a particular need for them. If we want to restrict the branches built in the `JuliaDocs` repo we could do that in the AppVeyor instead. However, if you think we should keep it there, I can drop the commit.

mortenpi:
Thanks for reviewing! I removed the few remaining `include(joinpath(...))` cases for consistency. Will merge once the tests pass.

MichaelHatherly:
Thanks for sorting this one out @mortenpi!

PR #302, opened 2016-10-06
fix partial documentation

abhijithch:
When all the exported items in a module are included in the `src/index.md` and in if actual only few are documented, then the output in `build/index.d` only consists of the list. It does not create the documentation for the functions which are documented. This PR fixes this.


MichaelHatherly:
Thanks! Could do with a test of some kind if, that's possible?

abhijithch:
To test this what I did was, 

1. Got some exported APIs from package `HypothesisTests` into a .md placed in `src` folder file as shown below,

```
## API reference for HypothesisTests

```@docs
ChisqTest
MannWhitneyUTest
MultinomialLRT
SignedRankTest
ci
pvalue
testname
.```
```

2. Created a `make.jl` file,

```
using Documenter, HypothesisTests

makedocs(
    modules = [HypothesisTests],
    clean = true,
    format = Documenter.Formats.Markdown,
    sitename = "HypothesisTests.jl",
    linkcheck = !("skiplinks" in ARGS),
    pages = Any[ # Compat: `Any` for 0.4 compat
        "Home" => "HypothesisTests.md"
    ]
)
```

3. Ran `julia make.jl`, which generated the `build` folder with output `HypothesisTests.md` with documentation for only `ci`, `pvalue`, `testname`. 

Let me know if I can present this in any better way.

abhijithch:
The tests seem to be failing with this command, 

```
curl -A "$CURL_USER_AGENT" -s -L --retry 7 'https://s3.amazonaws.com/julialang/bin/linux/x64/0.5/julia-0.5-latest-linux-x86_64.tar.gz' | tar -C ~/julia -x -z --strip-components=1 -f -
```

The error message is,

```
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```

This seems not to be related to any of the code, I think.

MichaelHatherly:
OK, I must have been confused by the initial description of this, sorry about that.

This isn't a bug at all, if any of the names listed in an `@docs` block are not documented then the whole block will be ignored &mdash; this is by design.

The fix is to remove those names that aren't documented. If you want to just include every available doc then use a `@autodocs` block instead.

mdpradeep:
Just to clarify the issue. If there are 5 exported APIs in a package, let's say 3 of the APIs have some documentation and the remaining 2 are not documented. In this case the expectation is that document is generated for at least the 3 APIs that have been documented. The current behaviour is that it generates no document for any of the 5 APIs. With this PR documentation will be generated for the 3 APIs that have been documented.

MichaelHatherly:
> 3 of the APIs have some documentation and the remaining 2 are not documented

Undocumented bindings should not be included in `@docs` blocks.

This makes it more likely that you'll catch any typos since the entire block will "fail" rather than just ignoring the typo which would be more difficult to notice.

It also makes it clearer to anyone reading the `.md` files that *all* the names listed in a block are actually documented rather than having to either build the docs or search through the `.jl` files manually to find out.

FWIW warnings are printed, i.e.

```
...
 !! No docs found for 'h'. [src/index.md]
...
```

for any name that is not documented. If those are not being printed then that's a bug, but otherwise the `@docs` blocks are behaving as expected.

The fix is just to remove any undocumented bindings until they *are* documented, then add them back.

abhijithch:
The context I have here is we are automating the process of documentation for a list of packages. In this process the `.md` reference files are created automatically, so it is not possible to filter out the names of functions which do not have docs. Is there any right way of achieving the same?

MichaelHatherly:
Thanks for clarifying that. Any autogenerated `.md` files should just use [`@autodocs`](https://juliadocs.github.io/Documenter.jl/latest/man/syntax.html#@autodocs-block-1) which will do all the filtering and sorting for you, and only add documented names.

If you're able to provide the script you're using to do this autogeneration I'd be happy to have a look at it to see if there's anything more appropriate than `@autodocs`.

We could also potentially add a more programmatic API to Documenter to allow you to write autogenerated docs in a more straightforward way than currently. That'll be a little way off once I've got the spare time.

abhijithch:
The script testapi.txt(please change the extension to .jl) which generates the `.md` file is attached. Create a directory `temp` in the same level of directory structure. You can run this in a REPL, and call the function `docgen("DataFrames", "temp/")`. This should create a `build` directory in `temp/docs` which should have the generated `DataFrames.md`. 

Case 1 : `@autodocs` and original version of `Documenter.jl`
In the function `docgen`, when I have `@autodocs` in line 26 instead of `@docs`, the output does not contain documentation for any of the functions. 

Case 2: `@docs` and this PR version of `Documenter.jl`
The generated output `DataFrames.md` contains documentation for functions for which it is available. 

[testapi.txt](https://github.com/JuliaDocs/Documenter.jl/files/526407/testapi.txt)

MichaelHatherly:
Thanks for the script @abhijithch, much clearer now.

`@autodocs` does not take a list of names like `@docs` does. It takes key/value pairs:

    ```@autodocs
    Modules = [Mod1, Mod2] # Any number of modules here.
    # other options...
    ```

Please see the docs for `@autodocs`, which I've linked to in https://github.com/JuliaDocs/Documenter.jl/pull/302#issuecomment-253197369 for the full description the available options other than `Modules`.

abhijithch:
Thank you @MichaelHatherly for helping me solve my problem!

abhijithch:
One last question related to this, can we detect modules that do not have any inline documentation and there by do not generate a output md file for this. It so happens that for packages where there are no documentation, I am getting only the header info in my final output.

MichaelHatherly:
You could check in your script whether `isempty(Base.Docs.meta(my_module)) == true`, which means nothing is documented in the module. (That's an unexported function but I'm not likely to be removing/changing it.)

abhijithch:
This perfectly solved my requirement. Thanks.

PR #304, opened 2016-10-06
deploydocs: check that `julia` is a string, repo matches repo_slug

timholy:
I spent some time debugging a puzzling failure-to-deploy, before finally noticing that in my `make.jl` I had specified `julia = 0.5` rather than `julia = "0.5"`.


timholy:
I should note that it wasn't clear to me whether the repo test should be a warning or an error; I decided on a warning, but an error would obviously be much easier to test for.

MichaelHatherly:
Thanks. I'd be fine with both being `error`s, rather than just `warn`ing the user about the mismatch.

timholy:
Done.

MichaelHatherly:
Thanks.

PR #311, opened 2016-10-11
Remove the Requests dep, use curl instead

MichaelHatherly:


src/DocChecks.jl
```diff
@@ -423,41 +423,47 @@ end
...
+                local regex = r"^HTTP/1.1 (\d+) "
+                ismatch(regex, res) ? parse(Int, match(regex, res).captures[1]) : ERROR
+            catch err
+                ERROR
+            end
         print(" "^5)
-        if status in (0, 200)
-            print(" "^5)
+        if status < 400
+            print_with_color(:green, " $(status) ", link.url, "\n")
```
tkelman: distinct sense of deja vu here, but redirect statuses should at least warn

MichaelHatherly: https://github.com/JuliaDocs/Documenter.jl/pull/313#discussion_r83139126 was in the backport to `release-0.5` while this one is against `master`, i.e. it's the same commit.

PR #313, opened 2016-10-11
Backport Requests removal

MichaelHatherly:
Since `linkcheck` isn't documented it's fine to change behaviour and backport it. (Much too soon to have to release a `0.6`!)

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -423,10 +423,10 @@ end
 # Link Checks.
 # ------------
 
-import Requests
+hascurl() = (try; success(`curl --version`); catch err; false; end)
 
 function linkcheck(doc::Documents.Document)
-    if doc.user.linkcheck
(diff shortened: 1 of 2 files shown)


src/DocChecks.jl
```diff
@@ -423,41 +423,47 @@ end
...
+                local res = readstring(`curl -sI $(link.url)`)
+                local regex = r"^HTTP/1.1 (\d+) "
+                ismatch(regex, res) ? parse(Int, match(regex, res).captures[1]) : ERROR
+            catch err
+                ERROR
+            end
         print(" "^5)
-        if status in (0, 200)
-            print(" "^5)
+        if status < 400
```
tkelman: redirects should at least warn

MichaelHatherly: Yes, warnings for the 300s would be good to add.

PR #322, opened 2016-10-17
Disable 'strict' mode

MichaelHatherly:
`curl` on Linux nightly is failing for `https` addresses, probably due
to the current SSL bug on Julia `master`. Use non-strict mode for the
docs build to allow it to finish.


mortenpi:
Doesn't `curl` come from the Travis image? You don't happen to have a ref to the SSL bug?

MichaelHatherly:
Possibly related to the CA certificate problems in #18693, but since they seem to be working fine locally now I'm hoping it's just going to be a transient nightly problem that will go away once a new one is put up. I'll check back on it in a few days to see.

PR #326, opened 2016-10-17
Print redirected URLs in yellow

MichaelHatherly:
Just prints redirected URLs in a different colour. @tkelman did you have something a little more distinctive in mind?

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -447,22 +447,36 @@ end
 
 function linkcheck(link::Base.Markdown.Link, doc::Documents.Document)
     if !haskey(doc.internal.locallinks, link)
-        const ERROR = typemax(Int)
-        local status =
-            try
-                local res = readstring(`curl -sI $(link.url)`)
-                local regex = r"^HTTP/1.1 (\d+) "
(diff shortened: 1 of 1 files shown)


tkelman:
it would be good to display the redirected location too, so people can easily see what they should change the link to

MichaelHatherly:
Sounds good to me.

src/DocChecks.jl
```diff
@@ -447,22 +447,36 @@ end
...
+        if ismatch(STATUS_REGEX, result)
+            local status = parse(Int, match(STATUS_REGEX, result).captures[1])
+            if status < 300
+                print_with_color(:green, INDENT, "$(status) ", link.url, "\n")
+            elseif status < 400
+                local LOCATION_REGEX = r"^Location: (.+)$"m
+                if ismatch(LOCATION_REGEX, result)
+                    local location = strip(match(LOCATION_REGEX, result).captures[1])
+                    print_with_color(:yellow, INDENT, "$(status) ", link.url, "\n")
+                    print_with_color(:yellow, INDENT, " -> ", location, "\n\n")
```
MichaelHatherly: The indent here is to line up the two URLs so that it's easier to spot any differences. I played around with some different layouts and this one did seem to be the easiest to read to me.

PR #328, opened 2016-10-17
Include `custom.css` in HTML build if it exists

### LINKED ISSUE #327: Feature request: Insert custom css in HTML output
I want to emulate a terminal window in my documentation and in order to do so it would be nice if I could define some stuff in the CSS file and then use the raw html macro. Would it be possible to be able to append some custom CSS to the already generated CSS file? Perhaps as a new kwarg to `makedocs`?

MichaelHatherly:
Fixes #327.

@mortenpi perhaps we should also just have a single `custom.js` that gets included if it exists rather than the whole `Vector` of them, is that doable or is there something I'm missing?

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -29,17 +29,11 @@ the subdirectories of `assets/` are not linked).
 For the **logo**, Documenter checks for the existence of `assets/logo.png`.
 If that's present, it gets displayed in the navigation bar.
 
-For **scripts**, every `assets/*.js` gets a `<script>` link in the `<head>` tag
-of every page (except if it matches one of Documenter's default scripts;
-the filtering is done in [`user_scripts`](@ref)).
-
-Note that only javascript files are linked to the generated HTML. Any related CSS
-must be loaded by the script. With jQuery this could be done with the following
-snippet
-
-```javascript
-\$('head').append(\$('<link rel="stylesheet">').attr('href', documenterBaseURL + "/assets/<file>.css"))
-```
+Additional JS and CSS assets can be included in the generated pages using the `assets`
+keyword for `makedocs`. `assets` must be a `Vector{String}` and will include each listed
+asset in the `<head>` of every page in the order in which they are listed. The type of
+the asset (i.e. whether it is going to be included with a `<script>` or a `<link>` tag)
+is determined by the file's extension -- either `.js` or `.css`.
 """
 module HTMLWriter
 
@@ -73,7 +67,7 @@ type HTMLContext
     search_js :: Compat.String
     search_index :: IOBuffer
     search_index_js :: Compat.String
-    user_scripts :: Vector{Compat.String}
+    user_assets :: Vector{Compat.String}
 end
 HTMLContext(doc) = HTMLContext(doc, "", [], "", "", "", IOBuffer(), "", [])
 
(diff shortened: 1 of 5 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -70,12 +70,13 @@ type HTMLContext
     scripts :: Vector{Compat.String}
     documenter_js :: Compat.String
     documenter_css :: Compat.String
+    custom_css::Compat.String
```
mortenpi: Spaces around `::`?

mortenpi:
About `custom.js` -- I believe that there's no easy way of including additional javascript within a javascript file, and complications with cross-origin policy. So it seems simpler to me to just explicitly include all `.js` files as `<script>` tags. Of course, maybe we should be consistent about it and have the same behaviour for CSS too.

MichaelHatherly:
> I believe that there's no easy way of including additional javascript within a javascript file

:( silly javascript.

> Of course, maybe we should be consistent about it and have the same behaviour for CSS too.

Or maybe we should only allow a single `custom.js` file.

mortenpi:
> Or maybe we should only allow a single `custom.js` file.

Would make it harder for people to bundle additional libraries if they can't or don't want to CDN them for some reason, I think.

It also actually seems that there are some arguments for using `<link>`s over `@import`s actually, that have to do with the way the CSS files get loaded (`<link>`s get downloaded in parallel, so they're faster). But it's probably not that important for us.

MichaelHatherly:
I'll just mirror the custom js approach then.

mortenpi:
There's always the option of having options too. There's precedent in MkDocs with `extra_css` and `extra_javascript`. That way the user could choose if they want to do `@import`s or `<link>`s etc.

So in this case, maybe have it as an option, but default to `custom.css` and `custom.js`? Could be the best of both worlds really. I think you don't _usually_ need more than one user-defined file, but it should be possible to do it easily. But just loading all the files into the HTML may not always be what you want either.

One small nuance is the handling of non-existent files. If the file is explicitly defined, I'd say we should throw an error, but we shouldn't of course throw if the default file doesn't exist. Changing the `.js` behavior would be a breaking change though, in principle, but worth it, I think.

MichaelHatherly:
Yeah, that's probably the most sane approach to take.

MichaelHatherly:
New approach using a `assets = ` keyword for `makedocs`, i.e.

```julia
makedocs(
    assets = [
        "assets/custom.css",
        "js/foo.js",
    ],
    # ...
)
```

No magically including files based on filename, we just include the exact list that the user provides. Also combining the `.js` and `.css` into a single list allows us to include some CSS, then some JS, then some more CSS. Additionally `src/assets` isn't that special anymore, paths in the `assets` vector are taken to be relative to the `docs/src` dir.

src/Writers/HTMLWriter.jl
```diff
@@ -29,17 +29,9 @@ the subdirectories of `assets/` are not linked).
...
-Note that only javascript files are linked to the generated HTML. Any related CSS
-must be loaded by the script. With jQuery this could be done with the following
-snippet
-
-```javascript
-\$('head').append(\$('<link rel="stylesheet">').attr('href', documenterBaseURL + "/assets/<file>.css"))
-```
+Additional JS and CSS assets can be included in the generated pages using the `assets`
+keyword for `makedocs`. `assets` must be a `Vector{String}` and will include each listed
+asset in the `<head>` of every page in the order in which they are listed.
```
mortenpi: Perhaps we should add something along the lines of: "the type of the asset (i.e. whether it going to be included with a `<script>` or a `<link>` tag) is determined by file's the extension?" Just to make it clear for the user.

Otherwise, looks great! :100:

PR #330, opened 2016-10-18
Fix hrefs on Windows

MichaelHatherly:
Discovered over on https://github.com/JuliaStats/DataFrames.jl/pull/1105. Windows path separators in link `href`s don't seem to work, so just replace them all with `/`. @mortenpi are there any other places this should be done?

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -576,7 +576,9 @@ Calculates a relative HTML link from one path to another.
 """
 function relhref(from, to)
     pagedir = dirname(from)
-    relpath(to, isempty(pagedir) ? "." : pagedir)
+    # The regex separator replacement is necessary since otherwise building the docs on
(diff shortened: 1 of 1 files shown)


mortenpi:
I don't think so. I think all the paths go through the `relpath` function, and I couldn't see/think of anything when I looked through `HTMLWriter` again.

Could we maybe put a small comment there though saying why the regex is necessary, for future reference? (Sorry, I feel I am being exceptionally picky today :laughing:.)

davidanthoff:
Just tried it, and it fixes the issue I originally reported.

MichaelHatherly:
> Could we maybe put a small comment there though saying why the regex is necessary, for future reference? (Sorry, I feel I am being exceptionally picky today :laughing:.)

Yup, will do.

> Just tried it, and it fixes the issue I originally reported.

Thanks @davidanthoff for confirming.

PR #331, opened 2016-10-19
Explicitly adding location of Example module to LOAD_PATH

sivark:
The default suggestion now makes the LOAD_PATH concept discoverable, and is
therefore newbie friendly. That tidbit does not affect someone who already
knows about LOAD_PATH.

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -19,8 +19,8 @@ to creating a simple document.
 ### Setting up the folder structure
 
 Firstly, we need a Julia module to document. This could be a package generated via
-`PkgDev.generate` or a single `.jl` script. For this guide we'll be using a package called
-`Example.jl` that has the following directory layout:
+`PkgDev.generate` or a single `.jl` script accessible via Julia's `LOAD_PATH`. For this
+guide we'll be using a package called `Example.jl` that has the following directory layout:
 
 ```
 Example/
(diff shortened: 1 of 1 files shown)


MichaelHatherly:
Thanks.

docs/src/man/guide.md
```diff
@@ -61,13 +61,15 @@ a single empty file at the moment, but we'll be adding to it later on.
 Add the following to your `make.jl` file
 
 ```julia
+push!(LOAD_PATH,"../src/")
```
MichaelHatherly: Looking back over this this morning I think it's probably best not put the `push!` line here, since further up the page we've explicitly stated that `Example` is a normal package, and rather add a [`!!! note` block](http://docs.julialang.org/en/latest/manual/documentation/#admonitions) after the code block that draws attention to needing to add `push!` or use `include` directly to load non-packages.

dickfeynman: Could you point me to the line(s) in guide.md that you're referring to (which state that `Example` is a normal package)?

MichaelHatherly: From [this](https://juliadocs.github.io/Documenter.jl/latest/man/guide.html#Setting-up-the-folder-structure-1) section: "For this guide we'll be using a package".

dickfeynman: I found that somewhat confusing when I read it the first time; I interpreted it as referring to a package `Example.jl` which we're creating, which could (in principle) be living anywhere and not necessarily in one of the locations at LOAD_PATH. Especially because the example mentions using Documenter.jl with an arbitrary (single) Julia script.

I now understand better the perspective from which the document was written, and I wonder if it would be helpful to have a clarification near the statement you pointed to, because I completely missed the significance of what is assumed there.

sivark: @MichaelHatherly BTW, @dickfeynman is an alternate (old) account of mine that I don't use these days. I didn't realize my comments were being made from that account! Sorry about the confusion; I'll have to figure out if I've messed up some login configuration.

MichaelHatherly: Option 2, just needs a note below this code block warning readers that they will need to manually setup their `LOAD_PATH` or use `include`.

MichaelHatherly:
Thanks for the contribution @sivark!

PR #333, opened 2016-10-21
Streamline CSS downloads

MichaelHatherly:


MichaelHatherly:
Everything seems to work smoothly enough on IE11, FF49, Chrome 54, and Opera 40 on both Windows and Linux. That's probably a sufficiently wide range of browsers.

src/Writers/HTMLWriter.jl
```diff
@@ -166,17 +165,23 @@ function render_head(ctx, navnode, additional_scripts)
     @tags head meta link script title
     src = get(navnode.page)
     page_title = "$(mdflatten(pagetitle(ctx, navnode))) · $(ctx.doc.user.sitename) documentation"
+    css_links = [
+        "https://cdnjs.cloudflare.com/ajax/libs/normalize/4.2.0/normalize.min.css",
+        "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/styles/default.min.css",
+        "https://fonts.googleapis.com/css?family=Lato|Ubuntu+Mono",
+        "https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.min.css",
+        relhref(src, ctx.documenter_css),
+    ]
```
mortenpi: How do you feel about having external etc. somewhere at the top of the file, instead of being hidden away in the code? I feel that it would be nice to have them in once place, which is also the reason I already have the `const requirejs_cdn` there.

mortenpi:
Ah, I didn't notice this at first -- you've removed Fantasque completely? When FontLibrary was down before I did find out that it should be possible to request for things to be added to cdnjs via [their GitHub repo](https://github.com/cdnjs/cdnjs). If we'd put together a PR then we could have it be hosted on a more reliable CDN (Fantasque's license should allow that I think).

MichaelHatherly:
The nice thing about getting all the fonts from googlefonts is that we can then do

```
https://fonts.googleapis.com/css?family=Lato|Ubuntu+Mono
```

rather than having two separate requests, which seems to be what's making it a little snappier I think.

> but does it actually make much difference?

It does seem to, at least for me, and it's definitely no worst than before.

MichaelHatherly:
Inconsolata is also quite a nice alternative to Ubuntu or Fantasque.

mortenpi:
Having switched back and forth between them, I think I have a very slight preference for Ubuntu Mono since Inconsolata seems to have quite thin lines. However, I think both are good options, so I'd be perfectly happy with Inconsolata as well.

MichaelHatherly:
Let's go with Ubuntu then.

MichaelHatherly:
Thanks for reviewing @mortenpi.

MichaelHatherly:
No backporting of this one. There's enough changes on `master` to have another minor I think.

Issue #335, opened 2016-10-31
Travis.genkeys error


Ward9250:
Hi, I'm trying to get documentation set up for: https://github.com/BioJulia/WAHVectors.jl

```julia
julia> using Documenter

julia> Pkg.clone("https://github.com/BioJulia/WAHVectors.jl.git")
INFO: Cloning WAHVectors from https://github.com/BioJulia/WAHVectors.jl.git
INFO: Computing changes...

julia> Travis.genkeys("WAHVectors")
Generating public/private rsa key pair.
Your identification has been saved in .documenter.
Your public key has been saved in .documenter.pub.
The key fingerprint is:
SHA256:rDzcViYoDvy871LcKPNKo81Mg/ahJmhjMUydgQ42mjQ axolotlfan9250@Bens-MacBook-Pro.local
The key's randomart image is:
+---[RSA 2048]----+
|  .              |
|.E .             |
|*.+ o            |
|o+ o   o         |
|o o ...oS o      |
| + *o++o.+       |
|. = X== o        |
|o=.X.=.o         |
|oo+ B=o          |
+----[SHA256]-----+
We need your GitHub login to identify you.
This information will not be sent to Travis CI, only to api.github.com.
The password will not be displayed.

Try running with --github-token or --auto if you don't want to enter your password anyway.

Username: Ward9250
Password for Ward9250: ********secretstuff********
Two-factor authentication code for Ward9250: ********secretstuff********
Validation Failed
for a full error report, run travis report
ERROR: failed process: Process(`travis login --auto`, ProcessExited(1)) [1]
 in pipeline_error(::Base.Process) at ./process.jl:616
 in run at ./process.jl:592 [inlined]
 in (::Documenter.Travis.##2#3{String})() at /Users/axolotlfan9250/.julia/v0.5/Documenter/src/Documenter.jl:545
 in cd(::Documenter.Travis.##2#3{String}, ::String) at ./file.jl:59
 in #genkeys#1(::String, ::Function, ::String) at /Users/axolotlfan9250/.julia/v0.5/Documenter/src/Documenter.jl:533
 in genkeys(::String) at /Users/axolotlfan9250/.julia/v0.5/Documenter/src/Documenter.jl:522

shell> travis report
System
Ruby:                     Ruby 2.3.1-p112
Operating System:         Mac OS X 10.12
RubyGems:                 RubyGems 2.6.8

CLI
Version:                  1.8.2
Plugins:                  none
[...]

PR #336, opened 2016-11-01
Remove dependency on travis-cli

### LINKED ISSUE #298: Don't Require Local Travis?
Documenter.jl didn't require a local Travis installation before. But now you need it just for the keygen step. Why doesn't the documentation have a way to do this without using a local Travis installation (i.e. for anyone on RHEL or CentOS).

MichaelHatherly:
Fixes #298 and #335.


mortenpi:
We still have the `.documenter.enc` file as far as I can tell? How about dropping it altogether and just storing the whole SSH key in the Travis variable (i.e. https://github.com/JuliaDocs/Documenter.jl/issues/298#issuecomment-252497392)?

MichaelHatherly:
I believe that variables have a limit, something like 100 chars, so a full private key wouldn't really fit. What approach were you thinking of using?

mortenpi:
Nope, I don't think they do. Or at least it should be more than enough for a private key: https://travis-ci.org/mortenpi/Hafta.jl/jobs/172273012#L157 with [this `envshow.jl`](https://github.com/mortenpi/Hafta.jl/blob/key-test/envshow.jl).

mortenpi:
It seems that the encrypted variables in `travis.yml` have a size limitation (https://github.com/travis-ci/travis.rb/issues/100#issuecomment-57127635), but environment variables should be fine, I think.

MichaelHatherly:
Cool, well I guess that will simplify things a bit.

MichaelHatherly:
Seems to work well as far as I can tell.

docs/src/man/hosting.md
```diff
@@ -54,54 +53,34 @@ Then call the [`Travis.genkeys`](@ref) function as follows:
...
-    openssl aes-256-cbc -K $encrypted_<something>_key -iv $encrypted_<something>_iv -in .documenter.enc -out .documenter -d
-
-Pro Tip: You can add it automatically by running with --add.
-```
-
-**Ignore this**. Documenter will handle it for you.
-
-Once complete you will need to add the public key displayed in the REPL to your repository -- just follow the instructions displayed in the REPL:
+where `"MyPackage"` is the name of the package you would like to create deploy keys for. The
+output will look similar the text below:
```
mortenpi: "similar _to_ the text below:" ?

MichaelHatherly:
The other methods of deploying docs are being deprecated as well here &mdash; `deploydocs` is way too complex with having to support all three methods. The deprecations will stay in place for two minor tags (probably removed in `0.8.0`), however long that happens to be from now.

docs/src/man/hosting.md
```diff
@@ -54,54 +53,34 @@ Then call the [`Travis.genkeys`](@ref) function as follows:
...
+    **`Key`** field.  Note that you should include **no whitespace** when copying the key.
+    Check **`Allow write access`** to allow Documenter to commit the generated documentation
+    to the repo.
 
-- `encrypted_e6b49e69746a_key`
-- `encrypted_e6b49e69746a_iv`
+ 2. Next add the long private key to the Travis settings page using the provided link. In
+    the **`Environment Variables`** section add a key with the name `DOCUMENTER_KEY` and the
+    value that was printed out. **Do not** set the variable to be displayed in the build
+    log. Then click **`Add`**.
```
mortenpi: Since it's a potential security hazard, I was thinking that maybe we should have a big red box here, just in case. Something like the following perhaps:

```
    !!! warning "Security warning"
        To reiterate: make sure that the "Display value in build log" option is **OFF** for
        the variable, so that it does not get printed when the tests run.
        This base64-encoded string contains the *unencrypted* private key that gives full
        write access to your repository, so it must be kept safe.
        Also, make sure that you never expose this variable in your tests, nor merge any code
        that does. You can read more about Travis environment variables in
        [Travis User Documentation](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings).
```

ChrisRackauckas:
Hey, where can we find the docs on this? Thanks for getting around to this!

MichaelHatherly:
Looks like travis is just being slow to build at the moment. You can see the un-rendered source file though https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/man/hosting.md#ssh-deploy-keys.

PR #343, opened 2016-11-02
Update DifferentialEquations.jl documentation link

ChrisRackauckas:
It was moved to a separate repository, DiffEqDocs.jl, to make it easier to manage the docs for the whole org. It's probably a good example for how to setup travis to do so too!


MichaelHatherly:
Thanks.

PR #348, opened 2016-11-04
Detect doctest exceptions for empty output

### LINKED ISSUE #347: Test the syntax used in doctests
For a module:

```jl
module A
    export foo
    foo(x) = "Hello, world!"
end
```

These doctests pass

\`\`\`jldoctest
julia> bar(1)
\`\`\`

even though `bar` is not defined. The thrown `ERROR: UndefVarError: bar not defined` is not let through unless there is a output comparison. If there is a comparison, the test fail:

\`\`\`jldoctest
julia> bar(1)
"Hello, world!"
\`\`\`

which throws

```
!! Test error in the following code block in 'src/test.md':

        julia> bar(1)
        "Hello, world!"

    in expression:

        bar(1)

    expected:

        "Hello, world!"

    returned:

        ERROR: UndefVarError: bar not defined
         ...
```

It would be useful if the error is let through, to compare syntax used in the docs. For instance if  `foo` previously was called `bar`, then it would be nice if the doctest fail.

MichaelHatherly:
Fixes #347 where exceptions were not caught when the expected result
was an empty string.

Also fix skipped doctest when the last `julia>` prompt appears right
at the end of a code block, i.e.

    ```jldoctest
    ...
    julia> x
    ```

instead of

    ```jldoctest
    ...
    julia> x

    ```

Additionally, remove unnecessary parts of the generated backtrace that reference internals of the Documenter package itself, similar to how the Julia REPL handles backtrace printing.

@fredrikekre, if you've got a moment to confirm that this sorts out the issue you reported that would be great, thanks.

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -182,7 +182,7 @@ function eval_repl(code, sandbox, meta::Dict, doc::Documents.Document, page)
             # Input containing a semi-colon gets suppressed in the final output.
             result.hide = ends_with_semicolon(str)
             (value, success, backtrace, text) = Utilities.withoutput() do
-                eval(sandbox, ex)
+                Core.eval(sandbox, ex)
             end
             result.value = value
             print(result.stdout, text)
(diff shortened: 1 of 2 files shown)


fredrikekre:
Nice, can confirm it works!

MichaelHatherly:
Thanks for confirming that.

PR #352, opened 2016-11-14
Allow deploying to subdirectories

### LINKED ISSUE #351: Travis CI hangs forever after deploying docs.
Hi,

I encountered a problem that Travis CI hangs forever after deploying docs. Here is the log of a build:
https://travis-ci.org/bicycle1885/EzXML.jl/jobs/175317593.

I'm not sure whether this is due to this package, but this started to happen just after started to use `deploydocs` to deploy docs to GitHub pages. Also, this does not happen on other platforms that does not deploy docs (I'm using Julia 0.5 on Linux in deployment). So, I thought this may be related to its process.

[.travis.yml](https://github.com/bicycle1885/EzXML.jl/blob/master/.travis.yml):
```yml
# Documentation: http://docs.travis-ci.com/user/languages/julia/
language: julia
os:
  - linux
  - osx
julia:
  - 0.5
  - nightly
matrix:
  allow_failures:
    - julia: nightly
notifications:
  email: false
# uncomment the following lines to override the default test script
#script:
#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
#  - julia -e 'Pkg.clone(pwd()); Pkg.build("EzXML"); Pkg.test("EzXML"; coverage=true)'
after_success:
  # push coverage results to Coveralls
  # - julia -e 'cd(Pkg.dir("EzXML")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
  # push coverage results to Codecov
  - julia -e 'cd(Pkg.dir("EzXML")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
  - julia -e 'Pkg.add("Documenter")'
  - julia -e 'cd(Pkg.dir("EzXML")); include(joinpath("docs", "make.jl"))'
```

[make.jl](https://github.com/bicycle1885/EzXML.jl/blob/master/docs/make.jl):
```julia
using Documenter
using EzXML

makedocs(
    format=:html,
    sitename="EzXML.jl",
    pages=["index.md", "manual.md", "references.md", "devnotes.md"])

deploydocs(
    repo="github.com/bicycle1885/EzXML.jl.git",
    julia="0.5",
    target="build",
    deps=nothing,
    make=nothing)
```

MichaelHatherly:
Additionally, fixes #351 by referencing the ssh key files directly rather than via `ssh-agent`.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -289,6 +289,7 @@ GitHub.
 function deploydocs(;
         root   = Utilities.currentdir(),
         target = "site",
+        dirname = "",
 
         repo   = error("no 'repo' keyword provided."),
         branch = "gh-pages",
@@ -374,10 +375,12 @@ function deploydocs(;
             end
             Utilities.log("pushing new documentation to remote: $repo:$branch.")
             mktempdir() do temp
+                dirname = isempty(dirname) ? temp : joinpath(temp, dirname)
+                isdir(dirname) || mkpath(dirname)
                 # Versioned docs directories.
-                latest_dir = joinpath(temp, "latest")
-                stable_dir = joinpath(temp, "stable")
-                tagged_dir = joinpath(temp, travis_tag)
+                latest_dir = joinpath(dirname, "latest")
+                stable_dir = joinpath(dirname, "stable")
+                tagged_dir = joinpath(dirname, travis_tag)
 
                 keyfile, _ = splitext(ssh_key_file)
                 target_dir = abspath(target)
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -386,84 +386,67 @@ function deploydocs(;
                 target_dir = abspath(target)
 
                 # The upstream URL to which we push new content and the ssh decryption commands.
-                upstream, ssh_script =
+                upstream =
                     if documenter_key != ""
-                        "git@$(replace(repo, "github.com/", "github.com:"))",
-                        """
-                        echo "$(Compat.String(base64decode(documenter_key)))" >> $keyfile
-                        chmod 600 $keyfile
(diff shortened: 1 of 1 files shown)


src/Documenter.jl
```diff
@@ -389,8 +392,6 @@ function deploydocs(;
                         """
                         echo "$(Compat.String(base64decode(documenter_key)))" >> $keyfile
```
mortenpi: Just a thought that crossed my mind -- would it perhaps be better to directly write the key to a file, instead of interpolating it into a script first? I [did that while debugging](https://github.com/JuliaDocs/Documenter.jl/blob/mp/deploy-hang-debug/src/Documenter.jl#L382) since I wanted to see the contents of the script in the log for debugging.

src/Documenter.jl
```diff
@@ -462,6 +465,11 @@ function deploydocs(;
                     # Remove the unencrypted private key.
                     isfile(keyfile) && rm(keyfile)
                 end
+
+                # Undo ssh config changes.
```
mortenpi: It turns out that you can customize git's SSH command using the [`GIT_SSH` variable](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#Miscellaneous). We could `mktemp` the `ssh_config` and set `GIT_SSH` to `ssh -F tmp_ssh_config` perhaps (I haven't tried it though, so not sure if it works).

MichaelHatherly: Couldn't seem to get `GIT_SSH` working correctly with `withenv`, but did manage to remove the whole external shell script hack so I think this is as simple as it's going to get until the older authentication methods are finally removed.

PR #368, opened 2016-12-12
update a `find_object` signature for jb/subtype

JeffBezanson:
This makes Documenter work on my branch, and should be backwards-compatible.


MichaelHatherly:
Thanks.

Issue #385, opened 2016-12-26
No syntax highlighting in `jldoctest` block


dpsanders:
I seem to get syntax highlighting in code blocks with `julia` but not with `jldoctest`.

MichaelHatherly:
Thanks. Should be fixed by the referenced commit.

TotalVerb:
I still don't see syntax highlighting in `jldoctest` blocks; this is also the case in the Base manual. Note the lack of highlighting in these blocks: https://docs.julialang.org/en/latest/manual/strings.html#String-Basics-1

mortenpi:
I don't think highlight.js support highlighting doctests, so the `language-jldoctest` class doesn't really do anything.

mortenpi:
Should be fixed by #486.

Issue #370, opened 2016-12-13
Print diff in failing doctest


KristofferC:
Throwing out an idea here, but sometimes it is hard to see why a doctest is failing, see for example https://github.com/JuliaLang/julia/pull/19569#issuecomment-266771012.

Maybe it could be possible to print a diff to explicitly show where the problem is. On the other hand, maybe in the vast majority of cases, the difference should be obvious.

MichaelHatherly:
Yes, would be good to have.

MichaelHatherly:
Some basic diff printing in the referenced commit. Probably will need some fine tuning of the output, but I'd say it's definitely an improvement over what gets printed at the moment.

PR #386, opened 2016-12-26
Doctest fixes

MichaelHatherly:
- Correct highlighting for `jldoctest` code blocks in HTML output (#385).
- Print a diff when a doctest fails (#370).

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -316,20 +316,28 @@ function sanitise(buffer)
     remove_term_colors(rstrip(Utilities.takebuf_str(out), '\n'))
 end
 
+import .Utilities.TextDiff
+
 function report(result::Result, str, doc::Documents.Document)
-    buffer = IOBuffer()
-    println(buffer, "Test error in the following code block in '$(result.file)':")
-    print_indented(buffer, result.code; indent = 8)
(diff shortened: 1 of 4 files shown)


src/DocChecks.jl
```diff
@@ -316,20 +316,28 @@ function sanitise(buffer)
...
+    println(buffer, "```")
+    if !isempty(result.input)
+        print_with_color(:cyan, buffer, "\n> Subexpression:\n")
+        print_indented(buffer, result.input; indent = 4)
     end
-    print_indented(buffer, "expected:")
-    print_indented(buffer, result.output; indent = 8)
-    print_indented(buffer, "returned:")
-    print_indented(buffer, rstrip(str); indent = 8) # Drops trailing whitespace.
+    local warning = Base.have_color ? "" : " (REQUIRES COLOR)"
```
tkelman: will it have leading +/- indicators even if it's not in color?

MichaelHatherly: Yes, line-based diffs have a +/- prefix per line. For these doctest diffs, I'm using a word-based diff at the moment which does require color output. We can fine tune the output if it turns out to still be difficult to read though.

PR #390, opened 2016-12-28
Allow named code literal `at-ref`s

MichaelHatherly:


tkelman:
should this have a unit test here?

MichaelHatherly:
At the moment these kind of changes aren't properly testable in a reasonably simple way, though that situation will be much improved with some larger restructuring that I'm currently working on and tests for things like this will definitely be included.

PR #403, opened 2017-01-05
Update examples.md

cormullion:
Updated path to https://github.com/JuliaGraphics/Luxor.jl.

I notice in the examples that some other's links are to "stable", some to "latest". I'd prefer a link to the main start page to allow a choice...?

### DIFF
--- a/docs/src/man/examples.md
+++ b/docs/src/man/examples.md
@@ -29,7 +29,7 @@ Packages that have tagged versions available in `METADATA.jl`.
 - [EzXML.jl](https://bicycle1885.github.io/EzXML.jl/latest/)
 - [Highlights.jl](https://juliadocs.github.io/Highlights.jl/latest/)
 - [IntervalConstraintProgramming.jl](http://dpsanders.github.io/IntervalConstraintProgramming.jl/latest/)
-- [Luxor.jl](https://github.com/JuliaGraphics/Luxor.jl)
+- [Luxor.jl](https://github.com/JuliaGraphics/Luxor.jl/stable/)
 - [MergedMethods.jl](https://michaelhatherly.github.io/MergedMethods.jl/latest)
 - [Mimi.jl](http://anthofflab.berkeley.edu/Mimi.jl/stable/)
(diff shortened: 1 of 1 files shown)


MichaelHatherly:
> I notice in the examples that some other's links are to "stable", some to "latest". I'd prefer a link to the main start page to allow a choice...?

I'd prefer links to point directly to examples of Documenter-generated sites, rather than requiring extra clicks. Latest and stable will probably look pretty similar in nearly every case so it shouldn't really make much difference I'd imagine.

cormullion:
Good point. And thanks for your help in getting this far!

docs/src/man/examples.md
```diff
@@ -29,7 +29,7 @@ Packages that have tagged versions available in `METADATA.jl`.
 - [EzXML.jl](https://bicycle1885.github.io/EzXML.jl/latest/)
 - [Highlights.jl](https://juliadocs.github.io/Highlights.jl/latest/)
 - [IntervalConstraintProgramming.jl](http://dpsanders.github.io/IntervalConstraintProgramming.jl/latest/)
-- [Luxor.jl](https://cormullion.github.io/Luxor.jl)
+- [Luxor.jl](https://github.com/JuliaGraphics/Luxor.jl/stable/)
```
MichaelHatherly: Should be `https://juliagraphics.github.io/Luxor.jl/latest/` instead?

cormullion: Oh, that's a point. For 'stable', the heading is "tagged versions available in METADATA", so I would expect a link to the documentation for that tagged version, rather than some in-development version? But that's just a guess. I'm sure you're right about the URL, I wouldn't know.

MichaelHatherly: Or `stable` instead of `latest`, I don't mind which really :)

MichaelHatherly:
Thanks.

cormullion:
Cheers. Fortunately, that's my last Documenter.jl project for the next few weeks... :)

PR #409, opened 2017-01-17
Sidebar updates

### LINKED ISSUE #406: Visually distinguish "latest" and "stable"?
If I understand correctly, the "stable" version refers to the more recently *released* version, and the "latest" version of the docs refers to the most recently *built* version. It would be good if it was easy to distinguish visually between the two.

Here's what two versions look like on my iMac in Safari:

screen shot 2017-01-10 at 16 26 36

It's very hard to see which is which—the drop-down menu has reverted to "Version", and the URL bar is too long (and quite small)

So I'm wondering whether there's something you can do when you build a "latest" version that can visually draw attention to the fact that this version is "in development" or "not the most reliable" or even "not for production use". A border (via CSS)? A footer line?

(Slightly related issue: #337)

mortenpi:
Style and behaviour updates for the sidebar. Site build [available here](http://mortenpi.eu/Documenter.jl/fix-style/).

- Improve contrast of the sidebar menu as discussed [here](https://discourse.julialang.org/t/improve-contrast-in-docs-julialang-org-menu/1370/11).
- Update the version selector style and behaviour. Basically to make fit a bit better with the rest of the style and also make sure that it displays the current version when loaded.

The WIP bit is to make `makedocs` automatically determine the current version name (e.g. `latest`, tag name) which is currently only done in `deploydocs` (so, right now it would just show the commit hash for all builds).

I am also not necessarily entirely happy with the version selector's style, so feedback & ideas would be greatly appreciated (cc @cormullion).

Ideally fixes #406.

### DIFF
--- a/assets/html/arrow.svg
+++ b/assets/html/arrow.svg
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="16.5mm"
+   height="8.6603003mm"
+   viewBox="0 0 58.464567 30.686103"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.91 r13725"
+   sodipodi:docname="arrow.svg">
+  <defs
+     id="defs4" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageopacity="0.0"
+     inkscape:pageshadow="2"
+     inkscape:zoom="11.2"
+     inkscape:cx="14.209234"
+     inkscape:cy="29.780479"
+     inkscape:document-units="px"
+     inkscape:current-layer="layer1"
+     showgrid="false"
+     inkscape:window-width="1920"
+     inkscape:window-height="1053"
+     inkscape:window-x="0"
+     inkscape:window-y="27"
+     inkscape:window-maximized="1" />
+  <metadata
+     id="metadata7">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <g
+     inkscape:label="Layer 1"
+     inkscape:groupmode="layer"
+     id="layer1"
+     transform="translate(0,-1021.6761)">
+    <path
+       style="fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:#000000;stroke-width:0;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+       d="m 0,1021.6761 35.433071,0 -17.716536,30.6861 z"
+       id="path4140"
+       inkscape:connector-curvature="0"
+       sodipodi:nodetypes="cccc" />
+  </g>
+</svg>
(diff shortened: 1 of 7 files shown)


cormullion:
Looks pretty good—works well on macOS/Safari. I might prefer it if the dropdown menu and default text in the search box (and the logo) was all centered... 

On iPad (and probably iPhone too but...) the navigation  bar isn't very usable, though —touch event handling or something. Perhaps that could be added as a separate issue...

MichaelHatherly:
> Update the version selector style and behaviour.

The selector doesn't display the dropdown arrow as far as see. Could we add that back since it's a useful indication of what it is.

> The WIP bit is to make `makedocs` automatically determine the current version name

`makedocs` isn't currently dependant on the "version" at all. Changing that would probably require merging `makedocs` and `deploydocs`. Best way to handle this in `deploydocs` would be to create a `version.js` file containing `var DOCUMENTER_VERSION = "<dirname>";` for each version directory. Then use that value for the selector.

mortenpi:
> The selector doesn't display the dropdown arrow as far as see. Could we add that back since it's a useful indication of what it is.

Yup, I agree that it would be nice to have. However, I removed it intentionally since it can't be styled and it is very inconsistent across platforms/browsers and doesn't really match with the rest of the style (especially in Firefox). But I should play around with it a bit more -- I think the standard workaround is to use background images on the `select` to show an arrow.

> Best way to handle this in `deploydocs` would be to create a `version.js` file containing `var DOCUMENTER_VERSION = "<dirname>";` for each version directory. Then use that value for the selector.

It would be nice if it wouldn't depend on javascript though and the current version would be displayed statically. But this sounds like a reasonable solution for the first iteration.

mortenpi:
> I might prefer it if the dropdown menu and default text in the search box (and the logo) was all centered... 

Everything except the text in the search box should be centered. Could you give me a screenshot of what you're seeing?

> On iPad (and probably iPhone too but...) the navigation  bar isn't very usable, though —touch event handling or something. Perhaps that could be added as a separate issue...

Mobile experience from both the layout and touch experience point of view has not received any attention so far really. Would definitely be something that could be improved.

MichaelHatherly:
> especially in Firefox

Looks alright for me on Firefox, I think it depends on what dm/wm and theme is being used.

2017-01-19-1484808444-crop

> It would be nice if it wouldn't depend on javascript

There's already lots that depends on js in the docs, so for a first pass it's fine to require it for this.

> Mobile experience from both the layout and touch experience point of view has not received any attention so far really. Would definitely be something that could be improved.

According to the site analytics mobile views account for less than 5% of all traffic to Documenter's docs, so probably not a critical requirement just yet.

cormullion:
On Safari I see this:

screen shot 2017-01-19 at 10 05 46

Artist's impression of slight improvement (_de gustibus non est disputandum_ ...):

screen shot 2017-01-19 at 10 05 46 copy copy

mortenpi:
For me on Firefox, when I re-enable the arrow, it looks like this:
documenter_ff

However, I agree that there should be something indicating that you can actually select a version there, so I'll look into that. Also, agreed, let's go with a `.js` file for the time being.

As for centering -- the `select` has `text-align: center` and `text-align-last: center`, which means it is nicely centered on all the browsers that I've tested (but I don't have access to OSX/Safari). What styling did you add, @cormullion?

The search box I would keep left-aligned since it's quite wide and then it matched the menu below.

Just FYI: I'll be travelling in the coming days though, so this PR will likely be on hold until the latter half of next week.

MichaelHatherly:
Looks like a 0.4-specific failure. You're welcome to drop 0.4 support if you'd like to &mdash; we've been keeping it going for long enough I think.

mortenpi:
It's a small type thing with `String`s -- as a workaround I think can force the string to be `Compat.UTF8String` so it works on all versions.

mortenpi:
Updated on my part, latest build should be [here](http://mortenpi.eu/Documenter.jl/latest/) (built and deployed on [Travis](https://travis-ci.org/mortenpi/Documenter.jl/jobs/195858084)). Changes:

* Unified select box triangle with a background image.
* Uses `siteinfo.js` to populate the selector's "current version".

I do feel that this needs more work in the long run. Generating `siteinfo.js` in `deploydocs` doesn't feel quite right, and the `latest`/`release` etc. tags should probably have the exact version shown as well (e.g. "v0.5.0 (latest)" instead of just "latest"). But all that should be in future PRs.

mortenpi:
Cleaned up and rebased, [latest build](http://mortenpi.eu/Documenter.jl/latest/) updated. I would say that it's ready to merge.

Only thing unaddressed is the Safari style, which I can't test myself. @cormullion ?

cormullion:
Looks good to me!

img_0074

mortenpi:
The text in the selector should be centered (and is on other platforms). What did you add before to make it centered in Safari, @cormullion?

cormullion:
My artist's impression above, you mean? That was image editing... I moved the stuff around manually. That was after a quick look at the Safari Inspector thing, which lets you fiddle with the HTML/CSS. (I think Safari on iPad and Mac will be the same. I'll have a look after breakfast...☕️🥐)

cormullion:
Safari on Mac:

screen shot 2017-02-05 at 10 24 05

mortenpi:
Ah, I thought you did it in the inspector. I added a note about this for future reference as it looks OK even if it is not centered.

PR #410, opened 2017-01-17
Newlines and indentation when printing HTML

MikeInnes:
Makes for much nicer diffs and general output debugging. Example output at the source for [this page](http://mikeinnes.github.io/Flux.jl/latest/).

### DIFF
--- a/src/Utilities/DOM.jl
+++ b/src/Utilities/DOM.jl
@@ -246,12 +246,13 @@ end
 attributes!(out, s::Symbol) = push!(out, tostr(s => ""))
 attributes!(out, p::Pair)   = push!(out, tostr(p))
 
-function Base.show(io::IO, n::Node)
+function Base.show(io::IO, n::Node; level = 0)
     if n.name === Symbol("#RAW#")
         print(io, n.nodes[1].text)
     elseif n.name === TEXT
         print(io, escapehtml(n.text))
     else
+        print(io, "  "^level)
         print(io, '<', n.name)
         for (name, value) in n.attributes
             print(io, ' ', name)
@@ -262,14 +263,18 @@ function Base.show(io::IO, n::Node)
         else
             print(io, '>')
             if n.name === :script || n.name === :style
-                println(io)
-                isempty(n.nodes) || print(io, n.nodes[1].text)
+                if !isempty(n.nodes)
+                  println(io)
+                  println(io, n.nodes[1].text)
+                  print(io, "  "^level)
+                end
             elseif !isempty(n.nodes)
                 println(io)
                 for each in n.nodes
-                    show(io, each)
+                    show(io, each; level = level + 1)
                     println(io)
                 end
+                print(io, "  "^level)
             end
             print(io, "</", n.name, '>')
         end
--- a/src/Utilities/DOM.jl
+++ b/src/Utilities/DOM.jl
@@ -247,12 +247,13 @@ attributes!(out, s::Symbol) = push!(out, tostr(s => ""))
 attributes!(out, p::Pair)   = push!(out, tostr(p))
 
 function Base.show(io::IO, n::Node; level = 0)
+    ws = n.name ∉ (:code, :pre)
     if n.name === Symbol("#RAW#")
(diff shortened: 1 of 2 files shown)


mortenpi:
Good idea I'd think!

I am wondering -- would it make sense to add a special case for these cases?
```
            <span class="fa">

            </span>
```
I.e. nodes with a single `TEXT` child, to make them be like this, so they'd look a bit prettier:
```
            <span class="fa"></span>
```

Otherwise, LGTM!

MikeInnes:
Is it ok to assume that text nodes won't contain `\n`?

I agree it could be nicer to look at; it may work well to simply apply indentation to all text nodes. That can handle multiple children, cases where the tag itself is long, and multi-line nodes (could even make sense to do wrapping here).

MichaelHatherly:
> Is it ok to assume that text nodes won't contain \n?

Text nodes may contain any chars, including newlines.

> Makes for much nicer diffs and general output debugging

How useful are diffs of the generated HTML files? All the content changes should just match up with changes in the markdown files or docstrings. I've just been using Inspect Element for debugging the generated output rather than View Source.

I'm not totally against this change though, just want to make sure the extra complexity is worth it.

MikeInnes:
For example, I wanted to test a tweak to the output (adding custom CSS) without investing time up front to work out Documenter's internals and see whether it was possible to do it properly. This worked but was a hassle due to everything being on one line, and broke with a merge conflict every time I pulled a new build. Making this kind of workflow easier would just lower the barrier to making contributions. This is also totally subjective, but I think that these kind of details really make the difference in making a package seem really polished.

If the complexity is an issue, a good compromise may be to simply add the newlines and not the indentation, which is just an extra `println(io)`. Let me know what works for you.

MichaelHatherly:
> I wanted to test a tweak to the output (adding custom CSS)

The workflow I've been using to do that is

- run `make.jl`;
- use Inspect/Inspect Element (rather than View page source) to view the pretty-printed tree in a browser and find what I want to change;
- add a `src/assets/custom.css` file and add `assets = ["assets/custom.css"],` to `makedocs`;
- make some changes in `custom.css` and rebuild.

The `assets` keyword isn't documented yet, but probably won't be going away I think. All this can be added to the docs somewhere, would probably help lower the barrier as you say.

MikeInnes:
Sure, I appreciate now that that workflow is possible. But I didn't know about it earlier, and probably wouldn't have bothered with the custom CSS at all if there was no way to quickly test if I liked it. It's just about making the obvious thing work.

Documenting that particular workflow is somewhat beside the point; "tweaking the CSS" is a stand-in for any other kind of change you might want to test on the HTML output.

mortenpi:
I've also been using the inspector when playing around with the HTML/CSS, but I think having a readable source would still be an improvement without any downsides. I can definitely see that the ability to reasonably diff different versions could be useful for someone at some point when debugging their own build.

The only argument against it that I can see would be that "minified" files would be slightly smaller and save bandwith etc., but Julia or package docs are hardly high-volume sites.

MichaelHatherly:
> I can definitely see that the ability to reasonably diff different versions could be useful for someone at some point when debugging their own build.

https://github.com/prettydiff/prettydiff/ works really well for that I've found. There's an online version, http://prettydiff.com/, but can also be used locally.

Just to repeat my earlier comment:

> I'm not totally against this change though, just want to make sure the extra complexity is worth it.

If there's enough support for doing this then that's fine, I'm won't be vetoing anything here. My only request is that we don't just go with a partial solution, i.e. pretty printed output should match (as close as possible) other well-known HTML formatters, such as `tidy`.

mortenpi:
prettydiff looks useful, but not really as convenient as `diff`, `git diff` or GitHub's diffs, so it still seems worth it to me at least. Having it be close to `tidy` sounds like a good idea though. Also, I might have LGTMed it a bit prematurely before, e.g. the `pre` thing. Perhaps we should also extend the tests, to make sure there aren't any slightly non-obvious bugs introduced?

odow:
Closing as abandoned. Lots has changed since this PR.

PR #413, opened 2017-01-24
Document how to label doctest blocks

fredrikekre:
Added a note on how to label doctest blocks. I am not sure it is correct though, its based on my own experiments...
From my understanding the labeled blocks does *not* have to be consecutive (eg. it works if I insert a non-labeled block in between two labeled ones), but they have to be in the same file to share scope?

### DIFF
--- a/docs/src/man/doctests.md
+++ b/docs/src/man/doctests.md
@@ -52,7 +52,6 @@ julia> c = 3;  # comment
 
 julia> a + b + c
 6
-
 ```
 ````
 
@@ -77,7 +76,6 @@ ERROR: DivideError: integer division error
 
 julia> div(1, 0)
 ERROR: DivideError: integer division error
-
 ```
 ````
 
@@ -88,7 +86,6 @@ If instead the first `div(1, 0)` error was written as
 julia> div(1, 0)
 ERROR: DivideError: integer division error
  in div(::Int64, ::Int64) at ./int.jl:114
-
 ```
 ````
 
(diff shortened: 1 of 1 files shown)


MichaelHatherly:
Thanks! I'd say it would probably be preferable to use named doctests as opposed to `@meta` blocks since they're slightly less of a pain to use, so might be worth reordering the two sections perhaps. Also, "variable scope" is really just separate modules, might be worth making that clear in the docs.

> From my understanding the labeled blocks does not have to be consecutive

Yes, they don't need to be consecutive and you can intersperse different named doctests between each other. They are evaluated sequentially from top to bottom of the file though.

> but they have to be in the same file to share scope?

Correct, each file is sandboxed from all others with regard to doctests/at-example/at-repl blocks.

fredrikekre:
Okay, I will fix later.

> Also, "variable scope" is really just separate modules, might be worth making that clear in the docs.

So same-named blocks is gathered in the same module I presume?

> Yes, they don't need to be consecutive and you can intersperse different named doctests between each other. They are evaluated sequentially from top to bottom of the file though.

I will add a note about this too.

MichaelHatherly:
> So same-named blocks is gathered in the same module I presume?

Yes, they just refer to the same `Module`.

fredrikekre:
Restructured. Can we come up with a better example for the `DocTestSetup` example? E.g. one where whats in the `DocTestSetup` is actually not used in a block.

docs/src/man/doctests.md
```diff
@@ -115,11 +115,53 @@ Doctesting can be disabled by setting the [`makedocs`](@ref) keyword `doctest =
 This should only be done when initially laying out the structure of a package's
 documentation, after which it's encouraged to always run doctests when building docs.
 
+## Preserve definition between blocks
```
fredrikekre: Note to self: definition**s**

mortenpi: Perhaps also "preserv**ing**"? Matches with the previous heading.

fredrikekre:
Updated a bit. Also moved the section about disabling doctests to the bottom, since it felt a bit squashed in between sections about how to set up doctesting.

docs/src/man/doctests.md
```diff
@@ -105,59 +102,102 @@ line where checking should stop, i.e.
...
+## Preserving definitions between blocks
+
+Every doctest block is evaluated inside its own `module`. This means that definitions
+(types, variables, functions etc.) from a block can *not* be used in the next block.
+A doctest block can be labeled in order to collect several blocks into the same module.
+All blocks with the same label (in the same file) will be evaluated in the same module,
+and hence share scope. This can be useful if the same definitions are used in more than
+one block, with for example text, or other doctest blocks, in between. Example:
+
+````markdown
```
mortenpi: As I was reading through the text, the illustrative example felt a bit long, so I personally might prefer having it be a bit more concise and divided up.

I would propose the following alternative: have something along these lines after the first two sentences to illustrate the problem:

````
```jldoctest
julia> foo = 42
42
```

```jldoctest
julia> println(foo)
ERROR: UndefVarError: foo not defined
```
````

And then show how you can solve the issue with labels:
````
```jldoctest mylabel
julia> foo = 42
42
```

```jldoctest mylabel
julia> println(foo)
42
```
````

I would also move all the explanatory text outside of the example.

File: docs/src/man/doctests.md [hunk 3]
```diff
@@ -105,59 +102,102 @@ line where checking should stop, i.e.
...
-    julia> df = DataFrame(A = 1:10, B = 2:2:20);
+````markdown
+Some text discussing a function `foo`...
 
-    ```
+The definition of `foo` is defined in a `@meta` block below, and thus not visible in the
+rendered documentation:
 
-    Some text discussing `df`...
+```@meta
```
mortenpi: Could also be a bit more concise maybe. I might keep only one definition and one doctest block.

docs/src/man/doctests.md
```diff
@@ -105,59 +102,102 @@ line where checking should stop, i.e.
...
+
+```@meta
+DocTestSetup = nothing
+```
+````
+
+!!! note
+
+    The `DocTestSetup` value is **re-evaluated** at the start of *each* doctest block
+    and no state is shared between any code blocks. The `DocTestSetup = nothing` is not strictly
```
mortenpi: I started thinking that the no-state statement is not entirely accurate if you allow for labelled blocks, e.g. as in the following example:

````
```@meta
DocTestSetup = quote
    if isdefined(:foo)
        foo += 10
    else
        foo = 42
    end
end
```

```jldoctest
julia> foo += 1
43
```

```jldoctest
julia> foo += 1
43
```

```jldoctest label
julia> foo += 1
43
```

```jldoctest label
julia> foo += 1
54
```
````

But I am not sure if it is worth complicating the language here.

**Comment by fredrikekre** [comment]
I updated a bit, thanks for the comments @mortenpi. I think this is much more concise and more clear.

**Comment by fredrikekre** [comment]
On another note, why would these changes cause this when I build the docs?

```
 !! Reference for 'Setup-Code' could no be found. [src/man/syntax.md]
```

File: docs/src/man/doctests.md [hunk 5]
```diff
@@ -105,59 +102,101 @@ line where checking should stop, i.e.
...
 Doctests may require some setup code that must be evaluated prior to that of the actual
-example, but that should not be displayed in the final documentation. It could also be that
-several separate doctests require the same definitions. For both of these cases a `@meta`
-block containing a `DocTestSetup = ...` value can be used as follows:
+example, but that should not be displayed in the final documentation. For this purpose a
+`@meta` block containing a `DocTestSetup = ...` value can be used as follows:
 
-    ```jldoctest
-    julia> using DataFrames
+````markdown
```
mortenpi: Something going on here as well. I guess this bit can be dropped and the two paragraphs should be merged?

docs/src/man/doctests.md
```diff
@@ -105,59 +102,101 @@ line where checking should stop, i.e.
...
+
+````markdown
+```jldoctest
+julia> foo = 42
+42
 ```
 ````
 
-## Skipping Doctests
+In the doctest block below The variable `foo` will not be defined in the next block:
```
fredrikekre: I was a bit quick here... Meant to remove everything before `"The..."

docs/src/man/doctests.md
fredrikekre: oops

mortenpi:
Finally got around to merging this. Thanks again for the contribution @fredrikekre!

PR #419, opened 2017-01-29
Disable doctest colors

MichaelHatherly:


.travis.yml
```diff
@@ -5,7 +5,6 @@ os:
   - osx
 
 julia:
-  - 0.4
```
tkelman: be sure to do this in REQUIRE too

PR #424, opened 2017-02-08
Fix file refrences since build occurs in a tmp directory

### LINKED ISSUE #411: LatexWriter cannot find images generated from examples
I have a few `@example` sections in the documentation, which generate a PNG using Gadfly. That PNG is then referenced in the doc in markdown: ``

This works fine for HTML documents. However, when using the LatexWriter, the images are not found during pdf generation. I believe this is due to the generation happening in a temporary directory : https://github.com/JuliaDocs/Documenter.jl/blob/3e5e4792e6de672df4378d7d91ee1f376fc66aff/src/Writers/LaTeXWriter.jl#L46

When I manually run `latexmk` in the build directory (copying documenter.sty), everything works. 

Not sure what is the best way to fix this?

aviks:
Transform image references to absolute paths
copy assets directory to temporary build directory

This may not be the most elegant, but this is the only way I could fix this issue without a wholesale re-write of the structure of this codebase. So not sure if this is acceptable, but I thought I'd put this out there in any case. 

Fixes #411.

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -43,11 +43,10 @@ const DOCUMENT_STRUCTURE = (
 )
 
 function render(doc::Documents.Document)
-    task_local_storage(:doc, doc)
     mktempdir() do path
-        cd(path) do
+        cp(joinpath(doc.user.root, doc.user.build), joinpath(path, "build"))
+        cd(joinpath(path, "build")) do
             local file = replace("$(doc.user.sitename).tex", " ", "")
-            cp(joinpath(doc.user.root, doc.user.source, "assets"), "assets")
             open(file, "w") do io
                 local context = Context(io)
                 writeheader(context, doc)
@@ -87,7 +86,6 @@ function render(doc::Documents.Document)
             end
         end
     end
-    task_local_storage(:doc, nothing)
 end
 
 function writeheader(io::IO, doc::Documents.Document)
@@ -407,11 +405,10 @@ function latexinline(io::IO, md::Markdown.Italic)
 end
 
 function latexinline(io::IO, md::Markdown.Image)
-    doc = task_local_storage(:doc)
     wrapblock(io, "figure") do
         _println(io, "\\centering")
         wrapinline(io, "includegraphics") do
-            _print(io, joinpath(doc.user.root, doc.user.build, md.url))
+            _print(io,  md.url)
         end
         _println(io)
         wrapinline(io, "caption") do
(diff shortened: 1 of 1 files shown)


src/Writers/LaTeXWriter.jl
```diff
@@ -404,10 +407,11 @@ function latexinline(io::IO, md::Markdown.Italic)
 end
 
 function latexinline(io::IO, md::Markdown.Image)
+    doc = task_local_storage(:doc)
```
mortenpi: I think we could pass `doc` as the last argument to `latex` and `latexinline` everywhere. That would be in line with surrounding code and is basically how we do it `HTMLWriter`. `task_local_storage` is basically global state which I think we'd like to avoid.

src/Writers/LaTeXWriter.jl
```diff
@@ -404,10 +407,11 @@ function latexinline(io::IO, md::Markdown.Italic)
...
     wrapblock(io, "figure") do
         _println(io, "\\centering")
         wrapinline(io, "includegraphics") do
-            _print(io, md.url)
+            _print(io, joinpath(doc.user.root, doc.user.build, md.url))
```
mortenpi: Couldn't we copy the whole `build` over and avoid absolute paths altogether?

src/Writers/LaTeXWriter.jl
```diff
@@ -43,9 +43,11 @@ const DOCUMENT_STRUCTURE = (
 )
 
 function render(doc::Documents.Document)
+    task_local_storage(:doc, doc)
     mktempdir() do path
         cd(path) do
             local file = replace("$(doc.user.sitename).tex", " ", "")
+            cp(joinpath(doc.user.root, doc.user.source, "assets"), "assets")
```
mortenpi: These are static user assets, right? I.e. this is not related to _generated_ images per say?

aviks: Yes, true. It is yet another issue caused by generating the pdf in a tmp directory.

src/Writers/LaTeXWriter.jl
aviks: Thanks for the review. Yes, copying the whole build over will work. I've been wondering what option to choose in fixing this. If you think copying the whole `build` directory is an acceptable solution, then things are quite simple.

aviks:
Ok, I've changed this implementation to copy the `build` directory to the temporary LaTeX directory at @mortenpi 's suggestion. So the code is much simpler now.

aviks:
Superseded by #431

PR #426, opened 2017-02-10
Improvements to deploydocs

mortenpi:
Make sure that deploydocs runs git commands in the source directory (in
case of out-of-tree build) and handles git-rev-parse failures properly.

Ref: JuliaLang/julia#20519

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -312,7 +312,20 @@ function deploydocs(;
     travis_julia        = get(ENV, "TRAVIS_JULIA_VERSION", "")
 
     # Other variables.
-    sha = readchomp(`git rev-parse --short HEAD`)
+    sha = cd(root) do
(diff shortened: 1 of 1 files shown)


tkelman:
any update here?

mortenpi:
This could be merged. I was just waiting for @MichaelHatherly to sign off on it. I didn't feel that we're in a hurry with this per say.

However, it has been a while now, so it might be worth merging a few bugfixes (this, #434 I think, and I should look into making the deployment error messages a bit more helpful) and make a new patch release with them.

tkelman:
He's been traveling, right? It will be very helpful to have #434 available in a release before 0.6.0 beta.

mortenpi:
I guess so, I have no idea when he'll be back though. In any case, I'll tag a patch release with a couple of these PRs this weekend to keep things moving.

PR #431, opened 2017-02-22
Fix LaTeX file references (v2)

### LINKED ISSUE #411: LatexWriter cannot find images generated from examples

mortenpi:
Based on #424. Sorry, @aviks, I noticed an issue with the paths and got sort of carried away. I've retained your original contribution, I just merged the commits, removed a space and made small edits to the commit messages. I hope it is fine that it gets merged this way.

Closes #424, fixes #411.

### DIFF
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -65,6 +65,14 @@ end
     @test OuterModule in Documenter.Utilities.submodules(OuterModule)
     @test OuterModule.InnerModule in Documenter.Utilities.submodules(OuterModule)
     @test length(Documenter.Utilities.submodules(OuterModule)) == 2
+
+    @test Documenter.Utilities.isabsurl("file.md") === false
+    @test Documenter.Utilities.isabsurl("../file.md") === false
+    @test Documenter.Utilities.isabsurl(".") === false
+    @test Documenter.Utilities.isabsurl("https://example.org/file.md") === true
+    @test Documenter.Utilities.isabsurl("http://example.org") === true
+    @test Documenter.Utilities.isabsurl("ftp://user:pw@example.org") === true
+    @test Documenter.Utilities.isabsurl("/fs/absolute/path") === false
 end
 
 end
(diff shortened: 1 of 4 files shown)


test/utilities.jl
```diff
@@ -65,6 +65,13 @@ end
     @test OuterModule in Documenter.Utilities.submodules(OuterModule)
     @test OuterModule.InnerModule in Documenter.Utilities.submodules(OuterModule)
     @test length(Documenter.Utilities.submodules(OuterModule)) == 2
+
+    @test Documenter.Utilities.isabsurl("file.md") === false
+    @test Documenter.Utilities.isabsurl("../file.md") === false
+    @test Documenter.Utilities.isabsurl(".") === false
+    @test Documenter.Utilities.isabsurl("https://example.org/file.md") === true
+    @test Documenter.Utilities.isabsurl("http://example.org") === true
+    @test Documenter.Utilities.isabsurl("ftp://user:pw@example.org") === true
```
aviks: Add a test for a filesystem absolute URL eg. `/home/me/myself.md` ??

aviks:
Thanks Morten.

aviks:
Bump. Anything we can do speed this along?

test/utilities.jl
mortenpi: That's actually a relative URL technically, but a test would be useful nevertheless.

The interpretation of such an URL should be that they're relative to the root of the document (and not relative to the filesystem root). I added a special case for that to the LaTeX code as well.

mortenpi:
I am planning to tag a patch version this weekend with a couple of the bugfix PRs and I'll include this as well. Sorry for being a bit slow on this.

PR #434, opened 2017-02-28
Replace backslashes in href paths with forward slashes

tkelman:
And Cygwin paths with Windows paths, if applicable

Working towards making the doc build of Base give the same
output whether run on Linux or Windows

Remaining discrepancies:
- build_sysimg source link isn't getting added yet on Windows
- method ordering differs

For code-signing reasons, I need to prepare release binaries from Windows. With the new Documenter-based docs, the source tarballs will therefore end up with html docs built from Windows. Many of the relative links in such a build have backslashes in them, which should probably be avoided in html href links.

I could use help in how to test this properly. It is possible to use Cygwin on AppVeyor, but a bit annoying. Maybe Base's tests would be a better place to do this? Would require some cross-CI coordination, and storing the built docs somewhere.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -363,12 +363,31 @@ end
 
 # Finding URLs -- based partially on code from the main Julia repo in `base/methodshow.jl`.
 
+function in_cygwin()
+    # check if in cygwin, if so we need to translate cygwin paths to windows paths
+    if is_windows()
+        try
+            return success(`cygpath -h`)
+        catch
+            return false
+        end
+    else
+        return false
+    end
+end
+
 function url(repo, file)
     file = abspath(file)
     remote = getremote(dirname(file))
     isempty(repo) && (repo = "https://github.com/$remote/tree/{commit}{path}")
+    # Replace any backslashes in links, if building the docs on Windows
+    file = replace(file, '\\', '/')
     commit, root = cd(dirname(file)) do
-        readchomp(`git rev-parse HEAD`), readchomp(`git rev-parse --show-toplevel`)
+        toplevel = readchomp(`git rev-parse --show-toplevel`)
+        if in_cygwin()
+            toplevel = readchomp(`cygpath -m "$toplevel"`)
+        end
+        readchomp(`git rev-parse HEAD`), toplevel
     end
     if startswith(file, root)
         _, path = split(file, root; limit = 2)
@@ -387,6 +406,8 @@ if VERSION >= v"0.5.0-dev+3442"
     function url(remote, repo, mod, file, linerange)
         remote = getremote(dirname(file))
         isabspath(file) && isempty(remote) && isempty(repo) && return Nullable{Compat.String}()
+        # Replace any backslashes in links, if building the docs on Windows
+        file = replace(file, '\\', '/')
         # Format the line range.
         line = format_line(linerange)
         # Macro-generated methods such as those produced by `@deprecate` list their file as
@@ -405,7 +426,11 @@ if VERSION >= v"0.5.0-dev+3442"
             )
         else
             commit, root = cd(dirname(file)) do
-                readchomp(`git rev-parse HEAD`), readchomp(`git rev-parse --show-toplevel`)
+                toplevel = readchomp(`git rev-parse --show-toplevel`)
+                if in_cygwin()
+                    toplevel = readchomp(`cygpath -m "$toplevel"`)
+                end
+                readchomp(`git rev-parse HEAD`), toplevel
             end
             if startswith(file, root)
                 if isempty(repo)
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -362,9 +362,42 @@ else
 end
 
 # Finding URLs -- based partially on code from the main Julia repo in `base/methodshow.jl`.
+#
+# Paths on Windows contain backslashes, so the `url` function needs to take care of them.
+# However, the exact formatting is not consistent and depends on whether Julia runs
+# separately or in Cygwin.
+#
+# We get paths from Julia's docsystem, e.g.
+#     Docs.docstr(Docs.Binding(Documenter,:Documenter)).data[:path]
+# and from git
+#     git rev-parse --show-toplevel
+#
+# * Ordinary Windows binaries (both Julia and git)
+#   In this case the paths from the docsystem are Windows-like with backslashes, e.g.:
+#       C:\Users\<user>\.julia\v0.6\Documenter\src\Documenter.jl
+#   But the paths from git have forward slashes, e.g.:
+#       C:/Users/<user>julia/v0.6/Documenter
+#
+# * Running under Cygwin
+#   The paths from the docsystem are the same as before, e.g.:
+#       C:\Users\<user>\.julia\v0.6\Documenter\src\Documenter.jl
+#   However, git returns UNIX-y paths using a /cygdrive mount, e.g.:
+#       /cygdrive/c/<user>/.julia/v0.6/Documenter
+#   We can fix that with `cygpath -m <path>` to get a Windows-like path with forward
+#   slashes, e.g.:
+#       C:/Users/<user>/.julia/v0.6/Documenter)
+#
+# In the docsystem paths we replace the backslashes with forward slashes before we start
+# comparing paths with the ones from git.
+#
 
+"""
+    in_cygwin()
+
+Check if we're running under cygwin. Useful when we need to translate cygwin paths to
+windows paths.
+"""
 function in_cygwin()
-    # check if in cygwin, if so we need to translate cygwin paths to windows paths
     if is_windows()
         try
             return success(`cygpath -h`)
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -136,7 +136,9 @@ function copy_asset(file, doc)
         ispath(dst) && Utilities.warn("Overwriting '$dst'.")
         cp(src, dst, remove_destination=true)
     end
-    normpath(joinpath("assets", file))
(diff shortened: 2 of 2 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -584,6 +588,8 @@ function domify_doc(ctx, navnode, md::Markdown.MD)
             ret = Any[domify(ctx, navnode, Writers.MarkdownWriter.dropheaders(markdown))]
             # When a source link is available then print the link.
             Utilities.unwrap(Utilities.url(ctx.doc.internal.remote, ctx.doc.user.repo, result)) do url
+                # Replace any backslashes in links, if building the docs on Windows
+                url = replace(url, '\\', '/')
```
mortenpi: Couldn't we move this into the 5-argument `url` function (related to the previous comment)?

src/Writers/HTMLWriter.jl
```diff
@@ -740,6 +746,8 @@ function mdconvert(link::Markdown.Link, parent)
         path = first(s)
         path = endswith(path, ".md") ? Formats.extension(:html, path) : path
         url = (length(s) > 1) ? "$path#$(last(s))" : Compat.String(path)
+        # Replace any backslashes in links, if building the docs on Windows
+        url = replace(url, '\\', '/')
```
mortenpi: I'd say that the `link.url` here should already be without backslashes. If that's not the case then there should be a `replace` somewhere earlier.

The only way we should get backslashes here is if the user has them in the URL, in which case it's an user error. We could then either warn, replace them but in the parsing step, or just ignore and assume that that's what the user wants for some reason.

tkelman: who's the "user" here? a lot of these come from file paths which have the backslashes in them

mortenpi: I meant the person writing the Markdown files and docstrings. Any generated URLs should be sanitized already. But tracking this down could be a bit of work, so it would be fine to keep this here as it is, and figure this detail out later.

src/Utilities/Utilities.jl
```diff
@@ -363,15 +363,34 @@ end
...
+end
+
 function url(repo, file)
     file = abspath(file)
     remote = getremote(dirname(file))
     isempty(repo) && (repo = "https://github.com/$remote/tree/{commit}{path}")
     commit, root = cd(dirname(file)) do
-        readchomp(`git rev-parse HEAD`), readchomp(`git rev-parse --show-toplevel`)
+        toplevel = readchomp(`git rev-parse --show-toplevel`)
+        if in_cygwin()
```
mortenpi: When running with the ordinary Julia Windows binary (i.e. not in cygwin) then the source links are still missing. Should we maybe have an `elseif is_windows(); replace(...); end` to emulate `cygpath` in this case?

tkelman: where's your git from? you often won't have git on the path in which case the rev-parse will fail.

mortenpi: From the ordinary Windows installer, with the minimal paths setup. It might make sense to check for the existence of `git` in `makedocs` and `deploydocs` because we implicitly assume it, and it can often be missing as you pointed out.

src/Writers/HTMLWriter.jl
tkelman: It looks like that would change links to the backslash function, which we don't want.

tkelman: Whoops, guess this current version is also doing that, but in fewer places.

src/Utilities/Utilities.jl
tkelman: what format does `git rev-parse --show-toplevel` output in that case?

src/Writers/HTMLWriter.jl
tkelman: fixed by making sure to only apply replacement to the left of any `#` link anchors

src/Utilities/Utilities.jl
mortenpi:
```
julia> run(`git rev-parse --show-toplevel`)
C:/Users/morten/.julia/v0.5/Documenter
```

tkelman: the `startswith` should probably work now then, since I moved the replacement to before that check

mortenpi: Agreed. However, I think I did come up with another special case: when the packages are installed under Cygwin (by changing `JULIA_PKGDIR`), then the paths in the docsystem are
```
julia> Docs.docstr(Docs.Binding(Documenter,:Documenter)).data[:path] |> print
\home\morten\julia\v0.6\Documenter\src\Documenter.jl
```
but `git rev-parse --show-toplevel` + `cygpath` adds a drive

```
shell> git rev-parse --show-toplevel
/cygdrive/c/home/morten/julia/v0.6/Documenter

shell> cygpath -m /cygdrive/c/home/morten/julia/v0.6/Documenter
C:/home/morten/julia/v0.6/Documenter
```
which should make `startswith` fail.

tkelman: Where does `\home\` with backslashes come from? Nothing in Julia should create that kind of path.

mortenpi: I am setting `JULIA_PKGDIR=$HOME/julia`. It seems that then Julia gets UNIX-y paths, but uses the Windows path separator to join them together, or something along those lines? Already the `Pkg.dir()` is weird:
```
$ JULIA_PKGDIR=$HOME/julia /cygdrive/c/Users/morten/julia/julia-bb76add105/bin/julia
julia> print(Pkg.dir())
\home\morten\julia\v0.6
```

tkelman: cygwin's `$HOME` is going to be a path style that Windows Julia isn't going to interpret as an absolute path, so this might be a variant of https://github.com/JuliaLang/julia/issues/20896

mortenpi: Ok, I just realized that Julia interprets this as `C:\home\morten\julia`, and not as the location of the Cygwin home (which in my case should be `C:\cygwin64\home\morten\julia`). Should we go for an issue in Base?

tkelman: I think that's correct, if confusing, behavior when you leave off the drive letter on a Windows path.

tkelman: If you want a version of Cygwin `$HOME` that Windows Julia can understand, you would do `JULIA_PKGDIR=$(cygpath -w $HOME)/julia`

mortenpi: Ah, yes, that works, and makes sense. I get proper full paths then. I guess my only "complaint" would be that perhaps Julia should add the drive letter back when it gets "\something" (or "/something", where I guess the slashes just get converted)?

src/Utilities/Utilities.jl
```diff
@@ -363,12 +363,31 @@ end
 
 # Finding URLs -- based partially on code from the main Julia repo in `base/methodshow.jl`.
```
mortenpi: I was thinking that it wouldn't hurt to document this path stuff here since it's not necessarily obvious what's going on in url function. I think it could be more concise and worded better, but I'd propose something along the lines:

```
# Paths on Windows contain backslashes that we need to get rid of in the URLs.
# However, the exact formatting is not consistent and depends on whether Julia runs separately
# or in Cygwin etc.
#
# We get paths from Julia's docsystem, e.g.
#     Docs.docstr(Docs.Binding(Documenter,:Documenter)).data[:path]
# and also from git
#     git rev-parse --show-toplevel
#
# * Ordinary Windows binaries (both Julia and git)
#   In this case the paths from the docsystem are Windows-like with backslashes (e.g. C:\Users\<user>\.julia\v0.6\Documenter\src\Documenter.jl),
#   but the paths from git have forward slashes (e.g. C:/Users/<user>julia/v0.6/Documenter).
#
# * Running under Cygwin, with files outside of Cygwin
#   The paths from the docsystem are still Windows-like with backslashes (e.g. C:\Users\<user>\.julia\v0.6\Documenter\src\Documenter.jl).
#   However, git returns a UNIX-y path using the /cygdrive mount (e.g. /cygdrive/c/<user>/.julia/v0.6/Documenter).
#   We fix that with `cygpath -m <path>` to get a Windows-like path with forward slashes (e.g. C:/Users/<user>/.julia/v0.6/Documenter)
```

src/Utilities/Utilities.jl
tkelman: I think there's a specific meaning for that, "relative to the current drive" or something like that - it's a bit obscure and not so useful, but if you propose changing it you'll likely have to convince `@vtjnash`.

mortenpi: I'll sleep on that. In any case, since this is unlikely to come up in practice, I think we can ignore this here (i.e. that `startswith` would fail in this case).

src/Utilities/Utilities.jl
tkelman: I'm likely to just copy-paste whatever you suggest, so easier to get the git attribution right if you edit as you please.

mortenpi:
I don't see any backslashes in the Windows builds anymore. Is there anything besides tests that should be done here?

tkelman:
Not that I can think of. We may want to open a separate issue about the method ordering differences.

tkelman:
Thoughts on the best way to test this? As I said, I'd really strongly prefer if this were available in a release before tagging a beta of 0.6.

mortenpi:
Not off the top off my head at the moment. But I'd be happy to merge this as is and tag a new version. We can open an issue for the ordering and tests.

By the way, any ideas what happens to attributions when I squash merge this? Or, alternatively, would you mind squashing your commits?

tkelman:
squash merge combines commit authorship last I checked

squashed my commits.

mortenpi:
Thanks @tkelman!

PR #435, opened 2017-03-01
Specify repo URL format for deploydocs()

christopher-dG:
I just spent much longer than I care to admit troubleshooting why `deploydocs` couldn't fetch from my repo... it turns out I absentmindedly set `repo` to `https://github.com/my/repo.git` which Documenter converted to `git@https://github.com:my/repo.git`. I thought it might be helpful to explicitly tell people to avoid that.

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -121,7 +121,8 @@ deploydocs(
 )
 ```
 
-where `USER_NAME` and `PACKAGE_NAME` must be set to the appropriate names.
+where `USER_NAME` and `PACKAGE_NAME` must be set to the appropriate names. Note that `repo`
+should not specify any protocol, i.e. it should not begin with `https://` or `git@`. 
 
 By default `deploydocs` will deploy the documentation from the `nightly` Julia build for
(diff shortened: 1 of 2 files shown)


mortenpi:
Thanks, always good to amend the docs!

Could you also add the same clarification to the [`deploydocs` docstring](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Documenter.jl#L250-L256)?

On a slight tangent, I've been wondering whether it would be better to use something like `repo = GitHub("JuliaDocs", "Documenter.jl")` instead, so it would be more clear what you're specifying and it would be easier for Documenter to interpret that. We could still fall back to strings for repositories that are hosted somewhere other than GitHub.

christopher-dG:
Done. About changing the format, I'm not really sure that the benefit outweighs the cost, especially since the problem it solves is already being more or less solved via documentation. Also, I like that it's currently consistent no matter where you host your repo.

mortenpi:
LGTM, thanks! Will merge in the not too distant future. Regarding the format -- yup, you're probably right.

PR #441, opened 2017-03-07
Use repo to set 'Edit on...' button for HTMLWriter

christopher-dG:
This:
gh
Becomes this:
gl

And so on for Bitbucket, dependent on the value of `repo` in `makedocs`.


mortenpi:
Thanks!

PR #446, opened 2017-03-14
Enable handling of 'where' expressions and fix deprecations

### LINKED ISSUE #445: [Julia 0.6] Wrong category for UnionAll types; Constant instead of Type
```julia
"""
This is a Blob
"""
struct Blob{T} end
```
On 0.5:
image

On 0.6:
image

helgee:
x-ref #445

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.5
-Compat 0.8
+Compat 0.17
 DocStringExtensions 0.2
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -300,6 +300,9 @@ doccat(b::Binding, ::Type)  = "Method"
 
 doccat(::Function) = "Function"
 doccat(::DataType) = "Type"
+if VERSION >= v"0.6.0-dev.2122"
+    doccat(::UnionAll) = "Type"
+end
 doccat(::Module)   = "Module"
 doccat(::Any)      = "Constant"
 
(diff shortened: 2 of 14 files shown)


src/Utilities/Utilities.jl
```diff
@@ -300,6 +300,9 @@ doccat(b::Binding, ::Type)  = "Method"
 
 doccat(::Function) = "Function"
 doccat(::DataType) = "Type"
+if VERSION >= v"0.6.0-dev.2122"
```
mortenpi: Not sure actually which is better practice, but `isdefined(Base, :UnionAll)` should work here as well, right?

helgee:
I couldn't resist to also fix the low-hanging deprecations.

helgee:
There is only the following deprecation warning left:
```
WARNING: deprecated syntax "inner constructor Diff(...) around /Users/helge/.julia/v0.6/Documenter/src/Utilities/TextDiff.jl:71".
Use "Diff{T}(...) where T" instead.
```
But I am not sure how to handle this one.

helgee:
I fixed the last deprecation warning.

helgee:
BTW I also added the requested tests. Anything left to do?

REQUIRE
```diff
@@ -1,3 +1,3 @@
 julia 0.5
-Compat 0.8
+Compat 0.18
```
tkelman: doesn't look like any of the type aliases are parametric, so 0.17 should be enough

helgee:
Fixed.

mortenpi:
Looks good, thanks for doing the 0.6 stuff as well. One last request -- could you squash your commits into two, one for the deprecations and one for the `where` thing?

helgee:
Done 👍

mortenpi:
Thanks!

PR #449, opened 2017-03-16
responsive side navigation

wookay:
It's too hard to read the docs on mobile,
I did just make nav.toc to be relative.

Here's an example.
https://wookay.github.io/Documentery/html/

thanks.

### DIFF
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -278,6 +278,10 @@ article {
 
 article > header {}
 
+article > header div#topbar {
+    display: none;
+}
+
 article > header nav ul {
     display: inline-block;
     list-style: none;
@@ -432,4 +436,106 @@ article section.docstring a.source-link {
 .hljs {
     background-color: transparent;
     padding: 0;
+}
+
+@media only screen and (max-width: 768px) {
+    nav.toc {
+        position: fixed;
+        overflow-y: scroll;
+        width: 16em;
+        left: -16em;
+        -webkit-overflow-scrolling: touch;
+        -webkit-transition-property: left; /* Safari */
+        -webkit-transition-duration: 0.3s; /* Safari */
+        transition-property: left;
+        transition-duration: 0.3s;
+        -webkit-transition-timing-function: ease-out; /* Safari */
+        transition-timing-function: ease-out;
+        z-index: 2;
+    }
+
+    nav.toc.show {
+        left: 0;
+    }
+
+    article {
+        margin-left: 0;
+        padding: 3em 0.9em 0 0.9em; /* top right bottom left */
+        overflow-wrap: break-word;
+    }
+
+    article > header {
+        position: fixed;
+        left: 0;
+        z-index: 1;
+    }
+
+    article > header nav, hr {
+        display: none;
+    }
+
+    article > header div#topbar {
+        display: block; /* is mobile */
+        position: fixed;
+        width: 100%;
+        height: 1.5em;
+        padding-top: 1em;
+        padding-bottom: 1em;
+        background-color: #fcfcfc;
+        box-shadow: 0 1px 3px rgba(0,0,0,.26);
+        top: 0;
+        -webkit-transition-property: top; /* Safari */
+        -webkit-transition-duration: 0.3s; /* Safari */
+        transition-property: top;
+        transition-duration: 0.3s;
+    }
+
+    article > header div#topbar.headroom--unpinned.headroom--not-top.headroom--not-bottom {
+        top: -4em;
+        -webkit-transition-property: top; /* Safari */
+        -webkit-transition-duration: 0.7s; /* Safari */
+        transition-property: top;
+        transition-duration: 0.7s;
+    }
+
+    article > header div#topbar span {
+        position: fixed;
+        width: 80%;
+        height: 1.5em;
+        margin-top: -0.1em;
+        margin-left: 0.9em;
+        font-size: 1.2em;
+        overflow: hidden;
+    }
+
+    article > header div#topbar a.fa-bars {
+        float: right;
+        padding: 0.6em;
+        margin-top: -0.6em;
+        margin-right: 0.3em;
+        font-size: 1.5em;
+    }
+
+    article > header div#topbar a.fa-bars:visited {
+        color: #3091d1;
+    }
+
+    article table {
+        overflow-x: auto;
+        display: block;
+    }
+
+    article div.MathJax_Display {
+        overflow: scroll;
+    }
+
+    article span.MathJax {
+        overflow: hidden;
+    }
+}
+
+@media only screen and (max-width: 320px) {
+    body {
+        font-size: 15px;
+    }
 }
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -376,7 +376,7 @@ end
 # ------------------------------------------------------------------------------
 
 function render_article(ctx, navnode)
-    @tags article header footer nav ul li hr span a
+    @tags article header footer nav ul li hr span a div
 
     header_links = map(Documents.navpath(navnode)) do nn
         title = mdconvert(pagetitle(ctx, nn))
@@ -398,7 +398,9 @@ function render_article(ctx, navnode)
     Utilities.unwrap(Utilities.url(ctx.doc.user.repo, getpage(ctx, navnode).source)) do url
         push!(topnav.nodes, a[".edit-page", :href => url](span[".fa"](logo), " Edit on $host"))
     end
-    art_header = header(topnav, hr())
+    page_title = string(mdflatten(pagetitle(ctx, navnode)))
+    topbar = div["#topbar"](span(page_title), a[".fa .fa-bars", :href => "#"])
+    art_header = header(topnav, hr(), topbar)
 
     # build the footer with nav links
     art_footer = footer(hr())
--- a/assets/html/documenter.js
+++ b/assets/html/documenter.js
@@ -9,6 +9,7 @@ requirejs.config({
     paths: {
         'jquery': 'https://code.jquery.com/jquery-3.1.0.js?',
         'jqueryui': 'https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.0/jquery-ui.min',
+        'headroom': 'https://cdnjs.cloudflare.com/ajax/libs/headroom/0.9.3/headroom.min',
         'mathjax': 'https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML',
         'highlight': 'https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/highlight.min',
         'highlight-julia': 'https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/languages/julia.min',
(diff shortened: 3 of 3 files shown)


mortenpi:
Thanks for tackling this. Mobile UX is certainly somethings that's on the todolist. A couple of things that I noticed:
- We can probably have it be 100% wide on mobile screens, and also drop the shadow thing on the right. I.e. that the menu would cover the whole screen
- How hard do you think it would be to have the page header show up when the user pulls the page down? I.e. that you wouldn't have to scroll all the way up to open the menu. E.g. something like the [Material theme does for MkDocs](https://juliaplots.github.io/install/).

wookay:
> We can probably have it be 100% wide on mobile screens, and also drop the shadow thing on the right. I.e. that the menu would cover the whole screen

it's not perfect, added the topbar. see the Julia Documentation example.
https://wookay.github.io/Documentery/en/

> How hard do you think it would be to have the page header show up when the user pulls the page down? I.e. that you wouldn't have to scroll all the way up to open the menu. E.g. something like the Material theme does for MkDocs.

then we need more javascript framework like bootstrap. mkdocs looks really good.

randy3k:
It looks very good. Is there any reason of not merging this (besides the merge conflict)?

mortenpi:
Sorry, I just haven't gotten around to reviewing this. I'll try to go through this in the next few days, hopefully to get this released as part of 0.10. Together with #409 this will be a nice set of updates to the frontend bit.

src/Writers/HTMLWriter.jl
```diff
@@ -160,18 +160,19 @@ end
...
     head = render_head(ctx, navnode)
+    topbar = render_topbar(ctx, navnode)
     navmenu = render_navmenu(ctx, navnode)
     article = render_article(ctx, navnode)
 
     htmldoc = DOM.HTMLDocument(
         html[:lang=>"en"](
             head,
-            body(navmenu, article)
+            body(topbar, div[".main"](div[".container"](navmenu, article)))
```
mortenpi: I would prefer avoiding unnecessary `div`s. They're not used in the style, so as far as I can tell they could be dropped. A `body(topbar, navmenu, article)` ought to do.

assets/html/documenter.css
```diff
@@ -428,3 +432,98 @@ article section.docstring a.source-link {
...
     padding: 0;
 }
+
+@media only screen and (max-width: 768px) {
+    div.topbar {
+        display: block; /* is mobile */
+        position: fixed;
+        top: 0px;
+        left: 0px;
+        right: 0px;
```
mortenpi: Should drop the units from the zeroes here.

src/Writers/HTMLWriter.jl
```diff
@@ -275,6 +276,20 @@ function render_search(ctx)
...
+# topbar for mobile
+# ------------------------------------------------------------------------------
+
+function render_topbar(ctx, navnode)
+    @tags div ul li span a i
+
+    page_title = string(mdflatten(pagetitle(ctx, navnode)))
+    topbarleft = ul[".float-left"](li(span(page_title)))
+    topbarright = ul[".float-right"](li(a[".btn", :href => "#"](i[".fa .fa-bars", Symbol("aria-hidden") => "true"])))
+    div[".topbar"](div[".container"](topbarleft, topbarright))
```
mortenpi: As above -- any reason we couldn't just drop the `.container` `<div>`?

src/Writers/HTMLWriter.jl
```diff
@@ -275,6 +276,20 @@ function render_search(ctx)
...
+
+# topbar for mobile
+# ------------------------------------------------------------------------------
+
+function render_topbar(ctx, navnode)
+    @tags div ul li span a i
+
+    page_title = string(mdflatten(pagetitle(ctx, navnode)))
+    topbarleft = ul[".float-left"](li(span(page_title)))
+    topbarright = ul[".float-right"](li(a[".btn", :href => "#"](i[".fa .fa-bars", Symbol("aria-hidden") => "true"])))
```
mortenpi: Why is the `aria-hidden` here? Based on [this article](https://www.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/) it seems that `aria-hidden` is not very useful and, if anything, we should use HTML5's `hidden` attribute.

Also, why use `ul`s here, this does not feel like a list to me? A simple `<div id="topbar">{PAGE_TITLE}<a class="btn fa fa-bars"></a></div>` ought to do the trick here.

mortenpi:
Playing around with it now I quite like the top bar and the ~80% width of the menu. But I do have a couple of items of feedback, so I hope you'd up for iterating on this PR a bit.

* Clicking on (touching?) the main page behind the menu should probably close the menu (e.g. how it's done [here](https://material.io/guidelines/))
* If the page is wider than the screen then it breaks a bit, [e.g. on the Examples page](http://mortenpi.eu/Documenter.jl/mobile-css-deploy/man/examples.html).
* Also, if the screen is very wide (e.g. phone in landscape), I feel that the menu probably shouldn't extend to 80%, but only be as wide as the text basically (e.g. the Material theme).
* The `.topbar` actually duplicates the `<header>` tag. Easiest could be to just hide the `header` in mobile mode. Or alternatively use the `header` tag for both things. I think it's fine if we drop the "Edit on GitHub" link on mobiles.
   
* If you're scrolled down and you click on the menu button (), it jumps back to the beginning of the page. Probably would be nice if wouldn't do that.
* The menu seems to scroll together with the page even if the menu is hidden. Would better if it didn't I think, since in practice this means that you are at the bottom of the menu every time you open it up.

wookay:
thanks for your reviews! I would investigate it tomorrow.

src/Writers/HTMLWriter.jl
wookay: just updated, used `span` tag to align the PAGE_TITLE in the css (`<span>{PAGE_TITLE}</span>`).

wookay:
> If you're scrolled down and you click on the menu button (), it jumps back to the beginning of the page. Probably would be nice if wouldn't do that.

I've been used headroom.js (http://wicky.nillia.ms/headroom.js) for this.

assets/html/documenter.css
```diff
@@ -428,3 +432,98 @@ article section.docstring a.source-link {
...
     padding: 0;
 }
+
+@media only screen and (max-width: 768px) {
+    div#topbar {
+        display: block; /* is mobile */
+        position: fixed;
+        width: 100%;
+        height: 2.7em;
+        background-color: white;
```
mortenpi: I would also give the `#topbar` a slight background color as well, probably the same as the menu `#fcfcfc`.

assets/html/documenter.css
```diff
@@ -428,3 +432,98 @@ article section.docstring a.source-link {
     background-color: transparent;
     padding: 0;
 }
+
+@media only screen and (max-width: 768px) {
+    div#topbar {
+        display: block; /* is mobile */
+        position: fixed;
+        width: 100%;
+        height: 2.7em;
```
mortenpi: I would make the `#topbar` a bit wider in the vertical direction.

If I may I would propose a way to set the vertical heights in the following way, which I think would be a bit more clear:

* Have `padding-top: 1em` and `padding-bottom: 1em` on the `#topbar`.
* Drop `margin-top` from `div#topbar span` and only keep `padding-right: 1em` for `div#topbar a.fa-bars`
* Have an explicit `font-size: 1.5em` on the `div#topbar a.fa-bars` (as opposed `x-large`) -- no other reason than to have an explicit number here.
* The navlink text will be smaller, so for the `#topbar` it should also be `height: 1.5em`.
* The total height of the `#topbar` will then be 3.5em, so we should also have a `padding-top: 3em` on the `article`, since the default 2em will not be enough.
* The headroom should also then probably be something like `top: -4em`.

assets/html/documenter.css
```diff
@@ -428,3 +432,98 @@ article section.docstring a.source-link {
...
+
+    article span.MathJax {
+        overflow: hidden;
+    }
+
+    article > header {
+        display: none;
+    }
+
+    article [id]:before {
```
mortenpi: Confused a bit here -- why do we need this one?

assets/html/documenter.css
```diff
@@ -428,3 +432,98 @@ article section.docstring a.source-link {
...
+
+    article {
+        margin-left: 0em;
+        padding-left: 0.9em;
+        padding-right: 0.9em;
+        overflow-wrap: break-word;
+        z-index: 1;
+    }
+
+    article table {
```
mortenpi: Also, why do we have this one exactly? Maybe you could add a few comments here to explain why/how we need to give some elements special treatment to make sure they render properly on mobile (i.e. so that someone not familiar with mobile UX could get up to speed quickly).

wookay:
```css
    article table {
        table-layout: fixed;
        width: 99%;
    }
```
used this to fix the table overflowing such as
https://wookay.github.io/Documentery/en/stdlib/linalg.html#Base.LinAlg.qrfact
screen shot 2017-04-25 at 9 35 49 pm

assets/html/documenter.css
wookay: oh, it doesn't need anymore, I'll fix.

wookay:
thanks a lot for the details!
I have applied above reviews, and
- move the `div#topbar` in the article header.
- `a.fa-bars` has padding around 0.6em to enlarge the touch area.

wookay:
I just found a tool, Sizzy (https://github.com/kitze/sizzy)
here are links to browse around.
https://sizzy.co/?url=https%3A%2F%2Fwookay.github.io%2FDocumentery%2Fen%2F
https://sizzy.co/?url=https%3A%2F%2Fwookay.github.io%2FDocumentery%2Fhtml%2F

assets/html/documenter.css
wookay: it was used to prepend the spacing of the article.
when touching the url with the fragment identifier such as 
https://wookay.github.io/Documentery/en/manual/strings.html#man-characters-1
unfortunately, sometimes h2 anchor gets hidden over by topbar.

jebej:
Is this ready to go?

assets/html/documenter.css
```diff
@@ -433,3 +437,105 @@ article section.docstring a.source-link {
...
+        margin-top: -0.6em;
+        margin-right: 0.3em;
+        font-size: 1.5em;
+    }
+
+    article > header div#topbar a.fa-bars:visited {
+        color: #3091d1;
+    }
+
+    article table {
```
mortenpi: Tables that should be wide become quite ugly on narrower screens at the moment (e.g. https://sizzy.co/?url=https%3A%2F%2Fwookay.github.io%2FDocumentery%2Fen%2Fstdlib%2Flinalg.html%23Base.LinAlg.qrfact).

Perhaps having `overflow-x: auto; display: block;` on a `table` would be better? This should allow the user to scroll wide tables instead. Any thoughts on this?

wookay: updated that table styles, it looks much better.

wookay:
rebased past commits :)

KristofferC:
This is awesome!

mortenpi:
Thank you very much for the contribution, for catering to all my requests and being patient with my slow reviews, @wookay! I think it's a really great addition and we can now finally strike this long standing item off the todo list :smile:

wookay:
you're welcome. it is really appreciated at all.
I'm happy to help the improvement.

PR #453, opened 2017-03-20
Backport #446 to 0.9

mortenpi:
For 0.9.2 release. Ref #446.


mortenpi:
@helgee is it fine with you if we keep the deprecation warnings in the 0.9 branch and get rid of them once we get 0.10 out (not sure when that will happen though)? It's seems to be a non-essential change, so I wouldn't backport it.

helgee:
Fine with me 👍

PR #454, opened 2017-03-21
Support for RawHTML type as a result in an  block

lucianolorenti:
If a an @eval block returns a RawHTML
```
```@eval
import Documenter.Utilities.DOM: Tag
import Documenter.Documents.RawHTML
a =  Tag(:figure)[:class=>"figure"]( Tag(:img)[:src => "my_image.png", alt="Alt text"],     Tag(:figcaption)("My Caption"))
RawHTML(string(a))
```

 it will output
```html
<figure class="figure"><img src="my_image.png"><figcaption>My Caption</figcaption></figure>
```
but not in an @example block. The HTML code is converted to string. Now boths blocks have the same behaviour. I think it was not possible to return an HTML code in an @example block before.
```
Now if you have an  @example block:
```@example
import Documenter.Utilities.DOM: DOM, Tag, @tags
import Documenter.Documents.RawHTML

a =  Tag(:figure)[:class=>"figure"]( Tag(:img)[:src => "my_image.png", alt="Alt text"], Tag(:figcaption)("My Caption"))
RawHTML(string(a))
```
It will output
```
[The code] 
```
and 
```html
<figure class="figure"><img src="my_image.png"><figcaption>My Caption</figcaption></figure>
```

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -17,10 +17,10 @@ import .Documents:
     DocsNode,
     DocsNodes,
     EvalNode,
-    MetaNode
+    MetaNode,
+    RawNode
 
 using Compat
-
 
 function expand(doc::Documents.Document)
     for (src, page) in doc.internal.pages
@@ -30,7 +30,6 @@ function expand(doc::Documents.Document)
         end
     end
 end
-
 
 # Expander Pipeline.
 # ------------------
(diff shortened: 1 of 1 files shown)


mortenpi:
I am still not sure whether `@example` blocks should special-case the `RawHTML` block. I feel that the block's behavior should be as simple as possible by default, which would mean that it should show whatever would show up in the REPL. That being said, we are already special-casing SVG images. The other thing would be that, forcing the user to wrap the outputs in `RawHTML` objects doesn't look very pretty, I think.

One alternative I could propose is a flag on the `@example` block, e.g.
````
```@example html
Something()
```
````
And then `Something` would be mimewritable to `text/html`. Documenter would then check for that and output the HTML. But I'm not sure whether that's actually a better approach.

Is your use case somewhere publicly accessible by the way?

lucianolorenti:
Unfortunately my use case is not publicly available yet. 

I think your approach is better. [I've implemented what you suggested](https://github.com/lucianolorenti/Documenter.jl/commit/8256fa68c776a06a240ff666cd286fcc6566473d) and is clearly superior. 
Now I can write

    ```@example html
        HTML("<div> aasdasd "</div>")
    ```
or 

    ```@example myname html
    HTML("<div> aasdasd "</div>")
    ```

And works as I expect.

fingolfin:
This PR was lasted updated in 2017, over 8 years ago.

Is there any point in keeping this open at this point? As in: is there still a usecase, is anyone interested in reviving this, is having the code in here going to be helpful, considering all that changed since then?

If this is  meant to be kept around, I'd like to voice my objections to the `@example html` syntax, it clashes with the existing `@example block-name` syntax and could in fact even be breaking for some projects. A different way to set a config setting here would be needed. Perhaps `@example block-name; html = true`  resp. `@example ; html = true` ?

lucianolorenti:
I think it should be closed.

PR #459, opened 2017-04-16
Allow whitelisting links in linkcheck

### LINKED ISSUE #447: Need a way of whitelisting certain url's in linkcheck
This was a feature sphinx had that we need in one or two places in base to get linkcheck to pass. There's a url on the kde bug tracker that always returns errors when you try to access it from the particular machine I do release preparation and testing on.

mortenpi:
@tkelman apologies for being a bit passive on the Documenter front at the moment. I assumed this to be a "nice to have".

I think adding another keyword argument would do the trick. This more or less copies how this is done in Sphinx as far as I can tell.

This will require a minor version bump, so I am thinking it might be time to merge a couple of the other PRs as well while we're at it. How fast would you like to have this available?

Fixes #447.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -7,7 +7,7 @@ A package for building documentation from docstrings and markdown files.
 !!! note
 
     Please read through the
-    [Documentation](http://docs.julialang.org/en/latest/manual/documentation.html) section
+    [Documentation](https://docs.julialang.org/en/latest/manual/documentation.html) section
     of the main Julia manual if this is your first time using Julia's documentation system.
     Once you've read through how to write documentation for your code then come back here.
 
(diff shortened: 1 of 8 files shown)


tkelman:
Cool, this looks like it'll do, thanks much. If you can get out a new release in a week or so, it can be part of rc2 instead of rc1 if needed.

mortenpi:
If I'd say I'll do a release by Friday, would that be fast enough for rc1?

tkelman:
rc1 will probably be before next friday, but doesn't absolutely need this to be available yet

tkelman:
Maybe that was ambitious. If someone hadn't broken the build on power, we'd probably have rc1 already.

Anyway, put to good use in https://github.com/JuliaLang/julia/pull/21473/commits/0afcf0539b9159896397e8e4050ae03eff5ed770.

Issue #461, opened 2017-04-21
Specify fonts on a per-project basis


ararslan:
Personally I quite dislike the default monospace font, Ubuntu Mono. I have no issues with the rest of the Documenter style, so what I'd like to be able to do is override a particular default, in this case just the font. (Unless of course you're open to a different default. 😉) Is this currently possible? Is there a way to e.g. inject
```css
pre, code {
  font-family: 'Source Code Pro', 'Andale Mono WT', courier, monospace !important;
}
```
?

mortenpi:
Hmm, "Source Code Pro" is a bit sharper, which I actually like. It should be at `0.9em` though in my opinion, otherwise it feels a bit too large relative to the surrounding text. We could consider changing the font.

As for overriding the font for a repository -- it's doable but should be discouraged, so that the Julia package ecosystem would have a consistent look and feel. That being said:
* You could roll your own `documenter.css` under `docs/src/assets` -- this would replace the one provided by the package, and might break in the future when there are changes to the HTML.
* So the better option would be to have a custom `.css` file and use the `assets` keyword. I think we do get the ordering right, so the CSS in user-provided files takes precedence. The folks at Invenia are making use of that, e.g. https://invenia.github.io/Dispatcher.jl/latest/

One thing to consider is that for custom fonts there should be a link to a CDN as well, since it's likely that users don't have the fonts installed. Ideally the CDN link should go into a `<link>` tag on each of the HTML pages.. and that could be doable via the `assets` keyword as well, although absolute URLs might be broken at the moment (if it is it then that should be fixed). Otherwise, an `@import` in the beginning of your custom CSS file should do the trick as well.

ararslan:
If you are interested in changing the font, I think Source Code Pro is a very nice choice; it has excellent Unicode support and I've yet to come across any character it didn't render properly. That's not the case for e.g. Inconsolata, which doesn't include Greek last I checked. Source Code Pro is also the font of choice for the Rust documentation, I believe.

> should be discouraged, so that the Julia package ecosystem would have a consistent look and feel

Hm, I'm not sure I agree, but that's neither here nor there.

Thanks for the heads up regarding Invenia's customization, I'll take a look at that.

ararslan:
[...]

PR #462, opened 2017-04-22
Use Source Code Pro as the monospace font

ararslan:
This PR replaces Ubuntu Mono with [Source Code Pro](https://fonts.google.com/specimen/Source+Code+Pro) as the monospace font, as @mortenpi and I discussed in #461. Per Morten's request, the font size is scaled down to 0.95em.

### DIFF
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -25,6 +25,10 @@ pre, code {
   font-size: 0.95em;
 }
 
+pre code {
+  font-size: 1em;
+}
+
 a {
     color: #2980b9;
     text-decoration: none;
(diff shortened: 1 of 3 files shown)


assets/html/documenter.css
```diff
@@ -21,7 +21,8 @@ body, input {
 }
 
 pre, code {
-  font-family: 'Ubuntu Mono', Monaco, courier, monospace;
+  font-family: 'Source Code Pro', Monaco, courier, monospace;
+  font-size: 0.95em;
```
mortenpi: I noticed that there should be an additional
```css
pre code {
  font-size: 1em;
}
```
selector as well. This is because we have `<pre><code>` blocks, and so the `.95` would get applied twice, yielding `.90` fonts in code blocks.

ararslan: Can do. 👍

PR #463, opened 2017-04-22
Use Roboto Mono as the monospace font

ararslan:
Competitor to #462

This PR replaces the current monospace font, Ubuntu Mono, with [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono). Currently I've scaled this down to match what was done for Source Code Pro in #462, but I'm not sure whether it's necessary for Roboto Mono.

cc @mortenpi

### DIFF
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -21,7 +21,12 @@ body, input {
 }
 
 pre, code {
-  font-family: 'Ubuntu Mono', Monaco, courier, monospace;
+  font-family: 'Roboto Mono', Monaco, courier, monospace;
+  font-size: 0.95em;
+}
+
+pre code {
+  font-family: 1em;
 }
 
 a {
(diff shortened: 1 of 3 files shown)


ararslan:
What do you think, @mortenpi? Also, I'm not super familiar with Codecov; is the project coverage decrease here and in #462 indicative of a problem?

mortenpi:
It's all good, I just figured I'd let it stay open for a few days in case there are any dissenting opinions. But I think we can merge this.

ararslan:
Looking at the most recent Documenter docs, which now have Roboto Mono, it looks like it might still be a bit too big compared to the regular font. What do you think, should we decrease the size a bit more, maybe 0.9 instead of 0.95?

mortenpi:
> maybe 0.9 instead of 0.95?

I'm on board with that -- it seems that with `.90em` the lowercase letters are exactly the same height in Lato and Roboto, which is probably what we want.

PR #466, opened 2017-04-25
Default deploydocs to release

ChrisRackauckas:
There are many times when a package won't work on nightly. But this means that by default the docs won't build for many packages. It's probably better to build on the current release.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -259,7 +259,7 @@ hide(root::AbstractString, children) = (true, nothing, root, map(hide, children)
         branch = "gh-pages",
         latest = "master",
         osname = "linux",
-        julia  = "nightly",
+        julia  = "release",
         deps   = <Function>,
         make   = <Function>,
     )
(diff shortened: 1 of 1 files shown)


djsegal:
looking forward to this

mauro3:
Bump: what's the status here?

mortenpi:
I am pretty sure that this will silently break all setups where `release` is not listed in `.travis.yml` (i.e. they have an explicit `0.6`). I unfortunately haven't had time to tally up how many packages might be affected.. if someone could give me a number or a list, that could help this along. 

In any case, package authors are recommended to use `julia = "release"` or `julia = "0.6"` argument explicitly in their `deploydocs` calls.

@tkelman It was mentioned in #526 that `release` might be discouraged. Could you confirm or deny that?

tkelman:
Yes, `release` has been removed from the generated .travis.yml template and is not actively supported, it will likely be deprecated or removed on the travis side.

mortenpi:
I think we should make `julia` a mandatory argument in this case.

PR #468, opened 2017-04-27
Bundle the highlight.js library

mortenpi:
Julia 0.6 related updates are not available in the latest release yet, so we can't use the CDN version for the time being.

This includes the highlight.js v9.11.0+ version build by hand from the highlight.js commit `35f2ff935e9d45c50feab6ee7538382a0f2412f6`. This includes the Julia 0.6 syntax updates.

cc @ararslan. It turns out that installing Node wasn't such a hassle after all on Ubuntu, so I think I've managed to build it for us. Could you give this branch a quick spin by any chance to make sure the highlighting works properly?

PR #475, opened 2017-05-04
Fix doctest highlighting #385

darwindarak:
When a REPL session doctest was processed, its `code.language` changes to `jlcon`.  This works for Pygments and the Latex output, but not highlights.js.  This PR changes `jlcon` to `julia` when generating HTML docs.


mortenpi:
There's already a fix for this in #468. We're just trying to bikeshed some details about the highlight.js library before merging it.

darwindarak:
Ok, sounds good!

PR #480, opened 2017-05-10
prepare for macrocall having a source location argument

vtjnash:
JuliaLang/julia#21746

(will need https://github.com/JuliaLang/Compat.jl/pull/355 to be merged and tagged first, then restart CI here)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.5
-Compat 0.17
+Compat 0.25
 DocStringExtensions 0.2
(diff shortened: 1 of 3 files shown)


mortenpi:
Since I'm not really familiar with the docsystem code -- this is an enhancement that will be available with 0.7 / master, right?

vtjnash:
Yes, this is for a PR against Julia master

mortenpi:
But 0.7 builds won't fail without this, right? I am wondering whether I should worry about backporting.

vtjnash:
They'll start to fail once https://github.com/JuliaLang/julia/pull/21746 is merged, which is why this needed to be done first

mortenpi:
Ah, I'll tag a patch release for Documenter asap then.

vtjnash:
Thanks :)

PR #485, opened 2017-05-14
Allow using pretty URLs

mortenpi:
Allows pretty URLs to be used in the HTML build, i.e. `dir/page.md` can be referenced as `dir/page/` instead of the current `dir/page.html`. This behaviour is consistent with MkDocs and Sphinx, and has been on the roadmap.

This is done by storing the actual page in `dir/page/index.html`. It does mean that this can't be used for local / offline builds though.

Currently it has to enabled via the `html_prettyurls` option and is disabled by default. It should become the default for the deploy-to-gh-pages builds for packages, even though it will involve breaking existing links.

@tkelman I think this would also be nice to have for 0.6. This way we won't have to swap to `.html` links for a release and then back again (assuming we want this in the long term), potentially breaking quite a lot of references to Base docs. Also, links that were referring to the 0.5 docs should start working again. For Base we should enable this feature explicitly and only for the `gh-pages` builds.

WIP because some internal links are still not being generated correctly, code needs cleanup and more testing.

Example build: http://mortenpi.eu/Documenter.jl/pretty-urls-deploy/

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Please open an [issue][issues-url] if you encounter any problems. If you have a 
 
 [gitter-url]: https://gitter.im/juliadocs/users
 
-[contrib-url]: https://juliadocs.github.io/Documenter.jl/latest/man/contributing.html
+[contrib-url]: https://juliadocs.github.io/Documenter.jl/latest/man/contributing/
 
 [docs-latest-img]: https://img.shields.io/badge/docs-latest-blue.svg
 [docs-latest-url]: https://juliadocs.github.io/Documenter.jl/latest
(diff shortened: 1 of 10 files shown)


tkelman:
ok good idea, but it may be too late to make rc2. will have to see about final.

mortenpi:
I should take another look at the code after a good night's sleep, but I believe it's done, and I don't think I have broken any old features even though I had to refactor a bit.

* Updated Documenter build: http://mortenpi.eu/Documenter.jl/pretty-urls-deploy/
* Julia manual: http://mortenpi.eu/julia/pretty-urls/

PR #513, opened 2017-06-06
detect named doctests as jldoctest

KristofferC:
Named doctests also want syntax highlighting :)

See e.g. https://docs.julialang.org/en/latest/manual/strings/#String-Basics-1 = no highlighting because the doctests are named

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -260,7 +260,7 @@ const LEXER = Set([
 ])
 
 function latex(io::IO, code::Markdown.Code)
-    language = if code.language == "jldoctest"
+    language = if startswith(code.language, "jldoctest")
         # When the doctests are not being run, Markdown.Code blocks will have jldoctest as
         # the language attribute. The check here to determine if it is a REPL-type or
         # script-type doctest should match the corresponding one in DocChecks.jl. This makes
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -260,7 +260,7 @@ const LEXER = Set([
 ])
 
 function latex(io::IO, code::Markdown.Code)
-    language = if startswith(code.language, "jldoctest")
+    language = if first(split(c.language)) == "jldoctest"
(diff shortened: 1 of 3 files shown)


mortenpi:
Good catch, LGTM. Thanks! Will merge when tests pass.

KristofferC:
I don't know if there is any reason to care about this but this PR would also match miss-spellings like `jldoctestx`. Could also do it as `split(c.language)[1] == "jldoctest"` to fix this.

mortenpi:
Good point. It would probably be best to have the correct solution, if you don't mind making the change. Could also be `first(split(c.language)) == ...`, but either way is fine for me.

KristofferC:
I YOLO:d with the in browser editor...

src/Writers/LaTeXWriter.jl
```diff
@@ -260,7 +260,7 @@ const LEXER = Set([
 ])
 
 function latex(io::IO, code::Markdown.Code)
-    language = if code.language == "jldoctest"
+    language = if first(split(c.language)) == "jldoctest"
```
mortenpi: Needs `code.language` here.

KristofferC: See! It is impossible to do things correctly with in browser editor...

KristofferC:
Updated

src/Writers/LaTeXWriter.jl
```diff
@@ -260,7 +260,7 @@ const LEXER = Set([
 ])
 
 function latex(io::IO, code::Markdown.Code)
-    language = if code.language == "jldoctest"
+    language = if first(split(code.language)) == "jldoctest"
```
mortenpi: It seems it needs an `isempty` check here as well (HTML already has it). Praised be the automatic tests :laughing:

KristofferC: let's try this then

KristofferC:
Good to go?

KristofferC:
Bump

mortenpi:
Whoops, apologies for missing the 0.6.0 boat with this. However, it's good to afaict. Thanks!

PR #514, opened 2017-06-08
WIP: Added glob-capabilities to `Pages` option in `@autodocs`

mauro3:
Now the `@autodocs` block supports glob-matching:
````
```@autodocs
Modules = [SomeModule]
Pages = ["*/SomeModule.jl", "*/sub-folder/*"]
```
````

Note that this is not backward compatible, but maybe could be made so.

Needs tests & test updates.  Let me know if this is of interest and I can finish it.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.5
 Compat 0.25
 DocStringExtensions 0.2
+Glob 1.1.1
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -20,7 +20,7 @@ import .Documents:
     MetaNode
 
 using Compat
-
+import Glob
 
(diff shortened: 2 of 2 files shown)


mortenpi:
Seems like a nice feature to have. Definitely needs extensive testing on Windows though.

As for backwards-compatibility -- I like the "explicit is better than implicit" mantra, so I think explicit globbing is probably better design for this, and so it might be worth breaking backwards compatibility. Not sure how to treat patterns that have no `*` in them though in that case. If it happens that those basically become meaningless then we could retain the current behaviour as well, with a deprecation warning perhaps.

mortenpi:
I think having globs for `Pages` would be neat, so if anyone wants it, please do open an issue or a PR. But let's close this PR since it's very stale.

PR #515, opened 2017-06-10
Add style for <kbd>

omus:
Added CSS styling for the `<kbd>` HTML tag which allows Markdown like:

````md
```@raw html
<kbd>Ctrl</kbd> + <kbd>D</kbd>
```
````

To render as:

<kbd>Ctrl</kbd> + <kbd>D</kbd>

Styling was based off GitHub

### DIFF
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -105,6 +105,20 @@ pre code {
     background-color: initial;
 }
 
+kbd {
+    font: 11px "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;
+    display: inline-block;
+    padding: 3px 5px;
+    line-height: 10px;
+    color: #444d56;
+    vertical-align: middle;
+    background-color: #fafbfc;
+    border: solid 1px #c6cbd1;
+    border-bottom-color: #959da5;
+    border-radius: 3px;
+    box-shadow: inset 0 -1px 0 #959da5;
+}
+
 /* Headers in admonitions and docstrings */
 .admonition h1,
 article section.docstring h1 {
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -20,7 +20,7 @@ body, input {
   text-rendering: optimizeLegibility;
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: assets/html/documenter.css [hunk 1]
```diff
@@ -105,6 +105,20 @@ pre code {
     background-color: initial;
 }
 
+kbd {
+    font: 11px "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;
+    display: inline-block;
+    padding: 3px 5px;
+    line-height: 10px;
```
mortenpi: It would be great to have these in `em`s I think, so that it would scale properly with the rest. The border/shadow ones should stay `px` though.

assets/html/documenter.css
```diff
@@ -105,6 +105,20 @@ pre code {
     background-color: initial;
 }
 
+kbd {
+    font: 11px "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;
```
mortenpi: Font should probably be consistent with the other monospace stuff. Instead of having it here we could add `kbd` to the the `pre` and `code` on line 33. Need to make sure it works with the rest of the styling though.

omus: Looks pretty good to me.

assets/html/documenter.css
omus: I switched these to `em`s.

mortenpi:
One more small thing -- could you add the example snippet in the OP to the end of `test/examples/src/man/tutorial.md`?

omus:
> One more small thing -- could you add the example snippet in the OP to the end of `test/examples/src/man/tutorial.md`?

Done. It may be worth adding some contribution documentation to explain "test/examples".

mortenpi:
LGTM, thanks!

> It may be worth adding some contribution documentation to explain "test/examples".

It should be turned into some sort of a proper page of the manual probably. Right now it's just a collection of snippets that I add to every now and then so that I wouldn't forget all the different tiny things we support.

PR #518, opened 2017-06-10
take more space above headings

bicycle1885:
I'd like to suggest allocating more space above headings to make a section visually separable to the next section as follows (left: current, right: proposed)

screen shot 2017-06-10 at 18 15 15

### DIFF
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -47,8 +47,8 @@ body {
 }
 
 h1 { font-size: 1.75em; }
-h2 { font-size: 1.50em; }
-h3 { font-size: 1.25em; }
+h2 { font-size: 1.50em; margin: 1.75em 0 0; border-bottom: 1px solid #ddd; }
+h3 { font-size: 1.25em; margin: 1.50em 0 0; }
 h4 { font-size: 1.15em; }
 h5 { font-size: 1.10em; }
 h6 { font-size: 1em; }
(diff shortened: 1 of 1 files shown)


KristofferC:
FWIW this looks better to me.

mortenpi:
Agreed, completely in favour of having a bit more space there.

* Could we maybe have the color be `#e5e5e5` so that it would be consistent with the `hr` tags? It's currently a bit darker than the lines on the top and bottom of the page.

* It's also possible to have `<h1>` tags in the middle of the page, so it would probably make sense to style those as well, but only if they're _not_ the second element (i.e. the page title). To make sure that we fall back to the non-styled version on older browsers, I came up with this:

  ```css
  /* Unless the <h1> the is very first thing on the page (i.e. the second element in the
   * <article>, * after the <header>, we add some additional styling to it to make it stand
   * out a bit more. This way we get a reasonable fallback if CSS3 selectors are not supported
   * in the browser.
   */
  article > h1:not(:nth-child(2)) {
    margin: 1.75em 0 0;
    border-bottom: 1px solid #e5e5e5;
  }
  ```
  Not sure if there's a cleaner way to do this.

bicycle1885:
> It's also possible to have <h1> tags in the middle of the page,

What about having a guideline on the top-level headings, or converting all h1 tags to h2 except the first one?

mortenpi:
I see no reason not to support additional `h1` headings. We're already treating them properly in the navmenu and when trying to detect the document outline.

mortenpi:
A friendly bump. Any thoughts on the `h1` handling?

bicycle1885:
Oh, sorry. What about this?

PR #522, opened 2017-06-22
Fix CurrentModule example in manual

nalimilan:
The module taken as an example throughout the guide is Example, not Documenter. (Or maybe I'm missing something.)


mortenpi:
Thanks!

PR #528, opened 2017-06-26
Merge Travis CI configuration lines into a single line

nalimilan:
This is cleaner when packages have another line for e.g. coverage.

Was there a particular reason to use two lines?

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -98,8 +98,7 @@ run your `docs/make.jl` file:
 
 ```yaml
 after_success:
-  - julia -e 'Pkg.add("Documenter")'
-  - julia -e 'cd(Pkg.dir("PACKAGE_NAME")); include(joinpath("docs", "make.jl"))'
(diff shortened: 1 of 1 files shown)


mortenpi:
I think which is cleaner is subjective :laughing: But having just one line for Documenter-related stuff makes sense, as you point out.

I actually prefer the look of `julia docs/make.jl` though. This of course assumes that the working directory does not change. But I think it could be generalized with `TRAVIS_BUILD_DIR`:
```
after_success:
  - julia -e 'Pkg.add("Documenter")'
  - julia ${TRAVIS_BUILD_DIR}/docs/make.jl
```
No way to merge into a single line though as far as I can tell (unless we could have [custom REQUIRE files](https://discourse.julialang.org/t/resolving-a-custom-require-when-calling-julia/4844)).

nalimilan:
As you prefer, that's not a big deal.

mortenpi:
But one can use `&&`. Could we compromise on
```yaml
after_success:
  - julia -e 'Pkg.add("Documenter")' && julia ${TRAVIS_BUILD_DIR}/docs/make.jl
```
by any chance?

PR #529, opened 2017-06-26
Render arbitrary markdown files

timholy:
Not sure if you want this, but I found it handy to preview a README.md file for another package. Yes, there are many markdown renderers out there, but Documenter is the one that counts :wink:.

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -345,3 +345,17 @@ See [Hosting Documentation](@ref) for further information on configuring MkDocs 
     of the navigation menu.
 
     **`pages`** defines the hierarchy of the navigation menu.
+
+## Rendering arbitrary markdown files
+
+You can use Documenter to preview arbitrary markdown files like this (example on Linux):
+
+```sh
+cd /tmp
+mkdir src/
+ln -s /path/to/README.md src/
+julia -e "using Documenter; makedocs(format = :html)"
+```
+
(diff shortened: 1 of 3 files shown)


docs/src/man/guide.md
```diff
@@ -345,3 +345,17 @@ See [Hosting Documentation](@ref) for further information on configuring MkDocs
...
+
+## Rendering arbitrary markdown files
+
+You can use Documenter to preview arbitrary markdown files like this (example on Linux):
+
+```sh
+cd /tmp
+mkdir src/
+ln -s /path/to/README.md src/
+julia -e "using Documenter; makedocs(format = :html)"
```
mortenpi: I'm getting an error when running this one-liner:
```
$ julia -e 'using Documenter; makedocs(format = :html)'
ERROR: syntax: incomplete: premature end of input
```
on v0.6.0 / Linux. It seems as if you can't have `using`/`import` statements in the command line input. Are you seeing the same?

mortenpi:
I am wondering whether it might make sense to have a separate API for this in principle as well, e.g. something like `render("Package/README.md")` which would be tuned for rendering single pages.

asinghvi17:
Bump?

mortenpi:
Let's close this. I don't think we actually want to approach rendering arbitrary Markdown files this way (it also needs a different Markdown parser to work well), and this PR has some major conflicts now.

PR #530, opened 2017-07-02
Correct length to endof in SubString call

bkamins:
`length` will fail on multibyte `Char`.
E.g.
```
x = "∀∀∀"
length(x) # 3 - results incorrect indexing
endof(x) # 7 - correct index of last Char
```

This is part of the fix needed for https://github.com/JuliaLang/julia/pull/22511.

Fixes #470.

PR #531, opened 2017-07-02
Improve handling of SubString selection

bkamins:
Second part of #530. Again we make sure that we use proper index into `SubString`.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -121,16 +121,17 @@ function parseblock(code::AbstractString, doc, page; skip = 0, keywords = true)
     code = last(split(code, '\n', limit = skip + 1))
     # Check whether we have windows-style line endings.
     local offset = contains(code, "\n\r") ? 2 : 1
-    local strlen = length(code)
+    local endofstr = endof(code)
     local results = []
     local cursor = 1
-    while cursor < strlen
+    while cursor < endofstr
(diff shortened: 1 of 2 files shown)


mortenpi:
Is this also holding back  JuliaLang/julia#22511?

bkamins:
This is the same type of problem as #530 - in the code `length` of string is used which is number of characters in a string and not last index in a string.
See for example:
```
julia> s= "∀∀∀"
"∀∀∀"

julia> i=length(s)
3

julia> s[i]
ERROR: UnicodeError: invalid character index
Stacktrace:
 [1] slow_utf8_next(::Ptr{UInt8}, ::UInt8, ::Int64, ::Int64) at .\strings\string.jl:172
 [2] next at .\strings\string.jl:204 [inlined]
 [3] getindex(::String, ::Int64) at .\strings\basic.jl:32
```

Similarly using `ncursor-1` is invalid with multibyte characters and `prevind` should be used.

When [JuliaLang/julia#22511](https://github.com/JuliaLang/julia/pull/22511) is merged the current code (without fixes) may produce wrong results when multibyte characters in strings are used in documentation.

mortenpi:
I squashed your commits and added a semi-related test (managed to create a case where it already fails with Unicode code) and a doc note. It's good to go when tests pass and will be part of the next patch release. Many thanks for finding and fixing these!

PR #540, opened 2017-07-27
add option to ignore stacktraces

KristofferC:
This makes it feasible to run doctests as a part of CI on e.g. Base Julai

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -174,6 +174,9 @@ ignored.
 **`strict`** -- [`makedocs`](@ref) fails the build right before rendering if it encountered
 any errors with the document in the previous build phases.
 
+**`doctest_stacktraces`** -- [`makedocs`](@ref) do not consider an erronous stacktrace
+for a doctest as a doctest failure.
+
 ## Non-MkDocs builds
 
 Documenter also has (experimental) support for native HTML and LaTeX builds.
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@
 test/examples/builds/
 test/missingdocs/build/
 test/errors/build/
+test/doctest_stacktrace/build/
 docs/build/
 docs/site/
--- a/docs/src/man/doctests.md
+++ b/docs/src/man/doctests.md
@@ -105,6 +105,9 @@ ERROR: DivideError: integer division error
 ```
 ````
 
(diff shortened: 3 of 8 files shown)


KristofferC:
I dont think failures on maser is because of this PR

src/Documenter.jl
```diff
@@ -174,6 +174,9 @@ ignored.
 **`strict`** -- [`makedocs`](@ref) fails the build right before rendering if it encountered
 any errors with the document in the previous build phases.
 
+**`doctest_stacktraces`** -- [`makedocs`](@ref) do not consider an erronous stacktrace
```
mortenpi: Typo: "erroneous"

KristofferC:
Having an extensible way of adding options to doctests-blocks seems like a good idea to me. Something like `jldoctest key1, key2=val2, key3` etc, where `val`s are potential settings associated with the key. This would of course only be worth it if we expect to add more of these options for doctests.

fredrikekre:
Named doctests could be changed to use `name = nameofblock` so there is one more option together with `stacktraces = {ignore, warn, error}`.

KristofferC:
Perhaps globaal doctest settings could be specified in `DocTestSetup`?

mortenpi:
Do you think we could have this as a regex pattern with the option from #549?

KristofferC:
I've been meaning to write a more extensible solution than this, where you can easily specify doctest-option on a doctest by doctest basis, or globally etc.If that is something that sounds good, I will probably have time for it in ~2 weeks.

mortenpi:
The per-doctest thing sounds great. I do think that it should ideally be specified at the doctest itself. But a global option is an OK stopgap as well I think.

But that's actually orthogonal to the question whether we should have a more generic API for dealing with changing doctests. We could

* have an option specific to stacktraces (as is here), and separate ones for other ones
* use regex patterns (i.e. #549)
* allow the user define arbitrary functions (that would remove parts that should not be checked from the expected and real doctest outputs?)

With the more generic approach, we could maintain a few more commonly used patterns/functions (e.g. one for stacktraces) in Documenter.

I am always leaning towards the most generic solution, but I am wondering whether I am over-engineering this.

cc @bicycle1885, because of your work on #549

KristofferC:
https://github.com/JuliaDocs/Documenter.jl/pull/573

PR #545, opened 2017-07-31
Allow HTMLWriter to proceed when not using git

danielmatz:
I have a few Julia packages that I use at work.  They are version controlled, but not with git.

Currently, Documenter's HTML writer will die when run on a package that isn't using git.  There are a few git commands that are used to generate URLs for links to the source code.  This PR allows the `Utilities.url` functions to return null results if the git commands fail, such that the HTML writer can proceed and just not include links.

I also broke out new utility functions for getting the path of a file relative to the repository root and the commit of a file.  The new functions redirect stderr, to hide the git errors.

### DIFF
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -196,6 +196,7 @@ struct User
     analytics::String
     version :: String # version string used in the version selector by default
     html_prettyurls :: Bool # Use pretty URLs in the HTML build?
+    html_disable_git :: Bool # Don't call git when exporting HTML
 end
 
 """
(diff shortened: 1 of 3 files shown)


mortenpi:
Non-git repos should definitely be supported, so thanks for this! I am slightly worried about Git failures passing unnoticed when they shouldn't though (main use case is still Travis builds with Git-based repos). Maybe we could add an option (`html_disable_git` or something) to `makedocs` which would allow the user to toggle this?

danielmatz:
Good point!  I'll work on adding the option.

danielmatz:
Sorry about that, I mucked up my branch and github closed the PR on me...  Hopefully I fixed it.

mortenpi:
I should've merged this ages ago. Thanks @danielmatz!

PR #546, opened 2017-07-31
avoid introducing binding for String

vtjnash:
this ensures that we only introduce package-local bindings for values
that we define in this package

ref JuliaLang/julia#22984

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.5
-Compat 0.25
+Compat 0.29
 DocStringExtensions 0.2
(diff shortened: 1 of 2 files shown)


REQUIRE
```diff
@@ -1,3 +1,3 @@
 julia 0.5
-Compat 0.25
+Compat 0.29
```
tkelman: this change doesn't seem like a requirement for this package to work

src/Utilities/DOM.jl
```diff
@@ -99,7 +99,7 @@ import ..Utilities
 using Compat
 
 if VERSION < v"0.5.0-dev"
```
tkelman: this is dead, would be better to delete this branch

mortenpi:
The whole code block will be removed by #555

PR #547, opened 2017-08-01
fix some MathJax links to the latest CDN provider

bicycle1885:
This fixes some links that still point to the old CDN.
https://www.mathjax.org/cdn-shutting-down/


mortenpi:
Thanks!

PR #549, opened 2017-08-05
add ignore_pointers option

bicycle1885:
My doctests have pointer values in the output and hence I cannot use `jldoctest` for my package. For example:
```
julia> using EzXML

julia> doc = XMLDocument()
EzXML.Document(EzXML.Node(<DOCUMENT_NODE@0x00007fa2ec190b70>))

```

The pointer values change from run to run, which results in failure when running doctests.

This pull request adds `ignore_pointers` option to `makedocs`. When `ignore_pointers=true`, the checker trims substrings that look like a pointer value and then compares the results. The default is `false`, so this does not change the behavior at all by default.

### DIFF
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -189,6 +189,7 @@ immutable User
     linkcheck_ignore::Vector{Union{String,Regex}}  # ..and then ignore (some of) them.
     checkdocs::Symbol         # Check objects missing from `@docs` blocks. `:none`, `:exports`, or `:all`.
     strict::Bool              # Throw an exception when any warnings are encountered.
+    ignore_pointers::Bool     # Ignore difference of pointers in doctests.
     modules :: Set{Module}    # Which modules to check for missing docs?
     pages   :: Vector{Any}    # Ordering of document pages specified by the user.
     assets  :: Vector{Compat.String}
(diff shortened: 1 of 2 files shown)


bicycle1885:
I started to think more generic approach like `ignore_pattern=r"@0x[0-9a-f]{16}"` is better.

mortenpi:
Similar to #540, I am wondering whether a per-block option would be better here?

bicycle1885:
Possibly. But it is tedious to add ignore pattern options to all doctests of a package. Having both a package-global option and a doctest-local option would be good.

src/Documents.jl
```diff
@@ -189,6 +189,7 @@ immutable User
     linkcheck_ignore::Vector{Union{String,Regex}}  # ..and then ignore (some of) them.
     checkdocs::Symbol         # Check objects missing from `@docs` blocks. `:none`, `:exports`, or `:all`.
     strict::Bool              # Throw an exception when any warnings are encountered.
+    pattern_ignore::Nullable{Regex}  # Ignore differences that match this pattern in doctests.
```
mortenpi: Should we make it an array perhaps, so that one could have multiple patterns? Should we call it `doctest_ignore`, to namespace it more clearly?

bicycle1885:
I'm sorry for being late. If #573 supersedes this, please feel free to close this pull request.

bicycle1885:
#573 looks a better approach, so I close this pull request. Thank you @KristofferC!

PR #552, opened 2017-08-18
External pages

### LINKED ISSUE #551: Allow linking of files not in docs directory as pages in makedocs
I'm not sure if this is possible, but I'd really love to do the following in my `make.jl`:

```julia
makedocs(
    # ...
    pages=[
        # ...,
        "Contibuting" => "../CONTRIBUTING.md",
        # ...,
    ],
    # ...
)
```

I tried it and received the following error:

```
ERROR: LoadError: '../CONTRIBUTING.md' is not an existing page!
Stacktrace:
 [1] walk_navpages(::Bool, ::String, ::String, ::Array{Any,1}, ::Void, ::Documenter.Documents.Document) at /home/degraafc/.julia/v0.6/Documenter/src/Builder.jl:154
 [2] walk_navpages(::Pair{String,String}, ::Void, ::Documenter.Documents.Document) at /home/degraafc/.julia/v0.6/Documenter/src/Builder.jl:171
 [3] collect_to!(::Array{Documenter.Documents.NavNode,1}, ::Base.Generator{Array{Any,1},Documenter.Builder.##3#4{Void,Documenter.Documents.Document}}, ::Int64, ::Int64) at ./array.jl:474
 [4] collect(::Base.Generator{Array{Any,1},Documenter.Builder.##3#4{Void,Documenter.Documents.Document}}) at ./array.jl:442
 [5] walk_navpages(::Array{Any,1}, ::Void, ::Documenter.Documents.Document) at /home/degraafc/.julia/v0.6/Documenter/src/Builder.jl:172
 [6] runner(::Type{Documenter.Builder.SetupBuildDirectory}, ::Documenter.Documents.Document) at /home/degraafc/.julia/v0.6/Documenter/src/Builder.jl:124
 [7] dispatch(::Type{Documenter.Builder.DocumentPipeline}, ::Documenter.Documents.Document) at /home/degraafc/.julia/v0.6/Documenter/src/Selectors.jl:164
 [8] cd(::Documenter.##2#3{Documenter.Documents.Document}, ::String) at ./file.jl:70
 [9] #makedocs#1(::Bool, ::Array{Any,1}, ::Function) at /home/degraafc/.julia/v0.6/Documenter/src/Documenter.jl:198
 [10] (::Documenter.#kw##makedocs)(::Array{Any,1}, ::Documenter.#makedocs) at ./<missing>:0
while loading /home/degraafc/code/PkgTemplates/docs/make.jl, in expression starting on line 3
```

Maybe I'll take a closer look at the code when I have time to make a PR.

christopher-dG:
Closes #551 

So far this ™️ *works on my machine* ™️, and the existing tests pass.

In the general case, to add a file that's in your repo but not in the documentation root directory to `pages`, use: `external("relative/path/from/repo/root")` or `"Page Title" => external("relative/path")`.
  
Still to do:

* [ ] Put `external` in the docs somewhere
* [ ] Add tests (Could you point me in the right direction of where I should put them?)


src/Documenter.jl
```diff
@@ -193,11 +193,19 @@ modules ([`Writers.HTMLWriter`](@ref), [`Writers.LaTeXWriter`](@ref)).
...
-function makedocs(; debug = false, args...)
-    document = Documents.Document(; args...)
+function makedocs(;
+    debug::Bool = false,
+    root::AbstractString = Utilities.currentdir(),
+    source::AbstractString  = "src",
+    pages::Vector = Any[],
+    args...
+)
+    pages = copy_external(pages, root, source)
```
mortenpi: I would avoid actually copying files. I would modify the parser part instead. Although that might require refactoring, so I could be OK with copying as well as a temporary solution.

src/Documenter.jl
```diff
@@ -750,4 +758,85 @@ function generate(pkgname::AbstractString; dir=nothing)
     nothing
 end
 
+const external_dir = "DOCUMENTER_EXTERNAL"
```
mortenpi: I might actually put these pages right under `src/` by default.

Furthermore, the main issue with these "external" pages is that Documenter does not know what the name (i.e. the URL) of the page should be. For the normal pages it is determined from the path (e.g. `src/dir/page.md` becomes `/dir/page/`).

So, I would actually propose `external(src, target)`. E.g. `external("../README.md", "info/readme.md")`. And `target` could be omitted, with default being the `src` root and the name of the file (i.e. `dirname(src)`).

I'm not sure what to do with relative links in the source files. E.g. say there's a link to an authors file in the readme: `[authors file](AUTHORS.md)`. We could try being smart about resolving these links (e.g. if `../README.md` goes to `/readme/` and `../AUTHORS.md` goes to `/pages/authors/`, then they would still resolve to each other). But for now, I guess the easiest thing would be that it's up to the user to make sure that the page structure is replicated with the `external` calls.

src/Documenter.jl
```diff
@@ -750,4 +758,85 @@ function generate(pkgname::AbstractString; dir=nothing)
...
+
+"""
+A file outside of the documentation directory.
+"""
+struct ExternalPage
```
mortenpi: This type could live in `Documents.jl` maybe. The `source` and `destination` functions as well.

christopher-dG:
Thanks for the feedback! I will address it when I have some time on my hands.

JeffBezanson:
:+1: This will be really useful for the new "default packages" structure.

christopher-dG:
Hey, sorry this has gone ignored for so long -- I'll make an effort to work on this this weekend.

src/Documenter.jl
christopher-dG: I think I'm going to keep it this way for now, it's just a bit more trouble to refactor than it's worth to me. The only downside of this approach is that if for some reason the build fails, there may be some leftover files in `docs/src`.

src/Documenter.jl
christopher-dG: I took your advice and added a `target` keyword, and by default they're put directly in `src`.

As for relative links, I added a warning in the `external` docstring, and we can leave the rest up to the user.

src/Documenter.jl
christopher-dG: `ExternalPage` has been moved. I ended up getting rid of the other two functions because they became a lot simpler after some refactoring.

src/Documenter.jl
```diff
@@ -750,4 +758,85 @@ function generate(pkgname::AbstractString; dir=nothing)
...
+    path::AbstractString
+    root::Nullable{AbstractString}
+end
+
+"""
+    external(path; root=nothing) -> ExternalPage
```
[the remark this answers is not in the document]
christopher-dG: Went with a single `repo_root` keyword (which defaults to `dirname(root)`, covering the average case); all external pages are relative to that.

christopher-dG:
Okay, this should be ready for another review.

src/Documenter.jl
```diff
@@ -736,4 +751,74 @@ function generate(pkgname::AbstractString; dir=nothing)
...
+        elseif isa(page, Pair) && isa(page.second, Documents.ExternalPage)
+            pages[i] = pages[i].first => copypage(page.second)
+            push!(externals, pages[i].second)
+        elseif isa(page, Pair) && isa(page.second, Vector)
+            for (j, el) in enumerate(page.second)
+                if isa(el, Documents.ExternalPage)
+                    pages[i].second[j] = copypage(el)
+                    push!(externals, pages[i].second[j])
+                end
+            end
```
mortenpi: This only goes down two levels, right? `pages` can (in principle) have an arbitrary number of levels, so I think this should be done recursively. Something like how `walk_navpages` does it in `Builder.jl`.

src/Documents.jl
```diff
@@ -52,6 +52,14 @@ function Page(source::AbstractString, build::AbstractString)
     Page(source, build, elements, ObjectIdDict(), Globals())
 end
 
+"""
+A file outside of the documentation directory.
+"""
+struct ExternalPage
+    path::AbstractString
+    target::Nullable{AbstractString}
```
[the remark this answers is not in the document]
christopher-dG: Anything wrong with using a `Nullable` other than preference? The constructor still works as we'd expect it to:
```julia
julia> struct X
           x::Nullable{Int}
       end

julia> x = X(nothing)
X(Nullable{Int64}())

julia> x = X(1)
X(Nullable{Int64}(1))
```

src/Documenter.jl
```diff
@@ -84,6 +85,9 @@ running `makedocs` from the Julia REPL like so:
 
     julia> makedocs(root = Pkg.dir("MyPackage", "docs"))
 
+**`repo_root`** is the top-level directory of the repository, assumed to be `root`'s
+parent. You only need to set this if you are making use of [`external`](@ref).
```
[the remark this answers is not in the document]
christopher-dG: I was operating under the assumption that people would only include external pages that were in the repo, but I suppose it's totally valid to take them from elsewhere, good point.

src/Documenter.jl
christopher-dG: Yeah, I just wasn't aware that going deeper was allowed.

src/Documents.jl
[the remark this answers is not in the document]
KristofferC: https://github.com/JuliaDocs/Documenter.jl/pull/576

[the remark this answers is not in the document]
christopher-dG: Ah, got it.

christopher-dG:
I've rebased this, we'll see if I can resurrect it :slightly_smiling_face:

src/Documenter.jl
```diff
@@ -785,4 +804,263 @@ function getenv(regex::Regex)
     error("could not find key/iv pair.")
 end
 
-end # module
+<<<<<<< HEAD
```
fredrikekre: Messed up rebase

christopher-dG: Huh, I spotted this last night and could have sworn I pushed the fix, but I guess not.

jlumpe:
I'm really interested in this, is it still being worked on?

christopher-dG:
Not by me at the moment, but maybe one day I'll come back to it... Or if someone wants to take over the branch they can feel free.

mortenpi:
I have a slightly updated version of this, but it also needs a bit more work (which is also unfortunately unlikely to happen any time too soon): https://github.com/JuliaDocs/Documenter.jl/issues/1074#issuecomment-515477974

fingolfin:
This sounds like a nice feature to have, but last discussions here were in 2019. Back then @mortenpi hinted he might have an alternative implementation...

@christopher-dG are you still interested in this feature, and/or in reviving this PR?

PR #554, opened 2017-08-20
Fix #553, @autodocs can't handle parameterized types

### LINKED ISSUE #553: @autodocs can't handle parameterized types
For the following source code:
```julia
module A

"""
A Foo foos.
"""
struct Foo
end

"""
A Bar bars.
"""
struct Bar{T}
end

end # module
```

and an `index.md` that looks like
````markdown
# A

```@autodocs
Modules = [A]
Order = [:type]
```
````
running `julia make.jl` results in a missing docstring warning for `A.Bar`, and only `Foo` shows up in the generated documentation. This is on Julia 0.6, with both version 0.11.2 and current master of Documenter.

tkoolen:
Handle `UnionAll` in `category`.

Also handle `UnionAll` in `doccat` in a more correct way (at least in my opinion).

Could use more tests; I see that there are currently no direct tests for `category`. Should I add some, and if so, where?

Fixes #553.

### DIFF
--- a/src/DocSystem.jl
+++ b/src/DocSystem.jl
@@ -373,6 +373,9 @@ function category(b::Docs.Binding)
 end
 category(::Function) = :function
 category(::DataType) = :type
+if isdefined(Base, :UnionAll)
+    category(x::UnionAll) = category(Base.unwrap_unionall(x))
+end
 category(::Module) = :module
 category(::Any) = :constant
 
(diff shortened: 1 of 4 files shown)


tkoolen:
Rebased.

mortenpi:
I think you're right that it's the correct way. LGTM anyhow. Thanks!

tkoolen:
Thank you!

PR #555, opened 2017-08-21
Drop julia 0.5

fredrikekre:
I enabled FemtoCleaner on my fork (https://github.com/fredrikekre/Documenter.jl/pull/1). I think it would be nice to drop 0.5 support; when building Base PR's there are ~3000 lines of deprecations.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -418,9 +418,9 @@ function url(repo, file)
         _, path = split(file, root; limit = 2)
         repo = replace(repo, "{commit}", commit)
         repo = replace(repo, "{path}", path)
-        Nullable{Compat.String}(repo)
+        Nullable{String}(repo)
     else
-        Nullable{Compat.String}()
+        Nullable{String}()
     end
 end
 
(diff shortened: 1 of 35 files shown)


src/Utilities/Utilities.jl
```diff
@@ -444,54 +427,48 @@ end
...
-
-                _, path = split(file, root; limit = 2)
-                repo = replace(repo, "{commit}", commit)
-                repo = replace(repo, "{path}", path)
-                repo = replace(repo, "{line}", line)
-
-                Nullable{Compat.String}(repo)
+function url(remote, repo, mod, file, linerange)
+    remote = getremote(dirname(file))
+    isabspath(file) && isempty(remote) && isempty(repo) && return Nullable{Compat.String}()
```
ararslan: This could just be `String` rather than `Compat.String`. (Same in another couple places.) It should be possible to ditch the Compat dependency entirely in this package, but I'm not certain.

ararslan:
Looks good to me aside from the note about `Compat.String`. Thanks!

fredrikekre:
Changed `Compat.String` to `String`.

src/Utilities/Utilities.jl
```diff
@@ -435,63 +418,57 @@ function url(repo, file)
...
+        Nullable{String}(repo)
     else
-        Nullable{Compat.String}()
+        Nullable{String}()
     end
 end
 
 url(remote, repo, doc) = url(remote, repo, doc.data[:module], doc.data[:path], linerange(doc))
 
 # Correct file and line info only available from this version onwards.
```
mortenpi: Should probably remove this comment as well.

fredrikekre: Done.

fredrikekre:
Download failure on nightly mac build. Perhaps restart? Otherwise I think this is good to go.

PR #565, opened 2017-09-22
Remove references to 0.5

PallHaraldsson:


PallHaraldsson:
Seems STABLE doc also outdated, say older versions/0.4 was supported/is in old code.

mortenpi:
For a package dealing with documentation, I should really take better care of its documentation. There's also a reference to 0.5 in the README -- could you also remove that? Docs for stable will update once 0.12 gets tagged.

fredrikekre:
Is this actually needed? Documenter still works on 0.5.

mortenpi:
`master` does not anymore. The current 0.11 branch is the last one that will support 0.5.

fredrikekre:
Well, yea, but there are still releases that work on 0.5. From the README the impression now is that it only works on 0.6 and above. Does not really matter though.

PR #570, opened 2017-10-02
use ends_with_semicolon from Base

KristofferC:
This will make sure that these are in sync (they currently are not).


KristofferC:
Nightly mac worker is at a newer commit than linux which is why build fails there.

KristofferC:
Thanks again for the review and merge!

PR #571, opened 2017-10-02
remove compats, unnecessary locals and fix doctests

KristofferC:


KristofferC:
Nightly mac worker is at a newer commit than linux which is why build fails there.

KristofferC:
Thanks for the review and mere!

PR #573, opened 2017-10-02
allow for custom filtering of doctests

KristofferC:
Replaces https://github.com/JuliaDocs/Documenter.jl/pull/540 and perhaps https://github.com/JuliaDocs/Documenter.jl/pull/549.

PR #574, opened 2017-10-03
fix deprecations on 0.7

KristofferC:
Removed usage of `current_module()` and just default to `Main`.

Made on top of https://github.com/JuliaDocs/Documenter.jl/pull/571


mortenpi:
This supersedes #556 as far as I can tell.

mortenpi:
Once this and #573 are rebased we could tag 0.12.

KristofferC:
I want to test some packages with this PR before we merge it. I'll try get to it today or tomorrow.

KristofferC:
Rebased this. The problem right now is the following:

```
julia> module AnonymousDoctestModule
       struct Point
           x
       end

       @show Point
       end;
Point = Main.AnonymousDoctestModule.Point
```

Previously, moduled defined in `Main` did not display with the prefix `Main.`. However, string replacing `"Main."` with `""` seems fragile, indeed.
One possible solution would be to evaluate everything into a module with a name that can be used as a sentinel value and replace that instead.
This would however no longer mean that modules are evaluated in `Main` like the current documentation says.
The question is, is evaluating into Main useful anymore when modules are no longer in main by default?
You can no longer do `using` somewhere and get access to the module from a doctest.

KristofferC:
Ah, I think I know what to do. I was stupid.

KristofferC:
Ok, doctests on Julia Base master only gives expected errors now. This is good to go from my point of view.

KristofferC:
Ok, this still has a problem:

```
```jldoctest
julia> g(x::Float64, y) = 2x + y
g (generic function with 1 method)

julia> g(x, y::Float64) = x + 2y
g (generic function with 2 methods)

julia> g(2.0, 3)
7.0

julia> g(2, 3.0)
8.0

julia> g(2.0, 3.0)
ERROR: MethodError: g(::Float64, ::Float64) is ambiguous. Candidates:
  g(x, y::Float64) in Main.Main at none:1
  g(x::Float64, y) in Main.Main at none:1
Possible fix, define
  g(::Float64, ::Float64)
```

The `Main.Main` are wrong.

KristofferC:
Ok, should be fixed and test added. This is ready for review again (hehe).

The only problem left on 0.7 is the following:

```
julia> x -> x+3
#1 (generic function with 1 method)
```

in combination with

https://github.com/JuliaDocs/Documenter.jl/blob/54902e4e1fb7ad7e0162fe91c19ce2cd6e2df149/src/DocChecks.jl#L335-L337

since that will just skip the output from defining the anonymous function.

What do you think is the best thing to do here?

KristofferC:
Perhaps add a special regex to match the anonymous function syntax:

` #[0-9]+ \(generic function with [0-9]+ method(s)?\)`

mortenpi:
Do we actually need to ignore these lines in the first place? But I could definitely live with that workaround.

KristofferC:
There is no need but it could be argued that it is a feature.

mortenpi:
Not sure it's a feature that should be used.. you can't have comments in the real REPL like this. But it could break some doctest somewhere.

Let's go with the workaround and add a TODO note to remove / refactor this in the future. At one point it would be good to do a backwards incompatible Documenter release anyway, so we can revisit it then.

KristofferC:
Something like this then?

src/DocChecks.jl
```diff
@@ -338,7 +339,9 @@ function repl_splitter(code)
         line = shift!(lines)
         # REPL code blocks may contain leading lines with comments. Drop them.
         # TODO: handle multiline comments?
-        startswith(line, '#') && continue
+        # ANON_FUNC_DECLATION deals with `x->x` -> `#1 (generic function ....)`on 0.7
+        # TODO: Remove this special case and just disallow lines with comments?
+        !ismatch(ANON_FUNC_DECLARATION, line) && startswith(line, '#') && continue
```
mortenpi: Typo with `ANON_FUNC_DECLARATION ` and missing space before "on" in the comment. Highly premature optimization, but we could flip the order around the `&&` to check regex only if `starswith` already matches.

But yes, this looks great. I'd take one more look through the changes tomorrow and then we can merge I think.

KristofferC: Fixed

test/examples/make.jl
```diff
@@ -1,13 +1,15 @@
+using Compat
+
 # Defines the modules referred to in the example docs (under src/) and then builds them.
 # It can be called separately to build the examples/, or as part of the test suite.
 #
 # It defines a set of variables (`examples_*`) that can be used in the tests.
 # The `examples_root` should be used to check whether this file has already been included
 # or not and should be kept unique.
-isdefined(:examples_root) && error("examples_root is already defined\n$(@__FILE__) included multiple times?")
+@isdefined(examples_root) && error("examples_root is already defined\n$(@__FILE__) included multiple times?")
```
mortenpi: `@isdefined` is not available on 0.6. We should have `isdefined(@__MODULE__, :examples_root)` instead. It should be possible to run the `make.jl` file separately as well.

src/DocChecks.jl
```diff
@@ -226,13 +226,17 @@ end
 
 # Regex used here to replace gensym'd module names could probably use improvements.
 function checkresult(sandbox::Module, result::Result, doc::Documents.Document)
-    mod_regex = Regex("(Symbol\\(\"$(sandbox)\"\\)|$(sandbox))[,.]")
+    sandbox_name = module_name(sandbox)
+    mod_regex = Regex("(Main\\.)?(Symbol\\(\"$(sandbox_name)\"\\)|$(sandbox_name))[,.]")
+    mod_regex_nodot = Regex(("(Main\\.)?$(sandbox_name)"))
```
mortenpi: Is "Main" optional in these regexes because of 0.6 / 0.7 differences? If so, we could maybe have a small note here saying we should clean it up once 0.6 support is dropped.

KristofferC: The difference is in e.g.

```
julia> module M
       struct P end
       @show P
       end;
```

**0.6**

```
P = M.P
```

**0.7**

```
P = Main.M.P
```

However, I was not sure if there are still places where `M.P` is printed so I left it optional.

mortenpi: Hmm.. and I guess JuliaLang/julia#23806 would revert back to just `M.P`, right?

KristofferC: No, I don't think so since that is only for types and the type `P` is not reachable from `Main`.

KristofferC: Yes, for example:

```
julia> module M
           sqrt(3)
           sqrt = 2
       end
ERROR: cannot assign variable Base.sqrt from module M
```

mortenpi: Right, never mind this comment then :smile:

test/examples/make.jl
KristofferC: I put an `@isdefined` definition in `runtests.jl` that defaults to `isdefined` on 0.6 (otherwise this would have failed on 0.6).

PR #577, opened 2017-10-06
smallest possible change to make tests pass on 0.7

KristofferC:
With this tests should pass on 0.7. I intend to merge this quickly since tests are needed to pass for other PRs and this almost only touches the tests (and fixes an exposed bug).

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -474,7 +474,7 @@ function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
 
     # Special-case support for displaying SVG graphics. TODO: make this more general.
     output = mimewritable(MIME"image/svg+xml"(), result) ?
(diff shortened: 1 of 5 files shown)


mortenpi:
Looks fine to me. A couple of quick follow up question though:

* `Base.invokelatest` is almost certainly the correct solution, but what changed that the tests started failing suddenly?
* Which PR was 0.7.0-DEV.1877?

By the way, I intend to take a closer look at the other PRs over the weekend. Thanks for all of them, you're on fire bro :laughing:

KristofferC:
* After I moved the definition for the module, it is now evaluated at the same as the call to the show method, so there is no opportunity to bump the world age counter.
* https://github.com/JuliaLang/julia/pull/23579

Thanks in advance  for looking over the PRs. My immidiate goal is to have Documenter run on 0.7 without the deprecation warnings and be able to filter stacktraces from doctests so we can enable them in CI for julia base.

PR #583, opened 2017-10-15
Allow for different line range formattings depending on repo type

cbecker:
I realized that the links to source code with line and line ranges bookmarks were not working for bitbucket because the syntax is different than the syntax of github.

This aims at fixing that, by using a LineRangeFormatting struct that adapts to the repository domain.  Let me know if you have any suggestions on how to improve this PR.

I can easily make another one for v0.5 if this one looks ok for you. Thanks for the hard work behind Documenter.jl, I find it a super useful package :)

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -500,11 +500,11 @@ end
 #      "https://bitbucket.org/xxx" => RepoBitbucket
 # If no match, returns RepoUnknown
 function repo_host_from_url(repoURL::String)
-    if contains(repoURL, "bitbucket.org")
+    if contains(repoURL, "bitbucket")
         return RepoBitbucket
-    elseif contains(repoURL, "github.com")
+    elseif contains(repoURL, "github")
         return RepoGithub
-    elseif contains(repoURL, "gitlab.com")
+    elseif contains(repoURL, "gitlab")
         return RepoGitlab
     else
         return RepoUnknown
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -425,7 +425,7 @@ function url(remote, repo, mod, file, linerange)
     # Replace any backslashes in links, if building the docs on Windows
     file = replace(file, '\\', '/')
     # Format the line range.
-    line = format_line(linerange, LineRangeFormatting(repo))
+    line = format_line(linerange, LineRangeFormatting(repo_host_from_url(repo)))
     # Macro-generated methods such as those produced by `@deprecate` list their file as
     # `deprecated.jl` since that is where the macro is defined. Use that to help
     # determine the correct URL.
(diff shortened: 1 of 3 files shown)


src/Utilities/Utilities.jl
```diff
@@ -501,11 +501,25 @@ function linerange(text, from)
...
 
-function format_line(range::AbstractRange)
-    top = format_line(first(range))
-    return length(range) <= 1 ? top : string(top, '-', format_line(last(range)))
+struct LineRangeFormatting
+    prefix::String
+    separator::String
+
+    function LineRangeFormatting(repoURL::String)
+        if (startswith(repoURL, "https://bitbucket.org") || startswith(repoURL, "http://bitbucket.org"))
```
cbecker: I am wondering whether there is a cleaner way to do this. Does anyone know of whether we can easily parse a URL/URI with julia's base packages?

cbecker: I just found this: https://github.com/JuliaDocs/Documenter.jl/blob/181a1b6bbded1904ef69767d0312ed7619bd448a/src/Writers/HTMLWriter.jl#L411 . so we could use `contains`, although it may match latter parts of the URI and lead to an unwanted behavior

mortenpi: I reckon it is probably good enough, so it would be fine if we stick to that. But I think it would be good if this check would be factored out into a separate function though (maybe returning `:github`, `:bitbucket`, `:gitlab`?), so that we wouldn't duplicate that logic.

cbecker: @mortenpi makes sense, I'll make the changes and ping you when done. Would you place that functionality in `Utilities.jl` and use it from other modules like `HTMLWriter.jl`? To know where to place them already.

mortenpi: > Would you place that functionality in `Utilities.jl` and use it from other modules like `HTMLWriter.jl`?

Yup

cbecker:
@mortenpi I did some changes and travis is throwing a strange error. Do you know if it's possible to re-trigger it? I get the error:

```
The command "curl -A "$CURL_USER_AGENT" -s -L --retry 7 'https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linux64.tar.gz' | tar -C ~/julia -x -z --strip-components=1 -f -" failed and exited with 2 during .
```

test/utilities.jl
```diff
@@ -80,6 +80,34 @@ end
...
+        expectedString = "L2-L5"
+
+        formatting = Documenter.Utilities.LineRangeFormatting(repoType)
+        @test Documenter.Utilities.format_line(lineRange, formatting) == expectedString
+    end
+
+    let
+        repoType = Documenter.Utilities.RepoBitbucket
+        lineRange = 2:5
+        expectedString = "2:5"
```
mortenpi: Same style note here.

src/Utilities/Utilities.jl
```diff
@@ -501,11 +521,25 @@ function linerange(text, from)
...
+        if host == RepoBitbucket
+            new("", ":")
+        else
+            # default is github-style
+            new("L", "-")
+        end
+    end
+end
+
+function format_line(range::Range, format::LineRangeFormatting)
```
mortenpi: `Range` is apparently deprecated. We should stick to `AbstractRange` I think.

```
WARNING: importing deprecated binding Base.Range into Utilities.
WARNING: Base.Range is deprecated, use Base.AbstractRange{T} where T instead.
  likely near /home/travis/.julia/v0.7/Documenter/src/Utilities/Utilities.jl:538
```

cbecker: Thanks, but this is for v0.7, right? In v0.6 we still have Range. What's the usual way you deal with these changes?

mortenpi: [Compat](https://github.com/JuliaLang/Compat.jl) provides `AbstractRange` on 0.6. Compat is the standard way of supporting multiple Julia versions at the same time.

cbecker: Thanks, I didn't know about it, that's great. I've just finished addressing all comments and pushed.

PR #584, opened 2017-10-17
Backports for 0.11.3: Allow for different line range formattings depending on repo type (for 0.5)

cbecker:
Same as https://github.com/JuliaDocs/Documenter.jl/pull/583 but for Julia v0.5 .

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -447,7 +447,7 @@ if VERSION >= v"0.5.0-dev+3442"
         # Replace any backslashes in links, if building the docs on Windows
         file = replace(file, '\\', '/')
         # Format the line range.
-        line = format_line(linerange)
+        line = format_line(linerange, LineRangeFormatting(repo_host_from_url(repo)))
         # Macro-generated methods such as those produced by `@deprecate` list their file as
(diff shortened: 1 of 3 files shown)


cbecker:
Travis is red but I think it's because it's checking against julia v0.6 and julia v0.7, which is not the goal of this PR. Let me know if you have any comments, I'm eager to merge this as I'm using bitbucket and I'd like to start using Documenter on my project :)

mortenpi:
This should target `release-0.11` I think. Julia 0.5 was dropped in Documenter 0.12, so there is no need to merge this into the 0.10 branch, right?

0.7 tests won't pass, but 0.6 and 0.5 ones should.

cbecker:
Thanks, yes, I noticed that I was using an outdated clone. I've updated the PR, I'll close and re-open the PR to trigger CI.

mortenpi:
The commit still looks outdated (the TLDs are still there). Probably best to simply [`cherry pick -x`](https://git-scm.com/docs/git-cherry-pick) the 860546f50bad1578e451a48437115681f84e32fc commit.

cbecker:
Thanks, I missed that. Should be alright now.

PR #586, opened 2017-10-18
Backports for 0.11.4: Fix url generation function to work with symlinked paths

cbecker:
Took me a while to find out why the `source` link was not appearing when I was building the docs.

Turned out that I had a symlink from `$HOME/.julia` to another path, and the `file` and `root` variables used in the `url` function (see commit diff) would then have different base names, as `cd()` translates the links to the real paths.

This fixes this issue. I will make the fix for master if you're ok with this one. 

---

Ref: #587

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -415,7 +415,7 @@ function in_cygwin()
 end
 
 function url(repo, file)
-    file = abspath(file)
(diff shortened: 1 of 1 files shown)


cbecker:
@mortenpi I've updated this PR. The tests pass on my local copy, let's see what happens with travis.

PR #587, opened 2017-10-18
Fix url generation function to work with symlinked paths

cbecker:
Took me a while to find out why the source link was not appearing when I was building the docs.

Turned out that I had a symlink from $HOME/.julia to another path, and the file and root variables used in the url function (see commit diff) would then have different base names, as cd() translates the links to the real paths.

This fixes this issue. I will make the fix for master if you're ok with this one.

[Backport here https://github.com/JuliaDocs/Documenter.jl/pull/586]

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -402,7 +402,7 @@ function repo_commit(file)
 end
 
 function url(repo, file; commit=nothing)
-    file = abspath(file)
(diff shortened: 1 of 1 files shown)


cbecker:
@mortenpi what do you think about this PR? I re-ran the tests, they pass now.

mortenpi:
Thanks! Will try to backport and tag quickly.

PR #588, opened 2017-10-20
regex needs to match both expected and actual output to filter

KristofferC:
Fixes a stupid bug...


mortenpi:
Thanks for catching this, LGTM!

PR #593, opened 2017-11-03
improve error reporting at parse failure

KristofferC:


mortenpi:
What is the actual difference? As far as I can tell, the outputs are identical. Or is it just that `ParseError` specifically does not have a nice `Base.showerror` implemented?

Should this perhaps go into the `warn` function on line 60 instead?

KristofferC:
There was a purpose to this (I noticed it when I was moving the Base parse function) but I'm not sure what the reason was anymore :D

PR #610, opened 2017-12-22
Fix Void deprecations

ararslan:


mortenpi:
Thanks @ararslan! I'll push a few more commits here soon to fix the broken string stuff and hopefully get the tests pass again.

ararslan:
Great! Looks like the tests are passing now so I think this should be good to go.

Issue #612, opened 2017-12-23
Undefined variable when using symbols


Datseris:
This is an example I used:

````markdown
```@example 1
using DynamicalSystems
using PyPlot
PyPlot.ioff() # hide

ds = Systems.logistic()
i = 1
pvalues = 2:0.001:4
ics = [rand() for m in 1:10]
n = 50
Ttr = 5000
parameter = :r 
output = orbitdiagram(ds, i, parameter, pvalues; n = n, Ttr = Ttr)

figure()
for (j, p) in enumerate(pvalues)
    plot(p .* ones(output[j]), output[j], lw = 0,
    marker = "o", ms = 0.5, color = "black")
end
xlabel("\$r\$"); ylabel("\$x\$")
savefig("logisticdiagram.png"); nothing # hide
```
````

The line `parameter = :r` gives an `UndefVarError [:r]` when I run make.jl . If I change this to not defining `parameter` and instead passing the value immediately to the function, I wasn't getting the error anymore:
`output = orbitdiagram(ds, i, :r, pvalues; n = n, Ttr = Ttr)`

mortenpi:
Can reproduce. My guess is that Documenter gets confused with the RHS, thinking the `Symbol` literal is a variable symbol.

mortenpi:
Fixed by #614.

PR #614, opened 2017-12-26
fix symbols on RHS of assignments in examples

KristofferC:
Possible fix to #612.

Would be nice if someone could try this on their docs, cc @Datseris


Datseris:
Hi, to try this I first have to checkout `Pkg.checkout("Documenter", "kc/fix_symbol")`, correct?

KristofferC:
Yeah, or you just use git from the command line like:

```
git fetch
git checkout kc/fix_symbol
```

Datseris:
I am sorry, but it doesn't work.

It is stuck on 
```
```
for quite some time now, long enough to make me believe that there is something wrong.

Datseris:
P.S.: I also had some issues while checking out (it was saying it cannot forward merge). I do not know how impactful this was.

KristofferC:
That's interesting. Can I try it out myself. Do you use the master branch of some package?

Datseris:
If you are talking about the packages I used when I encountered this, then no. The only thing you have to do is `Pkg.add("ChaosTools")` and then create a markdown file with:

```
using ChaosTools
using PyPlot

ds = Systems.logistic()
i = 1
pvalues = 2:0.001:4
ics = [rand() for m in 1:10]
n = 50
Ttr = 5000
output = orbitdiagram(ds, i, :r, pvalues; n = n, Ttr = Ttr)
```
and run it with `make.jl`. The problem arises when one replaces the last line with 
```
param = :r
output = orbitdiagram(ds, i, param, pvalues; n = n, Ttr = Ttr)
```

mortenpi:
```julia
y = :x
@show y
```
is a simple MWE as far as I can tell.

KristofferC:
Your MWE that shows this PR doesn't work? It seems to work ok locally:

screen shot 2017-12-26 at 21 12 39

mortenpi:
No, it works fine. I just meant that it would be a simple test, but I missed that you already had it in the examples. I tested it with the `ChaosTools` example as too, it works fine. LGTM!

You're worried that it might break some existing `@example` blocks? It would be good to test with some package that has a lot of them, just in case.

KristofferC:
I was thinking about the comment https://github.com/JuliaDocs/Documenter.jl/pull/614#issuecomment-353963447 explicitly saying that it doesn't work.

Datseris:
Yeah I am rerunning it now, maybe it was some bad install or something

Datseris:
```
Pkg.checkout("Documenter", "kc/fix_symbol")

INFO: Checking out Documenter kc/fix_symbol...
WARNING: Cannot perform fast-forward merge.
INFO: Pulling Documenter latest kc/fix_symbol...
WARNING: Cannot perform fast-forward merge.
INFO: No packages to install, update or remove
```

When I run `make.jl` it fails:
```
 !! failed to run code block.

UndefVarError(:r) [src\chaos\orbitdiagram.md]
```

Here is the code block with \`\`\`\@example header:
```
using DynamicalSystems
using PyPlot

ds = Systems.logistic()
i = 1
pvalues = 2:0.1:4
ics = [rand() for m in 1:10]
n = 50
Ttr = 5000
param = :r
output = orbitdiagram(ds, i, param, pvalues; n = n, Ttr = Ttr)
```

Datseris:
I am guessing it's just coz of that "fast-forward merge" thing? I am not really experienced with git.

mortenpi:
Yeah, I don't think it checked out the branch, which is why it is still failing. Could try Git directly to check out the branch? You should be able to do it within your Julia session:

```
julia> cd(Pkg.dir("Documenter"))
shell> git status
shell> git checkout kc/fix_symbol
```

Datseris:
I am sorry, but I really couldn't make it work. I don't know what to do, git always shows me that I am on kc/fix_symbol but yet I can never properly checkout there. 

Anyway I think its clear that this is an issue from my end, and right now I don't want to spend more time on fixing git. Since you both confirmed that the PR works, I think you can go ahead and merge.

PR #620, opened 2018-01-07
make Selectors.dispatch non-generated

shashi:
I really like organization of code that Selectors affords. But it's basically not extensible by other packages because `dispatch` is generated.

This patch is required for https://github.com/JuliaGizmos/DocServer.jl to work...


mortenpi:
Theoretically the generated function should be a bit faster if `dispatch` would get called multiple time, but the difference should be negligible in practice -- `dispatch()` for a particular selector typically only gets called once. :+1: to making it extensible, LGTM!

Could you also update the related docstring?

shashi:
Thanks. And yes I'll push the fixed docstrings soon.

mortenpi:
Thanks!

mortenpi:
By the way, DocServer looks interesting! If you don't mind me asking, where do you see DocServer mainly being used?

Is it OK for you to work with Documenter's `master` for a short while (a few weeks)?

shashi:
DocServer will help me create examples which talk to Julia using web widgets... For example https://github.com/JuliaGizmos/InteractNext.jl could be used to plot a function where the visitor can vary some parameters. Something like @ChrisRackauckas's http://app.juliadiffeq.org/ode could then be done within Documenter docs. Another package that will soon be able to interact with Julia this way Makie.jl

Yes I'm fine using `master`.

ChrisRackauckas:
Wooh, interesting. I'd like to migrate my app over to something less "low level" some time.

Issue #621, opened 2018-01-07
support rel="canonical" so search engines only index the latest stable version


tkluck:
Serving similar content (e.g. different versions of the same documentation) can make content hard to find in search engines. [The site readthedocs.io address this by using canonical URLs](https://docs.readthedocs.io/en/latest/canonical.html). This seems a useful feature for Documenter.jl's html output as well.

What do you think? I could maybe squeeze in a pull request if you support the idea.

mortenpi:
Good idea, a PR would be welcome!

The canonical URLs can't really be relative, right? So I don't really see any way of determining the canonical URL automatically, but an option like `html_canonical = "https://juliadocs.github.io/Documenter.jl/latest/"` to determine the root of the URL could be the way to go perhaps.

mortenpi:
Closed by #622.

PR #622, opened 2018-01-08
HTMLWriter: add html_canonical option

tkluck:
As discussed in #621, This allows search engines to point their users only to the most relevant version of a certain page. This is particularly useful when serving many similar pages (e.g. many versions of the same documentation).

See https://en.wikipedia.org/wiki/Canonical_link_element for details.

### DIFF
--- a/test/examples/make.jl
+++ b/test/examples/make.jl
@@ -157,3 +157,35 @@ examples_html_doc = makedocs(
     ],
     doctest = false,
 )
+info("Building mock package docs: HTMLWriter with pretty URLs and canonical links")
+examples_html_doc = makedocs(
+    debug = true,
+    root  = examples_root,
+    build = "builds/html-pretty-urls",
+    format   = :html,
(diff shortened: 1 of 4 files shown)


test/examples/make.jl
```diff
@@ -157,3 +157,35 @@ examples_html_doc = makedocs(
     ],
     doctest = false,
 )
+info("Building mock package docs: HTMLWriter with pretty URLs and canonical links")
+examples_html_doc = makedocs(
+    debug = true,
+    root  = examples_root,
+    build = "builds/html-pretty-urls",
```
mortenpi: The build should go into a separate directory. But I think it would actually be fine to just enable `html_canonical` for the normal "pretty URLs" build and not have a separate build. The `html_canonical = nothing` case will be tested by the other HTML build.

tkluck:
Thanks for the suggestions @mortenpi ! Just pushed new commits incorporating them.

PR #624, opened 2018-01-12
fix some deprecations

fredrikekre:
Building the docs on CI results in a huge amount of depwarn again, fixed most of them.

Needs https://github.com/JuliaLang/Compat.jl/pull/452 to pass on julia 0.6

It would be nice with a new release with this and 611aa82bfac3ea64306a5b8e8be9675cbc519eb8 which should silence most of the warnings.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.6
-Compat 0.41.0
+Compat 0.43.0
 DocStringExtensions 0.2
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.6
-Compat 0.43.0
+Compat 0.46.0
 DocStringExtensions 0.2
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -307,7 +307,8 @@ Code blocks may have some content that does not need to be displayed in the fina
 
 ````markdown
 ```@example
-srand(1) # hide
+using Documenter # hide
+Documenter.Random.srand(1) # hide
 A = rand(3, 3)
 b = [1, 2, 3]
 A \ b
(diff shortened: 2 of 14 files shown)


fredrikekre:
Tests pass now, but there was a newer nightly on macOS, which also included some recent `find*` changes, should be fixed with https://github.com/JuliaLang/Compat.jl/pull/457 so we could either wait for that or merge this and fix that in a separate PR.

JeffBezanson:
Thanks for this! I suspect these deprecations are slowing down the doc build in julia CI quite a bit, so would be good to get this released as soon as possible.

KristofferC:
Looks like DocStringExtensions need to be fixed as well? 

```
Stacktrace:
 [1] <(::Int64, ::Nothing) at ./operators.jl:227
 [2] >(::Nothing, ::Int64) at ./operators.jl:250
 [3] keywords(::Function, ::Method) at /home/travis/.julia/v0.7/DocStringExtensions/src/utilities.jl:252
.
.
```

on Travis

fredrikekre:
Yes, I am waiting for https://github.com/JuliaLang/Compat.jl/pull/457 for that

KristofferC:
Ah, of course. Sorry.

Keno:
Can we just take the most direct route possible to getting rid of the deprecations, even if that means an ugly hack here? All the deprecation warnings are causing Base CI to time out.

JeffBezanson:
Seconded. This is now top priority, as it is slowing down Base development quite a bit. In many cases CI times out because of it.

mortenpi:
The nightly test should pass now. Once it does, I'll merge and tag.

KristofferC:
Ok, this passed on 0.6 and nightly (linux)

PR #627, opened 2018-01-18
Silence of the depwarns

fredrikekre:
There is no end to the deprecations Julia master generates :slightly_smiling_face:


mortenpi:
LGTM, thanks! I'd be happy to tag it right away as well if you'd like.

KristofferC:
https://github.com/JuliaDocs/Documenter.jl/pull/628 could perhaps be sneaked in as well.

PR #640, opened 2018-02-15
HTMLWriter.jl: lowercase categories

### LINKED ISSUE #634: Casing of search result types should be consistent
Initially reported as part of #629. In the example there, a search for "markdown" (http://docs.julialang.org/en/latest/search/?q=markdown) produced the following results:

>- [Markdown syntax](https://docs.julialang.org/en/latest/manual/documentation/#Markdown-syntax-1)(section)
>- [Markdown](https://docs.julialang.org/en/latest/stdlib/Markdown/#)(page)
>- [Markdown](https://docs.julialang.org/en/latest/stdlib/Markdown/#Markdown-1)(section)
>- [Markdown Syntax Extensions](https://docs.julialang.org/en/latest/manual/documentation/#Markdown-Syntax-Extensions-1)(section)
>- [InteractiveUtils.varinfo](https://docs.julialang.org/en/latest/stdlib/InteractiveUtils/#InteractiveUtils.varinfo)(Function)
>- [Advanced Usage](https://docs.julialang.org/en/latest/manual/documentation/#Advanced-Usage-1)(section)
>- [Tables](https://docs.julialang.org/en/latest/manual/documentation/#Tables-1)(section)
>- [Links](https://docs.julialang.org/en/latest/manual/documentation/#Links-1)(section)
>- [Documentation](https://docs.julialang.org/en/latest/manual/documentation/#Documentation-1)(section)
>- [Unicode Input](https://docs.julialang.org/en/latest/manual/unicode-input/#Unicode-Input-1)(section)
>- [Language changes](https://docs.julialang.org/en/latest/NEWS/#Language-changes-1)(section)
>- [Deprecated or removed](https://docs.julialang.org/en/latest/NEWS/#Deprecated-or-removed-1)(section)

It's unclear why "(Function)" is in uppercase, while "(section)" and "(page)" are in lowercase. I would expect this to be consistent (all in lowercase).

This seems an easy enough issue that even someone not familiar with the codebase could fix, but I'd need some pointers regarding which files / lines to look at. Perhaps somewhere around here? https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Utilities/Utilities.jl#L280-L284

waldyrious:
this ensures that the category of entries in search results listings
(e.g. "function", "method", "module", "page", "section", etc.)
will be rendered consistently in lowercase.

Fix #634

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -556,7 +556,7 @@ function search_flush(sib)
         "location": "$(jsonescape(ref))",
         "page": "$(jsonescape(sib.page_title))",
         "title": "$(jsonescape(sib.title))",
-        "category": "$(jsonescape(string(sib.category)))",
+        "category": "$(jsonescape(lowercase(string(sib.category))))",
         "text": "$(jsonescape(text))"
     },
     """)


fredrikekre:
I think it makes sense to have `Function`, `Method`, and `Module` capitalized as is, since the docstring is for something of type `Function`, `Method` and `Module`. Perhaps instead change section etc to Section?

waldyrious:
I understand that reasoning, but in the context of the search results (and especially among the other categories that *aren't* Julia types), it's not clear that that's the reason why those are capitalized. Capitalizing everything, to my eyes, would look more like an instance of the style of capitalization that affects every word in a title, as opposed to those that lowercase words like "of", "and", etc.

As I mentioned in #634, the only way I can think of to make the reason for the capitalization more evident would be to either link to the documentation describing those types, or formatting those categories with a monospaced font (just like you did in your comment above to make it clear what you meant). Would any of these work for you?

mortenpi:
I would argue that the words here refer to concepts, not specific types, so they're plain English words. Also, I think people first and foremost associate "module", "functions" etc with the keywords, which are also lowercase. So I would merge this as is.

PR #641, opened 2018-02-16
Don't use std(in|out|err) as identifiers

ararslan:
Needed for https://github.com/JuliaLang/julia/pull/25959. In that PR, the lowercase names `stdin`, `stdout`, and `stderr` will be claimed by Base, so this avoids overwriting those names, which should hopefully allow documentation to build on that PR.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -570,8 +570,8 @@ where
 """
 function withoutput(f)
     # Save the default output streams.
-    stdout = STDOUT
-    stderr = STDERR
+    default_stdout = STDOUT
+    default_stderr = STDERR
 
     # Redirect both the `STDOUT` and `STDERR` streams to a single `Pipe` object.
     pipe = Pipe()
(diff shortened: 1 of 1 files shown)


src/DocChecks.jl
```diff
@@ -172,7 +172,7 @@ mutable struct Result
     file   :: String # File in which the doctest is written. Either `.md` or `.jl`.
     value  :: Any        # The value returned when evaluating `input`.
     hide   :: Bool       # Semi-colon suppressing the output?
-    stdout :: IOBuffer   # Redirected STDOUT/STDERR gets sent here.
+    io     :: IOBuffer   # Redirected STDOUT/STDERR gets sent here.
```
JeffBezanson: It shouldn't be necessary to change field names.

ararslan: Fair point, though arguably this is a better name anyway. ¯\\\_(ツ)\_/¯ I can change it back if that'd be preferable.

fredrikekre:
The failure on nightly is fixed in https://github.com/JuliaDocs/Documenter.jl/pull/643/commits/31044e5d5015154069630741d3e597c75aebccbb

ararslan:
Thanks for restarting CI, @mortenpi! Could you merge and tag?

PR #644, opened 2018-02-17
Escape all characters not allowed in JS strings

mortenpi:
@samoconnor thanks for the report and references! Could you test this branch to make sure it fixes your docs?


samoconnor:
I will try to get to this in a few days. Lots on at the moment.
If you'd like to try it for yourself, the docs are here:  https://github.com/JuliaCloud/AWSCore.jl/tree/gh-pages

mortenpi:
It seems to fix the AWSCore build, so we can have this in the next tag as well.

PR #646, opened 2018-02-20
Some more deprecation fixes

fredrikekre:


src/Documenter.jl
```diff
@@ -34,6 +34,10 @@ else
     using Base: ip_matches_func
 end
 
+@static if VERSION >= v"0.7.0-DEV.3656"
+    import Pkg
+end
```
mortenpi: Wouldn't `import Compat: Pkg` work?

fredrikekre:
Unrelated nightly windows failure, requires admin to merge.

mortenpi:
Thanks! You would also want a tag soon-ish I guess?

fredrikekre:
> You would also want a tag soon-ish I guess?

No hurry.

PR #647, opened 2018-02-22
Bring back color printing

fredrikekre:
also remove an old `takebuf_str` method.


fredrikekre:
Unrelated nightly windows failure, requires admin to merge.

mortenpi:
Definite :+1: to removing `takebuf_str`. But what's the actual behavior difference with the colors? As far as I can tell, `julia --color=yes docs/make.jl` has color and `julia docs/make.jl` doesn't, both with this PR and current Documenter master, across Julia versions.

mortenpi:
Ah OK, the difference is on Travis. The nightly builds still have less color than the 0.6 ones though, even with this PR, but that makes sense since this only touches the doctest printout. But why is Travis colorless with the nightly in the first place, even though `Pkg.test()` gets called with `--color=yes`?

fredrikekre:
See #25067, color printing are now governed by `get(io, :color, false)` rather than `Base.have_color` and in the case for this PR we print to an `IOBuffer` without color.

PR #650, opened 2018-02-27
support kwargs for doctests

fredrikekre:
This makes it possible to define doctest-local setups and filters in a nicer way, and closer the doctest inside the docstring. Without something like this it is very awkward to use filter for inline docstrings.
Examples:
````
```jldoctest; setup = :(foo(x) = x^2)
julia> foo(2)
4
```
````

````
```jldoctest; filter = r"[0-9\.]+ seconds \(.*\)"
julia> @time [1,2,3,4]
  0.000003 seconds (5 allocations: 272 bytes)
4-element Array{Int64,1}:
 1
 2
 3
 4
```
````

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -118,10 +118,10 @@ function __ans__!(m::Module, value)
 end
 
 function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page)
-    m = match(r"jldoctest[ ]?(.*)$", block.language)
-    if m !== nothing
+    lang = block.language
+    if startswith(lang, "jldoctest")
         # Define new module or reuse an old one from this page if we have a named doctest.
-        name = m[1]
+        name = match(r"jldoctest[ ]?(.*)$", split(lang, ';', limit = 2)[1])[1]
         sym = isempty(name) ? gensym("doctest-") : Symbol("doctest-", name)
         sandbox = get!(page.globals.meta, sym) do
             newmod = Module(sym)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/DocChecks.jl [hunk 1]
```diff
@@ -118,21 +118,33 @@ function __ans__!(m::Module, value)
 end
 
 function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page)
-    m = match(r"jldoctest[ ]?(.*)$", block.language)
-    if m !== nothing
+    lang = block.language
+    if contains(lang, r"jldoctest")
```
KristofferC: `startswith`?

src/DocChecks.jl
```diff
@@ -118,21 +118,34 @@ function __ans__!(m::Module, value)
...
         end
+        # parse keyword arguments to doctest
+        # kwargstr = split(lang, ';')
+        d = Dict()
+        idx = Compat.findfirst(equalto(';'), lang)
+        if idx !== nothing
+            kwargs = Meta.parse("($(lang[nextind(lang, idx):end]),)")
+            for kwarg in kwargs.args
+                @assert kwarg.head === :(=)
+                @assert isa(kwarg.args[1], Symbol)
```
mortenpi: I think it would be good if these errors would be a bit more informative. Since we don't have Markdown line number information, I'd say we should output the filename and the doctest when there's a typo in the arguments.

src/DocChecks.jl
```diff
@@ -118,21 +118,34 @@ function __ans__!(m::Module, value)
 end
 
 function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page)
-    m = match(r"jldoctest[ ]?(.*)$", block.language)
-    if m !== nothing
+    lang = block.language
+    if startswith(lang, "jldoctest")
         # Define new module or reuse an old one from this page if we have a named doctest.
-        name = m[1]
+        name = match(r"jldoctest[ ]?(.*)$", split(lang, ';')[1])[1]
```
mortenpi: `split(; limit=1)`?

fredrikekre:
Sure, we can go with the colon, I don't have a strong preference, but as you say the semicolon is used for keywords in julia functions and thats why I went with that.

src/DocChecks.jl
fredrikekre: I assume you mean `limit=2`?

src/DocChecks.jl
fredrikekre: Good idea

src/DocChecks.jl
mortenpi: I do indeed.

fredrikekre:
Updated with a better error message for invalid kwarg definition. I think that it might be nice to keep the semicolon as a hint to julia syntax for kwargs.

fredrikekre:
Also, when giving filters as
````
```@meta
DocTestFilters = [r"foo", r"bar"]
```
````
we specify it as a vector of filters, perhaps we should then also have
````
```jldoctest; filter = [r"foo", r"bar"]
```
````
It would be nice to support both I guess.

mortenpi:
> I think that it might be nice to keep the semicolon as a hint to julia syntax for kwargs.

Yup, let's go with the semicolon.

fredrikekre:
Clarified the docs a bit. This is good to go from my POV, but could use another review.

PR #654, opened 2018-03-01
use invokelatest when calling showerror, fix #653

### LINKED ISSUE #653: Printing of error in doctest
The following doctest, taken from julia base
````
```jldoctest
julia> struct MyException <: Exception
           msg::AbstractString
       end

julia> function Base.showerror(io::IO, err::MyException)
           print(io, "MyException: ")
           print(io, err.msg)
       end

julia> err = MyException("test exception")
MyException("test exception")

julia> sprint(showerror, err)
"MyException: test exception"

julia> throw(MyException("test exception"))
ERROR: MyException: test exception
```
````

fails on the last expression. Instead of `ERROR: MyException: test exception` Documenter evaluates to `MyException("test exception")` so somehow we never end up in the custom `showerror` function.

fredrikekre:


mortenpi:
Thanks! I'm not going to bother backporting this, rather just tag a minor release over the weekend which would include #650.

PR #656, opened 2018-03-05
add doctest = :fix option to makedocs

### LINKED ISSUE #448: Doctest "apply fix" flag
e.g. an official API that accomplishes the same thing as @vtjnash's patch from https://github.com/JuliaLang/julia/pull/20417 in a way you could request with a flag

fredrikekre:
This makes it easier to fix multiple outdated doctests in one command, which is useful to do e.g. when preparing a new release.

Not really sure how to test this.

fix #448

### DIFF
--- a/docs/src/man/doctests.md
+++ b/docs/src/man/doctests.md
@@ -263,6 +263,26 @@ julia> @time [1,2,3,4]
     The global filters, filters defined in `@meta` blocks, and filters defined with the `filter`
     keyword argument are all applied to each doctest.
 
+## Fixing outdated Doctests
+
+To fix outdated doctests, the `doctest` flag to [`makedocs`](@ref) can be set to
+`doctest = :fix`. This will run the doctests, and overwrite the old results with
+the new output.
+
+!!! note
+
+    The :fix option currently only works for LF line endings (`'\n'`)
+
+!!! note
+
+    It is recommended to `git commit` any code changes before running the doctest fixing.
+    That way it is simple to restore to the previous state if the the fixing goes wrong.
+
+!!! note
+
+    There are some corner cases where the fixing algorithm may replace the wrong code snippet.
+    It is therefore recommended to manually inspect the result of the fixing before committing.
+
 
 ## Skipping Doctests
 
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -96,7 +96,7 @@ the document generation when an error is thrown. Use `doctest = false` keyword i
 [`Documenter.makedocs`](@ref) to disable doctesting.
 """
 function doctest(doc::Documents.Document)
-    if doc.user.doctest
+    if doc.user.doctest === :fix || doc.user.doctest
         println(" > running doctests.")
         for (src, page) in doc.internal.pages
             empty!(page.globals.meta)
@@ -131,7 +131,7 @@ function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page
         end
 
         # Normalise line endings.
-        code = replace(block.code, "\r\n" => "\n")
+        block.code = replace(block.code, "\r\n" => "\n")
 
         # parse keyword arguments to doctest
         d = Dict()
@@ -147,7 +147,7 @@ function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page
                         in '$(meta[:CurrentFile])'
 
                         ```$(lang)
-                        $(code)
+                        $(block.code)
                         ```
                         """
                         )
@@ -162,11 +162,11 @@ function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page
             Meta.isexpr(expr, :block) && (expr.head = :toplevel)
             eval(sandbox, expr)
         end
-        if contains(code, r"^julia> "m)
-            eval_repl(code, sandbox, meta, doc, page)
+        if contains(block.code, r"^julia> "m)
+            eval_repl(block, sandbox, meta, doc, page)
             block.language = "julia-repl"
-        elseif contains(code, r"^# output$"m)
-            eval_script(code, sandbox, meta, doc, page)
+        elseif contains(block.code, r"^# output$"m)
+            eval_script(block, sandbox, meta, doc, page)
             block.language = "julia"
         else
             push!(doc.internal.errors, :doctest)
@@ -175,7 +175,7 @@ function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page
                 Invalid doctest block. Requires `julia> ` or `# output` in '$(meta[:CurrentFile])'
 
                 ```$(lang)
-                $(code)
+                $(block.code)
                 ```
                 """
             )
@@ -194,23 +194,23 @@ end
 # Doctest evaluation.
 
 mutable struct Result
-    code   :: String # The entire code block that is being tested.
-    input  :: String # Part of `code` representing the current input.
-    output :: String # Part of `code` representing the current expected output.
+    block  :: Markdown.Code # The entire code block that is being tested.
+    input  :: String # Part of `block.code` representing the current input.
+    output :: String # Part of `block.code` representing the current expected output.
     file   :: String # File in which the doctest is written. Either `.md` or `.jl`.
     value  :: Any        # The value returned when evaluating `input`.
     hide   :: Bool       # Semi-colon suppressing the output?
     stdout :: IOBuffer   # Redirected stdout/stderr gets sent here.
     bt     :: Vector     # Backtrace when an error is thrown.
 
-    function Result(code, input, output, file)
-        new(code, input, rstrip(output, '\n'), file, nothing, false, IOBuffer())
+    function Result(block, input, output, file)
+        new(block, input, rstrip(output, '\n'), file, nothing, false, IOBuffer())
     end
 end
 
-function eval_repl(code, sandbox, meta::Dict, doc::Documents.Document, page)
-    for (input, output) in repl_splitter(code)
-        result = Result(code, input, output, meta[:CurrentFile])
+function eval_repl(block, sandbox, meta::Dict, doc::Documents.Document, page)
+    for (input, output) in repl_splitter(block.code)
+        result = Result(block, input, output, meta[:CurrentFile])
         for (ex, str) in Utilities.parseblock(input, doc, page; keywords = false)
             # Input containing a semi-colon gets suppressed in the final output.
             result.hide = Documenter.REPL.ends_with_semicolon(str)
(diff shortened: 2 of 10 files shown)


fredrikekre:
Will do something for https://github.com/JuliaDocs/Documenter.jl/issues/448#issuecomment-304837875 too.

fredrikekre:
Added a commit that looks for the complete code-snippet before doing the replacement instead of just searching for the result. This should essentially eliminate the risk of double replacement as described in https://github.com/JuliaDocs/Documenter.jl/issues/448#issuecomment-304837875

fredrikekre:
Actually, this will still turn 
````
```jldoctest
julia> 1 + 1
3

julia> 1 + 2
3
```
````
into
````
```jldoctest
julia> 1 + 1
2

julia> 1 + 2
2
```
````

**Comment by fredrikekre** [comment]
Okay, this version should work in all (?) cases:
First search in the file for the whole code snippet (rather than just the output string), then search within the code block for the input string, and last replace the old output with the new one.

Will add some tests, but feel free to review the implementation.

**Comment by fredrikekre** [comment]
Turns out the method I described in the last comment still failed in for the following case:
````
```jldoctest
julia> 1 + 2
2

julia> 1 + 2
2
```
````
because when we try to fix the second fail in this block we can not find that block in the file, since we replaced it with
````
```jldoctest
julia> 1 + 2
3

julia> 1 + 2
2
```
````

I solved this by storing the full `Markdown.Code` object in `Result` and then when replacing the content in the file we also replace the content in the `Markdown.Code` object such that the internal representation mathces that in the file.

I have now also added tests with the corner cases I have found when working on this.

File: src/DocChecks.jl [hunk 1]
```diff
@@ -343,8 +353,8 @@ function report(result::Result, str, doc::Documents.Document)
     println(ioc)
     printstyled(ioc, "> File: ", result.file, "\n", color=:cyan)
     printstyled(ioc, "\n> Code block:\n", color=:cyan)
-    println(ioc, "\n```jldoctest")
-    println(ioc, result.code)
+    println(ioc, "\n```$(result.block.language)")
```
fredrikekre: Storing the full `Markdown.Code` object also allows to print the actual `language` field here, including `name`, `filter` and `setup`. Previously it was a bit confusing that those things were left out from the error printing.

docs/src/man/doctests.md
```diff
@@ -263,6 +263,12 @@ julia> @time [1,2,3,4]
     The global filters, filters defined in `@meta` blocks, and filters defined with the `filter`
     keyword argument are all applied to each doctest.
 
+## Fixing outdated Doctests
+
+To fix outdated doctests, the `doctest` flag to [`makedocs`](@ref) can be set to
+`doctest = :fixup`. This will run the doctests, and overwrite the old results with
+the new output.
+
```
mortenpi: Might be good to recommend here that the user should `git commit` their code changes, just in case, before running the doc build with this option on. That way, should the find-and-replace algorithm go a bit wonky and completely mess up the source file, the user can easily just `git reset --hard` to their state.

test/doctests.jl
```diff
@@ -0,0 +1,77 @@
...
+    # fix up
+    makedocs(source = srcdir, build = builddir, doctest = :fixup)
+    # test that strict = true works
+    makedocs(source = srcdir, build = builddir, strict=true)
+    # also test that we obtain the expected output
+    open(joinpath(srcdir, "index.md"), "r") do f
+        str = read(f, String)
+        @test str == good_str
+    end
+end
```
mortenpi: This should also fix doctests in `.jl` files too right? Could perhaps also test that to make sure that the file paths get resolved correctly etc.

fredrikekre:
Yea thats the a corner case I also thought of. But it should be pretty rare(?) that the same input yields different output. Will add the recommendation to `git commit` and also recommend that the result should be inspected manually.

mortenpi:
> But it should be pretty rare(?) that the same input yields different output.

Agreed. As you said, a note in the docs saying that there are potential edge cases etc. should be enough.

KristofferC:
Great stuff to have!

fredrikekre:
Hmm, this will not actually work for script-doctests....

fredrikekre:
Ok, we are good I think.

src/DocChecks.jl
```diff
@@ -232,14 +232,14 @@ function eval_repl(code, sandbox, meta::Dict, doc::Documents.Document, page)
...
-function eval_script(code, sandbox, meta::Dict, doc::Documents.Document, page)
+function eval_script(block, sandbox, meta::Dict, doc::Documents.Document, page)
     # TODO: decide whether to keep `# output` syntax for this. It's a bit ugly.
     #       Maybe use double blank lines, i.e.
     #
     #
     #       to mark `input`/`output` separation.
-    input, output = split(code, "\n# output\n", limit = 2)
-    result = Result(code, "", output, meta[:CurrentFile])
+    input, output = split(block.code, "\n# output\n\n", limit = 2)
```
fredrikekre: Any dangers with this? The problem was that `output` here became e.g. `"\n2" while the result string became "3", thus we replaced 
````
```jldoctest
1 + 2

# output

2
```
````
with 
````
```jldoctest
1 + 2

# output
3
```
````
instead of 
````
```jldoctest
1 + 2

# output

3
```
````

Comment (fredrikekre) [reply src/DocChecks.jl]: Perhaps it would be safer to do
```julia
input, output = split(block.code, "# output\n", limit = 2)
input = rstrip(input, '\n')
output = lstrip(output, '\n')
```
?

fredrikekre: From the manual:
> Note that the amount of whitespace appearing above and below the # output line is not significant and can be increased or decreased if desired.

So we should strip `'\n'` instead

fredrikekre:
Perhaps we should change `:fixup` -> `:fix`. I like `doctest = (true|false|:fix)` better I think. Thoughts?

fredrikekre:
`doctest = (:yes|:no|:fix)` might be even better, but w/e. Changed `:fixup` -> `fix`.

src/DocChecks.jl
```diff
@@ -366,6 +378,34 @@ function print_indented(buffer::IO, str::AbstractString; indent = 4)
...
 
+function fix_doctest(result::Result, str, doc::Documents.Document)
+    code = result.block.code
+    open(Base.find_source_file(result.file), "r+") do f
+        content = read(f, String)
+        seekstart(f)
+        truncate(f, 0)
+        # first look for the entire code block
+        codeidx = findnext(code, content, 1)
+        @assert codeidx != 0:-1 # this should never happen
```
mortenpi: If you have CRLF line endings then this throws. It also leaves the source file empty on error right now.. so even if we fix this particular case, I think it would probably be a good idea to `try ... catch` all of this and restore the original file when are any errors (and then rethrow).

fredrikekre: Good catch. We could potentially try to detect the line ending style, unless there could sometimes be a mix of the two?
Probably good to wrap in try catch anyway

fredrikekre:
Pushed a new version that should fail more gracefully if we don't find the code block. Also leaves the source file untouched in case of failure.

src/DocChecks.jl
```diff
@@ -366,6 +378,43 @@ function print_indented(buffer::IO, str::AbstractString; indent = 4)
...
+    # output stream
+    io = Compat.IOBuffer(sizehint = sizeof(content))
+    # first look for the entire code block
+    codeidx = Compat.findnext(code, content, 1)
+    if codeidx === nothing
+        Utilities.warn("Could not find code block in source file")
+        return
+    end
+    # write everything up until the code block
+    write(io, content[1:prevind(content, first(codeidx))])
```
mortenpi: I was able to trigger a

```
ERROR: LoadError: BoundsError: attempt to access "# Test\r\n\r\n```jldoctest\r\njulia> 2 + 2\r\n0000\r\n```\r\n"
  at index [0]
Stacktrace:
 [1] prevind(::String, ::Int64, ::Int64) at ./strings/basic.jl:426
 [2] prevind at ./strings/basic.jl:421 [inlined]
 [3] fix_doctest(::Documenter.DocChecks.Result, ::SubString{String}, ::Documenter.Documents.Document) at /home/morten/.julia/v0.7/Documenter/src/DocChecks.jl:395
 [4] checkresult(::Module, ::Documenter.DocChecks.Result, ::Dict{Symbol,Any}, ::Documenter.Documents.Document) at /home/morten/.julia/v0.7/Documenter/src/DocChecks.jl:305
```

with [CRLFs (this file)](https://gist.githubusercontent.com/mortenpi/8ad8460373222d24869ed7eb9cbd86fa/raw/86a595fcbac03ab4c960352bc0333688094c444a/index.md) right now. So this manages to avoid your checks. Sorry for keeping on breaking this :laughing:

src/DocChecks.jl
```diff
@@ -366,6 +378,43 @@ function print_indented(buffer::IO, str::AbstractString; indent = 4)
...
+    filename = Base.find_source_file(result.file)
+    # read the file containing the code block
+    content = open(f -> read(f, String), filename)
+    # output stream
+    io = Compat.IOBuffer(sizehint = sizeof(content))
+    # first look for the entire code block
+    codeidx = Compat.findnext(code, content, 1)
+    if codeidx === nothing
+        Utilities.warn("Could not find code block in source file")
+        return
```
mortenpi: This check (and the one below)  _should_ still never happen, right? Perhaps we should keep `throw`ing/`error`ing/`assert`ing here so that in case they are triggered, they don't go unnoticed.

test/doctests.jl
```diff
@@ -0,0 +1,165 @@
...
+1 + 2
+
+# output
+
+3
+```
+\"\"\"
+foo() = 1
+end # module
+"""
```
mortenpi: Having a test with CRLFs wouldn't hurt. We could also keep these in separate files and just `cp` them over to `tempdir` to avoid problems with differing line endings in a single file (although a simple `replace(..., '\n' => "\r\n")` works as well I guess).

src/DocChecks.jl
fredrikekre: My thought here was that since we have replaces `\r\n` with `\n` in the code block https://github.com/JuliaDocs/Documenter.jl/blob/b64a4d7f2200b1a3ca4b051c54eade54bce77bce/src/DocChecks.jl#L133-L134  we would obtain `nothing` here since we have not performed the replacement in the file where we search.

src/DocChecks.jl
fredrikekre: Hmm.. How? I simply get:
```
 !! Could not find code block in source file
```

fredrikekre: Could it be that you don't have https://github.com/JuliaLang/Compat.jl/commit/5ff98924e8f010c9cf9f83758655a7f8025dbeb3 ?

mortenpi: Ah, I think you're right, my bad. I happened to be running this with my 0.7 binary and I think it is a bit too old (9 days).

fredrikekre:
It is a bit tricky since the byte index in the string will change after we have replaced the line endings. I added a note that this currently only works for LF line endings for now. 

Also added some better tests, including multiline output which was missing before.

fredrikekre:
I have tried this with Julia base and some other projects and it works well for real projects too, not just the trivial test cases in the tests, so if you are happy with the current state, feel free to merge and we can work out possible problems as we find them.

fredrikekre:
I should just mention that it does not quite work together with doctest filters, but perhaps easier to review the changes to make that work in a separate PR.

mortenpi:
Yup, let's iterate on this in future PRs, if the need arises. Thanks a lot for this, this is a really neat feature!

fredrikekre:
This is a pretty useful feature when updating packages, doing this https://github.com/KristofferC/Tensors.jl/pull/86/commits/213ac2c10a14878de72bbeb74db88137bfb01ebb manually would take some time.

PR #657, opened 2018-03-06
more deprecation silencing

fredrikekre:
needs https://github.com/JuliaLang/Compat.jl/pull/512

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -668,7 +668,7 @@ function genkeys(package; remote="origin")
 
         # Prompt user to add public key to github then remove the public key.
         let url = "https://github.com/$user/$repo/settings/keys"
-            info("add the public key below to $url with read/write access:")
+            Compat.@info("add the public key below to $url with read/write access:")
             println("\n", read("$filename.pub", String))
             rm("$filename.pub")
         end
@@ -677,7 +677,7 @@ function genkeys(package; remote="origin")
         # *not* encoded for the sake of security, but instead to make it easier to
         # copy/paste it over to travis without having to worry about whitespace.
         let url = "https://travis-ci.org/$user/$repo/settings"
-            info("add a secure environment variable named 'DOCUMENTER_KEY' to $url with value:")
+            Compat.@info("add a secure environment variable named 'DOCUMENTER_KEY' to $url with value:")
             println("\n", base64encode(read(".documenter", String)), "\n")
             rm(filename)
         end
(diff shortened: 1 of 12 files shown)


src/Utilities/Utilities.jl
```diff
@@ -552,6 +553,12 @@ newlines(other) = 0
 
 # Output redirection.
 # -------------------
+@static if VERSION < v"0.7.0-DEV.3951"
+    link_pipe!(pipe; reader_supports_async = true, writer_supports_async = true) =
+        Base.link_pipe(pipe, julia_only_read = reader_supports_async, julia_only_write = writer_supports_async)
+else
+    import Base: link_pipe!
+end
```
fredrikekre: I don't think this would be very useful in Compat, so made this local fix here instead.

src/Expanders.jl
```diff
@@ -411,6 +411,7 @@ end
 
 function Selectors.runner(::Type{EvalBlocks}, x, page, doc)
     sandbox = Module(:EvalBlockSandbox)
+    eval(sandbox, :(import Compat.Markdown))
```
mortenpi: Why do we need this?

src/Documenter.jl
```diff
@@ -668,7 +668,7 @@ function genkeys(package; remote="origin")
 
         # Prompt user to add public key to github then remove the public key.
         let url = "https://github.com/$user/$repo/settings/keys"
-            info("add the public key below to $url with read/write access:")
+            @info("add the public key below to $url with read/write access:")
```
mortenpi: These are in a submodule, so needs an `import Compat: @info`.

src/Expanders.jl
fredrikekre: Not sure, but it silenced a bunch of warnings. I think because previously `Markdown` was always available by default in all modules since they all have implicit `using Base` in them.

fredrikekre:
```
WARNING: importing deprecated binding Base.Markdown into EvalBlockSandbox.
WARNING: Base.Markdown is deprecated, run `using Markdown` instead
  likely near /home/fredrik/.julia/v0.7/Documenter/test/examples/make.jl:91
```

fredrikekre: I guess we `eval` things into the module which references `Markdown`.

fredrikekre: Ah, good catch, should be fixed now.

Issue #661, opened 2018-03-09
deployment experience on Gitlab


tpapp:
I have been exploring Gitlab and summarized my solutions in a minimal repository [GitlabJuliaDemo.jl](https://gitlab.com/tkpapp/GitlabJuliaDemo.jl) that demonstrates CI + docs generation + docs deployment.

Some issues I encountered:

1. Documenter.jl needs a `git` binary, so one needs to be installed, either via `apt` or the Docker image. The first option is easy but adds considerably to the build time. I wonder if `Libgit2` could be used instead? But I realize this is a major overhaul.

2. I follow the [suggestion](https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Public-Interface-1) in the manual and use
```julia
repo = "https://gitlab.com/tkpapp/GitlabJuliaDemo/blob/{commit}{path}#L{line}"
```
but get URLs like
```
https://gitlab.com/tkpapp/GitlabJuliaDemo/blob/master/docs/src/index.md#L{line}
```
(ie `{line}` literally}), which also seems to be incorrect, but I don't know what the correct one is. Did the template syntax change?

The bright side: deployment is very easy, just moving the files to `public`, no key setup or anything.

mortenpi:
I'll close this issue, since 2. has been addressed and I opened a dedicated issue for 1. (switching to LibGit2 is a good idea I think). Thanks, by the way, for playing around with Documenter on GitLab and writing these things up -- this is really helpful feedback!

PR #662, opened 2018-03-10
Fixes for line ranges in URLs

mortenpi:
@tpapp: this should fix the issue with the `{line}`s in URLs your reported in #661. Also, as far as I can tell, the formatting of the line range was not correct either (GitLab is `#Lxx-yy` not `#Lxx-Lyy`). Could you by any chance test this branch with your docs?

You should use `repo = "https://gitlab.com/tkpapp/GitlabJuliaDemo/blob/{commit}{path}#{line}"`, since the `L` is now included automatically.

I am a bit uncertain about BitBucket right now. I don't think the format we are implementing right now (`#{xx}-{yy}`) actually works (anymore?), and they appear to use different formats in different versions (?).

* On their main site they appear to use `{filename}-{xx}:{yy}` (e.g. [../merger.py#merger.py-5:26](https://bitbucket.org/tutorials/merger/src/792fc12e175fdf9d0d0ea284bd8d19dd7aa4ec78/merger.py#merger.py-5:26))

* Whereas NASA's BitBucket repo uses a simple `#{xx}-{yy}` (e.g. [`../edsc-nlp/browse/Dockerfile#10-12](https://git.earthdata.nasa.gov/projects/EDSC/repos/edsc-nlp/browse/Dockerfile#10-12))

@cbecker: Since you implemented this in #583, do you perhaps have any thoughts on how to handle BitBucket?


tpapp:
@mortenpi: thanks for the quick fix. I can confirm that
```julia
repo = "https://gitlab.com/tkpapp/GitlabJuliaDemo.jl/blob/{commit}{path}#{line}"
```
works on Gitlab (I missed the `.jl` part before even though it is part of the repo name, my mistake).

I am looking forward to having this merged to `master`, and eventually released, it would simplify my scripts.

fredrikekre:
What's the status here? Rebase, merge and release v0.15? 😃

mortenpi:
I am not going to touch the BitBucket line ranges for now. But an issue would be welcome if it is not working for anyone.

PR #667, opened 2018-03-11
no rounding of search and version selector

KristofferC:
I must admit that I find the rounding of the search and version selector quite ugly. This changes it to be just rectangular.

After:

screen shot 2018-03-09 at 22 25 57

### DIFF
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -199,7 +199,6 @@ nav.toc select {
     margin: 0 auto;
     font-size: .83em;
     border: 1px solid #c9c9c9;
-    border-radius: 1em;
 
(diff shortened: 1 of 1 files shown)


mortenpi:
I'm basically neutral on the matter. This choice is from the original Sphinx theme. I am slightly sceptical of making these small subjective changes since you can go back and forth indefinitely. But since there seems to be support for this I propose let's leave this open for a week or two and, if there are no dissenting opinions, merge then.

The slight rounding of the selector you have seems to be platform dependent. For reference, this is what is looks like on Ubuntu / Gnome:

rounding

KristofferC:
Alright, I'll just configure these small things in my own css from now on. I think we can close this.

PR #671, opened 2018-03-14
print line number range for failed doctests

fredrikekre:
````
=====[Test Error]==============================

> Location: /home/fredrik/.julia/v0.7/TestPackage/src/TestPackage.jl:12-19        <--- Line info here is new.

> Code block:

```jldoctest; filter = r"[0-9\.]+ seconds \(.*\)"
julia> @time [1, 2, 3]
  5.0 seconds (5 allocations: 272 bytes)
3-element Array{Int64,1}:
 2
 4
 6
```

> Subexpression:

    @time [1, 2, 3]

> Output Diff:

  5.0 0.000002 seconds (5 allocations: 272 bytes)
3-element Array{Int64,1}:
 1
 2
 4
 63

=====[End Error]===============================
````

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -88,6 +88,20 @@ sigs(::Any) = Type[Union{}]
 # Julia code block testing.
 # -------------------------
 
+# helper to display linerange for error printing
+function find_codeblock_in_file(code, file)
+    content = read(Base.find_source_file(file), String)
+    content = replace(content, "\r\n" => "\n")
+    blockidx = Compat.findfirst(code, content)
+    if blockidx !== nothing
+        startline = countlines(IOBuffer(content[1:prevind(content, first(blockidx))]))
+        endline = startline + countlines(IOBuffer(code)) + 1 # +1 to include the closing ```
+        return ":$(startline)-$(endline)"
+    else
+        return ""
+    end
+end
+
 """
 $(SIGNATURES)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/DocChecks.jl [hunk 1]
```diff
@@ -140,11 +153,12 @@ function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page
             kwargs = Meta.parse("($(lang[nextind(lang, idx):end]),)")
             for kwarg in kwargs.args
                 if !(isa(kwarg, Expr) && kwarg.head === :(=) && isa(kwarg.args[1], Symbol))
+                    file = meta[:CurrentFile]
+                    lines = find_codeblock_in_file(block.code, file)
```
mortenpi: It should have trouble finding the line numbers in CRLF files though, right? Because `block.code` has the newlines normalized? I am wondering if we should just move the normalization into the `eval_*` functions and not modify the `.code` field at all.

fredrikekre:
Good idea. I have updated the code to only modify line endings where needed. I will add some tests later.

fredrikekre:
Eh... it only works for the code blocks coming from `.md` files. Julia seems to change to LF endings when loading `.jl` files.
```
julia> str = "\"```\r\njulia> 1 + 1\r\n2\r\n```\"\r\nfoo() = 1"
"\"```\r\njulia> 1 + 1\r\n2\r\n```\"\r\nfoo() = 1"

julia> write("testfile.jl", str);

julia> include("testfile.jl");

julia> str = @doc foo;

julia> str.content[1].content[1].content[1].code
"julia> 1 + 1\n2"
```

src/DocChecks.jl
```diff
@@ -88,6 +88,20 @@ sigs(::Any) = Type[Union{}]
 # Julia code block testing.
 # -------------------------
 
+# helper to display linerange for error printing
+function find_codeblock_in_file(code, file)
+    content = read(Base.find_source_file(file), String)
+    content = replace(content, "\r\n" => "\n")
```
fredrikekre: We can do this to make this work with CLRF since the number of lines does not change.

PR #684, opened 2018-03-23
Mind indentation when looking for source code

fredrikekre:


fredrikekre:
> I love the regex magic 😆

Yea its fun to play around with new stuff once in while 😄

src/DocChecks.jl
```diff
@@ -280,8 +285,8 @@ function filter_doctests(strings::NTuple{2, AbstractString},
     meta_block_filters == nothing && meta_block_filters == []
     doctest_local_filters = get(meta[:LocalDocTestArguments], :filter, [])
     for r in [doc.user.doctestfilters; meta_block_filters; doctest_local_filters]
-        if all(occursin.(r, strings))
-            strings = replace.(strings, r => "")
+        if all(occursin.((r,), strings))
+            strings = replace.(strings, (r => "",))
```
martinholters: Am I being stupid or did this break filtering on Julia 0.6? I think broadcasting over tuples of unequal length is different on 0.6 than on 0.7, always taking the length of the first tuple as the output length. So knowing that `strings` has length two, this should rather be
```julia
       if all(occursin.((r, r), strings))
            strings = replace.(strings, (r => "", r => ""))
```

fredrikekre: Hmm, I get a length-2 tuple on both versions.
`v0.6.2`:
```
julia> using Compat; occursin.((r"f",), ("foo", "bar"))
(true, false)
```
`master`:
```
julia> occursin.((r"f",), ("foo", "bar"))
(true, false)
```

Do you have an example that is broken?

martinholters: Oops, just noticed that I'm still 0.6.1 where indeed I get
```julia
julia> using Compat; occursin.((r"f",), ("foo", "bar"))
(true,)
```
So I guess updating to 0.6.2 will resolve my problems, will report back if not. Sorry for the noise.

fredrikekre: https://github.com/JuliaLang/julia/pull/23887 was backported to v0.6.2

PR #685, opened 2018-03-23
only use p tags for loose lists

KristofferC:


src/Writers/HTMLWriter.jl
```diff
@@ -925,6 +925,12 @@ mdconvert(list::Markdown.List, parent; kwargs...) = (Markdown.isordered(list) ?
 
 mdconvert(paragraph::Markdown.Paragraph, parent; kwargs...) = Tag(:p)(mdconvert(paragraph.content, paragraph; kwargs...))
 
+const list_has_loose_field = :loose in fieldnames(Markdown.List)
```
mortenpi: I'd add a small comment here noting that this is for 0.6 compat. That way we'll hopefully notice it when we drop 0.6 support.

KristofferC:
added a comment

PR #690, opened 2018-03-26
fix #689

### LINKED ISSUE #689: Documenter in combination with Revise
Building the docs for a `Revise`d docstring causes the following:
```
WARNING: Method definition foo(Any) in module TestPackage at /home/fredrik/.julia/v0.6/TestPackage/src/TestPackage.jl:9 overwritten at /home/fredrik/.julia/v0.6/TestPackage/src/TestPackage.jl:9.
WARNING: replacing docs for 'TestPackage.foo :: Tuple{Any}' in module 'TestPackage'.
 > checking for missing docstrings.
 > running doctests.
 > checking footnote links.
Documenter: populating indices.
ERROR: LoadError: MethodError: no method matching dirname(::Void)
Closest candidates are:
  dirname(::AbstractString) at path.jl:146
Stacktrace:
 [1] url(::String, ::String, ::Module, ::Void, ::UnitRange{Int64}) at /home/fredrik/.julia/v0.6/Documenter/src/Utilities/Utilities.jl:381
 [2] url(::String, ::String, ::Base.Docs.DocStr) at /home/fredrik/.julia/v0.6/Documenter/src/Utilities/Utilities.jl:378
 [3] (::Documenter.Writers.HTMLWriter.##29#30{Documenter.Writers.HTMLWriter.HTMLContext,Documenter.Documents.NavNode,Documenter.Utilities.DOM.Tag})(::Tuple{Base.Markdown.MD,Base.Docs.DocStr}) at /home/fredrik/.julia/v0.6/Documenter/src/Writers/HTMLWriter.jl:714
[...]
```

because `file` here https://github.com/JuliaDocs/Documenter.jl/blob/16ce61cfb385f991c4f186bf6cfe1b403193d225/src/Utilities/Utilities.jl#L381 is `nothing`. This works on julia master though, so something like

fredrikekre:
julia v0.6 in combination with Revise sometimes causes `file` to be `nothing`.

This adds a check for this, and a fast return if so, fix #689.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -378,6 +378,7 @@ end
 url(remote, repo, doc) = url(remote, repo, doc.data[:module], doc.data[:path], linerange(doc))
 
 function url(remote, repo, mod, file, linerange)
+    file === nothing && return nothing # needed on julia v0.6, see #689
     remote = getremote(dirname(file))
     isabspath(file) && isempty(remote) && isempty(repo) && return nothing
 


tpapp:
I don't think this fixes #689: the error goes away, but changes in docstrings don't show up in the generated docs unless Julia is restarted.

fredrikekre:
> I don't think this fixes #689: the error goes away, but changes in docstrings don't show up in the generated docs unless Julia is restarted

Works perfect for me. If the docstring is not updated it is not Documenters fault, Documenter just queries the base help system for the docstring.

fredrikekre:
Just to make sure, you  are doing `using Revise` before `using YourPackage` right?

tpapp:
`using Revise` is in my `.juliarc.jl`, so it is always loaded. The problem was doing 
```sh
julia -i make.jl
```
which somehow prevents it from working. But if I just run everything interactively it works fine.

PR #695, opened 2018-04-03
add continued = (false|true) kwarg to at-example

### LINKED ISSUE #694: Incremental examples
It would be very useful if there was an easy way to divide ```` ```@example ```` blocks and delay evaluation of the content. Example
````
```@example test
A = []
for i in 1:10
```

Some text

```@example test
    push!(A)
end
A
```
````
This will not work because the first block is not a complete julia expression. It is possible to kinda simulate this with

````
```@example
A = []
for i in 1:10
    push!(A) # hide
end # hide
A # hide
nothing # hide
```

Some text

```@example test
A = [] # hide
for i in 1:10 # hide
    push!(A)
end
A
```
````
but copy-pasting like that is rarely a good solution.

### PR DESCRIPTION
**Pull request by fredrikekre**
fix #694 
perhaps easier to review as https://github.com/JuliaDocs/Documenter.jl/pull/695/files?w=1

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -447,28 +447,42 @@ end
 # --------
 
 function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
-    matched = match(r"^@example[ ]?(.*)$", x.language)
-    matched === nothing && error("invalid '@example' syntax: $(x.language)")
     # The sandboxed module -- either a new one or a cached one from this page.
-    name = matched[1]
+    name = match(r"^@example[ ]?(.*)$", first(split(x.language, ';', limit = 2)))[1]
     sym  = isempty(name) ? gensym("ex-") : Symbol("ex-", name)
     mod  = get!(page.globals.meta, sym, Module(sym))::Module
+
+    # "parse" keyword arguments to example (we only need to look for continued = true)
+    continued = occursin(r"continued\s*=\s*true", x.language)
+
     # Evaluate the code block. We redirect stdout/stderr to `buffer`.
     result, buffer = nothing, IOBuffer()
-    for (ex, str) in Utilities.parseblock(x.code, doc, page)
-        (value, success, backtrace, text) = Utilities.withoutput() do
-            cd(dirname(page.build)) do
-                eval(mod, Expr(:(=), :ans, ex))
+    if !continued # run the code
+        # check if there is any code wating
+        if haskey(page.globals.meta, :ContinuedCode)
+            code = page.globals.meta[:ContinuedCode] * "\n" * x.code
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Expanders.jl [hunk 1]
```diff
@@ -447,27 +447,45 @@ end
...
+
     # Evaluate the code block. We redirect stdout/stderr to `buffer`.
     result, buffer = nothing, IOBuffer()
-    for (ex, str) in Utilities.parseblock(x.code, doc, page)
-        (value, success, backtrace, text) = Utilities.withoutput() do
-            cd(dirname(page.build)) do
-                eval(mod, Expr(:(=), :ans, ex))
+    if execute
+        # check if there is any code wating
+        if isdefined(mod, :__WAITING__CODE__)
```
mortenpi: I guess the "waiting code" could also be stored directly in e.g. `page.globals.meta`, so that we wouldn't create arbitrary stuff in the module, but either way is fine.

If we do go with a variable in the module, I think it would be good to namespace it even more (e.g. `__DOCUMENTER__WAITING__CODE__`). Would make any potential conflict even less likely and it would make it clear where this variable comes from, e.g. when someone does `names(@__MODULE__)` in an at-example. I would also mention it in the manual that Documenter sets that variable.

fredrikekre: > I guess the "waiting code" could also be stored directly in e.g. page.globals.meta, so that we wouldn't create arbitrary stuff in the module, but either way is fine.

Yes, that seems much cleaner.

fredrikekre:
> If I read it right, with this the continued = true blocks do not honor block names anymore.

Yea, that was an oversight.

> I'd suggest that we either don't allow `continued = true` blocks to have a name in the first place, 

I think we probably should require them to be names? Otherwise we will never use that code since the module will have a gensym name.

> or make sure that the names match (minimally, there should at least be a warning if there is a mismatch I think).

We will not know this until we reach the end of the page though? If we disallow  things like 
```
@example A; continued = true
@example B
@example A
```
then we can warn when evaluating `B`.

I think that we should 1. require them to be named and 2. warn at the end of the page if there is stored code that has not been used. Additionally we could disallow the pattern above, but I don't really see the point -- let the users use bad style if they want :stuck_out_tongue: 

The other option would be to say that they shouldn't be named, and say that `continued = true` simply delays the code until the next `@example` and evaluates there. But then you lose the ability for patterns like
```
@example A
@example A; continued = true
@example A
```
which would be a shame.

mortenpi:
> > I'd suggest that we either don't allow `continued = true` blocks to have a name in the first place, 
>
> I think we probably should require them to be names? Otherwise we will never use that code since the module will have a gensym name.

```

PR #700, opened 2018-04-05
Add EditURL key to at-meta block

fredrikekre:
This lets the user specify the link for `"Edit on GitHub"` which can be useful if the page is, for example, generated during doc build and you want to point the link to the source from where the page was generated.

### DIFF
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -205,8 +205,18 @@ Use [`for i = 1:10 ...`](@ref for) to loop over all the numbers from 1 to 10.
 ## `@meta` block
 
 This block type is used to define metadata key/value pairs that can be used elsewhere in the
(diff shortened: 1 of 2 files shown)


mortenpi:
So a theoretical use case would be that you have, say, a manual page in reST and you convert them into `.md` files before running `makedocs`? And of course, you'd like the links to point to the reST files, as opposed to the ephemeral `.md` files?

I would guess that the main use here would be in-repo links though? But in that case the full URL should still be generated dynamically, including honoring `html_edit_branch`. At the same time, allowing external absolute URLs would be good too I think. Which of the two it is can probably determined from whether `EditURL` is relative or absolute (there's `Utilities.isabsurl` for this).

I think it would also be good to explain `EditURL` and its use case a bit in the docs of the at-meta block.

Issue #701, opened 2018-04-09
Subpages


fredrikekre:
It would be nice if Documenter's HTML renderer could support the following pattern:
```
page.md
   |------  subpage1.md
   |------  subpage2.md
```
where the subpages where displayed as H2-pages in the navigation bar.

After discussing with @mortenpi we have converged to the following suggestion:

Introduce a `subpages` function, very similar to `hide` which could be used in `pages` like:

```julia
pages = ["index.md", subpages("page.md", ["subpage1.md", "override heading" => "subpage2.md"])]
```

What would `subpages` do?
The idea is that the only thing `subpages` does is that it changes the rendering of those pages in the navigation bar - from a H1 page to a H2 page.

Currently we always have the matching between a `.md` file and the `.html` file, and this will be preserved for subpages too, they will be their own `.html` pages.

Currently, by default, all html-pages are visible in the navigation bar (unless you use `hide`). The idea is that this should also be preserved, such that subpages are always visible in the navigation bar, even though they are on the H2 level.

We also think that it makes sense to leave the rendering of subpages alone (i.e. not change H1 to H2 for example). This hopefully makes implementation a bit simpler, since the only thing `subpages` will do is to change the navigation bar.

Some issues:
- How should we determine the name of a subpage?
  Currently Documenter looks for H1 headings to determine the name of the page. This will be done for subpages too, but since they are marked as subpages, it makes sense to also allow H2 heading as the highest level in the subpage. Probably we should first look for H1, and if none is found, look for H2. But should we do the same for non-subpages too? I'm guessing implementation is simplified a lot if we do they same lookup for normal pages too.

- If we are at a subpage, should H3 headings be shown in the navigation bar, the same way H2 headings are shown if we are on a normal page?
  Probably not, this just introduces a lot of clutter.

- If `"page.md"` in the example above have H2 headings, should they suddenly pop-up in the navigation bar when we are at `page.md` or should we consider the subpages to be the H2 headings for the main page?
  I think that it would be awkward if the H2-headings suddenly where spliced in between `page.md` and the subpages if we are at the `page.md`, but this can be worked around simply by not including any H2 headings on `page.md` if you don't want that behaviour. Let's just go with whatever makes implementing this simpler IMO.

Did I miss something @mortenpi ?

mortenpi:
There were also two additional small things that came up. Although they are orthogonal to the `subpages` idea, I'll just write them down here:

[...]

PR #702, opened 2018-04-10
Allow Markdown in page title overrides

fredrikekre:
See point 3 in https://github.com/JuliaDocs/Documenter.jl/issues/701#issuecomment-379916726


fredrikekre:
Hmm, this is not quite right:
screenshot from 2018-04-10 10-27-44


`Home` should be right after `>>` rather than below. Not quite sure why, perhaps because the parsed markdown introduces a new paragraph?

mortenpi:
Took the liberty to add an error check. LGTM!

PR #703, opened 2018-04-10
[WIP] search for h2 headings if we don't find h1 on the page

fredrikekre:
Part of  #701

If we don't find a H1 heading, look for the first H2 heading. I think we can do this for all pages, instead of only for subpages. Also changes the fallback to `basename(source)` if there are no H1 or H2 headers.

Edit: This does not quite work: We should probably not render the children for a page with H2 as max level. Currently we get this:

screenshot from 2018-04-10 09-55-45

where `Bar` is H2.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -390,7 +390,7 @@ function navitem(ctx, current, nn::Documents.NavNode)
     end
     item = (nn === current) ? li[".current"](link) : li(link)
 
-    # add the subsections (2nd level headings) from the page
+    # add the subsections (1st and 2nd level headings) from the page
     if (nn === current) && current.page !== nothing
         subs = collect_subsections(ctx.doc.internal.pages[current.page])
         internal_links = map(subs) do s
@@ -813,19 +813,17 @@ function pretty_url(ctx, path::AbstractString)
 end
 
 """
-Tries to guess the page title by looking at the `<h1>` headers and returns the
-header contents of the first `<h1>` on a page (or `nothing` if the algorithm
-was unable to find any `<h1>` headers).
+Tries to guess the page title by first looking at the `<h1>` headers and next at `<h2>`
+headers. Returns the contents of the first `<h1>`/`<h2>` on the page.
+If no `<h1>` or `<h2>` headings are found we fall back to the filename of the page.
 """
 function pagetitle(page::Documents.Page)
-    title = nothing
-    for element in page.elements
-        if isa(element, Markdown.Header{1})
-            title = element.text
-            break
-        end
-    end
-    title
+    elements = page.elements
+    idx = Compat.findfirst(x -> isa(x, Markdown.Header{1}), elements)
+    idx !== nothing && return elements[idx].text
+    idx = Compat.findfirst(x -> isa(x, Markdown.Header{2}), elements)
+    idx !== nothing && return elements[idx].text
+    return basename(page.source)
 end
 
 function pagetitle(ctx, navnode::Documents.NavNode)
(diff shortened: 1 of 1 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -813,30 +813,27 @@ function pretty_url(ctx, path::AbstractString)
 end
 
 """
-Tries to guess the page title by looking at the `<h1>` headers and returns the
-header contents of the first `<h1>` on a page (or `nothing` if the algorithm
-was unable to find any `<h1>` headers).
+Tries to guess the page title by first looking at the `<h1>` headers and next at `<h2>`
+headers. Returns the contents of the first `<h1>`/`<h2>` on the page.
+If no `<h1>` or `<h1>` headings ar found we fall back to the filename of the page.
```
mortenpi: "`<h1>` or `<h1>`" -> "`<h1>` or `<h2>`"

"ar" -> "are"

src/Writers/HTMLWriter.jl
```diff
@@ -813,30 +813,27 @@ function pretty_url(ctx, path::AbstractString)
...
     navnode.title_override === nothing || return navnode.title_override
 
     if navnode.page !== nothing
-        title = pagetitle(getpage(ctx, navnode))
-        title === nothing || return title
+        return pagetitle(getpage(ctx, navnode))
     end
 
-    "-"
+    "-" # unreachable?
```
mortenpi: In practice it should be since `navnode.page === nothing` can only happen when there's a `title_override`. Could perhaps invert the `if` and have a nice `error()` if `navnode.page === nothing` (in case someone runs into case when hacking on the internals etc).

mortenpi:
[`collect_subsections`](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Writers/HTMLWriter.jl#L842-L864) already discards the first `h1`. Should probably do the same with the first `h2`.

The edge case I am wondering about though is when you are actually using `title_override` and have  content before the first `h2`. So maybe only discard `h2` / use for page heading when it's exactly the first (non at-meta?) block?

mortenpi:
Sorry for flipping on this, but I am now actually thinking that perhaps we shouldn't change the heading behavior for normal pages. This became a bit of a TL;DR, but it's probably good to have the logic documented somewhere. Happy to discuss it though.

---

I think we should think in terms of the abstract page outline (similarly to LaTeX `\section{}`/`\subsection{}` and [the HTML outline](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines)). And we should have a strict correspondence between the heading levels and levels in the page outline, to keeps things simple and explicit.

Within this picture, the first highest level heading is interpreted to be a _page heading_ (i.e. the first H1). If you don't have any H1s on the page, Documenter should assume that it is your intent that there are no H1 levels in the page outline. If you want your H2+ to be a page heading, you should change it to a H1.

So there are two distinct things Documenter has to do:

1. Determining the page outline and whether there is a _page heading_ (i.e. every page may have a heading that gets special treatment).
   - For normal pages the page heading is the first H1 on the page, assuming there are no other headings H2 before it.
   - You are allowed to have non-heading content and H3+ headings before the page heading.
   - Headings within other blocks (docstrings, admonitions, block quotes etc.) are ignored when scanning for the page outline. 
   - The page heading does not get displayed in the navigation menu. But non-page-heading H1s will be, with special styling.
   - (I.e. the current behavior.)

2. Determining the page title that is displayed in the navigation menu, in `<title>` etc.
   - If the title is specified in `makedocs`'s `pages` argument (i.e. `title_override`), use that.
   - If not, use the _page heading_.
   - If there is no page heading, print a warning and default to `basename(source)`.

The subpages of #701 would would be different -- the point there is that the subpage headings correspond to H2 headings / levels in the outline. Therefore
  - If there are no H1s, use the first H2 as the page heading.
  - If there are H1s present on the page, determine the page heading like for a normal page.
  - H3s (or H2s, if there are H1s on the page) don't get displayed in the navigation menu since their place is already taken by the page headings.
  - You are allowed to have H3s before the page heading H2.
  - In the HTML output, the H1 and H2 headings retain their styling, leaving the user some freedom to determine how the subpages are rendered.

Allowing H1s to determine the subpage heading goes slightly against the principle that there should be a strict correspondence between the levels and how their used, but I feel that it's acceptable here.

---

A few additional notes:

* Theoretically we _could_ do what the HTML outline algorithm does and ignore absolute heading levels altogether, and create an outline that always steps deeper one level at a time. So, e.g. if you have only H4s and H6s on the page, they become the first and seconds level elements in the outline, determining the page heading and navbar, respectively.

  But I think we should allow the user to have the power to determine their own outlines which may have levels missing etc (e.g. suppose you want a page with headings, but don't want them to show up in the navbar -- you can use H3 for that right now).

* Having multiple H1-level headings on a page is actually a bit weird in terms of the outline. It's a bit tricky to know how to interpret this. LaTeXWriter treats it as if it were a separate page, but in HTMLWriter we just apply special styling in the navigation menu, so it feels like it's somewhere between a page and a H2 heading.

* A few notes on how MkDocs handles headings:
  - It ignores additional H1s and everything that comes after them.
  - If there are no H1s on the page, it actually adds the _page title_ as the first H1 on the page.
  - If you have H2-H1-H2 in it, it doesn't show anything in the navigation bar.

fredrikekre:
Yes, that makes sense. Thanks for writing it up.

PR #706, opened 2018-04-16
stop deploying to the release- folder

fredrikekre:
This should actually be safe since Documenter don't remove the old directories, so links should still work.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -264,6 +264,10 @@ deploydocs(
 )
 ```
 
+When building the docs for a tag (i.e. a release) the documentation is deployed to
+a directory with the tag name (i.e. `vX.Y.Z`) and to the `stable` directory.
+Otherwise the docs are deployed to the `latest` directory.
+
 # Required keyword arguments
 
 **`julia`** is the version of Julia that will be used to deploy generated documentation.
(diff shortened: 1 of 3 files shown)


mortenpi:
Code LGTM. I would leave this open for a short while to see if we gather any feedback.

You're right that this is essentially non-breaking. But I think I should nevertheless ANN the release with this change on Discourse, possibly together with a script / automated PRs that would remove the old `release-x.y` directories from packages.

mortenpi:
Also, I was wondering that perhaps we should move the `release-x.y` ones last in `generate_version_file` in this PR? I.e. in `for group in (named_folders, release_folders, tag_folders)`. Or, I guess we could remove them altogether.. this way they will stay on the `gh-pages` branch, but will be hidden from the reader.

fredrikekre:
Yea lets move them last. If people don't want them to show they can purge the gh-pages branch from these. Then some time in the future we can remove them from generate_version_file, since at that time Documenter has not build those branches in a while anywa.

fredrikekre:
Actually, lets remove them from the version selector but leave the old `release-` folders on the branch. That way links will still work as you say.

fredrikekre:
Rebased.

PR #707, opened 2018-04-16
sort correctly by version number

fredrikekre:
Currently in e.g. Documenters manual `v0.6.x` sort before `v0.10.x`. This sorts tagged dirs and release- dirs according to version number instead.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -486,10 +486,17 @@ function generate_version_file(dir::AbstractString)
         occursin(r"release\-\d+\.\d+", each) ? push!(release_folders, each) :
         occursin(Base.VERSION_REGEX, each)   ? push!(tag_folders,     each) : nothing
     end
+    # put stable before latest
+    sort!(named_folders, rev = true)
+    # sort tags by version number
+    sort!(tag_folders, lt = (x, y) -> VersionNumber(x) < VersionNumber(y), rev = true)
+    # sort release- folders by version number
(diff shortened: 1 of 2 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -486,10 +486,17 @@ function generate_version_file(dir::AbstractString)
         occursin(r"release\-\d+\.\d+", each) ? push!(release_folders, each) :
         occursin(Base.VERSION_REGEX, each)   ? push!(tag_folders,     each) : nothing
     end
+    # put stable before latest
+    sort!(named_folders, rev = true)
+    # sort tags by version number
+    sort!(tag_folders, lt = (x, y) -> VersionNumber(x) < VersionNumber(y), rev = true)
+    # sort release- folders by version number
+    vnum(x) = VersionNumber(match(r"release\-(\d+\.\d+)", x)[1])
```
KristofferC: Is it worth being careful about if someone has made a release that is not parsable as a ´VersionNumber`?

fredrikekre: Lets just not sort these then. I think those should be removed anyways (#706). But then perhaps put tagged version before release- version?

KristofferC: Yeah, just put anything not parsable as a version in the end?

mortenpi: Actually, the `release_folders` folders variable will only have valid ones, because it is also regexed above, no?

fredrikekre:
Thanks.

Issue #482, opened 2017-05-14
Better handling of tagged versions in deploydocs


mortenpi:
E.g. `-rc*` tags shouldn't deploy to `stable/`.

Ref: JuliaLang/julia#21846

huckl3b3rry87:
Is this related to easily being able to delete old tagged versions?

For instance in my package, there are several tagged versions of the docs
https://juliampc.github.io/MPCDocs.jl/stable/

and I would like to delete all of them but one. Is this possible?

mortenpi:
Not really related. Although we might have

If you are certain you want to get rid of the old versions of the docs then you could edit the `gh-pages` branch of the repo yourself and just delete them. You should also remove them from the `versions.js`, although this would get updated automatically on the next deploy as well.

mortenpi:
Also, when a bugfix is released on an older minor branch, it overwrites `stable/` (e.g. say the highest tag is 0.5.1, but 0.4.5 gets released). Thanks @fredrikekre for bringing this up.

mortenpi:
Fixed by #708.

PR #708, opened 2018-04-16
only deploy to stable if we build the latest tag

fredrikekre:
https://github.com/JuliaDocs/Documenter.jl/issues/482#issuecomment-381499222

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -377,7 +377,7 @@ function deploydocs(;
         error("julia must be a string, got $julia ($(typeof(julia)))")
     end
     if !isempty(travis_repo_slug) && !occursin(travis_repo_slug, repo)
-        warn("repo $repo does not match $travis_repo_slug")
+        Compat.@warn("repo $repo does not match $travis_repo_slug")
     end
 
     # When should a deploy be attempted?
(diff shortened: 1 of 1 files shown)


src/Documenter.jl
```diff
@@ -512,8 +512,14 @@ function git_push(
                     gitrm_copy(target_dir, latest_dir)
                     Writers.HTMLWriter.generate_siteinfo_file(latest_dir, "latest")
                 else
-                    gitrm_copy(target_dir, stable_dir)
-                    Writers.HTMLWriter.generate_siteinfo_file(stable_dir, "stable")
+                    # only push to stable if this is the latest stable release
+                    versions = filter!(x -> occursin(Base.VERSION_REGEX, x), readdir(dirname))
+                    maxver = mapreduce(x -> VersionNumber(x), max, v"0.0.0", versions)
+                    vtag = VersionNumber(tag)
```
mortenpi: I think here we do run into the problem that `tag` might not be a well-formed version number.

mortenpi: Putting an `if` around this checking the regex should be good enough.

fredrikekre: And  not deploy at all if the tag is bad?

mortenpi: I would still deploy `tagged_dir`, but not to stable. The release-x.y actually has this check already.

fredrikekre: I think we should just skip deployment then, since it would not be picked up in `generate_version_file` anyway if it is not a valid VersionNumber

mortenpi: Fair enough, they should be consistent indeed.

src/Documenter.jl
```diff
@@ -511,19 +511,23 @@ function git_push(
...
-                        release = "release-$(version.major).$(version.minor)"
-                        gitrm_copy(target_dir, joinpath(dirname, release))
-                        Writers.HTMLWriter.generate_siteinfo_file(joinpath(dirname, release), release)
-                    end
+                    # Build a `release-*.*` folder as well
+                    release = "release-$(version.major).$(version.minor)"
+                    gitrm_copy(target_dir, joinpath(dirname, release))
+                    Writers.HTMLWriter.generate_siteinfo_file(joinpath(dirname, release), release)
+                else
+                    Compat.@warn("tag `$tag` not a valid VersionNumber")
```
mortenpi: I warn already in the beginning of `deploydocs` though, where we do the other checks. Can keep this one as well though, for good measure -- it might help when debugging.

PR #711, opened 2018-04-19
rethrow InterruptException, fix #687

### LINKED ISSUE #687: Specialcase InterruptException in order to rethrow them from code blocks being evaluated?
It is quite hard to interrupt a running doc generation because the interrupt exceptions are caught.

```
InterruptException() [src/examples/juliatypes.md]
^C^C !! failed to run code block.

InterruptException() [src/examples/juliatypes.md]
^C !! failed to run code block.

InterruptException() [src/examples/juliatypes.md]
^C !! failed to run code block.
```

etc etc.

Perhaps InterruptExceptions should be rethrown.

fredrikekre:


src/Utilities/Utilities.jl
```diff
@@ -576,6 +576,9 @@ function withoutput(f)
         try
             f(), true, Vector{Ptr{Cvoid}}()
         catch err
+            # InterruptException should never during normal doc-testing (#687).
```
KristofferC: should never what?

fredrikekre: Left as an excercise for the reader.

Issue #652, opened 2018-02-28
Make named doctests share repl / setup / example blocks


jebej:
This would be useful for testing an example, and then also saving a plot from data generated by the test.

mortenpi:
My feeling is that the doctests should be a bit separate from the other at-blocks. One reason for this is that you want doctests to be as self-contained as possible. But, of course, we already allow doctests to share variables and you can also have setup code with `@meta` blocks. So allowing sharing between doctests and the at-blocks would not necessarily be that big of a stretch, and I understand the use case.

jebej:
Has thing been touched at all? I saw in #718 that you were considering something functionally equivalent, that is to have a doctest block with options eg `compare', and 'output' (it seems the `output` option exists now).

A second use case I have for this is for doctesting a bunch of expressions, and then only displaying another result, but without comparing because it displays some floating point values which might fluctuate.

fingolfin:
On the one hand, this should be pretty easy to implement now.

On the other hand, I worry that the execution context is not quite well-defined: named `@repl` etc. blocks are defined per-page: i.e., the names and shared context are per-page.

But docstrings often come from a different file, and in fact a docstring can in principle also be put into multiple pages. So now you get different "shared" context, and what is or isn't there isn't local anymore.

Issue #252, opened 2016-09-05
Make rendering of doctest output optional


tlycken:
I guess it's a matter of style, but the output section in most of my documentation snippets is very uninteresting, and mostly clutters the reading experience. It would be nice to be able to suppress the output of `# output` and anything following :)

MichaelHatherly:
Is this for docstrings within Julia source files or in the external markdown files?

We can't do any suppression within docstrings since that content should always be able to be viewed correctly independent of `Documenter`.

For doctests in the external markdown files perhaps see if the `@example` and `@repl` blocks are of any use for you needs, since you don't have to worry then about manually writing out the expected output for each block. These blocks allow for the output of individual lines to be hidden using `# hide` comments. (There's some examples of there use in the [Syntax](https://juliadocs.github.io/Documenter.jl/latest/man/syntax.html) section of the manual.)

tlycken:
I'm using it only in markdown files for now, so even if suppression within docstrings is off the table there is some value for me in it :)

`@example` blocks might very well be good enough, though, but I really liked how the doctests share no context at all - if I understand the documentation correctly, all `@example` blocks essentially build up a single, continuous Julia session?

MichaelHatherly:
> blocks essentially build up a single, continuous Julia session?

No, they are also sandboxed in anonymous modules, though you can share state by "naming" the `@example` or `@repl` with `@example name` and `@repl name`.

fredrikekre:
Perhaps
````
```jldoctest; output=(true|false)
foo(a, b) = a + b
foo(2, 3)

# output

5
```
````
which would still test the result, but hide the output in the rendered docs? I think `julia>` style doctests should be unaffected by `output` since you can just append a semicolon instead.

fredrikekre:
Note that this suggestion would still render "correctly" even if used in docstrings, but of course the output would still be there (which is fine).

PR #718, opened 2018-05-05
doctest improvements

fredrikekre:
Two changes to make doctest block more like `@example` blocks:
* Remove `# hide` lines from doctest-blocks
* Completely removes the `# output`  block if it is empty

We might also consider to split script-doctest input/output in two different `Markdown.Code`, that way we never see the `# output` comment in the output.


mortenpi:
LGTM, but also needs updates for the LaTeXWriter and MarkdownWriter. In the former the relevant part is [`LaTeXWriter.jl#L264-L275`](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Writers/LaTeXWriter.jl#L264-L275).

Not exactly sure about MarkdownWriter.. the current behavior seems to be to keel ` ```jldoctest ` as the language? While not ideal, we shouldn't change this behavior I think, since it would probably break existing builds.

fredrikekre:
> Not exactly sure about MarkdownWriter.. the current behavior seems to be to keel ```jldoctest as the language?

I am confused, with `doctest = true` we already do update the language (for all the writers):
https://github.com/JuliaDocs/Documenter.jl/blob/b9b63a73cfe9cce2b16b72eb7677a124cb4c1cd9/src/DocChecks.jl#L187
and
https://github.com/JuliaDocs/Documenter.jl/blob/b9b63a73cfe9cce2b16b72eb7677a124cb4c1cd9/src/DocChecks.jl#L190

mortenpi:
Ah, you're right. The handling of `jldoctest` is a bit inconsistent with the MarkdownWriter -- if doctests are on you get `julia`/`julia-repl`, but with doctests off you keep `jldoctest`. I agree, we should always give `julia`/`julia-repl` to the writers and let them handle it if need be.

MarkdownWriter should output `julia`/`julia-repl` as well I guess. I don't think there are any highlighters that would recognize `jldoctest` (highlight.js used to but that was changed to `julia-repl`). Pygments needs `jlcon`. So nothing should break.

I would separate this from the `doctest` function though (it is not related to verifying/fixing doctests). I'd say the `doctest_replace(doc)` call should go into the `Builder.Populate` step, before `populate!`, and the `doctest_replace` functions should also probably move into `Documents.jl`.

mortenpi:
Hmm.. not sure I agree with the `Walkers` change. What prompted you to move `walk` into `Utilities`?

In my head, `Utilities` are all such functions that could in principle be taken out of Documenter and used elsewhere. The `Walkers.walk` function is very closely tied to the internal workings of Documenter though, and so it's more appropriate to have it together with the other core files, directly in `src/`. It makes undocumented assumptions about the `meta` variable, the Markdown AST tree etc. (A some sort of generic walker API would be suitable for `Utilities`, but I don't think there is any need to generalize this at the moment.)

We could have it in `Documents`, if you'd prefer getting rid of the `Walkers` module. That would arguably be a more appropriate place for it.

test/examples/src/index.md
```diff
@@ -228,6 +228,28 @@ julia> f()
...
+
+```jldoctest
+A = 1
+A += 1 # hide
+A
+
+# output
+
+2
+```
```
mortenpi: This example made me question for an instant if we should allow `#hide` in doctests in the first place. I do think it is good to be consistent with at-example and it can be useful for hiding some repetitive initialization etc. But the other side of it is that the users should probably be able to run the doctests as they appear in the manual.

As a compromise, should we maybe add a note in the docs suggesting to the user to use `#hide` sparingly, to make sure they're not hiding relevant parts of the input?

fredrikekre:
> Hmm.. not sure I agree with the `Walkers` change. What prompted you to move `walk` into `Utilities`?

I needed the `Walker ` module in `Documents` since we moved the replace function there, and it was a mess re-arranging the files to make that work since `Walkers` depended on other modules etc. I can move it to `Documents` if you prefer.

src/Expanders.jl
```diff
@@ -621,4 +621,10 @@ function prepend_prompt(input)
     rstrip(String(take!(out)))
 end
 
+# add methods to Utilities.walk
+Utilities.walk(f, meta, block::Expanders.DocsNodes) = Utilities.walk(f, meta, block.nodes)
+Utilities.walk(f, meta, block::Expanders.DocsNode)  = Utilities.walk(f, meta, block.docstr)
+Utilities.walk(f, meta, block::Expanders.EvalNode)  = Utilities.walk(f, meta, block.result)
+Utilities.walk(f, meta, block::Expanders.MetaNode)  = (merge!(meta, block.dict); nothing)
```
mortenpi: These should also go into `Documents`, since the `*Node`s are defined there.

test/examples/src/index.md
fredrikekre: I agree that this is a bad example and this is not the usecase I have in mind. I am not a fan of disallowing things though, if the user wants to write bad example that is fine with me :smile: But yea, we can add a note about that.

The real usecase I have for this is that I essentially want a doctestsetup that is not hidden. Consider something like:

````
```jldoctest
foo(a, b) = a + b
bar(a, b) = a * b
```

Something something, with the definitions above we can do this cool stuff:

```jldoctest
julia> foo(1, 2)
3

julia> bar(2, 3)
6
```
````

So in the first block there I don't want to compare the output (since it would be very ugly...). This is where I imagined appending
```
nothing # hide

# output

```
to the first block, we can remove the `hide` line, and since we return `nothing` it will compare correctly against the empty output. And then we remove empty output sections...

While I am writing this I feel this might not be the best approach. Perhaps we can allow `jldoctest` blocks without `# output`? Such a block would simply evaluate into the named doctest module, but not capture, or compare, any output. Another option would be to do #652, and then you can use a `@example` block with trailing `nothing # hide` to this end. Not sure...

fredrikekre: Yea, do you compare the output still?

fredrikekre: Perhaps https://github.com/JuliaDocs/Documenter.jl/issues/252#issuecomment-389412041

PR #726, opened 2018-05-15
eval(m, expr) -> Core.eval(m, expr)

fredrikekre:
fixes incoming deprecations from https://github.com/JuliaLang/julia/pull/27098

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -131,8 +131,8 @@ function doctest(doc::Documents.Document)
 end
 
 function __ans__!(m::Module, value)
-    isdefined(m, :__ans__!) || eval(m, :(__ans__!(value) = global ans = value))
-    return eval(m, Expr(:call, () -> m.__ans__!(value)))
+    isdefined(m, :__ans__!) || Core.eval(m, :(__ans__!(value) = global ans = value))
+    return Core.eval(m, Expr(:call, () -> m.__ans__!(value)))
 end
 
 function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page)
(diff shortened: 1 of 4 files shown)


src/DocChecks.jl
```diff
@@ -144,8 +144,7 @@ function doctest(block::Markdown.Code, meta::Dict, doc::Documents.Document, page
         sym = isempty(name) ? gensym("doctest-") : Symbol("doctest-", name)
         sandbox = get!(page.globals.meta, sym) do
             newmod = Module(sym)
-            eval(newmod, :(eval(x) = Core.eval($newmod, x)))
-            eval(newmod, :(eval(m, x) = Core.eval(m, x)))
+            Core.eval(newmod, :(eval(x) = Core.eval($newmod, x)))
```
mortenpi: The second one is not necessary anymore? Why do we need to do this here in the first place?

mortenpi:
Will release 0.18 when this and #718 are in.

src/DocChecks.jl
fredrikekre: It is needed for things like
````
```jldoctest
julia> a = 1; b = 2;

julia> eval(:(a + b))
3
```
````
similar to https://github.com/JuliaLang/julia/pull/27098/files#diff-8d2b8a22d475d81e3a1944aedb57ad42R391

Although, the docstring for single arg `eval` states:

> Evaluate an expression in the current module. Every `Module` (except those defined
with `baremodule`) has its own 1-argument definition of `eval`, which evaluates
expressions in that module.

So presumably the sandbox module should already have its own `eval` defined, which should evaluate into the sandbox, but I get
```
┌ Warning: `eval(e)` is deprecated, use `Core.eval(Main, e)` instead.
│   caller = top-level scope
└ @ Core :0
ERROR: UndefVarError: a not defined
Stacktrace:
 [1] top-level scope
 [2] eval at ./boot.jl:316 [inlined]
 [3] eval(::Expr) at ./deprecated.jl:55
 [4] top-level scope

```
on your branch @JeffBezanson.

mortenpi: Might be good to add a small comment there to say what the `eval`s do, for future reference.

PR #731, opened 2018-05-25
Change `strip` behaviour

simonbyrne:
For https://github.com/JuliaLang/julia/pull/27232


mortenpi:
Since the old signature is only deprecated, JuliaLang/julia#27232 is not held back by this, right? It should land on Julia `master` and in Compat before this is merged.

mortenpi:
What's the status here? I see that the PR in Base has been merged. Does this need to be updated or is it good to go?

simonbyrne:
No, the existing behaviour was kept.

PR #733, opened 2018-05-30
Update index.md

cormullion:
broken link

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -7,7 +7,7 @@ A package for building documentation from docstrings and markdown files.
 !!! note
 
     Please read through the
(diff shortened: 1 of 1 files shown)


mortenpi:
The `/` links are the correct ones. `.html` ones were removed in January. (JuliaLang/julia#25623, 
https://github.com/JuliaLang/julia/tree/gh-pages/en/latest/manual)

But GitHub Pages has some sort of buggy caching, which means sometimes people still see the `.html` files. This has come up before. Would be great if you could help with debugging the requests. Could you run `curl` against the different links and report what it gives?

cormullion:
Yes, it works now. It's annoying when it doesn't though... :)

PR #738, opened 2018-06-04
Capture PNG output in addition to SVG

### LINKED ISSUE #728: Capture other representations than text/plain and image/svg+xml
Basically we need to extend
https://github.com/JuliaDocs/Documenter.jl/blob/6abfaeee9f488f0487a022551634807daf4c2515/src/Expanders.jl#L500-L503
to other representations and decide on a default hierarchy. It would be especially nice to capture `png`s.

davidanthoff:
I've put it in with a lower priority for now, that seems the save way to introduce this feature.

Fixes #728.

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -497,10 +497,14 @@ function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
     content = []
     input   = droplines(x.code)
 
-    # Special-case support for displaying SVG graphics. TODO: make this more general.
-    output = showable(MIME"image/svg+xml"(), result) ?
-        Documents.RawHTML(Base.invokelatest(stringmime, MIME"image/svg+xml"(), result)) :
+    # Special-case support for displaying SVG and PNG graphics. TODO: make this more general.
+    output = if showable(MIME"image/svg+xml"(), result)
+        Documents.RawHTML(Base.invokelatest(stringmime, MIME"image/svg+xml"(), result))
+    elseif showable(MIME"image/png"(), result)
+        Documents.RawHTML(string("<img src=\"data:image/png;base64,", Base.invokelatest(stringmime, MIME"image/png"(), result), "\" />"))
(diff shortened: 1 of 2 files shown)


davidanthoff:
I added things to ``/test/examples/src/man/tutorial.md`` as you suggested.

I don't have the time to add other MIME types, all I need for my use-case is the PNG case (for now), so would be great if we could merge that for now.

mortenpi:
Hmm.. the 0.7 failures look related? I'll look into fixing the master 0.7 builds.

mortenpi:
Sorry, I haven't had the time to check why this is failing on 0.7, and it might be a few weeks I get to this. The tests on master are fixed now though, so it has to be something in this PR. But once this is working on 0.7, it would be good to go as far as I am concerned, and we could also tag a release with this in.

fredrikekre:
I think test error on 0.7 is https://github.com/JuliaLang/julia/commit/15a345bf74a27429bc4bb0c33b99a9151d65a735#r29588023

davidanthoff:
Does that mean the test failures on 0.7 are due to a bug in julia, not this PR? Anything I need to do at this point to get this PR merged?

fredrikekre:
Yea I think it is only the Julua bug, I will submit a PR to fix it soon. We can wait until that has propagated to a nightly build and then this should be good to go.

fredrikekre:
https://github.com/JuliaLang/julia/pull/27927

fredrikekre:
Seems to pass on nightly now, after the Julia fix. OK to disable testing on Julia `beta` for now? `nightly` is closer to what `0.7` will be than `beta` anyways.

mortenpi:
Yup, the 0.7 targets should go into allowed failures for now.

mortenpi:
Since the nightly win64 binaries are lagging behind, let's put that into allowed failures as well to get a green badge and have this merged.

Issue #543, opened 2017-07-31
Floating point numbers in doc strings cause method error


samoconnor:
Somehow the dollar amounts in the docstring below get parsed as floats and lead to a method error for `mdflatten` (and `mdconvert`).

```julia
"""
## `BonusAmount::String` -- *Required*
The Bonus amount is a US Dollar amount specified using a string (for example, "5"
represents \$5.00 USD and "101.42" represents \$101.42 USD). Do not include
currency symbols or currency codes.

"""
```

This patch fixes the problem, but perhaps the root-cause is in the parser?

```diff
diff --git a/src/Utilities/MDFlatten.jl b/src/Utilities/MDFlatten.jl
index a336ca2..0c1fa6e 100644
--- a/src/Utilities/MDFlatten.jl
+++ b/src/Utilities/MDFlatten.jl
@@ -64,6 +64,7 @@ function mdflatten(io, t::Table, parent)
 end

 # Inline nodes
+mdflatten(io, n::Number, parent) = print(io, n)
 mdflatten(io, text::AbstractString, parent) = print(io, text)
 mdflatten(io, link::Link, parent) = mdflatten(io, link.text, link)
 mdflatten(io, b::Bold, parent) = mdflatten(io, b.text, b)
diff --git a/src/Writers/HTMLWriter.jl b/src/Writers/HTMLWriter.jl
index 9a2881e..d92eec0 100644
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -803,6 +803,8 @@ The `parent` argument is passed to allow for context-dependant conversions.
 """
 mdconvert(md; kwargs...) = mdconvert(md, md; kwargs...)

+mdconvert(n::Number, parent; kwargs...) = DOM.Node(string(n))
+
 mdconvert(text::AbstractString, parent; kwargs...) = DOM.Node(text)

 mdconvert(vec::Vector, parent; kwargs...) = [mdconvert(x, parent; kwargs...) for x in vec]
```

mortenpi:
This indeed comes from the parser. The dollar sign gets interpreted as a form of variable interpolation (not the usual one, but one than happens when the Markdown in docstrings gets parsed). There's a [very brief note in the manual](https://docs.julialang.org/en/latest/manual/documentation/#Syntax-Guide-1) about this. The `doc""` string macro should help get around this.

[...]

PR #739, opened 2018-06-07
Add a note on double-escaping the dollar sign

odow:
I just ran into https://github.com/JuliaDocs/Documenter.jl/issues/543 and had no idea how to track down the missing `mdflatten` error. 

I ended up adding each line at a time until I found that it was the dollar signs. 

Reading the discussion, I'm not sure what the right answer is, but at the very least this would have been helpful to read in the docs.

### DIFF
--- a/docs/src/man/latex.md
+++ b/docs/src/man/latex.md
@@ -36,6 +36,14 @@ This is the binomial coefficient.
 func(x) = # ...
 ```
 
+A related issue is how to add dollar signs to a docstring. They need to be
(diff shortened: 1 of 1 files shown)


mortenpi:
LGTM, thanks!

The same comment should probably be added to [the corresponding Julia manual section](https://docs.julialang.org/en/latest/manual/documentation/#Markdown-syntax-1) as well, since I think it's relevant for docstrings even if you don't use Documenter.

PR #763, opened 2018-07-20
fix pdf generation in prescence of hidden pages

KristofferC:
`hide(page)` inserts a four-component tuple into `doc.user.pages`. This case is unhandled in the `files!` function of the `LaTeXWriter`. This means that generating pdf for the manual in Julia Base is currently broken since it uses hidden pages.

This should hopefully fix that. With this patch, the pdf is generated. FWIW a "naked" four-component tuple like this feels a bit brittle.

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -506,6 +506,14 @@ function files!(out::Vector, v::Vector, depth)
     return out
 end
 
+# Tuples comde from `hide(page)` with either
+# (visible, nothing,    page,         children) or
+# (visible, page.first, pages.second, children)
+function files!(out::Vector, v::Tuple, depth)
+    files!(out, v[2] == nothing ? v[3] : v[2] => v[3], depth)
+    files!(out, v[4], depth)
+end
+
 files!(out, s::AbstractString, depth) = push!(out, ("", s, depth))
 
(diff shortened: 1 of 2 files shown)


src/Writers/LaTeXWriter.jl
```diff
@@ -506,6 +506,14 @@ function files!(out::Vector, v::Vector, depth)
     return out
 end
 
+# Tuples comde from `hide(page)` with either
```
tpapp: typo comde => come

src/Writers/LaTeXWriter.jl
```diff
@@ -506,6 +506,14 @@ function files!(out::Vector, v::Vector, depth)
...
+# (visible, nothing,    page,         children) or
+# (visible, page.first, pages.second, children)
+function files!(out::Vector, v::Tuple, depth)
+    files!(out, v[2] == nothing ? v[3] : v[2] => v[3], depth)
```
tpapp: just a suggestion: for v0.7-only code,
```julia
function files!(out::Vector, 
                (_, firstpage, secondpage, children)::NamedTuple{(:visible, :firstpage, :secondpage, :children)},
                depth)
    files!(out, firstpage == nothing ? secondpage : firstpage => secondpage, depth)
    files!(out, children, depth)
end
```
could be more informative without requiring an explicit type (with the same change in the caller).

PR #764, opened 2018-07-23
Support inline html output

JackDunnNZ:
As suggested in https://github.com/JuliaDocs/Documenter.jl/issues/728#issuecomment-400273235

I'm unsure about the best order of the various outputs, but if `text/html` is being used for interactive graphics it should probably come ahead of static images in case both are supported.


mortenpi:
Thanks! I think this is a good thing to have, although I am wondering if this should be an opt-in? Is there a case where you might want to show `text/plain` instead of `text/html`, even if it has a corresponding `show` method defined?

* As for the ordering -- `text/html` should be the very first one, I guess? E.g. for interactive plots. Also, I believe Jupyter/IJulia has this hierarchy somewhere defined as well, so it might be a good idea to be consistent with them. 
* There should be a test for this in `test/examples/`. You can look at #738 / #759 to see how it was done for the other formats. Some simple SVG+HTML+JS combo might be best as a proof of concept.

JackDunnNZ:
Yeah, I am not sure about the overall organization and fallbacks etc, your opinion would be useful. I think IJulia asks you to use `display` to show a different mime, so maybe that is a good approach to take here, i.e. finishing the block with

```
display("text/plain", x)  # hide
```

I have updated to match the ordering in IJulia as listed [here](https://github.com/JuliaLang/IJulia.jl/blob/bb438dc04003039a3168fff9ca1a4625f3b08e2c/src/inline.jl#L7)

Also added a simple example SVG with some JS interactivity

fredrikekre:
I think this is fine, most objects does not define a show for html anyway, and those who do probably want them to be used.

mortenpi:
Agreed. If there are some corner cases, we can deal with them then.

Issue #761, opened 2018-07-10
git not found on windows even though its there


Datseris:
I am trying to create deploy keys for documenter. I am on windows and not on linux.

I have to say it is tremendously difficult. I have spend _several hours_ now, and still had very little progress. The documentation is really not helpful for windows users, there is no information whatsoever. 

So far I have tried to create keys using `puttygen` but travis hangs, requesting passphrase. Seems like the only way to create working keys is using documenter itself.

But that is impossible on windows:

```
julia> using Documenter

julia> Travis.genkeys("JuliaMusic_documentation")

 in module Travis
in #2 at C:\Users\datseris\.julia\packages\Documenter\eULE\src\Travis.jl
ERROR: 'git' not found.
Stacktrace:
 [1] error at .\error.jl:33 [inlined]
 [2] #genkeys#1(::String, ::Function, ::String) at C:\Users\datseris\.julia\packages\Documenter\eULE\src\Travis.jl:49
 [3] genkeys(::String) at C:\Users\datseris\.julia\packages\Documenter\eULE\src\Travis.jl:48
 [4] top-level scope at none:0

shell> git
usage: git [--version] [--help] [-C <path>] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]

```
latest julia, latest documenter versions.

As a side note, I really have to express that this also makes me feel very unwelcomed: why would there be not a single sentence of support for someone that doesn't use linux? (in the documentation)

fredrikekre:
What is
```
shell> which git
```

mortenpi:
As a workaround, as long as you have `ssh-keygen` available, you should be able to run these commands on Windows to get the keys:

```julia
shell> ssh-keygen -N "" -f privatekey
julia> read("privatekey", String) |> base64encode |>  println # this should become a secure environment variable on Travis called DOCUMENTER_KEY
julia> read("privatekey.pub", String) |> println # this should be added as a Deploy Key on GitHub (https://github.com/$user/$repo/settings/keys)
```

mortenpi:
Closing this in favor of JuliaDocs/DocumenterTools.jl#5

Issue #567, opened 2017-09-27
Windows genkey error


Crghilardi:
Hello, 

I wanted to open up an issue following [this discussion about people in a windows env](https://discourse.julialang.org/t/documenter-jl-travis-genkeys-which-which-error-on-windows/5881/6).

I have made some progress but I stil have not gotten everything working correctly. 
I am not sure how much of this is able to be corrected in the codebase vs just adding to documentation outlining the process for windows more specifically. 
[I found this windows specific issue already open](https://github.com/JuliaDocs/Documenter.jl/issues/444) but I was not sure exactly how what I am running into fits with this. 

See the above linked Discourse post for steps taken so far:

So far I have 

1. Travis.genkeys() is not working 

Workaround: Generate keys manually from Gitshell using ssh-keygen and manually copy and paste strings to Travis and Github. 

Secondary problem: Travis requires escaping characters so I had to manually add backslashes to the string? Not sure if this is necessary? 
I also see .documenter.pub mentioned in various threads, do I need to create a file named that and move the physical .pub file anywhere?

I am now stuck on travis hanging on a passphrase on linux [as mentioned in this issue](https://github.com/JuliaDocs/Documenter.jl/issues/429) and completing the build successfully on Mac but not deploying the documentation. I have added the DOCUMENTER DEBUG keyword as well.

[Here](https://travis-ci.org/Crghilardi/ForestBiometrics.jl) is the link to my most recent attempt.

In conclusion:
What can I do to get around the passphrase error and get the documents building?

Is there any way to improve the windows genkey/deploy experience?
If not, I can put a more detailed write up together outlining what I had to do once I actually figure out how to get it working.

Thank you for your help

mortenpi:
I had forgotten that we [`base64encode` the whole private key](https://github.com/JuliaDocs/Documenter.jl/blob/0fa59b1806e88776c1cb2c3d818fdb163443f1e7/src/Documenter.jl#L657) (and decode it when reading the variable on Travis). So this would have to be done by hand as well. No need to escape anything then. And I think this is related to the passphrase as well -- decoding a non-encoded key results in gibberish and `ssh` assumes that it's encrypted.

[...]

PR #767, opened 2018-07-26
Travis.genkeys() now supports Windows

zundertj:
Bumped into issues similar to [#761](https://github.com/JuliaDocs/Documenter.jl/issues/761), and decided to fix it. Have to admit the fix is not very nice, as it basically becomes an if-else on `is_windows()` and needs a dependency. However, it really helps Windows users a lot.

I am aware that a solution is to add the git .exe to the `$PATH` variable, but the standard Git installation on Windows does not (see also [#567](https://github.com/JuliaDocs/Documenter.jl/issues/567)).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.6
 Compat 0.62.0
 DocStringExtensions 0.2
+WinReg 0.3.1
--- a/src/Travis.jl
+++ b/src/Travis.jl
@@ -7,6 +7,7 @@ module Travis
 
 using Compat, DocStringExtensions
 import Compat.Pkg
+import WinReg
 
 export genkeys
 
@@ -44,10 +45,12 @@ julia> Travis.genkeys("/path/to/target/directory")
 ```
 """
 function genkeys(package; remote="origin")
-    # Error checking. Do the required programs exist?
-    success(`which which`)      || error("'which' not found.")
-    success(`which git`)        || error("'git' not found.")
-    success(`which ssh-keygen`) || error("'ssh-keygen' not found.")
+    if !is_windows()
+        # Error checking. Do the required programs exist?
+        success(`which which`)      || error("'which' not found.")
+        success(`which git`)        || error("'git' not found.")
+        success(`which ssh-keygen`) || error("'ssh-keygen' not found.")
+    end
 
     directory = "docs"
     filename  = ".documenter"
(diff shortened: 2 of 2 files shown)


Datseris:
awesome thanks!

mortenpi:
This is great, thanks a lot for tackling this! @Datseris did you test this branch by any chance?

I am a little bit sceptical of introducing the WinReg dependency to Documenter though. But I am thinking that it might be best to move this type of code out of Documenter into a separate package anyway (DocumenterTools?). The `Travis` stuff and the `generate` function could be moved in my opinion. And there we could have as many dependencies we want for this type of stuff. Any thoughts?

zundertj:
I agree that the Travis functionality does not really belong in this package. In the end, this is a Julia wrapper around ssh keygen, maybe that should live in its own package?

Datseris:
Hi all,

I kind of disagree with both of you and I think this functionality should stay in documenter.jl. I can see why this becomes difficult with the new dependency, but here are my comments:

1. " this is a Julia wrapper around ssh keygen " : This is "incorrect" in the following way: It was impossible for me to generate the highly special pair of keys documenter wants. I have tried 5 different ways of generating a key-pair (literally 5, e.g. PuTTy, windows pre-installed way, etc.) and not a single one of them worked. It is obvious for me that documenter does a special extra step that is not documented at the moment. 
2. In my eyes, the documenter package has 2 uses: 1 is to create markdown files, the other is to deploy the documentation. It is no chance that there are (mainly) 2 functions used `makedocs` and `deploydocs`.
3. I cannot see how `deploydocs` can be used without the Keys, and making `deploydocs` "unusable" (at least for its first run) means that you cut 50% of documenter's functionality. If you want to move the key generation function somewhere else, then why would `deploydocs` stay here? 

This raises the question: Should documenter have the new package as dependency? My answer is yes. Why: It is not possible for Documenter to deploy without configuring the keys once. 

Notice that here: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#SSH-Deploy-Keys-1 there is no mention of other possible ways besides the Documenter one.

I can see why you would want to have this "generate keys" functionality elsewhere. But before this choice is made you have to consider the user as well. Does it benefit them in anyway? And if this move is indeed made, please be sure that the users can still do what they could do before, i.e. document the changes.

PS> @mortenpi Unfortunately I haven't had time to test this yet. I will do it either during or after JuliaCon, for now I really need to prepare for there :)

mortenpi:
You make reasonable points @Datseris, but I still think that we need to go with a separate package. The way I see it:

* The main purpose of the Documenter package is to be installed on Travis and other CI systems where it builds and deploys the docs. The key here is that it should be possible to do so without including unnecessary dependencies.

  E.g. a practical problem with including additional deps is that it would introduce another package dependency to the core language, which has to updated when merging PRs into `JuliaLang/julia`. While after 1.0 there should not be that many breaking changes, I think it is still better to avoid dependencies as much as possible. 

* DocumenterTools would be a supplementary core package, kept up to date with Documenter. But normally it would be installed on the users dev computers and not on CI systems. Ideally we would not have to split it up like this, but this is the only way to satisfy the requirement of avoiding dependencies.

* I do agree that we must not hide away how one can generate the keys, which are indeed integral to `deploydocs`. Hence, the necessary docs would still remain in Documenter's repo and it will just refer to the DocumenterTools package. I guess DocumenterTools will also deploy some docs, but it will probably just be the API docs / docstrings.

  In the end the only difference for the user are the two additional commands:

  ```
  pkg> add DocumenterTools 
  julia> using DocumenterTools
  ```

_Side note:_ the encoding scheme we use for `DOCUMENTER_KEY` should also be documented.

---

@zundertj I think we will not merge this PR. I will put up a new repo and move the current Travis code there. If you would be willing to then make a PR against that repo, that would be much appreciated :slightly_smiling_face:

fredrikekre:
I agree with Morten; setting up the build is a one-time thing and those tools can live outside. We should also move out Documenter.generate there.

Datseris:
Sounds good! As long as the docs are clear 1 more statement is not a big
deal

zundertj:
Sounds good, just ping me once you have the new repo ready, and I will submit another PR.

PR #769, opened 2018-07-27
Allow HTTP/2 in link-checks

timholy:
This supports `HTTP/2` headers when checking the validity of links.


mortenpi:
LGTM, thanks!

PR #774, opened 2018-07-31
Break out doctesting / refactor

### LINKED ISSUE #198: Provide hook to run doctests as part of normal test suite
It would be great if there was a function that I could call in my normal `runtests.jl` script that runs all the doctests, for example like this:

``` julia
Documenter.run_doctests(packagename)
```

This would have two benefits, as far as I can see:
1. Errors in the doctests would actually show up as a broken build (right now an error in the doctest is actually not breaking a travis build)
2. doctests could be run in all PRs, branches etc. My sense is that one would want to run doctests essentially whenever one wants to run normal tests, but of course doesn't want the docs to be build/deployed in all of these situations.

### LINKED ISSUE #535: Separate function just to run doctests
Would be useful for unit testing purposes

### LINKED ISSUE #756: Doc-testing without makedocs
I asked on [StackOverflow](https://stackoverflow.com/questions/50763456/julia-can-i-run-doctests-without-makedocs?noredirect=1#comment88541681_50763456) about doc-testing without makedocs. @fredrikekre there mentioned that this was a possible feature in future versions. I could not find an issue for it and am adding one.

mortenpi:
This refactors a few parts of the build pipeline to make it possible to run doctests without having to compile the whole document (close #198, close #535, close #756).

* Moves doctesting to an earlier stage in the build.
* Implements `doctest = :only` to `makedocs`, which only runs doctests and skips the later build steps.
* When fixing doctests you probably do not care about the output either, so `doctest = :fix` also skips compilation and rendering of the docs now.
* Creates an API for storing module-level documentation metadata (analogue to an at-meta block), via the exported `DocMeta` module (e.g. `DocMeta.setdocmeta!(module, :DocTestSetup, :(using Foo)`). Only supports `:DocTestSetup` at the moment.
* Adds an exported `doctest(::Vector{Module})` function to easily run doctests in a module.

The changes are a bit more extensive than they necessarily should be, but I am also trying to clean up and refactor the internals in the process. [There is a bit more info about where I want to go with these changes in the wiki](https://github.com/JuliaDocs/Documenter.jl/wiki/Documenter.jl-roadmap).

~This is still very much a WIP and I think I might have to rethink some decisions here. Also needs testing, tests etc. Any drive-by comments will be much appreciated though.~ This is now ready for review.

- [x] Update the manual page on doctesting


mortenpi:
So.. this should work now.. theoretically.

One hurdle I came across as I was trying to test it with the Julia manual, is that it actually breaks nearly all the doctests _in docstrings_. This is because they usually rely on `DocTestSetup`s in the Markdown files that no longer have any effect on the doctesting of the docstrings.

Now, this is not necessarily a bad thing. The whole idea of this is to be able to run doctest without having to go through all the hoops of figuring out which docstring ends where etc. It is also weird that the validity of a doctest depends on some metadata far away in a manual `.md` file. But having to put at-meta blocks into each docstring (which is now technically supported) is not optimal either.

So I've prototyped a way to add doctest metadata to modules instead. Essentially, every module gets a special `gensym`ed variable that contains the metadata (it's how docstrings themselves are handled). It does require Documenter -- there is the `Documenter.DocTests.doctestsetup!` function which declares an `Expr` object as a `DocTestSetup` of a module.

The question is: when should you actually add the metadata? Ideally, it would be close to the docstrings.

* You could make your package depend on Documenter and just call `Documenter.DocTests.doctestsetup!(@__MODULE__, :( ... ))` in the module. But we don't want packages to depend on Documenter.

* You can add the metadata in the `make.jl` file by just making all the `doctestsetup!` calls after you import your package. In a sense, this is similar to how it's done now, where the metadata lives in the `.md` files under `docs/`. This might be a good solution for Base and the standard libraries.

* But I am also thinking about making it possible to define the metadata using a global constant in your module, e.g. `const _DOCUMENTER_DOCTESTSETUP = :( ... )`. Documenter can then just walk through all the modules passed with `modules` keyword (these are the ones that get tested anyway) and check for this variable. The downside is that it introduces a visible variable into the module namespace.

mortenpi:
Ok, this is close to being finished now I think. I've updated the OP with a summary of what this adds and changes (Documenter will also export a `doctest` function and `DocMeta` module with this).

Bikeshedding invitation: I can't decide if it should be `DocMeta` / `DocMeta.getdocmeta` / `DocMeta.setdocmeta!` or `DocsMeta` / `DocsMeta.getdocsmeta` / `DocsMeta.setdocsmeta!` though (or maybe something else altogether).

Project.toml
```diff
@@ -14,14 +14,17 @@ REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
+Compat = "2.1"
```
fredrikekre: I don't think you can put compat on extras right now.

mortenpi: I just did :laughing: But yeah, best to remove if it does not have any effect on the resolver.

It's again just for `isnothing` though, so might be worth dropping it altogether here as well?

fredrikekre: Yea, I like `=== nothing` better anyway.

PR #777, opened 2018-08-01
A few deprecation fixes

mortenpi:


timholy:
Thanks! Can this be tagged? Packages that build docs using "nightly" will fail now that nightly means 1.0 and the deprecation is now an error.

fredrikekre:
Tagged https://github.com/JuliaLang/METADATA.jl/pull/16396

PR #784, opened 2018-08-09
Remove Compat dependency and references to Pkg.dir

ararslan:
Also require 0.7

The diff is big

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,12 @@ os:
   - osx
 
 julia:
-  - 0.6
   - 0.7
+  - 1.0
   - nightly
 
 notifications:
   email: false
 
 after_success:
-  - julia -e 'cd(Pkg.dir("Documenter", "test")); include("coverage.jl")'
+  - julia -e 'import Documenter; cd(joinpath(pathof(Documenter), "..", "..", "tests")); include("coverage.jl")'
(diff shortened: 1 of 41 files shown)


ararslan:
FYI this is needed fairly urgently to get 1.0 docs for Julia itself

mortenpi:
Ok, this is a bit of a surprise. I somehow assumed that a 0.6-Compatted package would _just work_ on 1.0. I'll merge and tag #782, and then we can deal with this next.

fredrikekre:
@mortenpi it's chill, I built docs locally and pushed to gh pages for now so there is no rush.

mortenpi:
This is not actually necessary for 0.7/1.0 deployments, right? Documenter should actually be working fine on 1.0 at the moment.

`Pkg.dir` does need to be fixed, but that only affects `Travis.genkeys` and `generate`, neither of which are important for doc builds.

mortenpi:
The deprecation that failed the deployment was [fixed in v0.19.3](https://github.com/JuliaDocs/Documenter.jl/pull/777/files#diff-2838712d7a9d8b22fbc48aecb3bf150dL533), but unfortunately Documenter hadn't been updated in Base. Ref: JuliaLang/julia#28559

fredrikekre:
#795

PR #789, opened 2018-08-10
Remove Travis.genkeys and Documenter.generate

fredrikekre:
Will push them to: https://github.com/JuliaDocs/DocumenterTools.jl


docs/src/man/hosting.md
```diff
@@ -33,63 +33,12 @@ The following sections outline how to enable this for your own package.
...
-
-        To reiterate: make sure that the "Display value in build log" option is **OFF** for
-        the variable, so that it does not get printed when the tests run. This
-        base64-encoded string contains the *unencrypted* private key that gives full write
-        access to your repository, so it must be kept safe.  Also, make sure that you never
-        expose this variable in your tests, nor merge any code that does. You can read more
-        about Travis environment variables in [Travis User Documentation](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings).
+!!! warning
+    Read the `DocumenterTools` documentation carefully. SSH keys give push access
+    to the repository, and should be handled with care.
```
mortenpi: I would rather keep the tutorial here and have DocumenterTools refer to this repo. This way the user is not forced to jump between the docs of different packages.

fredrikekre: Sure, seems like a good idea.

fredrikekre:
This now includes the docs from DocumenterTools here.

fredrikekre:
I'll merge this now to avoid further conflicts and we can improve it if needed during the 0.20-dev cycle.

mortenpi:
Documenter's own docs were built as part of the tests so that the CI would fail in case Documenter's own manual doesn't build (or deploy). But we might be able to put it into a build stage..

Once DocumenterTools gets tagged, `docs/Manifest.toml` goes away, right?

fredrikekre:
> Documenter's own docs were built as part of the tests so that the CI would fail in case Documenter's own manual doesn't build (or deploy).

Yea I know, but we can't load `DocumenterTools` from testing, unless adding it as a test dependency, which we can't do since it is not registered yet. (Well, we could `add` it in the test script but ...).

> Once DocumenterTools gets tagged, docs/Manifest.toml goes away, right?

Yea, just need it to find the url for now.

We should definitely fix this before the next Documenter release, just wanted to get this in for now.

PR #792, opened 2018-08-13
Make the navigation bar automatically scroll to the selected item

dalum:
I am not sure if I made the change in the right place, but it works locally.  If the navigation bar is really long (as in base), I find it slightly inconvenient that the navigation view jumps to the top when clicking a menu item.  This PR shifts the navigation view to the current item at the end of the `document.ready` event call.


mortenpi:
Thanks! This would be great to have!

My only concern is that it always hides the logo and the search box (you can scroll up, but you need to know that there's something there in the first place). Perhaps only the actual menu should scroll? A more involved version could then hide the the logo (but perhaps not the search) when the user starts scrolling down in the menu (like we do on mobiles with headroom).

dalum:
It took a bit of CSS fiddling to get it to work, but now it always shows the logo and search.  I could try to make scrolling the menu hide them, but I feel that it's more of a gimmick than a feature, and might be annoying to some.

mortenpi:
What do you think about subtracting `1.5*$(".current > a").outerHeight()` from the `scrollTop`, in order to have a bit of the menu above the current item visible, giving the user another visual cue that there is something up there?

Actually, alternatively, we wouldn't have to keep the logo and stuff fixed at all if we'd try to make sure that the gray box is symmetrically in the middle. "Home" is unlikely to be very tall, so it would keep the logo visible.  Any thoughts?

dalum:
@mortenpi Sorry, I haven't looked at this in a while, but I tried subtracting the `1.5*...` offset, and I have to admit I don't like it that much.  Same with the symmetrical placement.  To me it looks like the offset computation errored somehow. :smirk: I was also just reading the Rust book (https://doc.rust-lang.org/stable/book/2018-edition/) and noticed that their table of contents scrolls so the currently selected item is at the top, and in my opinion, it looks better and gives a feeling of "momentum" or progress, if you are reading the manual linearly.  Which of course isn't always the best way to read a manual, so that is worth considering.  To solve the search bar issue, the Rust book has it nicely tucked away as a drop–down in the main document.

I think the least invasive compromise is probably to make just the search bar hover when scrolling down.

dalum:
Good to go! :blush:

mortenpi:
It seems that this broke the version selector (ref #855), which I missed since the local build didn't have a selector. Somehow the `<select>` gets resized, even though it has an explicit `height`.

dalum:
Oh, my local build also didn't have the selector.  I can try and look into a fix.

PR #795, opened 2018-08-14
Bump Julia require to 0.7, stop testing on 0.6.

fredrikekre:
I think it might be easier to first get this in, then clean up after, since there will be quite a lot of changes for the next Documenter release. Thoughts?

I also removed PackageEval results from README. We can add it back when that actually works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,16 +3,18 @@
 
 *A documentation generator for Julia.*
 
-| **Documentation**                                                               | **PackageEvaluator**                                                                            | **Build Status**                                                                                |
-|:-------------------------------------------------------------------------------:|:-----------------------------------------------------------------------------------------------:|:-----------------------------------------------------------------------------------------------:|
-| [![][docs-stable-img]][docs-stable-url] [![][docs-latest-img]][docs-latest-url] | [![][pkg-0.6-img]][pkg-0.6-url] [![][pkg-0.7-img]][pkg-0.7-url] | [![][travis-img]][travis-url] [![][appveyor-img]][appveyor-url] [![][codecov-img]][codecov-url] |
+| **Documentation**                                                               | **Build Status**                                                                                |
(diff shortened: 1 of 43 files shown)


README.md
```diff
@@ -23,9 +25,9 @@ julia> Pkg.add("Documenter")
 
 ## Project Status
 
-The package is tested against Julia `0.6` and *current* `0.7-dev` on Linux, OS X, and Windows.
+The package is tested against Julia `0.7`, `1.0` and *current* `1.1-dev` on Linux, OS X, and Windows.
 
-Support for Julia `0.4` and `0.5` has been dropped in the latest version, but older versions of Documenter may still work with those Julia versions (the `0.8` and `0.11` branches for either Julia version, respectively).
+Support for Julia `0.4`, `0.5` and `0.6` has been dropped in the latest version, but older versions of Documenter may still work with those Julia versions (the `0.8` and `0.11` branches for either Julia version, respectively).
```
mortenpi: Should mentioned 0.19 for 0.6.

mortenpi:
Yeah, I think we should drop 0.6 before we do any other substantial changes. But also remove compat stuff, right? Should we apply FemtoCleaner here?

src/DocChecks.jl
```diff
@@ -76,7 +75,7 @@ meta(m) = Docs.meta(m)
 nameof(x::Function)          = typeof(x).name.mt.name
 nameof(b::Base.Docs.Binding) = b.var
 nameof(x::DataType)          = x.name.name
-nameof(m::Module)            = Compat.nameof(m)
+nameof(m::Module)            = nameof(m)
```
ararslan: Won't this stack overflow? As of 0.7 this local `nameof` definition should be able to be removed entirely, since it's in Base. (Though we might still need to add a definition like `Base.nameof(b::Base.Docs.Binding) = b.var`.)

PR #801, opened 2018-08-16
WIP remove stable

fredrikekre:
Trying to do this in small steps. This PR removes `stable`.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -45,7 +45,7 @@ makedocs(
     ],
     # Use clean URLs, unless built as a "local" build
     html_prettyurls = !("local" in ARGS),
-    html_canonical = "https://juliadocs.github.io/Documenter.jl/stable/",
+    html_canonical = "https://juliadocs.github.io/Documenter.jl/v0/",
 )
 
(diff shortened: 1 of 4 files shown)


fredrikekre:
What should we recommend that people set as their `html_canonical`? The docs to the latest major release? In case of Documenter that would be `https://juliadocs.github.io/Documenter.jl/v0` then? Which symlinks to the latest `v0.X`?

mortenpi:
Actually.. maybe for packages it makes sense to have `stable/` just for this reason, always pointing to the latest released version?

fredrikekre:
So maybe keep deploying to `stable` as an alias for latest release then maybe? We can let people use a configurable version selector to exclude `stable` from that instead.

fredrikekre:
But it could be a symlink instead of an exact copy.

mortenpi:
We should definitely make it a symlink. Let's keep it as a default for packages then, but we need an option to disable it (e.g. for base).

fredrikekre:
(I am reworking this for another PR instead).

PR #802, opened 2018-08-16
Rename latest to dev.

fredrikekre:
I think this might be easier to do before removing `stable`.


fredrikekre:
~25 packages uses the latest kwarg.

fredrikekre:
Result from running with this PR: https://fredrikekre.github.io/Literate.jl/stable/

fredrikekre:
Maybe call it `devel` by default? `dev` looks a bit short in the version selector from above.

src/Documenter.jl
```diff
@@ -325,14 +329,23 @@ function deploydocs(;
...
+        latest::Union{String,Nothing} = nothing, # deprecated
 
         osname = "linux",
         julia  = error("no 'julia' keyword provided."),
 
         deps   = Deps.pip("pygments", "mkdocs"),
         make   = () -> run(`mkdocs build`),
+
+        devbranch = "master",
+        devfolder = "dev",
```
mortenpi: I think "directory" is the technically more correct work here, over "folder". So `devdir`? `devdirectory`?  Or even maybe `devurl` or `devpath`, since it's not even a directory here what we're setting, but the URL?

src/Documenter.jl
```diff
@@ -312,6 +310,12 @@ deps = Deps.pip("pygments", "mkdocs")
 **`make`** is the function used to convert the markdown files to HTML. By default this just
 runs `mkdocs build` which populates the `target` directory.
 
+**`devbranch`** is the branch that "tracks" the in-development version of the  generated
+documentation. By default this value is set to `"master"`.
+
+**`devfolder`** the folder that in-development version of the docs will be deployed.
+Defaults to `"dev"`.
```
mortenpi: Hmm.. this is now a breaking change, and will be somewhat chaotic for the users that their URLs suddenly change. Not sure how to handle it at the moment though.

src/Writers/HTMLWriter.jl
```diff
@@ -478,22 +478,23 @@ function render_topbar(ctx, navnode)
...
     end
-    # put stable before latest
-    sort!(named_folders, rev = true)
     # sort tags by version number
-    sort!(tag_folders, lt = (x, y) -> VersionNumber(x) < VersionNumber(y), rev = true)
+    sort!(folders, lt = (x, y) -> VersionNumber(x) < VersionNumber(y), rev = true)
+
+    # include stable first, and dev at the bottom
+    "stable" in all_folders && pushfirst!(folders, "stable")
+    "dev" in all_folders && push!(folders, "dev")
```
mortenpi: I would keep "dev" at the top, since that's usually the one you want. At least as long as we have all the patch versions in the selector.

mortenpi:
As for `dev` vs `devel`: `dev/` looks a bit better in the URL I'd say. As a hack we could print `dev/` as "devel" or "development" in the version selector.

src/Documenter.jl
fredrikekre: `devbranchurl` :stuck_out_tongue:

src/Documenter.jl
fredrikekre: But then we also have to remove the `latest` directory if it exists so idk.

mortenpi:
It's exactly the links I am worried about. This is a pretty big behaviour change I'd say and shouldn't happen silently if we can avoid it. OTOH, the whole point of 0.20.0 is to break backwards compatibility, so we should go forward with it. But we need to figure out what to do with the packages. I see two options at the moment:

* As you suggested, have `latest/` still be a symlink to `dev/`. This would only have to run once.
* Do automated PRs to the whole ecosystem to fix Documenter to v0.19.
* (Or a combination of both)

src/Documenter.jl
```diff
@@ -325,14 +329,23 @@ function deploydocs(;
...
 
         deps   = Deps.pip("pygments", "mkdocs"),
         make   = () -> run(`mkdocs build`),
+
+        devbranch = "master",
+        devurl = "dev",
     )
+    # deprecation of latest kwarg (renamed to devbranch)
+    if latest !== nothing
+        Base.depwarn("The `latest` keyword argument have been renamed to `devbranch`.")
```
mortenpi: "has been renamed"

src/Documenter.jl
```diff
@@ -325,14 +329,23 @@ function deploydocs(;
...
+        devbranch = latest
+        latest = nothing
```
mortenpi: Not really necessary?

src/Documenter.jl
```diff
@@ -325,14 +329,23 @@ function deploydocs(;
...
         deps   = Deps.pip("pygments", "mkdocs"),
         make   = () -> run(`mkdocs build`),
+
+        devbranch = "master",
+        devurl = "dev",
     )
+    # deprecation of latest kwarg (renamed to devbranch)
+    if latest !== nothing
+        Base.depwarn("The `latest` keyword argument have been renamed to `devbranch`.")
+        devbranch = latest
```
mortenpi: One silly edge case: if the user defines both `latest` and `devbranch` simultaneously. Might be good to error here just in case.

src/Documenter.jl
```diff
@@ -500,10 +514,15 @@ function git_push(
...
+                # Copy docs to `devurl`, or `stable`, `<release>`, and `<version>` directories.
                 if isempty(tag)
-                    gitrm_copy(target_dir, latest_dir)
-                    Writers.HTMLWriter.generate_siteinfo_file(latest_dir, "latest")
+                    gitrm_copy(target_dir, devurl_dir)
+                    Writers.HTMLWriter.generate_siteinfo_file(devurl_dir, devurl)
+                    # symlink "latest" to devurl to preserve links (remove in some future release)
+                    if devurl != "latest"
+                        rm(joinpath(dirname, "latest"); recursive = true, force = true)
+                        cd(dirname) do; symlink(devurl, "latest"); end
```
mortenpi: Let's print a huge warning here as well that we're creating this symlink.

src/Documenter.jl
fredrikekre: No, just wanted this not to propagate.

src/Documenter.jl
fredrikekre: I don't think we can know this, if the user have set `devbranch` to the defaul. I'll just add an `@info` here saying which branch we choose.

PR #804, opened 2018-08-17
Use a generic description for non-GitHub hosts

tribut:
The `HTMLWriter` currently uses

GitHub-Icon Edit on GitHub

for the source link, even if the repo points to a private host (such as a self-hosted GitLab instance). This PR changes the default to simply

File-Icon Edit

in case no known host is detected.


mortenpi:
This is a good idea. The current implementation will take away all GitHub links though, since the default repo template is defined in the `url` function. It should probably be moved away from there and made the default argument for `doc.user.repo`. I don't know if doing that will change any other default behaviour though.

https://github.com/JuliaDocs/Documenter.jl/blob/c58e7e0fc0d62dc4ce631487002930813e4c51b3/src/Utilities/Utilities.jl#L386-L389

One other minor thing -- having just the work "Edit" looks somehow too narrow. What do you think about having something like "Edit source"?

tribut:
"Edit source" seems fine, I've update the PR accordingly.

However, I do not understand how my changes would take away Github links. Looking through the code, the url function is called twice, namely:

https://github.com/JuliaDocs/Documenter.jl/blob/b598cb077c11c279bc093cfcda11a362a48391bc/src/Writers/HTMLWriter.jl#L445

and

https://github.com/JuliaDocs/Documenter.jl/blob/b598cb077c11c279bc093cfcda11a362a48391bc/src/Writers/HTMLWriter.jl#L725

In both cases I cannot see how the repo argument would be different after my changes?

mortenpi:
I was a bit imprecise: GitHub links get the generic icon. They don't disappear.

And the reason is that `url` gets called after

https://github.com/JuliaDocs/Documenter.jl/blob/b598cb077c11c279bc093cfcda11a362a48391bc/src/Writers/HTMLWriter.jl#L427

and so at that point `ctx.doc.user.repo` is an empty string and does not match the GitHub repo.

The URL handling is a bit hectic and should probably be refactored at some point. But for now, to work around this issue, just make the GitHub part of `Utilities.repo_host_from_url` into the following:

```julia
elseif occursin("github", repoURL) || isempty(repoURL)
    return RepoGithub
```

`Utilities.url` seems to use the same heuristic to determine that it should create GitHub URLs.

tribut:
Oh, yes. Empty repo would have had that effect, yes. I've updated the PR as you suggested.

mortenpi:
LGTM, thanks!

PR #808, opened 2018-08-18
Don't define setcolor! in __init__

### LINKED ISSUE #806: WARNING: eval from module Base to 'my_package name'
I asked about this a few weeks ago on Discourse [Usage] and thought it was resolved in Documenter 0.19.5 but it isn't. Not sure if I'm doing something wrong or if below warning is just a consequence from moving to Julia 0.7 and 1.0:
```
WARNING: eval from module Base to CmdStan:    
Expr(:->, :x, Expr(:block, #= Symbol("/Users/rob/.julia/packages/Documenter/bK0Xg/src/DocChecks.jl"):651 =#, :y = :have_color, #= Symbol("/Users/rob/.julia/packages/Documenter/bK0Xg/src/DocChecks.jl"):651 =#, Expr(:global, :have_color = :x), #= Symbol("/Users/rob/.julia/packages/Documenter/bK0Xg/src/DocChecks.jl"):651 =#, :y))
  ** incremental compilation may be broken for this module **
```

Any insights or tips appreciated.

fredrikekre:
Don't define setcolor! in __init__, does not seem to be precompilable, fix #806.


fredrikekre:
Does not seem to be precompilable in a nested environment, `CmdStan` loads `Documenter` in the module, and I suspect that this only showed up now because precompile is now default, and `CmdStan` did not have `__precompile__()` and thus did not precompile earlier.

fredrikekre:
Marking as backport, since we are probably a bit away from releasing 0.20.

PR #809, opened 2018-08-19
Backports for 0.19.6

fredrikekre:
Ref: #808


mortenpi:
Should be merged with a merge commit. At least that's how I usually do it.

PR #813, opened 2018-08-20
Add selector argument to deploydocs

fredrikekre:
Add `selector` argument to `makedocs`. The `selector` argument decide the content, and order, of the version selector. The argument is a vector, defaulting to `["stable", "v#.#", devurl]` where the entries
expand according to:
 - `"stable"` will point to the latest release.
 - `"v#.#"` expands to one doc per minor release.
 - `devurl` maps to the development docs (#802).

The following arguments are also valid to `selector`:
 - `"v#"` expands to one doc per major release.
 - `"v#.#.#"` expands one doc per patch release.
 - A string which directly maps to a folder.
 - A pair, which will put `p.first` in the selector, and map it to `p.second`.

This lets users opt out of `stable` (#801) if they choose to.


src/Documenter.jl
```diff
@@ -529,20 +539,29 @@ function git_push(
...
-                    # only push to stable if this is the latest stable release
+                    # only symlink stable to latest release if this is the latest release
+                    # which is not a prerelease
                     versions = filter!(x -> occursin(Base.VERSION_REGEX, x), readdir(dirname))
                     maxver = mapreduce(x -> VersionNumber(x), max, versions; init=v"0.0.0")
-                    if version >= maxver && version.prerelease == () # don't deploy to stable for prereleases
-                        gitrm_copy(target_dir, stable_dir)
-                        Writers.HTMLWriter.generate_siteinfo_file(stable_dir, "stable")
+                    if version >= maxver && version.prerelease == () && in("stable", selector)
+                        cd(dirname) do; symlink(tag, "stable"); end
```
fredrikekre: Note to self: This should be removed.

fredrikekre:
Succesful deploy here: https://fredrikekre.github.io/Literate.jl/v1/ with this: https://github.com/fredrikekre/Literate.jl/blob/6b9dd249cfbc4d19f96e961c525d83568adb50dd/docs/make.jl#L62

src/Documenter.jl
```diff
@@ -527,22 +537,22 @@ function git_push(
...
+                # Expand the users `selector` vector
+                entries, symlinks = Writers.HTMLWriter.expand_selector(dirname, selector)
+
+                # Create the versions.js file containing a list of `entries`.
+                # This must always happen after the folder copying.
+                Writers.HTMLWriter.generate_version_file(dirname, entries)
+
+                # generate the symlinks
+                cd(dirname) do
+                    foreach(kv -> symlink(kv.second, kv.first), symlinks)
```
fredrikekre: Need to remove existing links here. And I guess error if what exist there is not an old symlink?

fredrikekre:
Ok, things seem to work now, so this should be ready for review.

test/htmlwriter.jl
```diff
@@ -21,28 +33,52 @@ import Documenter.Writers.HTMLWriter: jsescape, generate_version_file
     @test jsescape("policy to  delete.") == "policy to\\u2028 delete."
 
     mktempdir() do tmpdir
-        versions = ["stable", "dev", "v0.2.6", "v0.1.1", "v0.1.0"]
+        versions = ["stable", "dev",
+                    "2.1.1", "v2.1.0", "v2.0.1", "v2.0.0",
+                    "1.1.1", "v1.1.0", "v1.0.1", "v1.0.0",
+                    "0.1.1", "v0.1.0"] # note no `v` on first ones
```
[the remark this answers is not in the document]
fredrikekre: `0.1.1` is a valid tag so there might be folders in there with that naming convention. Just include that  here to make sure we dont symlink to something `v0.1.1` in this case, but to `0.1.1`

src/Documenter.jl
```diff
@@ -316,6 +317,16 @@ documentation. By default this value is set to `"master"`.
 **`devurl`** the folder that in-development version of the docs will be deployed.
 Defaults to `"dev"`.
 
+**`selector`** determines content and order of the resulting version selector in
+the generated html. The following entries are valied in the `selector` vector:
+ - `"stable"`: points to the latest released documentation.
```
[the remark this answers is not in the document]
fredrikekre: Hmm, maybe introduce a token like `"v^.^.^"` to indicate maximum version, and the default here could be `"stable" => "v^.^.^"` ?

fredrikekre:
Thanks for the review. I pushed a much simplified (?) version. Now only the following things are allowed:
 - `"v#"` one per major version
 - `"v#.#"` one per minor version
 - `"v#.#.#"` one per patch
 - `"v^"` the latest version
 - `"first" => "second"` mapping first to second, with special-casing  when second is `"v^"`, to be able to do e.g. `"stable" => "v^"`

Thus, removed the special casing of `"stable"`, and also removed the string only (replacing `"dev"` with `"dev" => "dev"`)

src/Documenter.jl
```diff
@@ -316,6 +317,18 @@ documentation. By default this value is set to `"master"`.
 **`devurl`** the folder that in-development version of the docs will be deployed.
 Defaults to `"dev"`.
 
+**`selector`** determines content and order of the resulting version selector in
+the generated html. The following entries are valied in the `selector` vector:
```
mortenpi: Just bikeshedding here now -- since the `selector` also defined which symlinks (i.e. which public URLs) get deployed, maybe it should be called something like `versions`? It's not only the selector.

I would also mention here that it creates the URLs the person reading the docs sees and should use.

src/Documenter.jl
```diff
@@ -316,6 +317,18 @@ documentation. By default this value is set to `"master"`.
...
+ - `"v#"`: includes links to the latest documentation for each major release cycle
+   (i.e. `v1.1`, `v2.0`).
+ - `"v#.#"`: includes links to the latest documentation for each minor release cycle
+   (i.e. `v1.0`, `v1.1`, `v2.0`).
+ - `"v#.#.#"`: includes links to all released versions.
+ - `"v^"`: includes a link to the maximum version.
```
mortenpi: What does it put in the selector though? Or can you just use this in a pair?

src/Documenter.jl
```diff
@@ -316,6 +317,18 @@ documentation. By default this value is set to `"master"`.
...
+ - A pair, e.g. `"first" => "second"`, which will put `"first"` in the selector,
+   with the same content as `"second"`. The second argument can be `"v^"`, to
+   point to the maximum version docs (as in e.g. `"stable" => "v^"`).
```
mortenpi: Ditto as above w.r.t. URLs. It not only puts something into the selector, but also defines the URL you can access "second" from.

src/Documenter.jl
fredrikekre: I like `versions`.

src/Documenter.jl
fredrikekre: It will put `vX.Y` for latest `X` and `Y`, and symlink to `vX.Y.Z` for latest `Z`.

fredrikekre:
Eh... failed deployment: https://travis-ci.org/JuliaDocs/Documenter.jl/builds/420718372?utm_source=github_status&utm_medium=notification will push a fix shortly.

PR #818, opened 2018-08-26
Remove the link when we are symlinking

fredrikekre:
Something else is wrong, I wonder if `islink` does not work on the cloned repo or something...
Trying to debug here.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -551,9 +551,17 @@ function git_push(
                 # This must always happen after the folder copying.
                 Writers.HTMLWriter.generate_version_file(joinpath(dirname, "versions.js"), entries)
 
-                # generate the symlinks
+                # generate the symlinks, make sure we don't overwrite devurl
                 cd(dirname) do
-                    foreach(kv -> rm_and_add_symlink(kv.second, kv.first), symlinks)
+                    for kv in symlinks
+                        i = findfirst(x -> x.first == devurl, symlinks)
+                        if i === nothing
+                            rm_and_add_symlink(kv.second, kv.first)
+                        else
+                            throw(ArgumentError(string("link `$(kv)` cannot overwrite ",
+                                "`devurl = $(devurl)` with the same name.")))
+                        end
+                    end
                 end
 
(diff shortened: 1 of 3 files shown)


fredrikekre:
Ok, the problem is that `stable` is a directory that exists... We can remove that, from Documenter of course, since Documenter will symlink it, but this could be bad if a user have set `devurl = "stable"` for some weird reason. I guess we can special case for now the Pair `"stable" => "v^"` and remove in only that case. I suspect that, at least to start with, most people will just use the defaults for `versions = ...` anyway...

mortenpi:
We could add a check though, to make sure that the version selector does not try to create `devurl`, and warn/fail the build if it does.

fredrikekre:
Not sure I understand what you mean; I removed the special casing of  `"stable"` and we now always remove the link. It should be fine to trust the user input here. We also filter out things like `"dev" => "dev"` https://github.com/JuliaDocs/Documenter.jl/blob/58e1883e183dacf7ecda39c24e76df765aa085db/src/Writers/HTMLWriter.jl#L551 so we never accidentally remove what we are trying to link to.

mortenpi:
What will happen if `devurl = "stable"` and you also have `"stable" => "v^"`? Won't it delete the `master` docs you just deployed?

fredrikekre:
Ah good catch. We should assert that a link does not point to another link.

mortenpi:
> Or should we check explicitly that we never remove the `devurl` folder?

I don't think it would hurt to have that check, just in case. It's a contradiction in `deploydocs`' arguments. Just a small loop in `git_push!` that checks that none of the `symlinks` firsts are equal to `devurl`?

fredrikekre:
Yea, might as well catch that corner case.

fredrikekre:
This could be annoying if anyone wants to keep `latest` in the selector(`versions = ["latest" => "dev"]`), but use `devurl = "dev"`. But I guess one has to manually remove stuff from `gh-pages` then.

mortenpi:
LGTM! I'll merge this to get the master doc builds passing again.

Issue #713, opened 2018-04-21
Breaking release


mortenpi:
It looks like we're heading towards a breaking release. We should try doing as many breaking changes together as possible, to make our lives easier. Right now I can think of these (but additional suggestions are very welcome):

* [x] Changing `makedocs(format=:html)`
* [x] Changing `deploydocs(julia=...)`. (#526 / #466; depwarn in #712)
* [x] Don't use the `release-x.y` directories. (#706)
* [x] Other versioning changes (ref: JuliaLang/julia#26825,  #813)
* [x] Remove `julia` and `osname` arguments to `deploydocs()` and promote Travis build stages instead ( #816)
* [x] ~Update message (https://discourse.julialang.org/t/displaying-an-update-message/12577)~ -- problematic with new Pkg.

Additional non-coding tasks we need to do before we can release this:

* [x] Automated PRs against METADATA packages using Documenter.
* [x] Need to ANN this on Discourse, ideally a reasonable time before the release hits METADATA.

The automated PRs could just add an upper bound to Documenter's `Pkg.add`. We could also consider actually making the necessary fixes to `make.jl` etc.

---

We also need to tag the new packages:

* [x] DocumenterTools.jl
* [x] DocumenterMarkdown.jl
* [x] DocumenterLaTeX.jl

fredrikekre:
> The automated PRs could just add an upper bound to Documenter's Pkg.add. We could also consider actually making the necessary fixes to make.jl etc.

I think with we should recommend people to make a doc-build project in e.g. docs/Project.toml, and commit its Manifest.toml. That way one needs to explicitly call up to update Documenter.

mortenpi:
I would go with the middle ground and have user have a `[compat]` section in their `Project.toml` to fix the major version. That way they would still get bugfixes.

mortenpi:
TODO list for the mass PRs for setting upper bounds -- summarizing our discussion @fredrikekre:

- [ ] (1) Draft the issue, PR and discourse PSA
- [ ] (2) Gather a list of packages we need to PR and see what we need to do exactly.
- [ ] (3) Create an issue on Documenter for tracking the PRs. We should also lock the conversation there and suggest that people subscribe to that issue to be notified when v0.20 is released. The only thing that should be posted there is the announcement that v0.20 has hit METADATA. Other conversation should happen here in #713 (or Discourse).
- [ ] (4) Make all the PRs. Make sure they refer to the issue created in (3).
[...]

PR #826, opened 2018-09-03
Change default output format to HTML

### LINKED ISSUE #394: Make HTML the default backend
Hi, I just yesterday explored this great package. From reading through the documentation it would make from my perspective make sense to make the html backend the default. The regular user will be pretty satisfied with it and having to not bother with extra tools is IMHO a huge win.

mortenpi:
Docs still need a bit of work.

Fixes #394, ref #713.

---

Has two breaking changes:

* `format` now defaults to `:html` (previously `:markdown`).
* `html_prettyurls` is now enabled by default. So Documenter will now generate `build/foo/index.html` instead of `build/foo.html` for `src/foo.md` by default.


fredrikekre:
How WIP is this?

mortenpi:
Definitely shouldn't be merged yet. Will hopefully have time over the weekend to finish it off though.

PR #829, opened 2018-09-13
Slightly tweak recommended travis setup.

fredrikekre:
Otherwise we update the registry twice etc and no reason not to do this in the same session.


mortenpi:
I feel like we could use a command line argument that would instantiate a project right away. I.e. that we could do away with `Pkg.instantiate()`.

PR #837, opened 2018-09-19
Clarify some of the debug info

### LINKED ISSUE #836: I get told to  to set DOCUMENTER_DEBUG for more info even if it is already set
I think this is caused by
https://github.com/JuliaDocs/Documenter.jl/blob/2a986d0fc03568d3ade30ece6b065282922e6ff8/src/Documenter.jl#L473

Not actually checking if it is already set.

It is a bit confusing being told to do something I have already done.
Unless there is a check, and I have just screwed up setting the environment variable in Travis somehow.

oxinabox:
Closes #836 

Also the use of match in 

```
 ?? TRAVIS_REPO_SLUG       = "oxinabox/DataDeps.jl"
 ??   should match "github.com/oxinabox/DataDeps.jl.git" (kwarg: repo)
```

lead  me down the wrong path.
As I interpretted 
match as "equal"

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -470,7 +470,7 @@ function deploydocs(;
     else
         Utilities.log("skipping docs deployment.")
         if get(ENV, "DOCUMENTER_DEBUG", "") != "true"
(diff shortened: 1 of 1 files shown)


mortenpi:
LGTM, thanks!

Issue #827, opened 2018-09-06
Generating ssh key on Windows


odow:
I just spent the last wee while bashing my head against this. I had done it a while ago and didn't think to write it down because it seemed obvious. When I went to do it again it wasn't so obvious.

I'm busy this afternoon but I will try to add a section to the docs going through this when I get the chance.

Here are the steps I took:
## Generate a key
1.	Open PuTTY Key Generator. ([Download link](https://www.ssh.com/ssh/putty/windows/puttygen).)
![image](https://user-images.githubusercontent.com/8177701/45131792-a09a6b80-b1e2-11e8-885b-499237651c29.png)
2.	Click Generate, and follow the instructions, moving the mouse around to create randomness.

## Add the public key to Github
3.	To go https://github.com/[YOUR_USER_NAME]/[YOUR_REPO_NAME]/settings/keys and click `Add deploy key`
![image](https://user-images.githubusercontent.com/8177701/45131985-4fd74280-b1e3-11e8-9d95-46539845ee3a.png)
4.	Make the title `documenter`
5.	Copy and paste the highlighted text from PuTTY beginning with ssh-rsa into the key 
![image](https://user-images.githubusercontent.com/8177701/45131800-a728e300-b1e2-11e8-8d40-dbb4fa383ff5.png)
6.	Click allow write access
7.	Click Add key

## Add the private key to travis
8.	Go back to the PuTTY Key Generator. Under the `Conversions` tab, click `Export OpenSSH key`. Click through the prompt that says you are saving without a password (Travis needs a key with no passphrase). Save that key somewhere.
![image](https://user-images.githubusercontent.com/8177701/45131813-b4de6880-b1e2-11e8-82bd-ec8ceb44a976.png)
9.	Run `Julia> Documenter.base64encode(readstring(“path/to/the/private/key”))`
10.	Open https://travis-ci.org/[YOUR_USER_NAME]/[YOUR_REPO_NAME]/settings 
![image](https://user-images.githubusercontent.com/8177701/45132034-801ee100-b1e3-11e8-8152-b407f0e89770.png)
11.	Under the `Environment Variables` heading, add an environment variable. The name should be `DOCUMENTER_KEY` and the value should be the output of the Julia command from step 9 (make sure to remove the surrounding quotes).
12.	Check that the “Display value in build log” is switched off and then click Add.

mortenpi:
[...]

PR #838, opened 2018-09-21
Add a more detailed walkthrough for SSH deploy-keys

odow:
This PR adds a verbose walk-through for generating an SSH key manually (i.e., not using `Travis.genkeys()`) and uploading the keys to Github and Travis.

See https://github.com/JuliaDocs/Documenter.jl/issues/827 for the preceding discussion. I've linked to the screenshots in that issue rather than upload them as assets.

Closes https://github.com/JuliaDocs/Documenter.jl/issues/827.


mortenpi:
This is great, thanks a lot @odow! I have a upcoming PR with some documentation updates, so I'll merge this right away.

The only request I would have is whether you could provide a larger screengrab for the third image (the one about exporting the private key)? I think it would be nice to have them all be the same size.

PR #840, opened 2018-09-23
Replace puttygen-export-private-key.png with bigger image.

odow:
As requested by @mortenpi in https://github.com/JuliaDocs/Documenter.jl/pull/838#issuecomment-423722598, here is a larger image. 

Not sure how I even took the smaller screengrab in the first place...


mortenpi:
Cheers!

PR #843, opened 2018-09-23
Replace DocChecks.nameof with Base.nameof

mortenpi:
Following up on https://github.com/JuliaDocs/Documenter.jl/pull/795#discussion_r210376821

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -73,10 +73,7 @@ end
 
 meta(m) = Docs.meta(m)
 
-nameof(x::Function)          = typeof(x).name.mt.name
-nameof(b::Base.Docs.Binding) = b.var
-nameof(x::DataType)          = x.name.name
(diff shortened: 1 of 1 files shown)


fredrikekre:
I don't think this is necessary; no need to expose this internal function to users, and it is technically type piracy.

mortenpi:
Well, the original `DocChecks.nameof` does stack overflow for `Module`s, so we need to do something. You are right about type piracy though. I didn't really check which methods we actually need in Documenter.. it does appear that we never call it with a `Module`.

Do you reckon we should just stick with a local name and fall back the `Base` version for the general case? I.e.

```julia
nameof(b::Base.Docs.Binding) = b.var
nameof(x) = Base.nameof(x)
```

fredrikekre:
> Do you reckon we should just stick with a local name and fall back the Base version for the general case? I.e.

Yes.

PR #844, opened 2018-09-24
Recommend develop instead of add.

fredrikekre:
@KristofferC pointed out that this is better since `add` does a (full ?) clone of the package, but we already have that clone so we can just point to it with `develop`.


mortenpi:
It also feels more correct actually -- a package checked out on Travis is more "dev" rather than "released" (`add` somewhat implies the latter, to me at least).

PR #848, opened 2018-10-02
set IO context for doctest output relative to sandbox module

JeffBezanson:
If we make a change like https://github.com/JuliaLang/julia/pull/29466, this will be needed. Should be safe now as well.


mortenpi:
LGTM, thanks!

I would suggest bumping Documenter in JuliaLang/julia#29470 and merging that to get this onto master. @fredrikekre?

But I will also backport this to 0.19.

PR #852, opened 2018-10-09
Fix Utilities.submodules for module aliases

### LINKED ISSUE #851: Module aliases confuse Utilities.submodules
```julia
julia> module M1
       using LinearAlgebra
       end
Main.M1

julia> module M2
       using LinearAlgebra
       const LA = LinearAlgebra
       end
Main.M2

julia> using Documenter

julia> Documenter.Utilities.submodules(M1)
Set(Module[Main.M1])

julia> Documenter.Utilities.submodules(M2)
Set(Module[LinearAlgebra.BLAS, Main.M2, LinearAlgebra, LinearAlgebra.LAPACK])
```

This is causing a lot of spurious doc check "docstrings potentially missing" warnings for me.

tkoolen:
Fix #851

### DIFF
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -35,6 +35,17 @@ export OuterModule
 end
 end
 
+module ModuleWithAlias
+using LinearAlgebra
+const LA = LinearAlgebra
+module A
+    module B
+    using Random
+    const R = Random
+    end
+end
+end
+
 @testset "Utilities" begin
     let doc = @doc(length)
         a = Documenter.Utilities.filterdocs(doc, Set{Module}())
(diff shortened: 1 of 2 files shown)


test/utilities.jl
```diff
@@ -35,6 +35,17 @@ export OuterModule
 end
 end
 
+module ModuleWithAlias
+using LinearAlgebra
```
fredrikekre: Create another test module instead? I believe this will pull in `LinearAlgebra` and `Random` as dependencies, which is not optimal.

tkoolen: Done. I do think there should be a module that's its own parent as a test case, so I'm using `Main` for that now.

mortenpi:
LGTM, thanks!

PR #853, opened 2018-10-09
add section for filtering docstrings

Datseris:


thofma:
There is already https://juliadocs.github.io/Documenter.jl/latest/man/guide/#Filtering-included-docstrings-1

fredrikekre:
Would still be good to add a link to that section from the `@docs` section.

mortenpi:
I'd say that the at-docs section should have the detailed description, as that page is like a reference for the at-blocks, and the Guide should have a simple example with a link to the reference.

PR #862, opened 2018-10-17
Fixes #855

### LINKED ISSUE #855: Version selector CSS broken
The styling of the version selector is broken for long sidebars due to #792:

selector-bug

dalum:
This seems to fix #855, at least in Firefox.


fredrikekre:
Good to go?

mortenpi:
I can confirm it also works on Chrome. Thanks!

Issue #75, opened 2016-06-01
CLI for make.jl


mortenpi:
This is a small RFC to discuss a possible extension of the `make.jl`

In a nutshell, the `make.jl` script could take arguments to allow the user have a bit more control over what it does. I am thinking basically how `mkdocs` works. So, you would call it like `julia make.jl <command>`, where the command could be `build`, `clean`, `deploy`, `serve` etc.

You could also have more arguments, to give even more power to the user, without them having to make changes to the script (e.g. specify the target format, `--debug` etc). Also, having a `serve` command would be a first step in addressing #11.

In the `make.jl` it would look something like this
```julia
using Documenter, PkgName
Documenter.run(ARGS; config...)
```

The implementation actually wouldn't necessarily conflict with the way the scripts work currently, since we can keep `makedocs` and `deploydocs` function (and would probably call them in `run`), even though perhaps in the long term they should be deprecated. Also, `build` could be the default command, so `julia make.jl` would work as it does currently.

One more benefit would be that some things would become a bit more explicit. For example, `deploydocs` currently semi-magically decides if it should do something or not. The explicitness could maybe help with testing as well.

MichaelHatherly:
Yes, probably worth doing I think.

PR #864, opened 2018-10-23
Allow arbitrary variables to be passed to be stored in User and Internal

adamslc:
This PR is the start of an attempt to make `Documenter` support passing arbitrary arguments from `makedocs` through to a `Writer`. So far, I have just removed a few of the HTML specific variables from `User`, but, if people this that this is a reasonable approach, I will continue to strip away the special case variables.

One downside to this approach as it is currently implemented is that it scatters the default values for variables throughout the code. But this could be resolved with a few `get!`s at the start of the `render` function in each writer.

Another potential downside is the performance hit resulting from the untyped dictionary. I haven't done any benchmarking to see if this is significant.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -16,6 +16,19 @@ module Documenter
 using DocStringExtensions
 import Base64: base64decode
 
+"""
+Any plugin that needs to either solicit user input or store information in a
+[`Documents.Document`](@ref) should create a subtype of [`Plugin`](@ref). The
+subtype, `T <: Documenter.Plugin`, must have an empty constructor `T()` that
+initialized `T` with the appropriate default values.
+
+To retrieve the values stored in `T`, the plugin can call [`Documents.getplugin`](@ref).
+If `T` was passed to [`makedocs`](@ref), the passed type will be returned. Otherwise,
+a new `T` object will be created.
+"""
+abstract type Plugin end
+
+
 # Submodules
 # ----------
 
@@ -33,13 +46,13 @@ include("Writers/Writers.jl")
 include("Deps.jl")
 
 import .Utilities: Selectors
+import .Writers.HTMLWriter: HTML
 
 
 # User Interface.
 # ---------------
 
 export Deps, makedocs, deploydocs, hide
-
 """
     makedocs(
         root    = "<current-directory>",
@@ -161,13 +174,14 @@ ignored.
 any errors with the document in the previous build phases.
 
 ## Output formats
+**`format`** allows the output format to be specified. The default value is `:html`. Other
+formats can be enabled by using other packages. For examples, see the `DocumenterMarkdown`
+and `DocumenterLaTeX` packages.
 
-**`format`** allows the output format to be specified. The default value is `:html`, but
-additional values are supported via plugins (see below).
-
-The different output formats may require additional keywords to be specified. The
-format-specific keywords for the default HTML output are documented at the
-[`Writers.HTMLWriter`](@ref) module.
+Documenter is designed to support multiple output formats. By default it is creates a set of
+HTML files, but the output format can be controlled with the `format` keyword. The different
+output formats may require additional keywords to be specified via plugins. The keywords for
+the default HTML output are documented for the [`Writers.HTMLWriter.HTML`](@ref) type.
 
 The default `:html` output format creates a set of HTML files, but Documenter is designed to
 support multiple output formats. Via plugin packages, Documenter also supports e.g. Markdown
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -178,8 +178,8 @@ support multiple output formats. Via plugin packages, Documenter also supports e
 A guide detailing how to document a package using Documenter's [`makedocs`](@ref) is provided
 in the [setup guide in the manual](@ref Package-Guide).
 """
-function makedocs(; debug = false, args...)
-    document = Documents.Document(; args...)
+function makedocs(components...; debug = false, kwargs...)
+    document = Documents.Document(components; kwargs...)
     cd(document.user.root) do
         Selectors.dispatch(Builder.DocumentPipeline, document)
     end
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -390,12 +390,12 @@ Retrieves the [`Plugin`](@ref) type for `T` stored in `doc`. If `T` was passed t
 [`makedocs`](@ref), the passed type will be returned. Otherwise, a new `T` object
 will be created using the default constructor `T()`.
 """
-function getplugin(doc::Document, type::Type{T}) where T <: Plugin
-    if !haskey(doc.plugins, type)
-        doc.plugins[type] = type()
+function getplugin(doc::Document, plugin_type::Type{T}) where T <: Plugin
+    if !haskey(doc.plugins, plugin_type)
+        doc.plugins[plugin_type] = plugin_type()
     end
 
-    doc.plugins[type]
+    doc.plugins[plugin_type]
 end
 
 ## Methods
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -13,7 +13,8 @@ import ..Documenter:
     Documenter,
     Anchors,
     Formats,
-    Utilities
+    Utilities,
+    Plugin
 
 using DocStringExtensions
 import Markdown
@@ -220,8 +221,6 @@ struct Internal
     errors::Set{Symbol}
 end
 
-abstract type DocumenterPlugin end
-
 # Document.
 # ---------
 
@@ -231,9 +230,10 @@ Represents an entire document.
 struct Document
     user     :: User     # Set by the user via `makedocs`.
     internal :: Internal # Computed values.
-    plugins  :: Dict{DataType, DocumenterPlugin}
+    plugins  :: Dict{DataType, Plugin}
 end
 
+Document(; kwargs...) = Document(nothing; kwargs...)
 function Document(plugins;
         root     :: AbstractString   = Utilities.currentdir(),
         source   :: AbstractString   = "src",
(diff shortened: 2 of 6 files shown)


adamslc:
This is in a state where it can be merged.

Now the question is if this is the right approach to allowing external components to interact with the internal Documenter pipeline. I'm not particularly tied to this solution, but I do think that *something* to this effect is needed.

mortenpi:
Apologies for not getting to this earlier and thanks for bringing this up! I do think this will require some bikeshedding. Here are some general comments that may or may not be relevant:

* We should do something about the `html_*` keywords. Although actually, this PR doesn't make any user-facing changes, so that is probably out of the scope for this PR.
* Helping out plugin packages is definitely something we want to do. Citations looks awesome and I am very keen to hear what roadblocks you run into and what you need from Documenter to implement such a package.
* It would be nice to have the default values defined in one place (ideally somewhere in the `HTMLWriter` module for the `html_*` variables).
* I have some thoughts about where the user API should be going. It is in no way a final design, but [I was playing around with some ideas at some point](https://gist.github.com/mortenpi/31f9d38fb6c7bc5d858ac5b86c4fb039). (Also [roadmap](https://github.com/JuliaDocs/Documenter.jl/wiki/Documenter.jl-roadmap#new-api-for-the-user) and #75 are tangentially related.)

So I think that I would ideally pass a `struct` like `HTML(canonical="...")` for the "non-global" variables. We could maybe then have another struct `Plugin`, which could simply be a wrapper around a `Dict` internally?

```julia
makedocs(
    sitename = "...",
    HTML(canonical = "..."), # equiv to current html_canonical="..."
    Plugin(foo = "bar"),
)
```

Or better yet, maybe a `struct` for every plugin? Documenter could define an abstract type for this which could be used to filter the `args` of `makedocs`. I.e. `CitationsPlugin <: DocumenterPlugin` and 

```julia
makedocs(
    sitename = "...",
    HTML(html_canonical = "..."),
    CitationsPlugin(foo = "bar"),
    SomeOtherPlugin(baz = "qux"),
)
```

I think there is some value in namespacing these things, as you might get name clashes otherwise if you have multiple plugins.

We also need some methods acting on the `Document` object for fetching configuration variables I think, to be used internally by the plugins. We shouldn't rely too much on the internal representation of the structs (especially since they should probably be refactored at some point).

I need to think about this a bit more though and would love to hear opinions on this.

adamslc:
The struct-per-plugin approach seems much better to me. I knocked together a prototype of that system. A few comments about this prototype:
1. The naming isn't particularly good. I especially don't like `plugin_settings`, but that can be decided on later.
2. The defaults for `HTMLWriter` have been consolidated once again (and arguably in a more sensible place). This also removes the need for the, IMHO ugly, `html_` prefixes.
3. It would be nice to remove the `format` keyword, and just build for each writer plugin provided to `makedocs`. We might need another type `WriterPlugin <: DocumenterPlugin` for this.

I also looked at your proposed interface, and I have a few thoughts about that:
1. I don't really like passing `ARGS` to `run`; it seems a little too magical. I think would be better to explicitly handle command line args in `make.jl`.
2. Making the whole process more modular would be a huge win in my mind. This scheme would eliminate the need for point (3) above.

src/Documents.jl
```diff
@@ -253,14 +253,9 @@ function Document(;
         sitename :: AbstractString   = "",
         authors  :: AbstractString   = "",
         analytics :: AbstractString  = "",
-        version :: AbstractString    = "",
-        html_prettyurls  :: Bool     = true,
-        html_disable_git :: Bool     = false,
-        html_edit_branch :: Union{String, Nothing} = "master",
-        html_canonical   :: Union{String, Nothing} = nothing,
```
mortenpi: We should keep these around with deprecation warnings though. I'd say:

* If `HTML` _and_ any of these are passed, throw an error.
* If only keywords are passed, construct an `HTML` object out of them, but also print a depwarn.

src/Documents.jl
```diff
@@ -307,9 +298,28 @@ function Document(;
...
+            throw("$(typeof(plugin)) is not a subtype of `DocumenterPlugin`.")
+        haskey(plugin_dict, typeof(plugin)) &&
+            throw("only one copy of $(typeof(plugin)) may be passed.")
+        plugin_dict[typeof(plugin)] = plugin
+    end
+
+    Document(user, internal, plugin_dict)
+end
+
+function plugin_settings(doc::Document, type::DataType)
```
mortenpi: Maybe `plugin` or `getplugin`?

src/Documents.jl
```diff
@@ -307,9 +298,28 @@ function Document(;
...
+    @assert type <: DocumenterPlugin
+    if !haskey(doc.plugins, type)
+        doc.plugins[type] = type()
```
mortenpi: We should document that a `DocumenterPlugin` needs to have `T()` construct available.

src/Documents.jl
```diff
@@ -223,6 +220,8 @@ struct Internal
     errors::Set{Symbol}
 end
 
+abstract type DocumenterPlugin end
```
mortenpi: This should probably move into the `Documenter` module.

Also, I wonder if it might be better to call it just `Plugin`. In this case, it shouldn't be exported, but it will be part of public API. The users should access it with `Documenter.Plugin`.

src/Documents.jl
```diff
@@ -253,14 +253,9 @@ function Document(;
...
-        version :: AbstractString    = "",
-        html_prettyurls  :: Bool     = true,
-        html_disable_git :: Bool     = false,
-        html_edit_branch :: Union{String, Nothing} = "master",
-        html_canonical   :: Union{String, Nothing} = nothing,
-        others...
+        version :: AbstractString    = ""
     )
-    Utilities.check_kwargs(others)
+    # Utilities.check_kwargs(others)
```
mortenpi: Let's keep this in for now. Theoretically, we currently allow passing arbitrary arguments without throwing an error, so this would be breaking.

src/Documents.jl
```diff
@@ -307,9 +298,28 @@ function Document(;
...
+        haskey(plugin_dict, typeof(plugin)) &&
+            throw("only one copy of $(typeof(plugin)) may be passed.")
+        plugin_dict[typeof(plugin)] = plugin
+    end
+
+    Document(user, internal, plugin_dict)
+end
+
+function plugin_settings(doc::Document, type::DataType)
+    @assert type <: DocumenterPlugin
```
mortenpi: Could do something like `plugin_settings(doc::Document, type::Type{T}) where T <: DocumenterPlugin`, right?

adamslc:
I think that I've updated all of the relevant docstrings, but it is entirely possible that I missed somethings. Also, it looks like there are a lot more things
that can find new homes in plugin types, but hopefully those can be tackled in later PRs?

src/Documenter.jl
```diff
@@ -34,13 +47,13 @@ include("Writers/Writers.jl")
...
 import .Utilities: Selectors
+import .Writers.HTMLWriter: HTML
 
 
 # User Interface.
 # ---------------
 
-export Deps, makedocs, deploydocs, hide
-
+export Deps, makedocs, deploydocs, hide, HTML
```
mortenpi: It turns out that `HTML` is already being exported by `Base`, so one needs to do explicit `Documenter.HTML`. But that's a bit long.. so I am wondering if there is a better way.

Maybe `export` `Writers` and have it be accessible by `Writers.HTML`:

```julia
makedocs(
  Writers.HTML(prettyurls=false),
  ...
)
```

src/Documents.jl
```diff
@@ -10,13 +10,17 @@ Defines [`Document`](@ref) and its supporting types
...
     Formats,
-    Utilities
+    Utilities,
+    # Writers,                  # Only needed for html keywords deprecation
+    Plugin
 
 using DocStringExtensions
 import Markdown
 using Unicode
+# import .Writers: HTMLWriter    # Only needed for html keywords deprecation
```
mortenpi: The commented out things should be removed I guess.

src/Utilities/Utilities.jl
```diff
@@ -159,6 +159,7 @@ function check_kwargs(kws)
     for (k, v) in kws
         println(out, "  ", k, " = ", v)
     end
+    println(out, "It is possible that the specified format will use these arguments.")
```
mortenpi: I guess this is not really necessary anymore?

src/Documenter.jl
adamslc: I opted for `Documenter.HTML` because pretty soon `Writers` will only have a single writer under it. Also, `make.jl` files shouldn't be edited too often, so the extra typing seems insignificant. 

I can change it if you feel strongly that `Writers.HTML` is the right way forward.

src/Documents.jl
adamslc: Yeah, sorry for the mess...

adamslc:
I just squashed this into two logical commits. I see you added a news label, but I'm not sure where the news should go? Is there anything else that needs to be done here?

mortenpi:
I would like to take another proper look at this and I should be able to get to it on the weekend. But I think it is most likely good to go, except for the conflict in `src/Documenter.jl`.

src/Documents.jl
```diff
@@ -307,9 +303,99 @@ function Document(;
...
+"""
+    getplugin(doc::Document, T)
+
+Retrieves the [`Plugin`](@ref) type for `T` stored in `doc`. If `T` was passed to
+[`makedocs`](@ref), the passed type will be returned. Otherwise, a new `T` object
+will be created using the default constructor `T()`.
+"""
+function getplugin(doc::Document, plugin_type::Type{T}) where T <: Plugin
+    if !haskey(doc.plugins, plugin_type)
+        doc.plugins[plugin_type] = plugin_type()
```
fredrikekre: Wouldn't it be better to return `nothing` here instead of assuming that `T` has a `T()` constructor?

fredrikekre:
Do you have an example implementation that make use of this @adamslc ?

src/Documents.jl
adamslc: The rational was to make the interface as easy as possible from the user's perspective. With this implementation, a user can just `using SomePlugin` in `make.jl` and things will Just Work (tm).

PR #867, opened 2018-10-27
Current package as a dep in docs/Project.toml

AsafManela:
The failure in the script is highly uninformative. Such a note would have saved me some time.

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -120,8 +120,8 @@ jobs:
       julia: 1.0
       os: linux
       script:
-        - julia --project=docs/ -e 'using Pkg; Pkg.instantiate();
-                                    Pkg.develop(PackageSpec(path=pwd()))'
+        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
+                                               Pkg.instantiate()'
         - julia --project=docs/ docs/make.jl
       after_success: skip
 ```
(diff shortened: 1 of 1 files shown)


fredrikekre:
I am not sure what you mean, can you clarify what you did and what failed?

fredrikekre:
Ah, thanks for the clarification! Turns out we can "fix" this by switching the order of the `instantiate` and the `develop`, so maybe we should apply
````diff
diff --git a/docs/src/man/hosting.md b/docs/src/man/hosting.md
index 1787afc..82fb811 100644
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -120,8 +120,8 @@ jobs:
       julia: 1.0
       os: linux
       script:
-        - julia --project=docs/ -e 'using Pkg; Pkg.instantiate();
-                                    Pkg.develop(PackageSpec(path=pwd()))'
+        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
+                                               Pkg.instantiate()'
       after_success: skip
 ```
````

and maybe we don't need this extra note added in this PR. Thoughts?

AsafManela:
Yes! That sounds more robust than the note.

fredrikekre:
Do you wanna update the PR?

fredrikekre:
Did you forget to `git add` the other change?

AsafManela:
Not a pro gitter yet... What exactly did you want me to do? I don't know how to merge just your last commit from an upstream branch that has other changes as well.

fredrikekre:
I was mostly wondering if you wanted to update this PR to include the change I suggested in the comment.

AsafManela:
OK, this should do the trick. Thanks!

PR #869, opened 2018-10-29
Document how to disable html_prettyurls for local builds

### LINKED ISSUE #868: Missing `index.html` with hyperlinks in Documenter v0.20
I have tried updating [MathOptInterface](https://github.com/JuliaOpt/MathOptInterface.jl) to Documenter v0.20: https://github.com/JuliaOpt/MathOptInterface.jl/pull/552
The transition went smoothly and the `make.jl` got largely simplified but when trying locally (I assume the same happens when it is built on Travis), I noticed that links do not work.
For instance, the `Manual` link from the index redirects to
```
file:///home/blegat/.julia/dev/MathOptInterface/docs/build/apimanual/#Manual-1
```
instead of
```
file:///home/blegat/.julia/dev/MathOptInterface/docs/build/apimanual/index.html#Manual-1
```
Looking at the the doc build with Documenter v0.19 at
http://www.juliaopt.org/MathOptInterface.jl/stable/index.html
it used to redirect to
http://www.juliaopt.org/MathOptInterface.jl/stable/apimanual.html#Manual-1
When looking at the Documenter doc, it seems that the link does not contain `index.html` but it works without it.

fredrikekre:
fix #868

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -124,9 +124,42 @@ build/
         documenter.js
         search.js
     index.html
-    search.html
+    search/index.html
     search_index.js
 ```
+
+!!! note
+
+    By default, Documenter has pretty URLs enabled, which means that `src/foo.md` is turned
+    into `src/foo/index.html`, instead of simply `src/foo.html`, which is the preferred way
+    when creating a set of HTML to be hosted on a web server.
+
+    However, this can be a hindrance when browsing the documentation locally as browsers do
+    not resolve directory URLs like `foo/` to `foo/index.html` for local files. You have two
+    options:
+
+    1. You can run a local web server out of the `docs/build` directory. If you have Python
+       installed, you can simple start one with `python3 -m http.server --bind localhost`
+       (or `python -m SimpleHTTPServer` with Python 2).
+
+    2. You can disable the pretty URLs feature by passing `prettyurls = false` with the
+       [`Documenter.HTML`](@ref) plugin:
+
+       ```julia
+       makedocs(..., Documenter.HTML(prettyurls = false))
+       ```
+
+       Alternatively, if your goal is to eventually set up automatic documentation deployment
+       with Travis CI (see [Hosting Documentation](@ref)), you can also use their environment
+       variables to determine Documenter's behavior in `make.jl` on the fly:
+
+       ```julia
+       makedocs(...,
+           Documenter.HTML(
+               prettyurls = get(ENV, "CI", nothing) == "true"
+           )
+       )
+       ```
 
 !!! warning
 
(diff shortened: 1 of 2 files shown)


mortenpi:
Thumbs down on this. I really don't think that `makedocs`'s behaviour should be depending on the environment like this. It should be explicit in `make.jl`.

I do understand that the "pretty URL" thing can be confusing though. Do you reckon recommending people to set `html_prettyurls` explicitly in the docs would be enough?

mortenpi:
Should we also maybe have a note on the Guide page? Also, the example file tree is slightly out of date actually -- it needs `search.html` -> `search/index.html`.

blegat:
In which case would it work locally ? In which situation would the user not want to do `html_prettyurls = get(ENV, "CI", nothing) == "true"` ?

mortenpi:
It's just that `makedocs`'s behaviour shouldn't magically depend on the environment. Especially since the `CI` variable is in principle specific to particular CI systems (although it is relatively universal, I guess). It would be much more confusing if people get `html_prettyurls = false` behavior when they're setting it up locally, but then suddenly, for no apparent reason Documenter will behave differently when it runs on Travis.

But `get(ENV, "CI", nothing) == "true"` is what you usually want, so it should definitely be documented and be part of the recommended default setup.

blegat:
I understand, thank you for detailing your choice.

mortenpi:
@fredrikekre I rebased this and slightly expanded the docs.

fredrikekre:
Thanks, LGTM.

docs/src/man/guide.md
```diff
@@ -124,10 +124,43 @@ build/
...
+    By default, Documenter has pretty URLs enabled, which means that `src/foo.md` is turned
+    into `src/foo/index.html`, instead of simply `src/foo.html`, which is the preferred way
+    when creating a set of HTML to be hosted on a web server.
+
+    However, this can be a hindrance when browsing the documentation locally as browsers do
+    not resolve directory URLs like `foo/` to `foo/index.html` for local files. You have two
+    options:
+
+    1. You can run a local web server out of the `docs/build` directory. If you have Python
+       installed, you can simple start one with `python3 -m http.server --bind localhost`
```
scls19fr: Isn't there a way to do the same with Julia?
https://github.com/JuliaWeb/HttpServer.jl/issues/143

mortenpi: Most likely and I would happily take a PR that documents such a workflow. But Python seems to be the easiest option at the moment since it's literally just one command on the command line, and it's what I happen to use.

Also, a more sophisticated implementation, more in the vein of `mkdocs serve`, could be implemented in DocumenterTools (JuliaDocs/DocumenterTools.jl#18).

PR #876, opened 2018-10-31
WIP Logging

fredrikekre:
Still need some clean up.


src/Documenter.jl
```diff
@@ -393,66 +393,50 @@ function deploydocs(;
...
-        Utilities.debug("TRAVIS_REPO_SLUG       = \"$travis_repo_slug\"")
-        Utilities.debug("  should occur in \"$repo\" (kwarg: repo)")
-        Utilities.debug("TRAVIS_PULL_REQUEST    = \"$travis_pull_request\"")
-        Utilities.debug("  deploying if equal to \"false\"")
-        Utilities.debug("TRAVIS_BRANCH          = \"$travis_branch\"")
-        Utilities.debug("TRAVIS_TAG             = \"$travis_tag\"")
-        Utilities.debug("  deploying if branch equal to \"$devbranch\" (kwarg: devbranch) or tag is set")
-        Utilities.debug("git commit SHA         = $sha")
-        Utilities.debug("DOCUMENTER_KEY exists  = $(!isempty(documenter_key))")
-        Utilities.debug("should_deploy          = $should_deploy")
```
mortenpi: Hmm, it might be good if we would still output all the environment variables and their values at the top of `deploydocs` (in addition to the warnings; "Here are all the environment variables affecting deploydocs:").

src/Builder.jl
```diff
@@ -76,7 +76,7 @@ Selectors.matcher(::Type{T}, doc::Documents.Document) where {T <: DocumentPipeli
 Selectors.strict(::Type{T}) where {T <: DocumentPipeline} = false
 
 function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
-    Utilities.log(doc, "setting up build directory.")
+    @info "Documenter: setting up build directory."
```
mortenpi: Hmm, now these lines are even more noisy.

```
[ Info: Documenter: setting up build directory.
[ Info: Documenter: expanding markdown templates.
[ Info: Documenter: building cross-references.
```

src/Documenter.jl
fredrikekre:
```
┌ Debug: Deployment criteria:
│ - ENV["TRAVIS_REPO_SLUG"]="JuliaLang/julia" occurs in repo="JuliaLang/julia" ✔
│ - ENV["TRAVIS_PULL_REQUEST"]="false" is "false" ✔
│ - ENV["TRAVIS_TAG"]="v0" is (i) empty or (ii) a valid VersionNumber ✔
│ - ENV["TRAVIS_BRANCH"]="master" matches devbranch="master" ✔
│ - ENV["DOCUMENTER_KEY"] exists ✔
│ Deploying: ✔
└ @ Main REPL[32]:1
```

src/Builder.jl
fredrikekre: Yea, I'm dropping the prefix.

src/Documenter.jl
mortenpi: Output looks good, but it won't show by default, right? Maybe best to always print this?

fredrikekre: Make it an `@info`?

mortenpi: Yep, for example.

src/Documenter.jl
```diff
@@ -463,10 +439,7 @@ function deploydocs(;
             end
         end
     else
-        Utilities.log("skipping docs deployment.")
-        if get(ENV, "DOCUMENTER_DEBUG", "") != "true"
-              Utilities.log("You can set DOCUMENTER_DEBUG to \"true\" in Travis to see more information.")
-        end
+        @info "skipping docs deployment. Turn on debug output for more diagnostics (see the docs)."
```
mortenpi: How do you actually do this easily in this case?

fredrikekre: `export JULIA_DEBUG=Documenter`

src/Documenter.jl
fredrikekre: It will typically be displayed for every doc-iteration locally, but its pretty short so its probably fine.

PR #878, opened 2018-11-02
Fix logic in should_deploy.

fredrikekre:
Broken by #876

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -393,25 +393,30 @@ function deploydocs(;
         end
     end
 
+    # Check criteria for deployment
+    ## The deploydocs' repo should match TRAVIS_REPO_SLUG
     repo_ok = occursin(travis_repo_slug, repo)
+    ## Do not deploy for PRs
     pr_ok = travis_pull_request == "false"
-    tag_ok = !isempty(travis_tag) && occursin(Base.VERSION_REGEX, travis_tag)
-    branch_ok = isempty(travis_tag) && travis_branch == devbranch
+    ## If a tag exist it should be a valid VersionNumber
+    tag_ok = isempty(travis_tag) || occursin(Base.VERSION_REGEX, travis_tag)
+    ## If no tag exists deploydocs' devbranch should match TRAVIS_BRANCH
+    branch_ok = !isempty(travis_tag) || travis_branch == devbranch
+    ## DOCUMENTER_KEY should exist
     key_ok = !isempty(documenter_key)
     should_deploy = repo_ok && pr_ok && tag_ok && branch_ok && key_ok
 
     marker(x) = x ? "✔" : "✘"
     @info """Deployment criteria:
-    - ENV["TRAVIS_REPO_SLUG"]="$(travis_repo_slug)" occurs in repo="$(repo)" $(marker(repo_ok))
-    - ENV["TRAVIS_PULL_REQUEST"]="$(travis_pull_request)" is "false" $(marker(pr_ok))
-    - ENV["TRAVIS_TAG"]="$(travis_tag)" is (i) empty or (ii) a valid VersionNumber $(marker(tag_ok))
-    - ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch)" $(marker(branch_ok))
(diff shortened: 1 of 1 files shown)


src/Documenter.jl
```diff
@@ -395,8 +395,8 @@ function deploydocs(;
 
     repo_ok = occursin(travis_repo_slug, repo)
     pr_ok = travis_pull_request == "false"
-    tag_ok = !isempty(travis_tag) && occursin(Base.VERSION_REGEX, travis_tag)
-    branch_ok = isempty(travis_tag) && travis_branch == devbranch
+    tag_ok = isempty(travis_tag) || occursin(Base.VERSION_REGEX, travis_tag)
```
fredrikekre: This should be ok if (i) the tag is empty or (ii) the tag is a valid version.

src/Documenter.jl
```diff
@@ -395,8 +395,8 @@ function deploydocs(;
...
+    branch_ok = !isempty(travis_tag) || travis_branch == devbranch
```
fredrikekre: And this should be true if (i) the tag is not empty (and thus checked for validity on the line above) or (ii) the branches match.

src/Documenter.jl
```diff
@@ -405,13 +405,12 @@ function deploydocs(;
...
     - ENV["TRAVIS_PULL_REQUEST"]="$(travis_pull_request)" is "false" $(marker(pr_ok))
     - ENV["TRAVIS_TAG"]="$(travis_tag)" is (i) empty or (ii) a valid VersionNumber $(marker(tag_ok))
-    - ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch)" $(marker(branch_ok))
+    - ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch) (if tag is empty) $(marker(branch_ok))
     - ENV["DOCUMENTER_KEY"] exists $(marker(key_ok))
     Deploying: $(marker(should_deploy))
     """
 
     if should_deploy
-        @info "deploying docs."
```
fredrikekre: Removing these since we display this on L410.

src/Documenter.jl
```diff
@@ -395,8 +395,8 @@ function deploydocs(;
```
mortenpi: Would you be willing, by any chance, to write a comment here with the logic of how we interpret the Travis variables? I think it would make it easier in the future to understand what these checks do exactly, without having to reverse engineer the logic again.

src/Documenter.jl
```diff
@@ -405,13 +405,12 @@ function deploydocs(;
     - ENV["TRAVIS_REPO_SLUG"]="$(travis_repo_slug)" occurs in repo="$(repo)" $(marker(repo_ok))
     - ENV["TRAVIS_PULL_REQUEST"]="$(travis_pull_request)" is "false" $(marker(pr_ok))
     - ENV["TRAVIS_TAG"]="$(travis_tag)" is (i) empty or (ii) a valid VersionNumber $(marker(tag_ok))
-    - ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch)" $(marker(branch_ok))
+    - ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch) (if tag is empty) $(marker(branch_ok))
     - ENV["DOCUMENTER_KEY"] exists $(marker(key_ok))
     Deploying: $(marker(should_deploy))
```
mortenpi: Completely unnecessary bikeshedding thing -- I started wondering if it might be better to have the markers before the string, e.g.

```
┌ Debug: Deployment criteria:
│ - ✔ ENV["TRAVIS_REPO_SLUG"]="JuliaLang/julia" occurs in repo="JuliaLang/julia"
│ - ✔ ENV["TRAVIS_PULL_REQUEST"]="false" is "false"
│ - ✔ ENV["TRAVIS_TAG"]="v0" is (i) empty or (ii) a valid VersionNumber
│ - ✔ ENV["TRAVIS_BRANCH"]="master" matches devbranch="master"
│ - ✔ ENV["DOCUMENTER_KEY"] exists
│ Deploying: ✔
└ @ Main REPL[32]:1
```

PR #879, opened 2018-11-02
Add support for FTP in linkcheck, disable other protocols

### LINKED ISSUE #873: `linkcheck` checks non-HTTP URLs but can only verify HTTP URLs
```julia
doc = Documenter.Documents.Document(; linkcheck=true)
link = md"[IANA](ftp://ftp.iana.org/tz/data/etcetera)".content[1].content[1]
Documenter.DocChecks.linkcheck(link, doc)
```
gives
```
 !! invalid result returned by `curl -sI ftp://ftp.iana.org/tz/data/etcetera`:

Last-Modified: Sat, 27 Oct 2018 12:10:09 GMT
Content-Length: 2790
Accept-ranges: bytes

false
```
This is a valid FTP response. I think `linkcheck` should either not handle non-HTTP(S) URLs, support FTP, or use some other measure of success.

iamed2:
Fixes #873 ?

This adds support for FTP as well as some tests, and causes unsupported protocols to fail when running curl vs at parse time.

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -146,8 +146,6 @@ footnote(other, orphans::Dict) = true
 
 hascurl() = (try; success(`curl --version`); catch err; false; end)
 
-const CURL_CMD = "curl -sI --proto =http,https,ftp,ftps"
-
 """
 $(SIGNATURES)
 
@@ -182,13 +180,15 @@ function linkcheck(link::Markdown.Link, doc::Documents.Document)
     end
 
     if !haskey(doc.internal.locallinks, link)
+        cmd = `curl -sI --proto =http,https,ftp,ftps $(link.url) --max-time 10`
+
         local result
         try
             # interpolating into backticks escapes spaces so constructing a Cmd is necessary
-            result = read(Cmd(String[split(CURL_CMD)..., link.url, "--max-time", "10"]), String)
+            result = read(cmd, String)
         catch err
             push!(doc.internal.errors, :linkcheck)
-            @warn "`$CURL_CMD $(link.url)` failed:" exception = err
+            @warn "$cmd failed:" exception = err
             return false
         end
         HTTP_STATUS_REGEX   = r"^HTTP/(1.1|2) (\d+) (.+)$"m
@@ -214,7 +214,7 @@ function linkcheck(link::Markdown.Link, doc::Documents.Document)
             @debug "linkcheck '$(link.url)': FTP success"
         else
             push!(doc.internal.errors, :linkcheck)
-            @warn "invalid result returned by `$CURL_CMD $(link.url)`:" result
+            @warn "invalid result returned by $cmd:" result
         end
     end
     return false
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -184,7 +184,8 @@ function linkcheck(link::Markdown.Link, doc::Documents.Document)
     if !haskey(doc.internal.locallinks, link)
         local result
         try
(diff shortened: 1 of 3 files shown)


src/DocChecks.jl
```diff
@@ -205,9 +208,12 @@ function linkcheck(link::Markdown.Link, doc::Documents.Document)
                 push!(doc.internal.errors, :linkcheck)
                 @error "linkcheck '$(link.url)' status: $(status)."
             end
+        elseif occursin(FTP_STATUS_REGEX, result)
+            # this regex is matched iff success
+            printstyled(INDENT, "FTP SUCCESS ", link.url, "\n", color=:green)
```
mortenpi: `INDENT` is missing. Also, it should be consistent with the `HTTP` success and use an `@debug`.

src/DocChecks.jl
```diff
@@ -182,15 +184,16 @@ function linkcheck(link::Markdown.Link, doc::Documents.Document)
     if !haskey(doc.internal.locallinks, link)
         local result
         try
-            result = read(`curl -sI $(link.url) --max-time 10`, String)
+            result = read(`curl -sI --proto =http,https,ftp,ftps $(link.url) --max-time 10`, String)
```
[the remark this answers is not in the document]
iamed2: Unfortunately interpolating into backticks escapes spaces, so it would try to call a program named "curl -sI --proto =http,https,ftp,ftps". I could do this instead:
```julia
Cmd(String[split(CURL_CMD)..., link.url, "--max-time", "10"])
```
which is more verbose but has no duplication. What do you think?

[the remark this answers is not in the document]
mortenpi: Good point, I didn't realize that. How about just locally setting a variable to the full `Cmd` object (``curl_cmd = `curl -sI --proto =http,https,ftp,ftps $(link.url) --max-time 10` ``) and then interpolating the `curl_cmd` into the error strings? Either way is fine though.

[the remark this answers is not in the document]
iamed2: That's much nicer, I've done that now

src/DocChecks.jl
```diff
@@ -180,17 +180,21 @@ function linkcheck(link::Markdown.Link, doc::Documents.Document)
     end
 
     if !haskey(doc.internal.locallinks, link)
+        cmd = `curl -sI --proto =http,https,ftp,ftps $(link.url) --max-time 10`
```
mortenpi: Is the space between `--proto` and `=` correct / necessary? Interestingly, it doesn't seem to affect the command.

iamed2: It is necessary, as it denies all other protocols. 

```
       --proto <protocols>
              Tells curl to limit what protocols it may use in the transfer. Protocols are evaluated left  to  right,  are  comma
              separated, and are each a protocol name or

              +  Permit this protocol in addition to protocols already permitted (this is the default if no modifier is used).

              -  Deny this protocol, removing it from the list of protocols already permitted.

              =  Permit  only this protocol (ignoring the list already permitted), though subject to later modification by subse-
                 quent entries in the comma separated list.

              For example:

              --proto -ftps  uses the default protocols, but disables ftps

              --proto -all,https,+http
                             only enables http and https

              --proto =http,https
                             also only enables http and https

       Unknown protocols produce a warning. This allows scripts to safely rely on being able  to  disable  potentially  dangerous
       protocols, without relying upon support for that protocol being built into curl to avoid an error.

       This  option  can  be  used  multiple  times, in which case the effect is the same as concatenating the protocols into one
       instance of the option.
```

mortenpi: Ah, ok, makes sense. Thanks!

PR #885, opened 2018-11-08
added filter function to @autodocs

timkittel:
Hi, I have a simple implementation of adding a filter function to `@autodocs` via a `Filter` keyword (as shortly discussed in [slack here and the corresponding thread](https://julialang.slack.com/archives/C66L48G1X/p1541515077072700)). See [here](https://github.com/timkittel/Documenter-test/blob/master/docs/src/index.md) for an example `index.md` where the filtering function is created [here](https://github.com/timkittel/Documenter-test/blob/master/docs/make.jl#L20-L22). Do you generally agree with that idea?

It seems to work if I try it with this example setup https://github.com/timkittel/Documenter-test but I don't really know how I would add the tests here. Could you give me some pointers?

Of course, I would add a sction to the documentation if there is positive feedback here (:


mortenpi:
Awesome, thanks! I don't think this really needs to be more complicated than this, so the implementation looks good.

Just needs docs as you mentioned. I am not entirely sure what are the types of the objects that will get passed to the filter function though -- it would be good to document that. For functions you probably get these `Binding` objects I assume.

Does it also work if you create a lambda function directly in the at-autodocs block?

For tests, I think we should add something to `test/examples/`.

timkittel:
Great!

Concerning lambda functions: No, unfortunately not. Providing a lambda function raises one of these world age errors, that I have never really understood. I assume there is somewhere a `@generated` function involved, but that would require to dig deeper into `Documenter.jl`.

I will add a description to the docs and my example to the tests. But that will take me a moment as I don't have access to a real computer at the moment (:

fredrikekre:
> Providing a lambda function raises one of these world age errors, that I have never really understood.

Probably just need to invoke it with `invokelatest` at the callsite.

src/Expanders.jl
```diff
@@ -346,15 +350,17 @@ function Selectors.runner(::Type{AutoDocsBlocks}, x, page, doc)
...
         public = get(fields, :Public, true)
         private = get(fields, :Private, true)
+        filterfunc = get(fields, :Filter, x -> true)
         results = []
         for mod in modules
             for (binding, multidoc) in Documenter.DocSystem.getmeta(mod)
                 # Which bindings should be included?
                 isexported = Base.isexported(mod, binding.var)
                 included = (isexported && public) || (!isexported && private)
+                filtered = filterfunc(Core.eval(binding.mod, binding.var))
```
fredrikekre: `filtered = Base.invokelatest(filterfunc, Core.eval(binding.mod, binding.var))`

timkittel: thanks, that worked like a charm

timkittel:
@mortenpi , I updated PR with lambda functions now working, an example in tests and a note in the documentation. Could you check whether you find it okai?

docs/src/man/syntax.md
```diff
@@ -85,6 +85,27 @@ docstrings. Note that page matching is done using the end of the provided string
 `a.jl` will be matched by *any* source file that ends in `a.jl`, i.e. `src/a.jl` or
 `src/foo/a.jl`.
 
+To filter out certain docstrings by your own criteria, you can provide function with them
```
mortenpi: s/them/the ?

mortenpi:
Hmm, I started wondering about what gets passed as the argument to the lambda function. E.g. if you add the following into the `Filter` module

```
"foobar"
function foobar end
```

you get this error

```
ERROR: LoadError: TypeError: in <:, expected Type, got typeof(Main.AutoDocs.Filter.foobar)
```

Could we have a short docs section on what are the types that will be used to call the filtering function and how to handle them?

Also, I was hoping that `Order = [:type]` would make the error go away, but it seems that the filtering does not work together with `Order`? Or `Order` is being done after `Filter`?

timkittel:
> Hmm, I started wondering about what gets passed as the argument to the lambda function. E.g. if you add the following into the `Filter` module
> 
> ```
> "foobar"
> function foobar end
> ```
> 
> you get this error
> 
> ```
> ERROR: LoadError: TypeError: in <:, expected Type, got typeof(Main.AutoDocs.Filter.foobar)
> ```
> 
> Could we have a short docs section on what are the types that will be used to call the filtering function and how to handle them?

As far as I understand, `Documenter.jl` simply goes through the whole module and checks for everything that has a doc string inside. The filter is just run on everything, that `Documenter.jl` provides, so your case, you try to check whether the function is a sub type of an abstract type and that fails.
To avoid misunderstandings, I added a check for the type of the provided element to the filter in the docs: 
(It's also implemented [here](https://github.com/timkittel/Documenter-test/blob/master/docs/src/index.md), if you simply want to check it quickly.)


And the description in the CHANGELOG.md sounds very good to me, thanks!

> Also, I was hoping that `Order = [:type]` would make the error go away, but it seems that the filtering does not work together with `Order`? Or `Order` is being done after `Filter`?

When trying that, I am stumbling over a an error that seems to be due to the parser of the `@autodocs` block. I will look into that and tell you when I know more ...

timkittel:
@mortenpi I modified the code so the `Filter` keyword is evaluated after the `Order` keyword. Now, providing `Order = [:type]` ensures that only types are given the the filter function. It is tested in the docs now as well, see https://github.com/JuliaDocs/Documenter.jl/pull/885/commits/8cd62b644aae2d48b0db59950d18605e13cf7fe8

That should conclude the PR?

On a side note: I would suggest to refactor the evaluation of the keywords a bit in order to avoid the for loops and if clauses which made it a bit hard for me to find the error. But that would be separate PR, I think. Still, would you generally be open for that, @mortenpi ?

mortenpi:
Yes, this is good to go. Thanks for iterating on this!

> I would suggest to refactor the evaluation of the keywords a bit in order to avoid the for loops and if clauses which made it a bit hard for me to find the error. But that would be separate PR, I think.

That sounds good to me. If you'd be willing to make a PR, that would be awesome!

PR #888, opened 2018-11-13
Replace deprecated `doc"""` syntax

tribut:
`doc"""` does no longer work in julia 0.7+. I've also added a note about the double-backticks being the official way to add math expressions to doc strings.

### DIFF
--- a/docs/src/man/latex.md
+++ b/docs/src/man/latex.md
@@ -5,31 +5,31 @@ documentation.
 
 ## Escaping Characters in Docstrings
 
-Since some characters used in ``\LaTeX`` syntax are treated differently in docstrings they
+Since some characters used in ``\LaTeX`` syntax, such as `$` and `\`, are treated differently in docstrings. They
 need to be escaped using a `\` character as in the following example:
 
 ```julia
 """
-Here's some inline maths: \$\\sqrt[n]{1 + x + x^2 + \\ldots}\$.
+Here's some inline maths: ``\\sqrt[n]{1 + x + x^2 + \\ldots}``.
 
 Here's an equation:
 
-\$\\frac{n!}{k!(n - k)!} = \\binom{n}{k}\$
+``\\frac{n!}{k!(n - k)!} = \\binom{n}{k}``
 
 This is the binomial coefficient.
 """
 func(x) = # ...
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/man/latex.md
```diff
@@ -21,10 +21,10 @@ This is the binomial coefficient.
 func(x) = # ...
 ```
 
-To avoid needing to escape the special characters the `doc""` string macro can be used:
+To avoid needing to escape the special characters the `@doc raw""` string macro can be used:
```
mortenpi: Hmm, technically they are two different macros. Saying something like "`raw""` string macro can be used, but then the docstring needs to be attached with the `@doc` macro" would be a bit more accurate.

docs/src/man/latex.md
```diff
@@ -36,6 +36,8 @@ This is the binomial coefficient.
 func(x) = # ...
 ```
 
+Do note however that the syntax using double-backticks described below is preferred over the traditional dollar-syntax for math expressions.
```
fredrikekre: Shouldn't we just update the examples in this file to use double backticks instead?

tribut: Not sure, then the whole point of the "escaping" section is moot, no? And I assume that a lot of people will still use `$...$` out of habit.

fredrikekre: No, you still need to escape the `\` in `\frac` etc.

docs/src/man/latex.md
tribut: I've updated the formulation.

docs/src/man/latex.md
```diff
@@ -21,10 +21,10 @@ This is the binomial coefficient.
 func(x) = # ...
 ```
 
-To avoid needing to escape the special characters the `doc""` string macro can be used:
+To avoid needing to escape the special characters the `raw""` string macro can be used, combined with `@raw`:
```
fredrikekre: `@raw` -> `@doc`.

docs/src/man/latex.md
tribut: True. I'll just update it, then.

docs/src/man/latex.md
tribut: Meh. Sorry.

mortenpi:
LGTM, thanks!

PR #891, opened 2018-11-23
docker for latex

fredrikekre:


fredrikekre:
Seems to be some file permission mismatch in the docker container:
`Cannot write file 'Documenter.jl.aux'`

src/Writers/LaTeXWriter.jl
```diff
@@ -83,19 +84,43 @@ function render(doc::Documents.Document)
...
+            str = replace(str, "{Writers}\n\\end{itemize}\n\\end{itemize}" =>
+                               "{Writers}\n\\end{itemize}")
+            write(file, str)
+
+            texcmd = `latexmk -f -interaction=nonstopmode -view=none -lualatex -shell-escape $file`
+
+            docker = Sys.which("docker") !== nothing
+            latexmk = Sys.which("latexmk") !== nothing
+            try
+                if docker
```
mortenpi: I am not sure if we should default to Docker when both are available. Also, I think it would be better if the "compiler backend" would be configured explicitly, rather then deciding which one to use automagically.

.travis.yml
```diff
@@ -9,12 +9,14 @@ julia:
   - 1.0
   - nightly
 
+services: docker
```
mortenpi: It feels like it would be better to have all the PDF-related stuff living in DocumenteLaTeX, including testing the Docker image. But with the writer living here currently, we can keep the testing here until it's moved.

fredrikekre:
Ok. This works now.

src/Writers/LaTeXWriter.jl
```diff
@@ -83,23 +84,64 @@ function render(doc::Documents.Document)
...
+            status = compile_tex(texfile)
+            status && cp(pdffile, joinpath(doc.user.root, doc.user.build, pdffile); force = true)
         end
     end
 end
 
+const DOCKER_IMAGE_TAG = "0.1"
+
+function compile_tex(texfile)
+    engine = get(ENV, "DOCUMENTER_LATEX_ENGINE", "latexmk") # TODO: make this configurable from makedocs
```
mortenpi: Having a `Documenter.LaTeX` plugin for this would be good. For simplicity, we could also implement it like `Documenter.HTML` for now, and change to `format = Documenter.HTML(...)` or `writer = Documenter.HTML(...)` later.

mortenpi:
I tried running the latexmk backend locally with `julia test/formats/latex.jl`, and I am getting the following:

```
[ Info: LaTeXWriter: using latexmk to compile tex.
┌ Error: LaTeXWriter: failed to compile tex with latexmk. Logs and partial output can be found in `/tmp/tmpv14bh5`.
│   exception = UndefVarError: file not defined
└ @ Documenter.Writers.LaTeXWriter ~/Julia/JuliaDocs/Documenter/src/Writers/LaTeXWriter.jl:107
Test Summary: | Pass  Total
LaTeX         |    1      1
```

The PDF looks fine though.

I am also not seeing any `latexmk` logs anywhere though. I see that they should just be going to stdout/-err?

Also, interestingly, it seems that `HTMLWriter` gets called too, because I end up with the `.html` files in the temporary directory too:

```
$ ls -Alh /tmp/tmpv14bh5/
total 516K
drwxrwxr-x  2 mortenpi mortenpi 4.0K Dec  2 20:53 assets
drwxrwxr-x  2 mortenpi mortenpi 4.0K Dec  2 20:53 contributing
-rw-rw-r--  1 mortenpi mortenpi    0 Dec  2 20:53 custom.sty
-rw-rw-r--  1 mortenpi mortenpi 251K Dec  2 20:53 Documenter.jl.pdf
-rw-rw-r--  1 mortenpi mortenpi  78K Dec  2 20:53 Documenter.jl.tex
-rw-rw-r--  1 mortenpi mortenpi 1.5K Dec  2 20:53 documenter.sty
-rw-rw-r--  1 mortenpi mortenpi 7.1K Dec  2 20:53 index.html
drwxrwxr-x  4 mortenpi mortenpi 4.0K Dec  2 20:53 lib
drwxrwxr-x 10 mortenpi mortenpi 4.0K Dec  2 20:53 man
drwxrwxr-x  2 mortenpi mortenpi 4.0K Dec  2 20:53 search
-rw-rw-r--  1 mortenpi mortenpi 152K Dec  2 20:53 search_index.js
```

I'll try to debug further tomorrow.

assets/latex/documenter.sty
```diff
@@ -75,3 +75,6 @@
 \setheaderspaces{1in}{*}{*}
 \checkandfixthelayout
 %
+
+% images etc.
+\usepackage{graphicx}
```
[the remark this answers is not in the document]
fredrikekre: Sure, will split out that.

fredrikekre:
> Also, interestingly, it seems that HTMLWriter gets called too

Maybe you had some old files around that got copied? This PR does not touch the dispatching code at all so...

src/Writers/LaTeXWriter.jl
```diff
@@ -53,12 +60,13 @@ const DOCUMENT_STRUCTURE = (
     "subparagraph",
 )
 
-function render(doc::Documents.Document)
+function render(doc::Documents.Document, fmt::LaTeX)
```
mortenpi: Currently, this needs to accept `Symbol` too, since `format` does not get replaced. It might be best to immediately in `Document()` to convert `:html` and `:latex` into the corresponding types. Together with a deprecation warning?

src/Writers/LaTeXWriter.jl
```diff
@@ -83,23 +91,65 @@ function render(doc::Documents.Document)
...
+
+function compile_tex(doc, fmt, texfile)
+    # engine = get(ENV, "DOCUMENTER_LATEX_ENGINE", "docker") # TODO: make this configurable from makedocs
+    engine = get(fmt.info, "engine", "docker")
+    if engine == "latexmk"
+        Sys.which("latexmk") === nothing && (@error "LaTeXWriter: latexmk command not found."; return false)
+        @info "LaTeXWriter: using latexmk to compile tex."
+        try
+            piperun(`latexmk -f -interaction=nonstopmode -view=none -lualatex -shell-escape $texfile`)
+            return true
```
mortenpi: With latexmk, the PDF is not getting copied back to `build/` at the moment, right?

mortenpi:
> Maybe you had some old files around that got copied? This PR does not touch the dispatching code at all so...

Right, my bad. My `docs/build/` was dirty and forgot that we copy that over to `/tmp`. With your latest updates, it seems to build fine, but `latexmk` still runs into some errors:

```
[ Info: LaTeXWriter: using latexmk to compile tex.
┌ Error: LaTeXWriter: failed to compile tex with latexmk. Logs and partial output can be found in `/tmp/tmpj3fw2C`.
│   exception = failed process: Process(`latexmk -f -interaction=nonstopmode -view=none -lualatex -shell-escape Documenter.jl.tex`, ProcessExited(12)) [12]
└ @ Documenter.Writers.LaTeXWriter ~/Julia/JuliaDocs/Documenter/src/Writers/LaTeXWriter.jl:115
Test Summary: | Pass  Total
LaTeX         |    1      1
```

And from `Documenter.log`:

```
! LaTeX Error: Something's wrong--perhaps a missing \item.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.139 \end{itemize}
```

I've put the full output here: https://gist.github.com/mortenpi/d103b0e4bbb19b6ef5a1e842e81e0b71

mortenpi:
Ah, it looks like if we have an empty index block (which happened right now, since for some reason Documenter can't find docstrings when running the `test/format/latex.jl` script), we get an empty `itemize`:

```
\subsection{Index}



A list of all internal documentation sorted by module.


\begin{itemize}
\end{itemize}


\hypertarget{3430565549354646931}{}
```

This (possibly stupid) patch fixes the problem:

```diff
diff --git a/src/Writers/LaTeXWriter.jl b/src/Writers/LaTeXWriter.jl
index 8253dc2c..5ef16de1 100644
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -250,6 +250,7 @@ end
 ## Index, Contents, and Eval Nodes.
 
 function latex(io::IO, index::Documents.IndexNode, page, doc)
+    length(index.elements) > 0 || return
     _println(io, "\\begin{itemize}")
     for (object, _, page, mod, cat) in index.elements
         id = string(hash(string(Utilities.slugify(object))))
@@ -263,6 +264,7 @@ function latex(io::IO, index::Documents.IndexNode, page, doc)
 end
 
 function latex(io::IO, contents::Documents.ContentsNode, page, doc)
+    length(contents.elements) > 0 || return
     depth = 1
     needs_end = false
     _println(io, "\\begin{itemize}")
```

src/Writers/LaTeXWriter.jl
fredrikekre: Why not? https://github.com/JuliaDocs/Documenter.jl/pull/891/files#diff-862e2c1cb8b709bb81be9bc357122322R97

(only if the compile was succesful though)

mortenpi: My bad, I didn't read the code carefully enough. And indeed, I had a non-successful compile.

src/Documenter.jl
```diff
@@ -176,26 +176,78 @@ ignored.
...
+    end
+    if html_canonical !== nothing
+        html_warn("html_canonical")
+        html_keywords[:canonical] = html_canonical
+    end
+    # deprecation of format as Symbols
+    function fmt(f)
+        if format === :html
+            Base.depwarn("`format = :html` is deprecated, use `format = Documenter.HTML()` instead.", :makedocs)
+            return Writers.HTMLWriter.HTML(html_keywords...)
```
mortenpi: Right now, if you set `makedocs(html_something = foo, format = Documenter.HTML())`, you would get the warning, but `html_something` would not have any effect, right? Most importantly, it affects the case where you are relying on the default value for `format`.

I think we should do the following:

* Have the default for `format` be `:html` (but keep docs as they are, this will just be an implementation detail). So, in case the user passes any of the `html_*` arguments without explicitly specifying `format`, we'll set the `Documenter.HTML` object fields appropriately.
* In the `html_*` checks above, if `format != :html` or maybe if `format isa Documenter.HTML`, print out a big warning saying that the arguments do not have any effect.

docs/make.jl
```diff
@@ -1,7 +1,14 @@
...
+using Documenter, DocumenterTools, DocumenterLaTeX, DocumenterMarkdown
 
 makedocs(
     modules = [Documenter, DocumenterTools],
+    # format = Documenter.HTML(
+    #     # Use clean URLs, unless built as a "local" build
+    #     prettyurls = !("local" in ARGS),
+    #     canonical = "https://juliadocs.github.io/Documenter.jl/stable/",
+    # ),
+    format = Documenter.Writers.LaTeXWriter.LaTeX(engine = "docker"),
```
mortenpi: Should we maybe set up a second Documentation build stage which deploys the PDFs to `gh-pages-pdf` (assuming updates to `deplydocs` are on the agenda)?

src/Documenter.jl
fredrikekre: Yea, you will still get the warnings, and if you have specified the `html_` ones they are probably the values you want.

fredrikekre:
I'm thinking that I update the CHANGELOG and then this is good to go from my side

.travis.yml
```diff
@@ -22,6 +25,13 @@ jobs:
       julia: 1.0
       os: linux
       script:
+        # html
         - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
         - julia --project=docs/ docs/make.jl
-      after_success: skip
+        # pdf
+        # - julia --project=docs/pdf/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
```
fredrikekre: Was running into https://github.com/JuliaLang/Pkg.jl/issues/913 here so yey for shooting myself in the foot.

.travis.yml
```diff
@@ -22,6 +25,13 @@ jobs:
...
+        - julia --project=docs/pdf/ -e 'using Pkg; Pkg.instantiate()'
+        - julia --project=docs/pdf/ docs/pdf/make.jl
+      after_success:
+        - julia --project=coverage/ -e 'using Pkg; Pkg.instantiate()'
```
fredrikekre: Not sure this is a good idea. Maybe just build the pdf during regular testing?

CHANGELOG.md
```diff
@@ -2,28 +2,50 @@
...
     ```
     makedocs(
-        Documenter.HTML(prettyurls = ...),
+        format = Documenter.HTML(prettyurls = ..., canonical = ...),
         ...
     )
     ```
 
-  - Packages extending Documenter can now define subtypes of `Documenter.Plugin`, which can be passed to `makedocs` as positional arguments to pass options to the extensions.
+* ![Feature][badge-feature] Packages extending Documenter can now define subtypes of `Documenter.Plugin`,
```
fredrikekre: I split up the above note on the kwargs with this new Plugin feature.

docs/src/lib/internals/formats.md
```diff
@@ -1,5 +0,0 @@
-# Formats
```
fredrikekre: The `Format` module can be completely removed in a follow up PR, I got some complaints about missing docs here so removed the doc-page for now.

src/Documents.jl
```diff
@@ -184,7 +184,7 @@ struct User
     root    :: String  # An absolute path to the root directory of the document.
     source  :: String  # Parent directory is `.root`. Where files are read from.
     build   :: String  # Parent directory is also `.root`. Where files are written to.
-    format  :: Vector{Symbol} # What format to render the final document with?
+    format  :: Vector{Any} # What format to render the final document with?
```
fredrikekre: Can probably be `Vector{Plugin}`

src/Writers/LaTeXWriter.jl
```diff
@@ -83,23 +105,63 @@ function render(doc::Documents.Document)
...
+            try; piperun(`docker stop latex-container`); catch; end
+        end
+    else
+        @error "LaTeXWriter: unrecognized engine: $(engine)"
+        return false
+    end
+end
+
+function piperun(cmd)
+    verbose = "--verbose" in ARGS
```
fredrikekre: Is this ok? running with verbose as default for the tex stuff is a pain since it results in huge amounts of output...

.travis.yml
mortenpi: I was thinking that we could have a separate build stage for this.

src/Writers/LaTeXWriter.jl
mortenpi: Maybe do it with an environment variable instead? That way you can also easily get the log on Travis.

test/runtests.jl
```diff
@@ -45,7 +45,7 @@ println("="^50)
 
     # Documenter package docs with other formats.
     include("formats/markdown.jl")
-    include("formats/latex.jl")
+    # include("formats/latex.jl")
```
mortenpi: Should be removed?

test/examples/tests.jl
```diff
@@ -46,7 +46,7 @@ end
         @test doc.user.source == "src"
         @test doc.user.build  == "builds/markdown"
         @test doc.user.clean  == true
-        @test doc.user.format == [:markdown]
+        @test doc.user.format == [Documenter.Writers.MarkdownWriter.Markdown()]
```
mortenpi: I just love how this looks. Javaland. :laughing:

.travis.yml
```diff
@@ -9,6 +9,9 @@ julia:
   - 1.0
   - nightly
 
+services:
+  - docker
```
mortenpi: Can we have the service for just the build stage?

docs/make.jl
```diff
@@ -2,6 +2,13 @@ using Documenter, DocumenterTools
 
 makedocs(
     modules = [Documenter, DocumenterTools],
+    format = Documenter.HTML(
+        # Use clean URLs, unless built as a "local" build
+        prettyurls = !("local" in ARGS),
+        canonical = "https://juliadocs.github.io/Documenter.jl/stable/",
+    ),
+    # format = LaTeX(engine = "docker"),
+    # format = Markdown(),
```
mortenpi: Remove these?

src/Documents.jl
mortenpi: Yeah, let's do that.

.travis.yml
mortenpi: If we can, let's do it. If not, let's leave LaTeX uncovered for now. Although, no deployment at the moment, right? In which case, we could indeed also just build the PDF during testing (as long as the build time does not blow up too much).

src/Writers/LaTeXWriter.jl
fredrikekre: Added `DOCUMENTER_VERBOSE`

fredrikekre:
I'd like to merge this to try it out and then fix any remainiing things in follow up PRs before the 0.21 release.

fredrikekre:
I'll have a look at coverage later.

PR #898, opened 2018-12-02
Misc fixes

fredrikekre:


assets/latex/documenter.sty
```diff
@@ -75,3 +75,7 @@
 \setheaderspaces{1in}{*}{*}
 \checkandfixthelayout
 %
+
+% images etc.
+\usepackage{graphicx}
```
mortenpi: Would be good to have this mentioned in the changelog too.

mortenpi:
> I still have this warning:
> 
> ```
> ┌ Warning: 1 docstring potentially missing:
> │ 
> │     Documenter.Documenter
> └ @ Documenter.DocChecks ~/dev/Documenter/src/DocChecks.jl:50
> ```
> but that feels like a false alarm, because that is included here https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/lib/public.md#L10 and shows up in the rendered docs.

Yeah, we've been getting this for a long time. I have been meaning to debug this, but haven't gotten around to it. It seems like something about the Julia docstring internals changed subtly at some point.

PR #904, opened 2018-12-09
LaTeX/PDF: Add version to title and filename.

fredrikekre:
Before:
oldtitle

After:
newtitle

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -76,8 +76,16 @@ function render(doc::Documents.Document, settings::LaTeX=LaTeX())
     mktempdir() do path
         cp(joinpath(doc.user.root, doc.user.build), joinpath(path, "build"))
         cd(joinpath(path, "build")) do
-            texfile = replace("$(doc.user.sitename).tex", " " => "")
-            pdffile = replace("$(doc.user.sitename).pdf", " " => "")
+            name = doc.user.sitename
+            let tag = get(ENV, "TRAVIS_TAG", "")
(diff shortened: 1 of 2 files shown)


src/Writers/LaTeXWriter.jl
```diff
@@ -76,8 +76,16 @@ function render(doc::Documents.Document, settings::LaTeX=LaTeX())
     mktempdir() do path
         cp(joinpath(doc.user.root, doc.user.build), joinpath(path, "build"))
         cd(joinpath(path, "build")) do
-            texfile = replace("$(doc.user.sitename).tex", " " => "")
-            pdffile = replace("$(doc.user.sitename).pdf", " " => "")
+            name = doc.user.sitename
+            let tag = get(ENV, "TRAVIS_TAG", "")
```
mortenpi: It would be nice if we didn't depend on the environment like this. We can keep it like this for now if you prefer, but I feel that passing an explicit argument to `LaTeX()` would be better.

fredrikekre: Maybe, but that is something that will be forgotten to be updated in 99% of cases, and then  you have a mismatch between the tag and the docs etc etc.

mortenpi: I agree that it should not be necessary be updated between versions, but I would probably have gone with `get(ENV, "TRAVIS_TAG", nothing)` as the recommended value for Travis builds.

In any case though, it would be nice if this behavior was documented, in e.g. the `LaTeX()` docstring.

PR #912, opened 2018-12-11
Remove (almost) unused internal Formats module.

fredrikekre:


src/Writers/MarkdownWriter.jl
```diff
@@ -23,9 +22,9 @@ end
 
 function render(doc::Documents.Document, settings::Markdown=Markdown())
     copy_assets(doc)
-    mime = Formats.mimetype(:markdown)
+    mime = MIME"text/plain"()
     for (src, page) in doc.internal.pages
-        open(Formats.extension(:markdown, page.build), "w") do io
+        open(string(splitext(page.build)[1], ".md"), "w") do io
```
mortenpi: It would keep the code a bit more readable if we would keep the `string(splitext(page.build)[1], ".md")` part in a small separate function. Especially since the same pattern is used in multiple places in this file. We'd only need a version of `Formats.extension` specialized for `.md` though, so :+1: for getting rid of `Formats` and the functions therein.

src/Writers/HTMLWriter.jl
```diff
@@ -930,7 +929,8 @@ function get_url(ctx, path::AbstractString)
         end
         isempty(d) ? "index.html" : "$d/index.html"
     else
-        Formats.extension(:html, path)
+        # change extension to .html
+        string(splitext(path)[1], ".html")
```
mortenpi: The other comment does not apply to this as the pattern is only used in this one place in `HTMLWriter`.

fredrikekre: What other comment?

fredrikekre: Oh, the review comment.

PR #917, opened 2018-12-20
Don't deploy documentation on cron jobs

christopher-dG:
It's not really useful to deploy the same commit repeatedly. 
Docs on the env variable can be found [here](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) (`TRAVIS_EVENT_TYPE`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Documenter.jl changelog
 
+## Version `v0.22.0`
+
+* ![Enhancement][badge-enhancement] Documentation is no longer deployed on Travis CI cron jobs. ([#917][github-917])
+
 ## Version `v0.21.0`
 
 * ![Deprecation][badge-deprecation] ![Enhancement][badge-enhancement] The symbol values to the `format` argument of `makedocs` (`:html`, `:markdown`, `:latex`) have been deprecated in favor of the `Documenter.HTML`, `Markdown` and `LaTeX`
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -463,6 +463,7 @@ function deploydocs(;
     travis_pull_request = get(ENV, "TRAVIS_PULL_REQUEST",  "")
     travis_repo_slug    = get(ENV, "TRAVIS_REPO_SLUG",     "")
     travis_tag          = get(ENV, "TRAVIS_TAG",           "")
+    travis_event_type   = get(ENV, "TRAVIS_EVENT_TYPE",    "")
 
 
     # Other variables.
@@ -492,7 +493,9 @@ function deploydocs(;
     branch_ok = !isempty(travis_tag) || travis_branch == devbranch
     ## DOCUMENTER_KEY should exist
     key_ok = !isempty(documenter_key)
-    should_deploy = repo_ok && pr_ok && tag_ok && branch_ok && key_ok
+    ## Cron jobs should not deploy
+    type_ok = travis_event_type != "cron"
+    should_deploy = repo_ok && pr_ok && tag_ok && branch_ok && key_ok && type_ok
 
     marker(x) = x ? "✔" : "✘"
     @info """Deployment criteria:
@@ -501,6 +504,7 @@ function deploydocs(;
     - $(marker(tag_ok)) ENV["TRAVIS_TAG"]="$(travis_tag)" is (i) empty or (ii) a valid VersionNumber
     - $(marker(branch_ok)) ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch) (if tag is empty)
     - $(marker(key_ok)) ENV["DOCUMENTER_KEY"] exists
+    - $(marker(type_ok)) ENV["TRAVIS_EVENT_TYPE"]="$(travis_event_type)" is not "cron"
     Deploying: $(marker(should_deploy))
     """
 


mortenpi:
LGTM, makes sense to have this. Would you mind adding a note about this to the CHANGELOG under a `v0.22-dev` heading?

Ideally, we would have the output be more deterministic. I think the sorting of methods / docstrings is a bit stochastic sometimes?

mortenpi:
Thanks!

dehann:
Hi, could you please point me to the right approach on how to deploy docs from Travis?  Since v0.22.x it [looks like the "cron" on master branch for deployment is no longer a viable way to do it](https://github.com/JuliaRobotics/Caesar.jl/blob/93d12ff8cbc9ed2e327eeaf110e9c4bb2c1b4466/.travis.yml#L30):
```julia
┌ Info: Deployment criteria:
│ - ✔ ENV["TRAVIS_REPO_SLUG"]="JuliaRobotics/Caesar.jl" occurs in repo="github.com/JuliaRobotics/Caesar.jl.git"
│ - ✔ ENV["TRAVIS_PULL_REQUEST"]="false" is "false"
│ - ✔ ENV["TRAVIS_TAG"]="" is (i) empty or (ii) a valid VersionNumber
│ - ✔ ENV["TRAVIS_BRANCH"]="master" matches devbranch="master" (if tag is empty)
│ - ✔ ENV["DOCUMENTER_KEY"] exists
│ - ✘ ENV["TRAVIS_EVENT_TYPE"]="cron" is not "cron"
└ Deploying: ✘
The command "julia --project=docs/ docs/make.jl" exited with 0.
```
I have looked at `Documenter.jl/.travis.yml` and the docs, but do not see where else you would set it up?

Reference:
- https://github.com/JuliaDocs/Documenter.jl/blob/f09537d3f9c0e34c5e120f3a1b228a516fbe8610/.travis.yml#L21
- https://github.com/JuliaDocs/Documenter.jl/blob/f09537d3f9c0e34c5e120f3a1b228a516fbe8610/docs/make.jl#L54 

Thanks in advance!

christopher-dG:
I took a look at your Travis log and it's really confusing.. It looks like your jobs are erroneously being marked as cron (i.e. a Travis bug). Maybe send their support an email to inquire.

christopher-dG:
Unless you indeed have some weird cron setup that runs only once on PR merge. Do you have any Travis settings tweaked from the defaults?

dehann:
Thanks for taking a look.

> It looks like your jobs are erroneously being marked as cron (i.e. a Travis bug).

The log is standard Travis `cron` job running once a week.  I would expect  `ENV["TRAVIS_EVENT_TYPE"]="cron"` to be correct behavior.  How else should docs be deploying, if not on cron builds?

> Unless you indeed have some weird cron setup that runs only once on PR merge.

FYI, I am not running tests on master commit, and maybe that is why?  My thinking a while back was test on PR and weekly "cron" builds only, thereby reducing Travis load.

> Do you have any Travis settings tweaked from the defaults?

I don't think so.  "Build pushed branches" is turned OFF.  "Build pushed pull requests" is ON.

dehann:
Will quickly try with  "Build pushed branches" ON.

mortenpi:
Yep, the deployment is meant to happen on "push to `master`" builds. However, if you do need deployment on cron builds, I'd happily take a PR that would allow for this to be configured in `deploydocs`.

christopher-dG:
In general, you should definitely be building on pushes to master ([here](https://bors.tech/essay/2017/02/02/pitch/) is some justification)). 

If you want to keep your existing CI setup, you can also manually set the environment variable `TRAVIS_EVENT_TYPE` to be something other than "cron" by updating the `julia` command in your documentation stage to be something like `TRAVIS_EVENT_TYPE="" julia -e ...`.

dehann:
That was it, thanks!
```julia
┌ Info: Deployment criteria:
│ - ✔ ENV["TRAVIS_REPO_SLUG"]="JuliaRobotics/Caesar.jl" occurs in repo="github.com/JuliaRobotics/Caesar.jl.git"
│ - ✔ ENV["TRAVIS_PULL_REQUEST"]="false" is "false"
│ - ✔ ENV["TRAVIS_TAG"]="" is (i) empty or (ii) a valid VersionNumber
│ - ✔ ENV["TRAVIS_BRANCH"]="master" matches devbranch="master" (if tag is empty)
│ - ✔ ENV["DOCUMENTER_KEY"] exists
│ - ✔ ENV["TRAVIS_EVENT_TYPE"]="push" is not "cron"
└ Deploying: ✔
Initialized empty Git repository in /tmp/tmp50WrnT/.git/
...
```
Think I initially went down the wrong path due to
```julia
│ - ✘ ENV["TRAVIS_EVENT_TYPE"]="cron" is not "cron"
```
rather than something like `ENV["TRAVIS_EVENT_TYPE"]="cron", will only deploy on "push" (master)`.  Either way not serious, was still able to find it.

Thanks again for the help!

christopher-dG:
I can definitely see how the notation might be confusing. I wonder if changing the order would be clearer:
```
- ENV["TRAVIS_EVENT_TYPE"]="cron" is not "cron":  ✘
```

dehann:
In my case, I don't think swapping the order would have worked for me.  I started searching for issues and hadn't found one yet -- hope this thread helps in the future.

I looked at the code a little and didn't want to add too many if statements for just that one line.  That would change the behavior for different cases of "is not cron" in different cases.  I think this issue discussion should be enough.  If this happens again and more people comment for "cron is not cron" to be little clearer, then perhaps worth it...

PR #919, opened 2018-12-22
Document env. variables affecting deploydocs

mortenpi:
Also contains docs for #917, so should be merged after that.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -419,7 +419,7 @@ of which are specific to the [Travis CI platform](https://travis-ci.com/).
  - **`TRAVIS_PULL_REQUEST`**: must be set to `false`.
 
    This avoids deployment on pull request builds. Note that there is no way to _safely_
-   enable builds on pull requests, since that would expose the SSH private key.
+   enable builds on pull requests, since that would expose the SSH private key
    (`DOCUMENTER_KEY`), giving anyone opening a pull request full write access to the repository.
 
  - **`TRAVIS_REPO_SLUG`**: must match the value of the `repo` keyword.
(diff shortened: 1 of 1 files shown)


src/Documenter.jl
```diff
@@ -409,6 +409,43 @@ the generated html. The following entries are valied in the `versions` vector:
...
+
+[`deploydocs`](@ref)'s behavior is influenced by the following environment variables, many
+of which are specific to the [Travis CI platform](https://travis-ci.com/).
+
+ - **`DOCUMENTER_KEY`**: must contain the Base64-encoded SSH private key for the repository.
+
+ - **`TRAVIS_PULL_REQUEST`**: must be set to `false`.
+
+   This avoids deployment on pull request builds. Note that there is no way to _safely_
+   enable builds on pull requests, since that would expose the SSH private key.
```
fredrikekre:
```suggestion
   enable builds on pull requests, since that would expose the SSH private key,
```

src/Documenter.jl
```diff
@@ -409,6 +409,43 @@ the generated html. The following entries are valied in the `versions` vector:
...
+
+ - **`TRAVIS_REPO_SLUG`**: must match the value of the `repo` keyword.
+
+ - **`TRAVIS_EVENT_TYPE`**: may not be set to `cron`.
+
+   This avoids the re-deployment of existing docs on builds that were triggered by a Travis
+   cron job.
+
+ - **`TRAVIS_BRANCH`**: unless `TRAVIS_TAG` is non-empty, this must have the same value as the
+   `devbranch` keyword
```
fredrikekre:
```suggestion
   `devbranch` keyword.
```

src/Documenter.jl
```diff
@@ -409,6 +409,43 @@ the generated html. The following entries are valied in the `versions` vector:
...
+
+   This makes sure that only the development branch (commonly, the `master` branch) will deploy
+   the "dev" documentation (deployed into a directory specified by the `devurl` keyword).
+
+ - **`TRAVIS_TAG`**: if set, a tagged version deployment is performed instead; the value must be
+   a valid version number (i.e. match `Base.VERSION_REGEX`)
```
fredrikekre:
```suggestion
   a valid version number (i.e. match `Base.VERSION_REGEX`).
```

src/Documenter.jl
mortenpi:
```suggestion
   enable builds on pull requests, since that would expose the SSH private key
```

mortenpi:
Thanks, @fredrikekre!

PR #926, opened 2019-01-23
Setting ans to a expression in REPLBlocks. 

### LINKED ISSUE #923: expressions in repl blocks are evaled
It seems like any `Expr` that's returned in a `@repl` block is evaled before it's printed.  This doesn't happen in an `@example` block though.  Here's an MWE:

# markdown input
````
```@repl 1
a = 1
:(a + 1)
```

```@example 1
:(a + 1)
```
````

# markdown output

````

```julia-repl
julia> a = 1
1

julia> :(a + 1)
2
```


```julia
:(a + 1)
```

```
:(a + 1)
```
````

fredrikekre:
Fix #923

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 * ![Enhancement][badge-enhancement] Documentation is no longer deployed on Travis CI cron jobs. ([#917][github-917])
 
+* ![Bugfix][badge-bugfix] `@repl` blocks now work correctly together with quoted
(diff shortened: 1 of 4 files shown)


mortenpi:
Seems to work, so LGTM. It might be good to test against Base doctests before the release, just in case.

Could use a CHANGELOG entry, but were thinking about adding them later all at once?

mortenpi:
Also, it might be good to drop the previously failing example into `test/examples`.

fredrikekre:
Done.

Issue #474, opened 2017-05-02
Completely fails if a function doesn't have a comment


essenciary:
If I accidentally add a function name within a `@docs` block and the function does not have a comment at all, everything fails and no documentation is generated for any of the functions. In my opinion it should output a warning to standard output and carry on.

mortenpi:
Warnings are easy to miss I'd say, especially in Travis builds. Names without docstrings cannot at all be spliced into the documentation, so having a reference to such a function is a pretty big error. So to me failing to build feels appropriate here and I don't see a big problem in having to clean up the docs after removing a docstring from a name.

essenciary:
Sure, but then what about providing a helpful error message? 

In my case it was a pretty bad first experience. It was my first time using the package in a fairly large and complex module with multiple includes. Nothing happened so I didn't know what expect... was Documenter broken, did I encounter a bug or limitation? In the end I went over the functions list, eliminating them one by one, to see what gives - and after a lot of trial and error I figured out which function had which problem.

essenciary:
Let me run some more tests and get back to you with more details. 

Till then, I just wanted to say that the package is amazing, in terms of features and flexibility. I'm now setting up doc hosting for my packages and I love it. Truly a gem in the Julia ecosystem! 👏

PR #928, opened 2019-01-23
Fix incomplete at-docs block

### LINKED ISSUE #799: All function docstrings are not rendered when one is failing in @docs macro
Hello,

I have a .md with

    ```@docs
    MyObj
    init
    send
    close
    ```

when compiling doc I get:

```
julia --color=yes make.jl
 !! Undefined binding 'init'. [src/index.md]
 > checking for missing docstrings.
 > running doctests.
 > checking footnote links.
Documenter: populating indices.
Documenter: skipping docs deployment.
  You can set DOCUMENTER_DEBUG to "true" in Travis to see more information.
```

and all function docstrings are not rendered (just because of one missing docstring: `init`)

It will be great (if possible) to have others functions docstring correctly rendered (ie here `MyObj`, `send`, `close`)

and  `init` docstring should be rendered with a red text showing that something is wrong here.

Thanks for Documenter.jl that's a great project

Kind regards

fredrikekre:
Include working doc strings instead of bailing completely in at-docs block.

Fix #799

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
   `@eval`, `@example` and `@setup` blocks now include information about the source location
   of the block. ([#929][github-929])
 
+* ![Enhancement][badge-enhancement] Docstrings from `@docs`-blocks are now included in the
+  rendered docs even if some part(s) of the block failed. ([#928][github-928])
+
 ## Version `v0.21.1`
 
 * ![Bugfix][badge-bugfix] `@repl` blocks now work correctly together with quoted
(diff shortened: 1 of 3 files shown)


mortenpi:
Big :+1: to including the docstrings that are possible to include.

However, now there is nothing about the failed ones in the generated output, which I think makes them easy to miss (unless you do strict builds). So I would still output something for the failed docstrings too. Could we maybe just push a generated `!!! warn` adminition into the doctree? Or maybe keep a list with the `DocsNodes` and have the writer interpret it?

fredrikekre:
also fixes #474

fredrikekre:
Now results in:
screenshot from 2019-01-28 14-45-41

from
````
```@docs
TestPackage.foo
TestPackage.baz
TestPackage.bar
```
````
when `TestPackage.baz` is missing.

fredrikekre:
> Have you tested it with all the output formats, to make sure all handle the updated DocsNodes correctly?

Yes.

> It might be good to have a test case in test/examples too (although, we might inadvertently already do).

Exists at the top of https://github.com/JuliaDocs/Documenter.jl/blob/91fc140e7adef2eb17b85d8a2848e805877062ba/test/errors/src/index.md.

Issue #78, opened 2016-06-01
Backtrace Information


ChrisRackauckas:
Currently the backtrace information does not tell you where the error occurred. It should tell the file and the line in order to be useful.

![error](https://cloud.githubusercontent.com/assets/1814174/15728705/1ae3d716-2813-11e6-9798-f32a1f10d02b.PNG)

MichaelHatherly:
Better info here would indeed be good. The markdown parser doesn't track line/column info currently though, so would probably need us to search through the original file contents in Documenter itself.

PR #929, opened 2019-01-23
Better log messages

fredrikekre:
Better log messages for failed at-meta, at-docs, at-autodocs, at-eval, at-example and at-setup blocks.


mortenpi:
Fixes #78?

fredrikekre:
> Fixes #78?

Yea, looks like it!

PR #934, opened 2019-01-31
Make linkcheck work on servers without HEAD, and add FTP status checking

iamed2:
`curl -I` uses `HEAD` requests, which don't work on all servers. Links really only need `GET`, so we can fallback to that if `HEAD` fails. 

I also added support for FTP status checking, once I found a way to get that information out. I needed to suppress the regular output of `curl`, which required redirecting to a file through `curl -o` (so I needed to use `/dev/null` instead of Julia's `devnull`). 

This ended up making the regex matching code simpler.

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -181,7 +181,7 @@ function linkcheck(link::Markdown.Link, doc::Documents.Document; method::Symbol=
 
     if !haskey(doc.internal.locallinks, link)
         null_file = @static Sys.iswindows() ? "nul" : "/dev/null"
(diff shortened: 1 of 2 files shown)


mortenpi:
It seems that `curl` on Travis is too old for this, hence the errors. It should be [7.35.0 in Ubuntu 14.04](https://packages.ubuntu.com/trusty-updates/curl), however, from the `curl` man page on `--write-out`:

```
scheme         The URL scheme (sometimes called protocol) that was effectively used (Added in 7.52.0)
```

Is there a way to avoid using `%{scheme}`?

iamed2:
> Is there a way to avoid using `%{scheme}`?

Likely; I'll figure something out

mortenpi:
Sorry for the delay, LGTM. Thanks! Will aim to backport and tag this shortly.

PR #938, opened 2019-02-04
Fix non-text/plain output for Markdown and LaTeX

### LINKED ISSUE #916: Output capturing for Markdown and LaTeX
This https://github.com/JuliaDocs/Documenter.jl/blob/1d06e35cca723573feda04f73de2eefe5fb208e3/src/Expanders.jl#L510-L525 is broken for `DocumenterMarkdown.Markdown()` and `DocumenterLaTeX.LaTeX()` output, see e.g. https://github.com/fredrikekre/Literate.jl/issues/49

We should evaluate all possible outputs, store them, and let the renderer choose what representation to include in the final document.

fredrikekre:
Fix non-text/plain output for Markdown
and LaTeX output by capturing several
output representations and let each writer
choose which one to render,
fixes  #916, fixes fredrikekre/Literate.jl#49.

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -316,6 +316,39 @@ function latex(io::IO, node::Documents.EvalNode, page, doc)
     node.result === nothing ? nothing : latex(io, node.result, page, doc)
 end
 
+# Select the "best" representation for LaTeX output.
+using Base64: base64decode
+function latex(io::IO, mo::Documents.MultiOutput)
+    foreach(x->Base.invokelatest(latex, io, x), mo.content)
+end
+function latex(io::IO, d::Dict{MIME,Any})
+    filename = String(rand('a':'z', 7))
+    if haskey(d, MIME"image/png"())
+        write("$(filename).png", base64decode(d[MIME"image/png"()]))
+        _println(io, """
+        \\begin{figure}[H]
+        \\centering
+        \\includegraphics{$(filename)}
+        \\end{figure}
+        """)
+    elseif haskey(d, MIME"image/jpeg"())
+        write("$(filename).jpeg", base64decode(d[MIME"image/jpeg"()]))
+        _println(io, """
+        \\begin{figure}[H]
+        \\centering
+        \\includegraphics{$(filename)}
+        \\end{figure}
+        """)
+    elseif haskey(d, MIME"text/latex"())
+        latex(io, Utilities.mdparse(d[MIME"text/latex"()]; mode = :single))
+    elseif haskey(d, MIME"text/plain"())
+        latex(io, Markdown.Code(d[MIME"text/plain"()]))
+    else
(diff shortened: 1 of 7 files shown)


fredrikekre:
> I suggest we put this in as an enhancement in 0.22

Definitely.

mortenpi:
A few more thoughts:

* It should not be a problem to also support WebP images in `MarkdownWriter` in the same way.
* It should actually be fine to have arbitrary HTML in the outputted Markdown -- MkDocs should handle it without issues (Documenter outputs some anyway, e.g. the anchor tags). So the `` approach should also work for `MarkdownWriter` too. It should also be possible to output `text/html`. Needs to be tested though.

To test the MkDocs build, `docs_dir` in `test/examples/mkdocs.yml` needs to be changed to `builds/markdown`.

src/Writers/MarkdownWriter.jl
```diff
@@ -138,6 +138,36 @@ function render(io::IO, mime::MIME"text/plain", node::Documents.EvalNode, page,
...
 
+# Select the "best" representation for Markdown output.
+using Base64: base64decode
+function render(io::IO, mime::MIME"text/plain", d::Documents.MultiOutput, page, doc)
+    foreach(x -> Base.invokelatest(render, io, mime, x, page, doc), d.content)
+end
+function render(io::IO, mime::MIME"text/plain", d::Dict{MIME,Any}, page, doc)
+    filename = String(rand('a':'z', 7))
+    if haskey(d, MIME"image/png"())
+        write("$(filename).png", base64decode(d[MIME"image/png"()]))
```
mortenpi: I believe this needs some path computation here. The files end up next to the `make.jl` script currently it seems.

fredrikekre: Yea, I was just looking at that. In this case it is trivial since `page` is an argument to this function, but we don't propagate that info this far for the other writers.

src/Writers/LaTeXWriter.jl
```diff
@@ -316,6 +316,39 @@ function latex(io::IO, node::Documents.EvalNode, page, doc)
...
+        """)
+    elseif haskey(d, MIME"image/jpeg"())
+        write("$(filename).jpeg", base64decode(d[MIME"image/jpeg"()]))
+        _println(io, """
+        \\begin{figure}[H]
+        \\centering
+        \\includegraphics{$(filename)}
+        \\end{figure}
+        """)
+    elseif haskey(d, MIME"text/latex"())
```
tkf: Doesn't it make sense to prefer `text/latex` in latex writer?  I don't use latex writer myself so I don't know what's the optimal order, though.

fredrikekre: I thought it made sense to prioritize more rich formats, such as images, before latex :man_shrugging:. I don't think it is very commong for things to be `showable` in both formats so probably doesn't matter.

PR #943, opened 2019-02-06
Add Project.toml file.

fredrikekre:


mortenpi:
We still can't get rid of `REQUIRE` though, right?

fredrikekre:
No, this is just for convenience when developing.

PR #946, opened 2019-02-07
Bump lunr.js and lodash to latest patch release

visr:


mortenpi:
Alright, if it's a CVE then I think it's worth backporting this and making a release.

visr:
It might. Not an expert so don't know if this applies to us at all.

PR #948, opened 2019-02-08
Render text/markdown in MarkdownWriter

tkf:
This PR lets you render `Markdown.MD` when the build format is markdown.  Previously (as of #938), it was rendered as an HTML.

Before:

``````
shell> cat make.jl
import DocumenterMarkdown
using Documenter
makedocs(format=DocumenterMarkdown.Markdown())

shell> cat src/index.md
# Example

```@example
using Markdown
Markdown.parse("""
![](https://raw.githubusercontent.com/JuliaGraphics/julia-logo-graphics/master/images/julia-logo-325-by-225.png)
""")
```

julia> include("make.jl")

shell> cat build/index.md

<a id='Example-1'></a>

# Example


```julia
using Markdown
Markdown.parse("""
![](https://raw.githubusercontent.com/JuliaGraphics/julia-logo-graphics/master/images/julia-logo-325-by-225.png)
""")
```

<div class="markdown"><p><img src="https://raw.githubusercontent.com/JuliaGraphics/julia-logo-graphics/master/images/julia-logo-325-by-225.png" alt="" /></p>
</div>
``````

After:

``````
shell> cat build/index.md

<a id='Example-1'></a>

# Example


```julia
using Markdown
Markdown.parse("""
![](https://raw.githubusercontent.com/JuliaGraphics/julia-logo-graphics/master/images/julia-logo-325-by-225.png)
""")
```

![](https://raw.githubusercontent.com/JuliaGraphics/julia-logo-graphics/master/images/julia-logo-325-by-225.png)
``````

### DIFF
--- a/src/Writers/MarkdownWriter.jl
+++ b/src/Writers/MarkdownWriter.jl
@@ -145,7 +145,9 @@ function render(io::IO, mime::MIME"text/plain", d::Documents.MultiOutput, page, 
 end
 function render(io::IO, mime::MIME"text/plain", d::Dict{MIME,Any}, page, doc)
     filename = String(rand('a':'z', 7))
-    if haskey(d, MIME"text/html"())
+    if haskey(d, MIME"text/markdown"())
+        println(io, d[MIME"text/markdown"()])
+    elseif haskey(d, MIME"text/html"())
         println(io, d[MIME"text/html"()])
     elseif haskey(d, MIME"image/svg+xml"())
         # NOTE: It seems that we can't simply save the SVG images as a file and include them
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -643,7 +643,7 @@ function display_dict(x)
     out[MIME"text/plain"()] = stringmime(MIME"text/plain"(), x)
     for m in [MIME"text/html"(), MIME"image/svg+xml"(), MIME"image/png"(),
               MIME"image/webp"(), MIME"image/gif"(), MIME"image/jpeg"(),
-              MIME"text/latex"()]
+              MIME"text/latex"(), MIME"text/markdown"()]
         showable(m, x) && (out[m] = stringmime(m, x))
     end
     return out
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by fredrikekre** [comment]
We can probably allow markdown shoable things for the other writers too.

**Comment by tkf** [comment]
OK, I added text/markdown branch just above text/plain.  I added a test for HTML but not sure how to do it with LaTeX.

File: src/Writers/MarkdownWriter.jl [hunk 1]
```diff
@@ -145,7 +145,9 @@ function render(io::IO, mime::MIME"text/plain", d::Documents.MultiOutput, page,
 end
 function render(io::IO, mime::MIME"text/plain", d::Dict{MIME,Any}, page, doc)
     filename = String(rand('a':'z', 7))
-    if haskey(d, MIME"text/html"())
+    if haskey(d, MIME"text/markdown"())
+        println(io, d[MIME"text/markdown"()])
+    elseif haskey(d, MIME"text/html"())
```
mortenpi: I wonder if we should actually prefer HTML output here since it would in general be richer than Markdown?

tkf: Since users can put their object in a wrapper object that is showable only as markdown or html (like `Base.Text` does), I think this is not a super big issue.  I'm in favor of trying to print text/markdown first because:

* Markdown files should be human-readable in principle.

    * Counter argument: We are using markdown as "intermediate representation" so it doesn't have to be human-readable.

* Rules for inlining html is non-trivial: https://spec.commonmark.org/0.28/#html-blocks (OK, this is commonmark, not markdown, but I suspect that commonmark compliant parser would be the main consumer of the MarkdownWriter output).  So, it would break markdown parser if `show` prints arbitrary html.

    * Counter argument: It is probably possible to post-process html so that it is treated as a single html block by commonmark parsers.

* It let us use some markdown extensions for downstream renderer (pandoc, gitlab wiki, etc.).

    * Counter argument: As I said above, it's not hard to choose preferred mime.

Since I can find counter argument to my reasoning, I don't really know which is better.  But I still prefer trying markdown first :)

mortenpi: My assumption in general is that it will be consumed by MkDocs, which uses [Python-Markdown](https://python-markdown.github.io/), which in turn implements [the original John Gruber Markdown](https://daringfireball.net/projects/markdown/).

However, I think you are right that using Markdown would be more reliable, especially as we have no control over how objects decide to encode themselves into HTML. And, as you say, if need be, this can be overriden by the user. So let's keep it as it is :slightly_smiling_face:

Issue #949, opened 2019-02-10
Some suggestions for the pdf manual first page


ViralBShah:
The pdf version of the manual is amazing to have. Thank you.

Here's a few suggestions:
- [x] Include the Julia version on the first page
- [x] Include the Julia logo
- [x] Start the table of contents on the next page
- [x] If formatted as a latex book, it would be good to have it as 3 part book - single pdf (manual, base API, dev docs). I understand that three separate PDFs would be difficult because of cross-linking issues.

odow:
I think these four points are all done. Is there anything still actionable?

PR #953, opened 2019-02-11
Update HTMLWriter logo handling and arguments

### LINKED ISSUE #308: Documentation: Adding a Logo to the Sidebar
I see in the Documenter.jl documentation that you have a logo on the top of the sidebar: how is this done? I didn't it specified in the `makedocs` command. I think specifics like these should be added to the documentation.

mortenpi:
Two slightly related changes (as in I got carried away when I was trying to document the logo).

* The logo can now also be an SVG/WebP/JPEG/GIF. Also swaps the manual logo over to an SVG version.
* The analytics and assets keywords to makedocs are deprecated in favour of the corresponding `HTML()` arguments. The reasoning is that those particular makedocs arguments are very `HTMLWriter`-specific.

Ref #212, close #308.


fredrikekre:
The logo could be used for the pdf output too though, e.g. https://github.com/JuliaDocs/Documenter.jl/issues/949 so maybe not move `assets` to `HTML`?

mortenpi:
The `assets` keyword only determines what gets included in the `<head>` part of the HTML pages. All the files are always copied.

src/Documenter.jl
```diff
@@ -194,21 +194,24 @@ function makedocs(components...; debug = false, format = HTML(),
                   html_disable_git::Union{Bool, Nothing} = nothing, # deprecated
                   html_edit_branch::Union{String, Nothing} = nothing, # deprecated
                   html_canonical::Union{String, Nothing} = nothing, # deprecated
-                  kwargs...)
+                  assets::Union{Vector{T}, Nothing} = nothing, # deprecated
```
fredrikekre:
```suggestion
                  assets::Union{Vector{<:AbstractString}, Nothing} = nothing, # deprecated
```

src/Documenter.jl
```diff
@@ -194,21 +194,24 @@ function makedocs(components...; debug = false, format = HTML(),
...
+                  analytics::Union{T, Nothing} = nothing, # deprecated
```
fredrikekre:
```suggestion
                  analytics::Union{<:AbstractString, Nothing} = nothing, # deprecated
```

src/Documenter.jl
```diff
@@ -194,21 +194,24 @@ function makedocs(components...; debug = false, format = HTML(),
...
+                  kwargs...) where {T <: AbstractString, S <: AbstractString}
```
fredrikekre:
```suggestion
                  kwargs...)
```

fredrikekre: Also, TIL that this doesn't error - `S` is not used in the signature.

mortenpi: The `S` was supposed to be for the second argument, but what you suggested is much cleaner. Thanks!

src/Documenter.jl
```diff
@@ -194,21 +194,24 @@ function makedocs(components...; debug = false, format = HTML(),
...
                   html_canonical::Union{String, Nothing} = nothing, # deprecated
+                  assets::Union{Vector{<:AbstractString}, Nothing} = nothing, # deprecated
+                  analytics::Union{<:AbstractString, Nothing} = nothing, # deprecated
                   kwargs...)
     # html_ keywords deprecation
     html_keywords = Dict()
     function html_warn(kw)
-        Base.depwarn("""
+        replace_with = startswith(kw, "html_") ? kw[6:end] : kw
+        @warn """
```
fredrikekre: Why the change from `depwarn` to `@warn`?

mortenpi: Ah, good point, would be good to have your opinion on this. What I noticed is that when you are using multiple deprecated arguments you only get one warning, whereas with `@warn` you get all applicable ones. Also, the line number in the depwarn was like `Core: -1`, whereas `@warn` points to the correct place. Otherwise, it didn't seem that there was any difference to the user.

fredrikekre: Ok yea, lets keep it like this then.

PR #956, opened 2019-02-12
Show no output if the result value is nothing.

fredrikekre:


fredrikekre:
Restores the same behavior as before #938.

PR #958, opened 2019-02-13
Use logging system for doctest-failure reporting.

fredrikekre:


src/DocTests.jl
```diff
@@ -258,34 +258,19 @@ end
...
-    printstyled(ioc, "\n> Output Diff", warning, ":\n\n", color=:cyan)
     diff = TextDiff.Diff{TextDiff.Words}(result.output, rstrip(str))
-    Utilities.TextDiff.showdiff(ioc, diff)
-    println(ioc, "\n\n", "=====[End Error]=", "="^30)
-    push!(doc.internal.errors, :doctest)
-    printstyled(String(take!(iob)), color=:normal)
-end
+    lines = Utilities.find_block_in_file(result.block.code, result.file)
+    @error("""
+        doctest failure in $(Utilities.locrepr(result.file, lines)).
```
mortenpi:
```suggestion
        doctest failure in $(Utilities.locrepr(result.file, lines))
```

mortenpi: I was also thinking it might look nicer to remove the backticks from `locrepr` here. I.e.

```
┌ Error: doctest failure in src/index.md:243-257
│ 
│ ```jldoctest
```

instead of

```
┌ Error: doctest failure in `src/index.md:243-257`.
│ 
│ ```jldoctest
```

src/DocTests.jl
```diff
@@ -258,34 +258,19 @@ end
...
-    println(buffer)
-    for line in split(str, '\n')
-        println(buffer, " "^indent, line)
-    end
+        ```$(result.block.language)
+        $(result.block.code)
+        ```
+
+        Subexpression:
+        $(result.input)
```
mortenpi: Could add an empty line here:

```suggestion

        $(result.input)
```

So

```
│ ```
│ 
│ Subexpression:
│
│ fullname(@__MODULE__)
│ 
│   diff = (:Main,)(:Main, Symbol("Main"))
```

instead of

```
│ ```
│ 
│ Subexpression:
│ fullname(@__MODULE__)
│ 
│   diff = (:Main,)(:Main, Symbol("Main"))
```

src/DocTests.jl
```diff
@@ -258,34 +258,19 @@ end
...
+
+        """, diff)
```
mortenpi: For consistency with the the rest, it might be nice to also put the diff in ourselves. E.g.:

```
│ Diff:
│ 
│  (:Main,)(:Main, Symbol("Main"))
``` 

instead of 

```
│   diff = (:Main,)(:Main, Symbol("Main"))
```

But I guess there might not be any easy way to paste the colored diff output into the error string?

mortenpi: Also, might it be worth having like `Output:` and `Expected output:` parts? Especially handy if you don't have color for some reason and the diff is unreadable.

OTOH, all of these things will make each doctest error longer and longer. But the `@error` block separates each one pretty well though.

src/Utilities/TextDiff.jl
```diff
@@ -87,15 +87,11 @@ end
 prefix(::Diff{Lines}, s::Symbol) = s === :green ? "+ " : s === :red  ? "- " : "  "
 prefix(::Diff{Words}, ::Symbol) = ""
 
-function showdiff(io::IO, diff::Diff)
+function Base.show(io::IO, diff::Diff)
+    get(io, :color, false) || println(io, "Warning: Diff output requires color.")
```
mortenpi: Any easy way we could maybe fall back to the line-by-line output if there is no color?

src/DocTests.jl
fredrikekre: Yea, but that does not work with the colors so...

src/Utilities/TextDiff.jl
fredrikekre: Hmm. I included `Output:` and `Expected output:` so witht those I think it is fine not to show the diff again.

mortenpi: Do you mean completely removing the outputting of the `diff` object? Or just when there is no color?

fredrikekre: No I just mean that in case of no color I don't think it helps much printing line by line when I already added both the output and expected output.

PR #960, opened 2019-02-13
fix #959: jldoctest output

### LINKED ISSUE #959: Doctest with a semicolon fails
Following doctest fails in Documenter 0.21.1--0.21.3 and 24ee0067847bacbc37bb29b3e5d0977e13509c25 but not in 0.21.0:

`````markdown
```jldoctest
julia> "hello"; "world"
"world"
```
`````

### PR DESCRIPTION
**Pull request by fredrikekre**
fix #959: bad interaction between ans and multiple expressions on same line.

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -537,9 +537,10 @@ function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
         for (ex, str) in Utilities.parseblock(code, doc, page; keywords = false)
             (value, success, backtrace, text) = Utilities.withoutput() do
                 cd(dirname(page.build)) do
-                    Core.eval(mod, Expr(:(=), :ans, ex))
+                    Core.eval(mod, ex)
                 end
             end
+            Core.eval(mod, Expr(:global, Expr(:(=), :ans, QuoteNode(value))))
             result = value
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Expanders.jl [hunk 1]
```diff
@@ -537,9 +537,10 @@ function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
         for (ex, str) in Utilities.parseblock(code, doc, page; keywords = false)
             (value, success, backtrace, text) = Utilities.withoutput() do
                 cd(dirname(page.build)) do
-                    Core.eval(mod, Expr(:(=), :ans, ex))
+                    Core.eval(mod, ex)
                 end
             end
+            Core.eval(mod, Expr(:global, Expr(:(=), :ans, QuoteNode(value))))
```
fredrikekre: It's a bit weird that we set `ans` for `@example` blocks but it was tested so...

mortenpi: :man_shrugging: I guess we consider every line in an at-example to be a REPL line, which is not necessarily unreasonable -- people will most likely paste it into the REPL or run it in Juno or something.

fredrikekre: Right, but we are not even doing it for script-style jldoctest.

PR #967, opened 2019-02-21
Fix depwarn for format

fredrikekre:
Fix depwarn for format; this accidentally created a closure over `format` which happened to work when it was not given as a vector.


mortenpi:
I think it's worth backporting this, but it could use a short CHANGELOG entry then.

fredrikekre:
Thanks!

PR #971, opened 2019-02-26
Small changes for GitLab

JackDunnNZ:
I have been experimenting with getting docs to build on GitLab and found a few things to fix up.

For reference and to help anyone else, I've been using the following hacked together script  that calls `git_push` to skip all of the travis-specific processing in `deploydocs`:

```julia
if haskey(ENV, "CI")
  tag = get(ENV, "CI_COMMIT_TAG", "")
  tag_ok = isempty(tag) || occursin(Base.VERSION_REGEX, tag)

  if tag_ok
    root = pwd()
    build_dir = "docs/build"
    target_repo = "gitlab.com/xxx/yyy"

    mktempdir() do temp
      devurl = "dev"
      Documenter.git_push(root, temp, target_repo;
          target=build_dir,
          tag=tag,
          host="gitlab.com",
          key=ENV["DOCUMENTER_KEY"],
          # Defaults from Documenter.deploydocs
          sha=readchomp(`git rev-parse --short HEAD`),
          devurl=devurl,
          versions=["stable" => "v^", "v#.#", devurl => devurl],
          forcepush=false,
      )
    end
  end
end
```

There are three commits each with a minor change:

1. Add a `host` option to `git_push` so that the url rewriting and temp ssh config file work for gitlab.com urls as well
2. Makes the .ssh directory if it doesn't already exist to avoid crashing
3. Wraps the identity file path in quotes in case of spaces in the filename

Happy to rework anything as required

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -606,8 +606,11 @@ function git_push(
 
     target_dir = abspath(target)
 
+    # Extract host from repo as everything up to first ':' or '/' character
+    host = match(r"(.*?)[:\/]", repo)[1]
+
     # The upstream URL to which we push new content and the ssh decryption commands.
-    upstream = "git@$(replace(repo, "github.com/" => "github.com:"))"
+    upstream = "git@$(replace(repo, "$host/" => "$host:"))"
 
     keyfile = abspath(joinpath(root, ".documenter"))
     try
@@ -626,9 +629,9 @@ function git_push(
         # Use a custom SSH config file to avoid overwriting the default user config.
         withfile(joinpath(homedir(), ".ssh", "config"),
             """
-            Host github.com
+            Host $host
                 StrictHostKeyChecking no
-                HostName github.com
+                HostName $host
                 IdentityFile $keyfile
                 BatchMode yes
             """
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -632,7 +632,7 @@ function git_push(
             Host $host
                 StrictHostKeyChecking no
                 HostName $host
-                IdentityFile $keyfile
+                IdentityFile "$keyfile"
                 BatchMode yes
             """
         ) do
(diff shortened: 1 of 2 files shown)


src/Documenter.jl
```diff
@@ -599,15 +599,15 @@ and when building docs for a tag they are deployed to a `vX.Y.Z` directory.
 function git_push(
         root, temp, repo;
         branch="gh-pages", dirname="", target="site", tag="", key="", sha="", devurl="dev",
-        versions, forcepush=false,
+        versions, forcepush=false, host="github.com",
```
fredrikekre: Are you planning to call `git_push` directly or where is this kwarg set? Couldn't we find the host from the `repo` argument to `deploydocs`?

src/Documenter.jl
```diff
@@ -624,12 +624,14 @@ function git_push(
 
     try
         # Use a custom SSH config file to avoid overwriting the default user config.
+        ssh_dir = joinpath(homedir(), ".ssh")
+        isdir(ssh_dir) || mkdir(ssh_dir)
```
fredrikekre: This should be cleaned up if it was created.

JackDunnNZ: I moved this logic into `withfile` to create and clean up the directory

src/Documenter.jl
JackDunnNZ: Yes currently I call `git_push` directly as in the first comment. We could definitely pull the host from the `repo` argument as something like `split(repo, "/")[1]` if you think that would be better - it's probably more robust.

mortenpi: Yeah, parsing `repo` seems better here.

I think currently we allow both `github.com:` and `github.com/` as the start of the `repo`. So we could formalize that a bit and say repo has to start with the domain name, then you have `/` or `:` as a separator, and then the rest of the URL follows. So we should also split by `:`, or probably use a regex to match the start.

mortenpi:
Other than the `repo`/`host` thing, LGTM! @JackDunnNZ could you also drop a small note into `CHANGELOG.md`?

JackDunnNZ:
Thanks both for the comments! I've changed to extract the host using regex and added the note to the changelog

mortenpi:
Thanks!

PR #976, opened 2019-03-03
add option to upload search index to Algolia

visr:
Still a work in progress, though opening here as a draft PR for visibility and discussion.

Continues after the refactoring in #966, by adding an option to upload the search index to Algolia. The uploading part should work already, given (copied from docstring):

> For this to work two environment variables need to be set;
`ALGOLIA_APPLICATION_ID` and `ALGOLIA_API_KEY`. Both can be found on the
"API Keys" page of your application on the Algolia website. `ALGOLIA_API_KEY`
refers to the Admin API Key, which is needed to create new indexes.

## User interface

However uploading the index and not using it is not very helpful, so I still plan to add this to the HTML & JS in this PR. I want to do this as simple as possible based on the example provided [here](https://github.com/algolia/algoliasearch-client-javascript/blob/27fa15a86d8ebe6854bac8c42acee091b1d07c12/examples/jquery.html). Leaving the UI as-is as much as possible, so we can leave that for a possible future PR (e.g. switching to using [InstantSearch.js](https://community.algolia.com/instantsearch.js/)  or @haampie's https://github.com/haampie/julia-docs-search-ui).

## Free plans and quotas

I  am slightly worried about this approach creating too many operations for the free plan (copied from docstring):

> Note that this approach is simple since it generates the full index from scratch,
and is therefore not reliant on updating a pre-existing index. However since
most search records are unlikely to change between each time this is run,
it would be more efficient to somehow hash the records and corresponding
Algolia object ID and only upload changes. Indexing operations count equally
as search operations towards the number of operations per month.

See also the discussion about this here: https://discourse.julialang.org/t/rfc-improved-search-for-documentation/8079/. The free [community plan](https://www.algolia.com/pricing/) will likely be too small for the largest projects, in which case they can also apply for the  [Algolia for Open Source (AOS) plan](https://www.algolia.com/for-open-source/). The records quota we can easily estimate and likely won't be the bottleneck, but the amount of operations is hard to estimate (not the number of upload operations, but the number of search operations). Though from the AOS page it seems that this limit can be discussed with Algolia (for base docs for instance).

##  Doc and search index versions

I don't know yet how to handle the different versions that are hosted simultaneously, but I also don't fully understand how versions work now. Is it true that `makedocs` always just makes one version, which is the latest master, which is deployed (`deploydocs`) to `dev`, `stable` or `vx.x` depending on what it is? So a new deploy will never touch old versions, is that right? At first I assumed it always builds all versions, in which case we could have a separate index per version, but now I'm not sure. The `versions` keyword is only available in `deploydocs` and not in `makedocs`, but `deploydocs` misses other data about the site name and `Vector{SearchRecord}` itself that I need. Not sure where I should upload the index, and how to handle versions. Ideas very welcome!

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1221,7 +1221,8 @@ fixlinks!(ctx, navnode, md) = nothing
 # ------------------------------------------------------------------------------
 
 # TODO create better settings, now all defaults
-"""Customized Algolia settings for optimal results
+"""
+Customized Algolia settings for optimal results
 
 https://www.algolia.com/doc/api-reference/settings-api-parameters/
 """
@@ -1238,7 +1239,8 @@ const ALGOLIA_SETTINGS = Dict{String, Any}(
                           c == '_' ||
                           (isascii(c) && (isletter(c) || isnumeric(c)))
 
-"""Create an index name based on sitename and version
+"""
+Create an index name based on sitename and version
 
 It is assumed that the
 
@@ -1257,7 +1259,8 @@ function indexname(sitename::AbstractString; temp::Bool=false)
     temp ? string(simplesite, "_temp_", randstr) : simplesite
 end
 
-"""Delete old temporary indexes
+"""
+Delete old temporary indexes
 
 This should normally speaking never find any, but in case a previous upload
 failed halfway we don't want to accumulate temporary indexes.
@@ -1290,7 +1293,8 @@ function checkstatus(d::AbstractDict)
     return d
 end
 
-"""Do a request to the Algolia REST API
+"""
+Do a request to the Algolia REST API
 
 For this to work two environment variables need to be set;
 `ALGOLIA_APPLICATION_ID` and `ALGOLIA_API_KEY`. Both can be found on the
@@ -1346,7 +1350,8 @@ function batch_write(index_temp::AbstractString, records::AbstractVector)
     end
 end
 
-"""Use the Algolia REST API to upload the search index
+"""
+Use the Algolia REST API to upload the search index
 
 This creates a temporary index which at the end is moved over the one used in
 the generated documentation website. This approach will not give any downtime,
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -286,7 +286,13 @@ end
 
 function add_search_algolia(ctx::HTMLContext, doc::Documents.Document, settings::HTML)
     render_search(ctx, settings.search)
-    upload_index(doc.user.sitename, ctx.search_index)
+    ctx.search_index_js = "search_index.json"
+    # store a JSON that has all information needed by deploydocs(), which will
+    # upload the search index to Algolia
+    d = Dict("sitename" => doc.user.sitename, "records" => ctx.search_index)
+    open(joinpath(doc.user.root, ctx.search_index_js), "w") do io
+        JSON.print(io, d)
+    end
 end
 
 """
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -40,6 +40,7 @@ then it is intended as the page title. This has two consequences:
 """
 module HTMLWriter
 
+import Random
 import Markdown
 import JSON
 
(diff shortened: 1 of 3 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -1183,4 +1217,172 @@ fixlinks!(ctx, navnode, md) = nothing
 # TODO: do some regex-magic in raw HTML blocks? Currently ignored.
 #fixlinks!(ctx, navnode, md::Documents.RawHTML) = ...
 
+# Algolia search
+# ------------------------------------------------------------------------------
+
+# TODO create better settings, now all defaults
+"""Customized Algolia settings for optimal results
```
mortenpi:
```suggestion
"""
Customized Algolia settings for optimal results
```

Minor style bikeshed, but the docstrings in Documenter usually start on a new line. Ditto for some of the others.

src/Writers/HTMLWriter.jl
```diff
@@ -419,14 +442,25 @@ function render_search(ctx)
...
-    )
+    else
+        htmldoc = DOM.HTMLDocument(
+            html[:lang=>"en"](
+                head,
+                body(navmenu, article),
+                script[:src => "https://cdn.jsdelivr.net/algoliasearch/3/algoliasearchLite.min.js"],
+            )
+        )
+    end
```
mortenpi: Either way is fine, but could slightly reduce code duplication by doing something like:

```julia
search_scripts = if search == "lunr"
    script[:src => relhref(src, ctx.search_index_js)], script[:src => relhref(src, ctx.search_js)]
else
    script[:src => "https://cdn.jsdelivr.net/algoliasearch/3/algoliasearchLite.min.js"],
end

htmldoc = DOM.HTMLDocument(
    html[:lang=>"en"](
        head,
        body(navmenu, article),
        search_scripts...,
    )
)
```

mortenpi:
This looks really awesome, thank you for tackling this! I did not test it yet, but the code looks great, so I don't really have much feedback to give there.

Regarding the UI: the search page is a separate page `.html` page created by `render_search`, so feel free to do whatever you need to include Algolia there.

Regarding versions etc.: 

* Yes, currently `makedocs` always creates the same thing, independent of versions. This is a limitation of the `makedocs` / `deploydocs` design -- you don't know anything about deployment / versions when you render. It's actually something that [should be changed at some point](https://github.com/JuliaDocs/Documenter.jl/wiki/Documenter.jl-roadmap) in my opinion.

* `master` builds deploy to `dev/`, and this gets continously overwritten as you update `master`. The versioned docs come from the CI runs for the tags, and are usually generated once (although will re-generate if you re-run the Travis build for the tag).

I guess ideally you would have a separate Algolia index for each version, so that the search would correspond to the actual docs?

If the Algolia rate limits are an issue, then another way would be to have Algolia only for the latest stable docs. But then we need a way to fall back to lunr for the old versions, ideally without re-generating them. Should be doable by some simple JS though, similar to how we handle the version selector.

One option might be to generate the search index as a `.json` file, and then let `deploydocs` decide what to do with it. Uploading it to Algolia should be handled by `deploydocs` anyway.

src/Writers/HTMLWriter.jl
```diff
@@ -132,15 +133,17 @@ struct HTML <: Documenter.Plugin
...
     function HTML(;
         prettyurls::Bool = true,
         disable_git::Bool = false,
         edit_branch::Union{String, Nothing} = "master",
         canonical::Union{String, Nothing} = nothing,
         assets::Vector{String} = String[],
-        analytics::String = "")
-        new(prettyurls, disable_git, edit_branch, canonical, assets, analytics)
+        analytics::String = "",
+        search::String = "lunr")
```
mortenpi: Minor bikeshed here, but I would change this to a `Symbol`.

visr:
Sorry for abandoning this. If anyone wants to pick it up, feel free. I'll leave it up to the maintainers if they prefer to close this or not.

Algolia would be a good option, but I think perhaps using the hosted search from JuliaHub would be an even better option, since it already exists and is made for searching Julia code / docs. I mentioned this in the link below and @StefanKarpinski thought it was a good idea. But given that JuliaHub is not open, someone at Julia Computing would probably need to make this happen, if they decide they want this.

https://discourse.julialang.org/t/ann-juliahub-explore-run-scale/41135/35

mortenpi:
I'll go ahead and close this. The branch should still remain available for anyone wanting to pick this up.

PR #987, opened 2019-03-23
Display stdout if nothing result in at-example

### LINKED ISSUE #986: stdout not being captured
On the current master, the following doesn't show any output being captured

    ```@example
    println("hello world")
    ```

Works as expected on 0.21.5

mortenpi:
Should restore the old behavior of at-example blocks with stdout/stderr output.

Fix #986

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -573,7 +573,8 @@ function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
     # Only add content when there's actually something to add.
     isempty(input)  || push!(content, Markdown.Code("julia", input))
     if result === nothing
-        push!(content, Markdown.Code(Documenter.DocTests.sanitise(buffer)))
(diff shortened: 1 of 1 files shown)


src/Expanders.jl
```diff
@@ -572,7 +572,11 @@ function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
 
     # Only add content when there's actually something to add.
     isempty(input)  || push!(content, Markdown.Code("julia", input))
-    isempty(output) || push!(content, output)
+    if result === nothing
+        push!(content, Markdown.Code(Documenter.DocTests.sanitise(buffer)))
```
fredrikekre: I guess we should only `push!` to `content` if `Documenter.DocTests.sanitise(buffer)` returns something non-empty?

mortenpi: This is also consistent with how it was before.

PR #989, opened 2019-03-24
Enable table column alignment in HTML

### LINKED ISSUE #511: Justification of Table Columns
It seems like the right-side justification on tables is not working.  For example see, https://lanl-ansi.github.io/PowerModels.jl/latest/results.html

Also I noticed that the bold directive does not seem to work on this page.  Both in the body of the text and in the table header.

The bold directive is recognized by github's markdown renderer, https://github.com/lanl-ansi/PowerModels.jl/blob/8d6ffb0dad80be64255d28d16afbff0c9c9e3c94/docs/src/results.md

mortenpi:
Note that tables that do not explicitly define alignment default to be right-aligned. Fix #511.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
 
 * ![Bugfix][badge-bugfix] `deploydocs` and `git_push` now support non-github repos correctly and work when the `.ssh` directory does not already exist or the working directory contains spaces. ([#971][github-971])
 
-* ![Bugfix][badge-bugfix] Tables now honor colunm alignment in the HTML output. If a column does not explicitly specify its alignment it defaults to being right-aligned, whereas previously all cells were left-aligned. ([#511][github-511], [#989][github-989])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -53,6 +53,8 @@
 
 * ![Bugfix][badge-bugfix] `deploydocs` and `git_push` now support non-github repos correctly and work when the `.ssh` directory does not already exist or the working directory contains spaces. ([#971][github-971])
 
+* ![Bugfix][badge-bugfix] Tables now honor colunm alignment in the HTML output. If a column does not explicitly specify its alignment it defaults to being right-aligned, whereas previously all cells were left-aligned. ([#511][github-511], [#989][github-989])
```
```suggestion
* ![Bugfix][badge-bugfix] Tables now honor column alignment in the HTML output. If a column does not explicitly specify its alignment, the parser defaults to it being right-aligned, whereas previously all cells were left-aligned. ([#511][github-511], [#989][github-989])
```

Issue #613, opened 2017-12-23
PyPlot.ioff() # hide does not actually hide the command


Datseris:
For example:

````markdown
```@example plot
using DynamicalSystems
using PyPlot
PyPlot.ioff() # hide

ds = Systems.logistic()
```
````

does not hide the line `PyPlot.ioff() # hide` when it is compiled.

Datseris:
Notice that other hide commands work for me. For example:
`savefig("logisticdiagram.png"); nothing # hide` is hidden in the output

joshday:
I've found something similar.  Some `@example` blocks hide lines correctly and some don't, and I haven't found a common pattern between them yet.

http://joshday.github.io/OnlineStats.jl/latest/visualizations.html

KristofferC:
Maybe fixed by https://github.com/JuliaDocs/Documenter.jl/pull/991?

mortenpi:
Yeah, this seems likely. I'll close this. If someone still runs into this (with Documenter 0.22 or higher), we can re-open.

PR #991, opened 2019-03-25
Consider CRLF when splitting code lines

notZaki:
I noticed that code lines ending with `# hide` were not getting hidden in the output. 
The problem went away if I switched from CRLF to LF line endings.
This might be related to issue #613

I think the issue is caused by `split(code, '\n')` (line 680 in expanders.jl) which assumes LF line ending. 
For CRLF input, code containing `... # hide` gets split to `... # hide\r` and this causes the regex on line 681 to not match---this is because `$` does not match `\r` [[ref](https://discourse.julialang.org/t/regex-doesnt-match-end-of-line-with-both-carriage-return-and-line-feed/5795)]

This PR fixes that by splitting code with `r"\r?\n"`.

---

Minimum working example:

- make.jl
```
using Documenter
makedocs(sitename="Example")
```
- src/index.md (using CRLF)
```
# Example

```@example
import Random # hide
Random.seed!(1) # hide
A = rand(3, 3)
b = [1, 2, 3]
A \ b
Random.seed!(2); nothing # hide
```

- Ouput using master (note that last line still gets hidden correctly):
image

- Output using PR
image

- If LF is used, then both master and PR have the same output where code is hidden correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@
 
 * ![Bugfix][badge-bugfix] Tables now honor column alignment in the HTML output. If a column does not explicitly specify its alignment, the parser defaults to it being right-aligned, whereas previously all cells were left-aligned. ([#511][github-511], [#989][github-989])
 
+* ![Bugfix][badge-bugfix] Code lines ending with `# hide` are now properly hidden for CRLF inputs. ([#991][github-991])
+
 ## Version `v0.21.5`
 
 * ![Bugfix][badge-bugfix] Deprecation warnings for `format` now get printed correctly when multiple formats are passed as a `Vector`. ([#967][github-967])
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -55,6 +55,8 @@
 
 * ![Bugfix][badge-bugfix] Tables now honor column alignment in the HTML output. If a column does not explicitly specify its alignment, the parser defaults to it being right-aligned, whereas previously all cells were left-aligned. ([#511][github-511], [#989][github-989])
 
+* ![Bugfix][badge-bugfix] Code lines ending with `# hide` are now properly hidden for CRLF inputs ([#991][github-991])
```
```suggestion
* ![Bugfix][badge-bugfix] Code lines ending with `# hide` are now properly hidden for CRLF inputs. ([#991][github-991])
```

mortenpi:
Some parallel processing here :laughing: Thanks! Will merge when tests go green.

notZaki:
Update changelog.
Didn't know if the policy was to squash commits or not, so I forced push the changes.

mortenpi:
I squash merge PRs usually, so no need to squash them by hand.

PR #995, opened 2019-04-02
document sitename.

### LINKED ISSUE #883: sitename missing from makedocs docstring
I think `makedocs(sitename = ...)` is still required, but is no longer mentioned at 
https://juliadocs.github.io/Documenter.jl/dev/lib/public/

tpapp:
Fixes #883.


mortenpi:
LGTM, thanks!

PR #999, opened 2019-04-12
Fix filepath in tex file on Windows.

fredrikekre:
See https://discourse.julialang.org/t/issue-with-documenters-pdf-output-pictures-not-found/23023/

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -535,6 +535,7 @@ function latexinline(io::IO, md::Markdown.Image)
         else
             normpath(joinpath(dirname(io.filename), md.url))
         end
+        url = replace(url, "\\" => "/") # use / on Windows too.
         wrapinline(io, "includegraphics") do
             _print(io, url)
         end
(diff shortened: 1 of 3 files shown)


src/Writers/LaTeXWriter.jl
```diff
@@ -533,7 +533,7 @@ function latexinline(io::IO, md::Markdown.Image)
             # URLs starting with a / are assumed to be relative to the document's root
             normpath(lstrip(md.url, '/'))
         else
-            normpath(joinpath(dirname(io.filename), md.url))
+            normpath(dirname(io.filename) * "/" * md.url)
```
fredrikekre: I guess there could also be `\\` in `dirname(io.filename)` and `md.url`? Maybe safest to do this:
```suggestion
            normpath(replace(dirname(io.filename) * "/" * md.url, "\\" => "/"))
```

cserteGT3:
With the latest version (` "\\" => "/" `) it's working.
(I'm still getting some error, but now I have a pdf with the pictures).
Thank you for the quick fix!

mortenpi:
LGTM. No need for a backport, right? Can just update changelog and `Project.toml` here, merge and tag?

fredrikekre:
> No need for a backport, right? Can just update changelog and Project.toml here, merge and tag?

Ah, right. Done.

PR #1002, opened 2019-04-22
Handle non-existing files in find_block_in_file

mortenpi:
Work around the issue reported [on Discourse](https://discourse.julialang.org/t/documenter-systemerror-while-trying-to-read-statistics-jl/23363/3). Still WIP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Documenter.jl changelog
 
+## Version `v0.22.4`
+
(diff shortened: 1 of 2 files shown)


fredrikekre:
Maybe just Maybe just `isfile(source_file) || return nothing`?

mortenpi:
Yep, I think you are right. I feel that we should maybe try to detect the situations that cause this issue, but I don't think is is the right place for the warning message.

PR #1005, opened 2019-04-28
Improve logo link URLs

mortenpi:
* Instead of pointing to potentially non-existent `/index.md`, the logo will now point to the top most page in the navigation menu.
* `index.html` no longer appears in the URL when pretty URLs are enabled.

This is technically slightly behavior-changing, so I'd leave it for 0.23.


mortenpi:
It technically changes behaviour if you don't have the first item pointing to `index.md` (although it is an extremely unlikely corner case). Will merge after #1004, so that there is no need to do backports.

PR #1015, opened 2019-05-08
Misc simplifications

fredrikekre:
See commit messages.


mortenpi:
Hmm. Should actually also update README and `docs/src/index.md`. They refer to 0.7.

PR #1016, opened 2019-05-08
Set version to 0.22.4

mortenpi:


mortenpi:
So.. it should work if I do..

@JuliaRegistrator register()

PR #1017, opened 2019-05-08
Use an at-eval block to parse correct Documenter version at build-time.

fredrikekre:


fredrikekre:
curl failure on AV 0.7

Issue #1013, opened 2019-05-07
Add option to change root path during build.


kescobo:
Assuming a standard project layout,

```
MyPackage
    ├── src
    │   ├── MyPackage.jl
    │   └── functions.jl
    ├── docs
    │   ├── makedocs.jl
    │   └── src
    │       ├── 01-foo.md
    │       ├── 02-bar.md
    │       └── index.md
    ├── test
    │   └── runtests.jl
    ├── Manifest.toml
    ├── Project.toml
    └── README.md
```

I'd like to be able to set an option such that `makedocs()` will execute code in `MyPackage/docs/src/01-foo.md` as if the current working directory is `MyPackage/`. 

I tried 

```
makedocs(
    ...
    root=joinpath(dirname(pathof(MyPackage)), ".."),
    source="docs/src",
    build="build")
```

thinking that the `root` option was what I wanted, but `pwd()` in a code block returns `MyProject/build/`. I also tried setting the `build="."`... oops #1012 

The reason for this desire is that I'm using the markdown files as interactive data analysis notebooks with Juno, and Juno sets the current working directory as the package root. I'd like to use Documenter to render html versions of the analysis notebooks, but would prefer not to go through and change all of the paths to the input data and output folders. 

Is this possible, even with some sort of hacky solution? I'm happy to work on a PR if I this would be useful for others, and if I can get a bit of direction about where to start.

mortenpi:
Hmm. How are you building the notebooks in Juno such that you're getting a different working directory? Are you just doing `include("docs/makedocs.jl")` or something else?

Would an at-setup block in each `.md` file where you `cd` to the correct directory work?

kescobo:
Actually, that doesn't seem to work - during the build, `MyPackage` is not defined. Do I have to add it to a doc `Project.toml` or something?

fredrikekre:
Yes, but most importantly you need to `using MyPackage` before using it.

kescobo:
In fact - that doesn't actually work. `MyPackage` is not defined during the build. I tried adding it to the `[deps]` in a `Project.toml` file inside `docs/`, but that did not help (guessing that only matters for travis??).

But I also tried something a bit more cloodgy and it didn't work either. I can't seem to get out of the build directory no mater what:

````
```@example 1
cd("../../../")# hide
pwd()
```

```@example 1
cd("..")
pwd()
```
````

[...]

PR #1025, opened 2019-05-21
Add option to change working directory for evaluation of example code

kescobo:
Attempt to address #1013 

- [x] Add `working_dir` option to `makedocs`
- [x] Successfully build test site with evaluation in non-build directory
- [x] Add tests
- [x] Add documentation

I've added a `working_dir` option to `makedocs()`, which is intended to allow a user to set where code in `@example` and `@repl` blocks is evaluated from. The default value is `:build` which recapitulates current behavior.


kescobo:
I'm testing on a document called `test.md`:

````md
# Testing

Here's a test

```@repl
pwd()
```
````

Default behavior works as expected, but if I pass strings like "src", "build" or "test", I get the same output for all of them:

```
julia> pwd()
"/Users/ksb/repos/julia_pkgs/DocumenterDev/docs"
```

But if I pass "." I get

```
julia> pwd()
ERROR: IOError: chdir docs: no such file or directory (ENOENT)
```

or if I pass ".." I get

```
julia> pwd()
ERROR: IOError: chdir DocumenterDev/docs: no such file or directory (ENOENT)
```

Clearly I'm missing something about how the paths are passed through - any help would be appreciated. Also happy to take other comments / suggestions of course

File: src/Documents.jl [hunk 1]
```diff
@@ -239,6 +245,7 @@ function Document(plugins = nothing;
         root     :: AbstractString   = Utilities.currentdir(),
         source   :: AbstractString   = "src",
         build    :: AbstractString   = "build",
+        working_dir::Union{Symbol, AbstractString}  = :build,
```
mortenpi: Even more bikeshed -- how about `workdir`? Just that all the other keywords are pretty short and this would make it more consistent with them I feel.

src/Documents.jl
```diff
@@ -61,12 +62,13 @@ end
...
-    order    :: Vector{Symbol} # What order should docs be listed in? Set by user.
-    build    :: String         # Path to the file where this index will appear.
-    source   :: String         # Path to the file where this index was written.
-    elements :: Vector         # (object, doc, page, mod, cat)-tuple for constructing links.
+    pages       :: Vector{String} # Which pages to include in the index? Set by user.
+    modules     :: Vector{Module} # Which modules to include? Set by user.
+    order       :: Vector{Symbol} # What order should docs be listed in? Set by user.
+    build       :: String         # Path to the file where this index will appear.
+    source      :: String         # Path to the file where this index was written.
+    working_dir :: Union{Symbol,String}         # Path to the directory where code in this index is executed..
```
mortenpi: I might be missing something, but why do we need to store it for every `IndexNode` and `ContentsNode` separately? Especially since those fields don't appear to be used?

src/Expanders.jl
```diff
@@ -478,7 +478,7 @@ end
 function Selectors.runner(::Type{EvalBlocks}, x, page, doc)
     sandbox = Module(:EvalBlockSandbox)
     lines = Utilities.find_block_in_file(x.code, page.source)
-    cd(dirname(page.build)) do
+    cd(dirname(page.working_dir)) do
```
mortenpi: Should probably remove `dirname` as `working_dir` should already be a directory? Necessary for `page.build` since it points to a file. This should hopefully fix the issues you were seeing.

Ditto for the other `runner`s.

src/Documents.jl
```diff
@@ -266,10 +273,21 @@ function Document(plugins = nothing;
         version = "git:$(Utilities.get_commit_short(root))"
     end
 
+    if working_dir == :build
+        # set working directory to be the same as `build`
+        working_dir = build
```
mortenpi: I think this changes behaviour? `pwd` will now always be `build/` right (by default)? Currently, it is `dirname(source_file)`, i.e. potentially different for each page if the pages live in subdirectories. And this is important when at-blocks create images or something that are to be included with the `` syntax, since you need to specify the relative path.

I would maybe store the `working_dir` option as a `Union{Symbol,String}` in `User` and do this check when creating the `Page` objects in `addpage!`.

src/Documents.jl
kescobo: I'm fine with this!

src/Documents.jl
kescobo: To be honest, I don't fully understand the pipeline, and just added this wherever I saw `build`. I can do a pass where I remove it everywhere it doesn't break things.

src/Expanders.jl
kescobo: Will give it a shot!

src/Documents.jl
kescobo: OK, will try that, thanks!

kescobo:
So, this now works locally. If I do `julia --project=@. --color=yes test/workdir/make.jl`, it correctly makes `workdir.txt` in the correct folder. 

Other tests all pass, but [this is](https://github.com/JuliaDocs/Documenter.jl/pull/1025/files#diff-fce720c43af3c52c862fd7451c7374b8R53) causing an error outside of tests:

```
Documenter: Error During Test at /Users/ksb/.julia/dev/Documenter/test/runtests.jl:18
  Got exception outside of a @test
  LoadError: IOError: open: no such file or directory (ENOENT)
```

I copied the code from `nongit`, but I'm not actually clear on what the test is in `nongit`. Is it just looking for a lack of error?

src/Documents.jl
kescobo: I removed from Index- and ContentsNodes. Everything still works!

src/Expanders.jl
kescobo: Did I accomplish what you were suggesting here?

src/Documents.jl
kescobo: Done

kescobo:
FWIW, other than #1027, this branch now works for my use-case 🎉

src/Builder.jl
```diff
@@ -105,9 +107,20 @@ function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
...
             dst = normpath(joinpath(build, relpath(root, source), file))
+
+            if workdir == :build
+                # set working directory to be the same as `build`
+                wd = normpath(joinpath(build, relpath(root, source)))
+            elseif typeof(workdir) <: Symbol
+                # Maybe allow `:src` and `:root` as well?
+                throw(ArgumentError("Unrecognized working directory option '$workdir'"))
+            else
+                wd = normpath(joinpath(root, workdir))
```
mortenpi:
```suggestion
                wd = normpath(joinpath(doc.user.root, workdir))
```
instead?

At the moment, relative paths become relative to the directory containing the `.md` file. I don't think that's intended? I.e. currently, if you do `workdir="."` in `make.jl`, it is equivalent to `:build`, while I'd expect it to be `workdir = @__DIR__`.

Somewhat related: how do you feel about `workdir = @__DIR__` or `workdir = joinpath(@__DIR__, "..")` being the recommended ways to set the path an absolute path for `workdir`?

mortenpi:
I took the liberty of re-organizing the tests a bit. I marked a few as broken though until it's clear how we want to handle relative paths.

src/Builder.jl
kescobo: Do I understand correctly that this would mean the path is no longer relative to `root`, but we're recommending people set the absolute path themselves? And is `@__DIR__` the directory in the `build` folder where it ends up, or the folder of the `.md` file in `src/`?

mortenpi: Depends on what you mean by "`root`". With my change, the working directory will always be set to a fixed path relative to `makedocs`' `root` argument.

Currently it is relative to `dirname("path/to/page.md")? So e.g. if `workdir="foo"`, then Documenter would `cd` into `build/foo` for `src/page.md`, but into `build/subdir/foo` for `src/subdir/page.md`?

kescobo:
That's awesome, thanks! I like your changes, and all of your tests now pass. How should I bring this up to date with `master` - do you have preferences about rebase vs merge commits?

src/Documenter.jl
```diff
@@ -107,6 +108,20 @@ written when [`makedocs`](@ref) is run. The name of the build directory is, by c
...
+executed. It can be either a path or the special value `:build` (default).
+
+If the `workdir` is set to a path, the working directory is reset to that path for each code
+block being evaluated. Relative paths are taken to be relative to `root`, but using absolute
+paths is recommended (e.g. `workdir = joinpath(@__DIR__, "..")` for executing in the package
+root for the usual `docs/make.jl` setup).
+
+With the default `:build` option, the working directory is set to a subdirectory of `build`,
+determined from the source file path. E.g. for `src/foo.md` it is set to `build/`, for
+`src/foo/bar.md` it is set to `build/foo` etc.
```
mortenpi: I took the liberty of editing the docs a little bit. Do you think this is clear enough do you think it would be better to have the slightly more explicit example you had before?

src/Documents.jl
```diff
@@ -433,7 +437,7 @@ doctest_replace!(block) = true
 
 function buildnode(T::Type, block, doc, page)
     mod  = get(page.globals.meta, :CurrentModule, Main)
-    dict = Dict{Symbol, Any}(:source => page.source, :build => page.build)
+    dict = Dict{Symbol, Any}(:source => page.source, :build => page.build, :workdir => page.workdir)
```
mortenpi: This is unnecessary, no?

```suggestion
    dict = Dict{Symbol, Any}(:source => page.source, :build => page.build)
```

src/Documenter.jl
kescobo: I think this is clearer than my example actually, thanks!

src/Documents.jl
kescobo: It appears to be, yes!

kescobo:
>I would probably just squash merge this, if that sounds reasonable to you.

That's fine with me! Please let me know if there's anything else you need, and thanks for helping me through this :-)

kescobo:
The appveyor failure only occurs on `julia 1.1 x64` and seems unrelated to this PR? I'm showing Travis as pending, but when I click through to the details it looks like everything is done and passing. @mortenpi Anything else to do here? (PS - I saw that you ended up adding this to `CHANGLOG.md`, thanks for that!)

fredrikekre:
I appreciate the work you have put into this PR. However, it feels a bit wrong with a global option for this. Would it be better with something in an `@meta` -block? E.g.
````
```@meta
WorkingDirectory = "..."
```
?

Also, regardless of this we should store the `pwd()` and re-use the one for blocks with the same name. That would make this pattern work: https://github.com/JuliaDocs/Documenter.jl/issues/1013#issuecomment-490563983

mortenpi:
> However, it feels a bit wrong with a global option for this. Would it be better with something in an `@meta` -block?

I do feel that per-file option would be better in principle. So I am thinking that it might be a good idea to  declare `workdir` an experimental keyword. That way @kescobo, you could use it for your use case, but we could also easily replace it when something better comes along.

mortenpi:
I marked it experimental so that we could go ahead and merge this, but also potentially revisit these decisions in the future.

As a side note, we should probably move some of the other "experimental" keywords out the experimental section -- e.g. `strict = true` is not really experimental anymore I think.

mortenpi:
I hope I didn't find one these compromises that makes everyone unhappy :laughing: 

Thank you @kescobo for your work on this! Are you fine with using Documenter on `master` for a while? I think it might be a few weeks before we do another (minor) release.

kescobo:
@mortenpi This is great! Thanks so much for shepherding me through this :-). Definitely happy to be on Documenter `master` for this project while waiting for a merge 👍

PR #1029, opened 2019-06-04
Allow build order of pages to be specified

### LINKED ISSUE #1027: Order and selection of built docs in Makedocs
Related to #1013 (since the order of code execution in the project is critical, otherwise I'm referencing files before they're created), and maybe to #936 ?

`make.jl`:

```
using Documenter

makedocs(
    sitename="Fakedocs",
    pages=[
    "Test" => [
        "Foo" => "foo.md",
        "Bar" => "bar.md",
        ],
    ],
 )
```

`src/foo.md`:

````
```@repl
@info "Foo"
isfile("test.txt")
```
````

`src/bar.md`:

````
```@repl
@warn "Bar"
touch("test.txt")
```
````

I would expect the answer to `isfile("test.txt")` should be `false`, since `"Foo"=> "foo.md"` is listed first in `pages`, but I get `true`.

In addition, if I add

`src/baz.md`:

````
```@repl
touch("baz.md")
```
````

the file is created even though it's not listed in `pages`.

mortenpi:
Adds `priority_expand` keyword to `makedocs`, which allows some pages to be evaluated before the others. Also, enforces a systematic ordering on the rest of the pages by sorting them by the path.

Should fix #1027. @kescobo, does this solve your use case? You can take a look at `docs/make.jl` to see the usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * Documenter v0.23 requires Julia v1.0. ([#1015][github-1015])
 
+* ![Feature][badge-feature] `makedocs` now accepts the `expandfirst` argument, which allows specifying a set of pages that should be evaluated before the other. ([#1027][github-1027], [#1029][github-1029])
+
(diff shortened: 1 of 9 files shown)


kescobo:
Yes, that looks great!

mortenpi:
Ok, I decided to call it `expandfirst`. Bikeshedding welcome though :slightly_smiling_face:

PR #1034, opened 2019-06-05
Insert video HTML tag for images ending with ".webm"

richiejp:
Markdown elements like `Alt video text` will be inserted
with a video HTML element in a similar way to Gitlab markdown.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1030,7 +1030,17 @@ mdconvert(h::Markdown.Header{N}, parent; kwargs...) where {N} = DOM.Tag(Symbol("
 
 mdconvert(::Markdown.HorizontalRule, parent; kwargs...) = Tag(:hr)()
 
-mdconvert(i::Markdown.Image, parent; kwargs...) = Tag(:img)[:src => i.url, :alt => i.alt]
+function mdconvert(i::Markdown.Image, parent; kwargs...)
+    @tags video img a
+
+    if occursin(r"(.webm|.mp4|.ogg|.ogm|.ogv|.avi)$", i.url)
+        video[:src => i.url, :controls => "true", :title => i.alt](
+            a[:href => i.url](i.alt)
+        )
+    else
+        img[:src => i.url, :alt => i.alt]
+    end
+end
 
 mdconvert(i::Markdown.Italic, parent; kwargs...) = Tag(:em)(mdconvert(i.text, i; kwargs...))
 
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1033,7 +1033,7 @@ mdconvert(::Markdown.HorizontalRule, parent; kwargs...) = Tag(:hr)()
 function mdconvert(i::Markdown.Image, parent; kwargs...)
     @tags video img a
 
-    if occursin(r"(.webm|.mp4|.ogg|.ogm|.ogv|.avi)$", i.url)
+    if occursin(r"\.(webm|mp4|ogg|ogm|ogv|avi)$", i.url)
         video[:src => i.url, :controls => "true", :title => i.alt](
             a[:href => i.url](i.alt)
         )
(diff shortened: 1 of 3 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -1030,7 +1030,11 @@ mdconvert(h::Markdown.Header{N}, parent; kwargs...) where {N} = DOM.Tag(Symbol("
 
 mdconvert(::Markdown.HorizontalRule, parent; kwargs...) = Tag(:hr)()
 
-mdconvert(i::Markdown.Image, parent; kwargs...) = Tag(:img)[:src => i.url, :alt => i.alt]
+mdconvert(i::Markdown.Image, parent; kwargs...) = if endswith(i.url, ".webm")
+    Tag(:video)[:src => i.url, :width => "800", :controls => "true", Symbol("data-title") => i.alt]
```
mortenpi: Unless I am missing something, would it not be better to style the `<video>` tag with CSS, rather than using the `width` attribute? I think we just should apply the same style as for `img`, i.e. just set `max-width`:

https://github.com/JuliaDocs/Documenter.jl/blob/6e35c14093577a23a78952e097b476bd95884543/assets/html/documenter.css#L82-L84

src/Writers/HTMLWriter.jl
```diff
@@ -1030,7 +1030,11 @@ mdconvert(h::Markdown.Header{N}, parent; kwargs...) where {N} = DOM.Tag(Symbol("
 
 mdconvert(::Markdown.HorizontalRule, parent; kwargs...) = Tag(:hr)()
 
-mdconvert(i::Markdown.Image, parent; kwargs...) = Tag(:img)[:src => i.url, :alt => i.alt]
+mdconvert(i::Markdown.Image, parent; kwargs...) = if endswith(i.url, ".webm")
```
mortenpi: Would be good to support other formats as well while we're at it I think. WebM (`video/webm`; `.webm`), MP4 (`video/mp4`; `.mp4`) and Ogg (`video/ogg`; `.ogg`, `.ogv`), going by the list [here](https://www.w3schools.com/tags/tag_video.asp)?

src/Writers/HTMLWriter.jl
mortenpi: I am wondering about `data-title` -- why not just use `title`? The latter is a [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title), so it should be correct to apply to a `<video>` tag too?

mortenpi: Not sure we really have browsers out there that do not support `<video>` anymore, but I guess it wouldn't hurt to have fallback content in the tag too, say with a link to the file?

richiejp:
Thanks for the feedback. I was on holiday, so couldn't respond straight away.

src/Writers/HTMLWriter.jl
```diff
@@ -1030,7 +1030,17 @@ mdconvert(h::Markdown.Header{N}, parent; kwargs...) where {N} = DOM.Tag(Symbol("
 
 mdconvert(::Markdown.HorizontalRule, parent; kwargs...) = Tag(:hr)()
 
-mdconvert(i::Markdown.Image, parent; kwargs...) = Tag(:img)[:src => i.url, :alt => i.alt]
+function mdconvert(i::Markdown.Image, parent; kwargs...)
+    @tags video img a
+
+    if occursin(r"(.webm|.mp4|.ogg|.ogm|.ogv|.avi)$", i.url)
```
mortenpi:
```suggestion
    if occursin(r"\.(webm|mp4|ogg|ogm|ogv|avi)$", i.url)
```

The unescaped `.` matches any character, so `foomp4` would match currently.

mortenpi: Just pushed this from the command line. It seems that you can't accept suggestions even if you have push rights to a PR :man_shrugging:

mortenpi:
@richiejp Thanks for the contribution! Will merge once the tests complete.

PR #1036, opened 2019-06-08
Remove deprecations from test/examples/make.jl

mortenpi:
Make sure it's `format=Markdown()`, not `format=:markdown` and that the `assets` keyword is passed to `HTML`, not `makedocs`. Also, remove the `builds/` directory every time `make.jl` runs to make sure we get clean builds.

@fredrikekre Just to make sure, having DocumenterMarkdown as a test dep won't affect normal `pkg> add` installations, right?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,6 +20,7 @@ JSON = "0.19, 0.20"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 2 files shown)


fredrikekre:
> Just to make sure, having DocumenterMarkdown as a test dep won't affect normal pkg> add installations, right?

Correct

mortenpi:
Thanks, @fredrikekre!

PR #1037, opened 2019-06-12
Add Random as test. dependency

mortenpi:
We do `using Random` in one of the tests. It seems that Julia has become stricter on `master` with standard libraries (see e.g. [this](https://travis-ci.org/JuliaDocs/Documenter.jl/builds/544540753))?


fredrikekre:
>  It seems that Julia has become stricter on master with standard libraries

We accidentally backported (to 1.0.4 and 1.1.1) a broken bugfix that let you use recursive deps, so it will also fail on e.g. Julia 1.0.3 and 1.1.0.

Issue #981, opened 2019-03-07
Switch to KaTeX


MikeInnes:
KaTeX is much nicer and faster than MathJax and it might be good to switch over. I suspect it's not missing many important features vs mathjax at this point, but if there are downsides we care about it might be good to list them out.

Mentioned in #626.

mortenpi:
We can probably default to KaTeX, although it would be good to have know how their features compare. But it would be nice to be able to configure this, so that people could still keep using MathJax if they need to, so this is slightly blocked by #982.

tlienart:
Re feature comparison, there's this nice site that gathers what does and doesn't work with KaTeX: https://utensil-site.github.io/available-in-katex/ it may be relevant if for instance using KaTeX by default and suggesting a way for users who would want some of the few non-supported feature to switch to MathJax.

mortenpi:
![](https://github.trello.services/images/mini-trello-icon.png) [Math rendering engine](https://trello.com/c/iLd2NGBA/7-math-rendering-engine)

briochemc:
I wholeheartedly support this! KaTeX looks so much nicer. It would make all math-intensive documentations a lot more attractive. Without KaTeX I am tempted to replace all the math with code, e.g., I'll prefer to display `f(x,y) = 0` rather than a misaligned, wrongly-sized <a href="https://www.codecogs.com/eqnedit.php?latex=\inline&space;f(x,y)=0" target="_blank"><img src="https://latex.codecogs.com/svg.latex?\inline&space;f(x,y)=0" title="f(x,y)=0" /></a> (using codecogs in this case). And looking at that list of things that don't render with KaTeX, I would say most are either  rare or have workarounds...

Anyway, I don't know if this is helpful but I saw that [JuDoc.jl](https://github.com/tlienart/JuDoc.jl) uses KaTeX to generate documentation. Maybe @tlienart has some useful advice on how to get this done?

Also is there any projected timeframe for this? 🙂

tlienart:
Given the label just before your message it looks like this is on the todo list for the google summer of code. 

In terms of insight, I doubt the guys here really need a hand. My only insights from JuDoc are firstly that katex's `auto-render` script is very useful to get all the math blocks rendered in one shot and secondly that all math elements can be pre-rendered to HTML using katex+node which removes the need for loading the javascript library on all pages containing maths. 

Pre-rendering is really nice I think with the downside that it requires users to have node installed on their machine.

[...]

Issue #618, opened 2018-01-04
Fonts and CSS for Unicode Art


samoconnor:
Here is a patch that improves rendering of ~~ASCII~~ Unicode Art:

```patch
--- a/assets/html/documenter.css
+++ b/assets/html/documenter.css
@@ -21,7 +21,8 @@ body, input {
 }

 pre, code, kbd {
-  font-family: 'Roboto Mono', Monaco, courier, monospace;
+  font-family: 'Menlo', 'Roboto Mono', Monaco, courier, monospace;
+  line-height: 1.3em;
   font-size: 0.90em;
 }
```

Maybe 'Roboto Mono' does not have glyphs for the arrow heads and is using a fallback font?

Before:
<img width="587" alt="roboto" src="https://user-images.githubusercontent.com/8266283/34551521-ef73e806-f16e-11e7-9127-36f80b171456.png">

After:
<img width="593" alt="menlo" src="https://user-images.githubusercontent.com/8266283/34551522-efbd5996-f16e-11e7-9f9d-fa878f6e16e5.png">

mortenpi:
Good catch, although I am not sure what to do with it. I believe you are right about this being a problem with fallback for some glyphs. Could you check what fonts get rendered on your system in the `<code>` tag?

If you have Chrome, you can check [rendered fonts with DevTools](https://developers.google.com/web/updates/2013/09/DevTools-answers-What-font-is-that) (note that it lives under the "Computed" tab and not "Styles" on recent Chrome versions). I get

> Roboto Mono—Network resource(737 glyphs)
> Liberation Mono—Local file(459 glyphs)
> DejaVu Sans Mono—Local file(1 glyph)

and on my system (Ubuntu Linux) the diagram renders fine.

Menlo does not appear to be available on any CDNs and is macOS-specific as far as I can tell (although the font appears to be an open font?). We could have Menlo as one of the fallback fonts (but not primary) to fix this for macs, but ideally we'd have a solution that ensures that it works across all platforms.

samoconnor:
When I load [this page](https://github.com/google/fonts/issues/1408) in Chrome and target the inspector on the `<code>` element it tells me:
> Menlo—Local file(1215 glyphs)

It says the style is`font-family: "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;`

Fontbook tells me that Menlo has 3,157 glyphs while Roboto Mono has 1,042.

[...]

PR #1043, opened 2019-06-24
Implement a new HTML front end with Bulma

### LINKED ISSUE #626: Bootstrap for styling
See also #615, it proves useful to have some basic Bootstrap styling when adding new features to the HTML part of Documenter.jl.

I've created a new repo (https://github.com/haampie/JuliaDocsAssets) with just some HTML and CSS, preliminary example here: https://haampie.github.io/JuliaDocsAssets/

There's some tooling involved, currently via node/npm packages (brunch and sass-brunch). Maybe it's therefore best to keep the assets in a separate repo? If someone wants to theme things, then it's possible to clone that repo, change the bootstrap variables, generate the CSS, and add the minified CSS to their own repo.

Features & todo:
- [x] Linkable paragraphs (useful for sharing and in search; immediately jump to the right paragraph) with pure CSS highlight animation (e.g. https://haampie.github.io/JuliaDocsAssets/#p6)
- [x] Modern CSS (flexbox for the layout & `position: sticky` for sidebars)
- [x] SVG logo rather than PNG.
- [x] Math via KaTeX
- [x] Code highlighting
- [ ] Mobile version
- [ ] Search field

### LINKED ISSUE #638: Navigation sidebar not usable on iPad
When browsing any site generated with Documenter on an iPad Pro 12.9 inch, the navigation sidebar on the left is hardly usable: Swiping over it doesn't smoothly scroll in the sidebar. Sometimes it scrolls (non-smoothly), and sometimes instead  the main content on the right scrolls; a rather frustrating experience.

### LINKED ISSUE #722: Default style overhaul
It is probably fair to say that the default style/CSS of Documenter’s HTML build could be improved. The current style is based on the old Sphinx style with minor modifications. Over time there have been various suggestions (e.g. #667, #668) and issues (e.g. #618, #638) reported that have to do with the style.

While improvements are definitenly most welcome, small design changes can be a bit problematic since whether or not they are an improvement is pretty subjective. Instead, doing a single bigger overhaul of the general look might be a better approach, since it would allow us to keep the whole design consistent etc.

This issue is very much up for grabs. Also, I'd be very happy to delegate the visual design decisions to someone else (but it should probably be a single person, since I don't think design-by-committee really works).

The philosophy is that Documenter’s default style should be the look-and-feel of the [main Julia manual](https://docs.julialang.org/). So this would directly affect the main manual in a pretty significant way.

**For potential contributors:** a good way to approach this would be to create HTML/CSS mockup first, based on e.g. a few pages in the Julia manual, which we could then iterate on.

There was also a suggestion to use [Bootstrap](https://getbootstrap.com/) as the underlying framework (#626), with some work done on that front. The use of a framework is, I think, a good idea, since it would do a lot of work for us (e.g. mobile UI/UX).

### LINKED ISSUE #724: “When was the documentation last updated?”
If I want to know when a documentation set was built, where should do I look? Say, is there anywhere where a string says “docs generated 2016-12-34” or something? “Stable” is OK, but is there anything in more detail?

(In earlier times, they used to include a [colophon](https://en.m.wikipedia.org/wiki/Colophon_\(publishing\)).)

There is a sneaky way of finding out, I just remembered. But it isn’t obvious...

### LINKED ISSUE #1030: Style info admonitions correctly
The [manual says that `info` and `note` are valid admonition categories and are equivalent](https://docs.julialang.org/en/v1/stdlib/Markdown/#Admonitions-1). But we style them differently in the HTML output (rather, `info` does not get styled and is therefore rendered in grey). We should probably apply the same blue note styling to it.

A grey `info` admonition in the wild: https://julialang.github.io/Pkg.jl/v1/compatibility/#**6.**-Compatibility-1

### LINKED ISSUE #1085: Note section in new darkly theme
Currently "Note" section under `darkly` theme uses the same colors as under `documenter` theme. This is not something critical, but maybe it would be better to make a bit darker (to match the general colosf or `darkly` theme). This is not a priority issue.

mortenpi:
Demo:

* Documenter's manual: https://mortenpi.eu/gsoc2019-mockups/bulma/v13/documenter/
* The Julia manual: https://mortenpi.eu/gsoc2019-mockups/bulma/v13/julia/

TODO:

- [x] Need to add the SCSS sources, requires basic APIs in DocumenterTools
- [x] Handle tables overflowing
- [x] Fix the search page
- [x] Revise sidebar menu styling, layout (close #638)
- [x] Try out a few other fonts
- [x] Footnote "back-references" (click on def, jump to location)
- [x] Colophon (for now, it's implemented as a settings dialog; close #724)
- [x] Flakiness when clicking on search results? (Seems fine now)
- [x] It seems that sidebar vertical height can be buggy on mobile when the _browser_ top bar disappears.
- [x] Page Down does not scroll the main body by default (since the scrollbar is actually for an inner `<div>`; won't fix now, but might revisit the use of bulma-dashboard)
- [x] Footer theming
- [x] Fiddle with the color scheme (close #1085)
- [x] Syntax highlighting in admonitions in the dark theme.
- [x] Different logo for dark theme (#1090)
- [x] Fix `note` admonition styling (close #1030)
- [ ] Link to point to the GitHub repo root
- [ ] List margins

Close #626, close #722. Also addresses #981, ~#618~.


assets/html/scss/documenter/layout/_sidebar.scss
```diff
@@ -0,0 +1,246 @@
...
+        text-rendering: auto;
+        line-height: 1;
+
+        font-size: 0.75*$size-normal;
+        margin-left: 1rem;
+        margin-top: auto;
+        margin-bottom: auto;
+        &::before {
+          @include font-awesome;
+          content: "\f054";
```
pfitzseb: If you make it specific to levels > 1, then you can put something like
```
margin-left: -1em;
margin-right: 0;
background: $sidebar-background;
```
here to get what I showed on Slack.

cormullion:
I love this, especially the Dark Mode, which is always interesting to design for. I wonder whether the Notes (if not the Warnings) could be dimmed a bit, so as not to be so distracting...?

docs-darkmode

Also, I think the Julia logo could be inverted? 

Anyway, it looks great!

Datseris:
Hello! Just came here to add exactly the same comment as @cormullion , but of course cormullion being cormullion did it much before me! I also thing the warning boxes should use a bit more saturated colors. It makes it easier on the eyes. Also, a light gray background over the text in the white version would help (as at the moment you cant really tell when the block ends)

mortenpi:
> Also, I think the Julia logo could be inverted?

Good idea, although I need to think about how to implement it (the logo and the theme are not technically related to each other). Do you have an SVG of the inverted logo by any chance though?

pfitzseb:
> Do you have an SVG of the inverted logo by any chance though?

See [here](https://github.com/JuliaDocs/DocumentationGenerator.jl/tree/master/landingpage/src/assets) or [here](https://github.com/JuliaLang/julia-logo-graphics/tree/master/images).

cormullion:
https://github.com/JuliaLang/julia-logo-graphics/tree/master/images

mortenpi:
I left the colors slightly more saturated than what @cormullion suggested at the moment:

admonitions

I feel that they are dimmed enough, yet nicely retain some colorfulness. Overall, I think I managed to take care of most color issues now (not the logo yet though). Feedback most welcome though. Here is the latest demo:

> https://mortenpi.eu/gsoc2019-mockups/bulma/v10/documenter/showcase/
> https://mortenpi.eu/gsoc2019-mockups/bulma/v10/julia/

cormullion:
Lower contrast is great now. 

(_de gustibus non est disputandum_) 😅

cormullion:
On the Settings menu, is that "generated by Documenter... " the date of the document's generation (fixing #724 yay!), or something to do with the settings?

mortenpi:
> On the Settings menu, is that "generated by Documenter... " the date of the document's generation (fixing #724 yay!), or something to do with the settings?

That's when the document was generated (you can hover to see the time too). I put it into the settings dialog since I couldn't really figure out how to make it look readable and unobtrusive in the footer.

cormullion:
Perhaps you could include the document name there, so that it's clear that it's this specific document, not the Documenter HTML theme, or whatever?

mortenpi:
In case anyone is brave enough to review any of this, then now would be the time. I will appreciate any feedback :slightly_smiling_face: Generally, I think it's in good enough shape to be merged (we'll probably have some fixes in follow-up PRs), but I'll leave it open for a few more days.

cormullion:
Looks really good! Worth upgrading for sure. Great job!

The version popup at the bottom doesn't always reflect the selection? I chose "Documenter", and it still says "Julia" - and there are some Markdown `#` there too...
 
Screenshot 2019-08-08 at 11 27 07

.travis.yml
```diff
@@ -20,9 +20,11 @@ after_success:
 
 jobs:
   include:
+    - stage: "Themes"
+      script:
+        - julia --project=test/themes -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd()))'
+        - julia --project=test/themes test/themes/themes.jl
     - stage: "Documentation"
-      julia: 1.0
-      os: linux
```
fredrikekre: Will this not deploy from multiple workers?

mortenpi:
> I noticed some Markdown `#` - is that correct?

Yup, those are correct. Although they don't come from Markdown -- they are just bullet points for the in-page headings, and intentionally look different from the rest of the sidebar in order to distinguish them from the inter-page links.

.travis.yml
mortenpi: Nope, build stages [by default "inherit the first value of the array that defines a matrix dimension"](https://docs.travis-ci.com/user/build-stages/#build-stages-and-build-matrix-expansion). So in our case `os: linux` and `julia: 1.0`.

ararslan:
Did the fonts change? The regular font seems lighter (is it gray maybe?) and the monospace looks like the fallback Courier kind of thing.

mortenpi:
The main text is still Lato, but the color is indeed slightly lighter. Foir the monospace I am indeed using browser default at the moment. Partly this might help with Unicode alignment problems (#618), but mainly it's because that is the Bulma default setting.

ararslan:
> but the color is indeed slightly lighter

Okay good, so I'm not going crazy after all 😄 

> mainly it's because that is the Bulma default setting

Are you thinking you'll revisit a centralized monospace font here before this goes live? The fallback fonts are a bit ugly.

mortenpi:
> The fallback fonts are a bit ugly.

What OS/browser are you on? Do you know what the fallback is in your case? You can quickly check that by looking at "rendered fonts" in Chrome or Firefox dev tools.

ararslan:
Seems to be Courier. At the moment I'm using Firefox on macOS.

mortenpi:
Hmm, might be better to go back to Roboto Mono then, but have a slightly better fallback behavior than before:

```css
font-family: 'Roboto Mono', 'SFMono-Regular', 'Menlo', 'Consolas', 'Liberation Mono', 'DejaVu Sans Mono', monospace;
```

Or, alternatively, we could try to target the OS default monospace fonts:

```css
font-family: 'SFMono-Regular', 'Menlo', 'Consolas', 'Liberation Mono', 'DejaVu Sans Mono', monospace;
```

I don't really have a very strong opinion myself. The alignment problem of #618 is pretty much unsolvable, and not really a problem for the Julia manual anyway, so we can just focus on which one looks better.

helgee:
> Yup, those are correct. Although they don't come from Markdown -- they are just bullet points for the in-page headings, and intentionally look different from the rest of the sidebar in order to distinguish them from the inter-page links.

Maybe a different bullet, i.e. one that is not the same as Julia's comment character, would be better? ;-)

BTW: This looks great 👍

tkf:
It seems that syntax highlighting is broken:

> image

from https://mortenpi.eu/gsoc2019-mockups/bulma/v7/documenter/man/syntax/#@ref-link-1

This happens if:

* You never visited the page (or cleared the cache?)
* And never switched to the dark theme.

Switching to the dark theme and then switching back to the default gives me:

> image

mortenpi:
@tkf Do you see the same issue with a newer version too?

> https://mortenpi.eu/gsoc2019-mockups/bulma/v11/documenter/man/syntax/#@ref-link-1

tkf:
Looks good to me.  I didn't realize that the link in the OP was an older version.

mortenpi:
> Maybe a different bullet, i.e. one that is not the same as Julia's comment character and Markdown headers, would be better? ;-)

I personally actually like that it mirrors the Markdown syntax since you create those headings literally with a `#` (well, with `##` normally).

Here are the latest builds (also updated the OP):

> https://mortenpi.eu/gsoc2019-mockups/bulma/v13/documenter/
> https://mortenpi.eu/gsoc2019-mockups/bulma/v13/julia/

helgee:
> I personally actually like that it mirrors the Markdown syntax since you create those headings literally with a `#` (well, with `##` normally).

I see your point and I certainly could get used to it. It just sticks out a bit especially if juxtaposed with other bulleted lists (see below) and like @cormullion, I first interpreted it as a rendering error.

grafik

ararslan:
I didn't even realize the hashes were intentional; I also thought they were an error.

mortenpi:
> I didn't even realize the hashes were intentional; I also thought they were an error.

Ok, three people reporting it means that it's no longer a matter of taste :laughing: Swapped it out for circles:

circles

mortenpi:
I merged this so that any follow-up changes could be made in separate, more manageable PRs. This does not mean that any of the style etc. is final yet -- I opened a dedicated issue to collect feedback: #1091.

PR #1047, opened 2019-06-28
Build non-pages pages too in HTMLWriter

### LINKED ISSUE #1031: Rendering of pages not in pages
Currently, if we have an `.md` file that is not included in the `pages` keyword, we still run the expanders etc., on it (e.g. executing all the at-blocks), but we don't render it in the HTML output. Would be good to be consistent with this -- either run & render, or do neither.

If we would enable rendering of these pages, they would end up as hidden HTML files that are outside of the normal navigation (since the nav. sidebar is defined by `pages`). Not sure what to do with them in the LaTeX output though -- create an appendix?

mortenpi:
They can then be reached via their URL, which is determined from the filename, as for the other pages. But they will not show up in the sidebar nor in the bottom navigation.

If there is no filter applied, these pages may still show up in an at-contents block. In fact, this actually fixes a bug where, in the HTML output, an at-contents block would list such pages, even though the writer does not produce any `.html` files, leading to broken links in the output.

My use case is to have hidden draft pages (that I actually want to use for my [GSoC blog](http://mortenpi.eu/gsoc2019/)).

Close #1031.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "0.22.4"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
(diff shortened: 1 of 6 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.22.4"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
fredrikekre: Is just for `isnothing`? Can't we just type out `i === nothing` instead? :)

PR #1048, opened 2019-06-28
Update to new donation URL.

fredrikekre:


ViralBShah:
I made a test donation to send $5 to Documenter. The credit card process went through.

Issue #1053, opened 2019-07-01
Integrate doctest with testsets


davidanthoff:
This issue is a follow up on https://github.com/JuliaDocs/Documenter.jl/pull/774 and https://github.com/JuliaDocs/Documenter.jl/issues/198, and what a nice public API for that functionality might look like.

My own preference would be something like this in my `runtests.jl`:
```julia
using Documenter, Test, MyPackage

@testset "MyPackage" begin

include("test_featureA.jl")

Documenter.run_all_doctests("MyPackage")

end
```

I.e. ideally I don't really want to deal with any subtleties at all, but would just like to call one function, and have it sort out everything. In terms of testset integration, my preference would be if `run_all_doctests` started a new (child) testset and then put each doctest into that testset.

The name for the function is not great, this is more an idea how the API should feel like.

mortenpi:
There are actually two separate ideas here: (1) provide a simplest possible API for `runtests.jl`, and (2) integrate it with testsets. I'll dedicate this issue for the testset integration and and leave the API discussion for #1051.

Regarding the testset integration: yep, I think it's a good idea. For the time being though, it's up for grabs -- it needs a slightly more sophisticate implementation for `doctest`.

fingolfin:
What exactly is left to be done here, given that `doctest(MyPackage)` runs all the doctests? We've been using that to run doctests inside of `runtests.jl` for several years now.

So perhaps it can be closed? Otherwise someone could perhaps clarify what is missing...

----

The main slightly annoying thing *I* know of is that a few global settings like `DocMeta.setdocmeta` and `doctestfilters` need to be duplicated between `docs/make.jl` and `test/runtests.jl`.

For us, we have essentially resolved this by establishing a pattern like this:
```julia
DocMeta.setdocmeta!(MyPackage, :DocTestSetup, MyPackage.doctestsetup(); recursive=true)
doctest(MyPackage)
```
where `MyPackage.doctestsetup` returns the code block.

This then has the added advantage that we can also reuse it in our `.md` files like so:
````
```@meta
CurrentModule = MyPackage
DocTestSetup = MyPackage.doctestsetup()
```
````
(It is still annoying that I have to remember to duplicate this in each `.md` file or else risk forgetting it in some).


I am not sure whether I should try to derive feature requests for this. But perhaps it would be worth documenting this workflow somewhere, perhaps some other people would benefit from it?

mortenpi:
[...]

PR #1054, opened 2019-07-02
Generalize doctest() API

### LINKED ISSUE #1051: Doctest manual pages too with doctest?
#774 brought the `doctest` function, but it can only check the doctests in docstrings at the moment. It would be worth looking into whether there are any issues with trying to doctest manual pages too in this lightweight manner.

### LINKED ISSUE #1052: Fix doctesting docs
The recommended way to include doctests into the test suite with `doctest = :only` actually breaks the docs (`deploydocs` runs now as part of the test suite, but has nothing to deploy, so it deletes the docs). So that needs to be rethought.

mortenpi:
Updates the API. You can now simply call `doctest(::Module)`, and it will doctest both docstrings _and_ manual pages, e.g.:

```julia
using Test, Documenter, MyPackage
@testset "MyPackage" begin
    doctest(MyPackage)
end
```

For more control, one can also use the `doctest(source, modules)` method, which does the heavy lifting (e.g. to test a set of modules, but no pages, you can call `doctest(nothing, [Module1, Module2, ...])`).

Finally, it also adds the `fix` keyword to the methods, to run the doctest fixing machinery. This actually means you can fix doctests pretty easily in the REPL now:
```
julia> using Documenter, MyPkg
julia> doctest(MyPkg, fix=true)
```

cc @davidanthoff You don't happen to have any suggestions for packages where we could test this in the wild?

Should also fix the `dev/` docs for Documenter, since we're now using the `doctest` function in `runtests.jl`.

Close #1051, close #1052

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 * ![Feature][badge-feature] `makedocs` now accepts the `doctest = :only` keyword, which allows doctests to be run while most other build steps, such as rendering, are skipped. This makes it more feasible to run doctests as part of the test suite (see the manual for more information). ([#198][github-198], [#535][github-535], [#756][github-756], [#774][github-774])
 
-* ![Feature][badge-feature] Documenter now exports the `doctest` function, which verifies the doctests in all the docstrings of a given module. This can be used to verify docstring doctests as part of test suite. ([#198][github-198], [#535][github-535], [#756][github-756], [#774][github-774])
+* ![Feature][badge-feature] Documenter now exports the `doctest` function, which verifies the doctests in all the docstrings of a given module. This can be used to verify docstring doctests as part of test suite, or to fix doctests right in the REPL. ([#198][github-198], [#535][github-535], [#756][github-756], [#774][github-774], [#1054][github-1054])
 
 * ![Feature][badge-feature] `makedocs` now accepts the `expandfirst` argument, which allows specifying a set of pages that should be evaluated before others. ([#1027][github-1027], [#1029][github-1029])
 
@@ -364,6 +364,7 @@
 [github-1034]: https://github.com/JuliaDocs/Documenter.jl/pull/1034
 [github-1037]: https://github.com/JuliaDocs/Documenter.jl/pull/1037
 [github-1047]: https://github.com/JuliaDocs/Documenter.jl/pull/1047
+[github-1054]: https://github.com/JuliaDocs/Documenter.jl/pull/1054
 
 [documenterlatex]: https://github.com/JuliaDocs/DocumenterLaTeX.jl
 [documentermarkdown]: https://github.com/JuliaDocs/DocumenterMarkdown.jl
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@
 
 * ![Experimental][badge-experimental] ![Feature][badge-feature] The current working directory when evaluating `@repl` and `@example` blocks can now be set to a fixed directory by passing the `workdir` keyword to `makedocs`. _The new keyword and its behaviour are experimental and not part of the public API._ ([#1013][github-1013], [#1025][github-1025])
 
+## Version `v0.22.5`
+
+* ![Maintenance][badge-maintenance] Fix a test dependency problem revealed by a bugfix in Julia / Pkg. ([#1037][github-1037])
+
 ## Version `v0.22.4`
 
 * ![Bugfix][badge-bugfix] Documenter no longer crashes if the build includes doctests from docstrings that are defined in files that do not exist on the file system (e.g. if a Julia Base docstring is included when running a non-source Julia build). ([#1002][github-1002])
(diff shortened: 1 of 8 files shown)


davidanthoff:
I'll create a branch in Query.jl to test this!

davidanthoff:
I tried it with https://github.com/queryverse/Query.jl/pull/269, and it all seems to work. The thing won't work for the CI builds because they don't pick up this branch here of Documenter, but locally it all looked great.

One thing I particularly like is that this does _not_ pick up anything from the `docs/Project.toml` or `docs/Manifest.toml`, but instead it all uses the env that is specified for the tests in the main package `Project.toml`. I think that is exactly right!

fredrikekre:
> I think that is exactly right!

Why?

davidanthoff:
I'm not checking in a `docs/Manifest.toml`, so I generally feel that shouldn't play a role.

I think one important aspect is that if my active env is say shared `foo`, and I have deved my package in that env (and potentially other packages as well), then I would want the doc test run to pick up any other packages that I've deved in `foo` for that test run, in the same way that a test run picks up specific versions of other packages that I have in `foo`. I'm actually not sure now whether that is happening with the current PR here for the doctests, but generally that is what I'd expect, and that seems somewhat easier if the doctests run in the same env as the general tests? Not sure.

fredrikekre:
So you would have to add all doc dependendencies as test dependencies then?

davidanthoff:
Yes, that is what I did for the Query.jl branch. I could see this all being more elegant with some of the sub-project stuff that is coming?  But for now this seems fine.

davidanthoff:
One weird thing I ran across: when I ran the doctests as part of `runtests.jl` locally, I had to change all the doctests so that the output looked like `1×2 DataFrame` instead of the `1×2 DataFrames.DataFrame`. I then opened a PR against Query.jl to change that in all doctests (https://github.com/queryverse/Query.jl/pull/268/files), but in the CI run these now all failed.

Not sure what is going on, and I never understood when I have to add the module name in the output, but it looks a little as if one needs to do this differently when the doctests are run from `runtests.jl`?

fredrikekre:
Thats because DataFrames was sometimes loaded in `Main`, see e.g. https://github.com/JuliaLang/julia/pull/29466

davidanthoff:
> Thats because DataFrames was sometimes loaded in Main, see e.g. JuliaLang/julia#29466, you should be able to work around it by loading that into Main unconditionally.

Cool, thanks for pointing me to that, I always wanted to know what is going on there :)

So the story then is that in `runtests.jl` I have a `using DataFrames`, but not in `docs/make.jl`, and that leads to different output of the doctests depending on how you run them.

The short term fix is simple: I'll just add `using DataFrames` to `docs/make.jl`. As a general solution not great, but I guess this really needs to be fixed in base...

tkf:
> ```julia
> using Test, Documenter, MyPackage
> @testset "Doctesting" begin
>     @test doctest(MyPackage)
> end
> ```

Can it be just `doctest(MyPackage)` instead of `@testset "Doctesting"; @test doctest(MyPackage); end`?  It would be nice to get N failures for N failing doctests, rather than just a binary output.

(PS: It's great to see that doctests is decoupled from documentation!)

mortenpi:
> Can it be just `doctest(MyPackage)` instead of `@testset "Doctesting"; @test doctest(MyPackage); end`? It would be nice to get N failures for N failing doctests, rather than just a binary output.

That's a good idea, but I'll leave the implementation for another time, since it's a bit of work to make it happen (#1053).

tkf:
How about defining it as `doctest(m::Module) = @test doctest_impl(m)` for now?  This way, users don't have to change anything at the call sites if you make test results more detailed at some point.

mortenpi:
> This way, users don't have to change anything at the call sites if you make test results more detailed at some point.

That's a fair point. I am inclined to keep the `doctest` function as is though, so that there would a simple function you can call that just gives you `true`/`false` (and allows fixing the doctests).

What about `@doctestset MyPackage`? Would then currently expand into:

```julia
@testset "Doctests: MyPackage" begin
    @test doctest(MyPackage)
end
```

tkf:
> a simple function you can call that just gives you `true`/`false`

My honest impression is that it's not clear what the target user of `doctest(::Module) :: Bool` is.  It's not handy for end-users because they need to use `@testset` and `@test`.  It's not really an API suitable for testing utility packages because the output is too coarse-grained.  But maybe this is not going to be a problem in practice.  Error reporting from `makedocs` has been already quite useful.

FYI, I created a test helper tool https://github.com/tkf/Aqua.jl#quick-usage (though not related to doctest at all).  I use the API like `doctest(m::Module) = @test doctest_impl(m)` I suggested above.  I've never needed `true`/`false` output and I'm enjoying the simplicity of the test setup.  Of course, this could be just due to my workflow.

tkf:
> What about `@doctestset MyPackage`?

I think it makes sense if you decide to go with `doctest(::Module) :: Bool`.

mortenpi:
@tkf You won me over. The `-> Bool` API would not actually be particularly useful indeed. The only other use I can see for the `doctest` function, outside of `runtests.jl`, is in the REPL and there is no problem if it also returns a testset there (in fact, it might be preferred).

tkf:
Great. Thank you for being open minded!

PR #1066, opened 2019-07-11
Use DejaVu fonts in PDF builds

### LINKED ISSUE #803: Unicode ∀, ∃ etc. missing in with Roboto Mono in the PDF
I suggest changing the monospace font used by the PDF documentation, so the special Unicode strings displayed in e.g. Manual -> Strings -> Unicode and UTF-8.

mortenpi:
DejaVu Sans Mono seems to be available out of the box on Linux (at least on Ubuntu), so no need to worry about installing it on the Docker image etc. It provides good Unicode coverage in my experience -- this is the fallback font on Ubuntu and I have not noticed any missing symbols. From a font quality perspective, I don't actually see much difference between it and Roboto Mono.

Fix #803. Once downstreamed will fix JuliaLang/julia#31048.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@
 
 * ![Enhancement][badge-enhancement] Videos can now be included in the HTML output using the image syntax (`![]()`) if the file extension matches a known format (`.webm`, `.mp4`, `.ogg`, `.ogm`, `.ogv`, `.avi`). ([#1034][github-1034])
 
+* ![Enhancement][badge-enhancement] The PDF output now uses DejaVu Sans Mono as the monospace font to provide better Unicode coverage. ([#803][github-803], [#1066][github-1066])
+
 * ![Bugfix][badge-bugfix] The HTML output now outputs HTML files for pages that are not referenced in the `pages` keyword too (Documenter finds them according to their extension). But they do exists outside of the standard navigation hierarchy (as defined by `pages`). This fixes a bug where these pages could still be referenced by `@ref` links and `@contents` blocks, but in the HTML output, the links ended up being broken. ([#1031][github-1031], [#1047][github-1047])
 
 * ![Bugfix][badge-bugfix] `makedocs` now throws an error when the format objects (`Documenter.HTML`, `LaTeX`, `Markdown`) get passed positionally. The format types are no longer subtypes of `Documenter.Plugin`. ([#1046][github-1046], [#1061][github-1061])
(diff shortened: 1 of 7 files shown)


mortenpi:
I also changed the sans-serif font to DejaVu Sans, since Lato is missing a lot of the necessary Unicode characters too. I would say that it looks good enough and my current feeling is that decent Unicode coverage is more important than having the most beautiful font.

Here is an example of the main manual compiled with the DejaVu fonts: [TheJuliaLanguage.pdf](https://github.com/JuliaDocs/Documenter.jl/files/3385228/TheJuliaLanguage.pdf)

Just a note for future: Tex, unfortunately, does not seem to support fallback fonts (i.e. that if a particular character is missing, it would render it from a different font). The closest you have is [ucharclasses](http://www.ctan.org/tex-archive/macros/xetex/latex/ucharclasses) for XeTeX, but it has two issues: (1) we are using LuaTex and it's XeTeX-only, and (2) it only allows setting different fonts for classes of characters but [does not provide fallbacks for individual characters](https://tex.stackexchange.com/questions/323575/how-can-i-specify-a-chain-of-fallback-fonts-in-xelatex).

mortenpi:
DejaVu Sans Mono does not render the ⊻ symbol, so I actually created our homebrew fallback system for this case. Instances of that character get replaced by the `\unicodeveebar` command in the TeX source, which then renders it DejaVu Sans instead.

In principle, this is pretty easily extensible to other missing Unicode characters as well, should they come up.

Here is an example build which shows off all the Unicode characters reported in #803: [Unicodetest.pdf](https://github.com/JuliaDocs/Documenter.jl/files/3388304/Unicodetest.pdf)

@fredrikekre Any thoughts about this hack?

Ronneesley:
Hello @mortenpi,

           I noticed that in [PDF manual version 1.2](https://raw.githubusercontent.com/JuliaLang/docs.julialang.org/assets/julia-1.2.0.pdf), some characters aren't appearing, like:

**Function composition and piping**

The symbol: ∘

```julia
(sqrt ∘ +)(3, 6)
```

**Repeated Evaluation: Loops**

The symbol: ∈

```julia
for s ∈ ["foo","bar","baz"]
     println(s)
end
```

       Is it resolved? Why new manual's version doesn't include the solution? Thanks.

mortenpi:
The 1.2.0 manual was still built with an older Documenter, so it did not have these fixes. With the latest Documenter, they render fine:

symbols

We could potentially backport JuliaLang/julia#32376 to 1.2.1. Otherwise, 1.3.0 will definitely have this fixed.

Issue #988, opened 2019-03-24
Deployment on other CI systems


mortenpi:
Currently we officially only support deploying GitHub repositories on Travis CI to GitHub pages. `deploydocs` also relies on Travis-specific environment variables mostly. We should probably support more Git hosting and CI systems (documentation + generalized API where necessary). The following come to mind:

- [ ] GitLab: has its own CI and Pages
- [ ] BitBucket: has its own CI (Pipelines) and Pages

In the meantime, there are a few examples of how to work around current limitations:

* An example Documenter setup on BitBucket using BitBucket Pipelines and BitBucket Pages: https://bitbucket.org/mortenpi/documenterexample.jl/
* An example of Julia and Documenter on GitLab: https://gitlab.com/tkpapp/GitlabJuliaDemo.jl

tkf:
FYI I added GitLabPages plugin to use Documenter.jl in GitLab repository created by PkgTemplates.jl: https://github.com/invenia/PkgTemplates.jl/pull/54

However, it only publishes `dev` document.  It would be nice if Documenter.jl supports versioned documentation on GitLab pages.

mortenpi:
How does GitLab handle tags? Can you run CI on tags?

tkf:
Yes.  It looks like you can detect it with `CI_COMMIT_TAG` https://docs.gitlab.com/ee/ci/variables/README.html#predefined-environment-variables

I think a harder (?) issue is that GitLab pages is not a git branch as in GitHub.  You create a directory `public` in your repository and the whole content is reflected to the GitLab pages (IIUC).  It's actually convenient for publishing a single document (see how it's done in PkgTemplates: https://github.com/invenia/PkgTemplates.jl/blob/901a0a3d7139588819bad19acfc891a63bda7637/defaults/gitlab-ci.yml#L20)  However, I think it's a bit more challenging to update a part of it (e.g., `public/v1.2`) since you need to store the whole content somewhere and retrieve it in the CI job.  I guess you can create (say) `gl-pages` branch like `gh-pages` in GitHub and check it out at `public`.  But I'm not familiar enough with the whole GitLab interface to know if there is a better way to do it.

marekdedic:
I was just about to create a new issue suggesting adding support for Circle CI....

I think some of the environment variable checks could be done in the CI configuration - e. g. checking for pull requests isn't necessary if you only run the deploy job on the master branch...

christopher-dG:
[...]

Issue #210, opened 2016-08-19
Streamlining the deployment experience


mortenpi:
As I was setting up Documenter for a package, the following thoughts crossed my mind:

- [ ] It should be possible to add the GitHub deploy key automatically using GitHub's API.
- [ ] In `genkeys`, `travis` asks for the password even if you are already logged in. We could check `travis whoami` to decide when to skip `travis login`.
- [ ] The branch (e.g. `gh-pages`) should be created automatically if it doesn't exist (`git checkout --orphan` + cleanup).
- [ ] Would be nice if you could also easily deploy from a local machine. It shouldn't see that much use, but it would still be convenient for people who don't use or have disabled Travis, etc.

MichaelHatherly:
Yes, improving on the current experience would be good.

MichaelHatherly:
Ref #144: for some extra discussion on this.

tkelman:
Is the plan to deploy docs to a separate gh-pages or other branch at some point, it just hasn't been done yet? Many packages are starting to use this and generate docs on master under a docs folder, and it adds a fair amount of noise to reviewing code changes.

mortenpi:
I'd say that Documenter is already designed to push to the `gh-pages` of the same repository. This is what all the packages I've seen do and what Documenter's manual recommends.

Do you mean that people are starting use the new feature where GitHub will host doc pages from master's `docs/` directory? If so, then I think that should be discouraged and people should use `gh-pages` instead.

tkelman:
Yes, I've seen several packages doing what looks like hosting html out of master.

MichaelHatherly:
Yeah, using `gh-pages` is the only recommended way of hosting the generated docs. Committing generated docs, whether html or md, to `master` really shouldn't be done. As far as I can tell there aren't too many packages doing that. All the well-known packages with any significant number of stars are doing things correctly, so hopefully that is enough to encourage new packages to do things the right way.

tkelman:
> hopefully that is enough to encourage new packages to do things the right way.

It isn't, I'm seeing an increasing number of packages put their built docs in master. The only way people will do otherwise is if the documentation here explicitly tells them to not do that, I suspect.

MichaelHatherly:
> It isn't, I'm seeing an increasing number of packages put their built docs in master.

[...]

PR #1067, opened 2019-07-13
Check env vars for non-Travis CI platforms

asinghvi17:
Special cases are now Travis, Gitlab, Drone, Cirrus and Appveyor. 

Not all of these CI services provide the same information as Travis 
does, or in the same way, so some massaging of variable contents is 
required.  Appveyor is arguably one of the worse ones here.

This is untested on most of these CI systems at the time this was 
committed; hence, any feedback is appreciated.


asinghvi17:
We will need some way to check which CI service is the "preferred uploader of documentation".  Perhaps we could introduce a keyword argument to `deploydocs`?  `uploader = "TRAVIS"`, could be any of the CI environment variable names.

src/Documenter.jl
```diff
@@ -573,28 +655,30 @@ function deploydocs(;
...
+    should_deploy = repo_ok && pr_ok && tag_ok && branch_ok && key_ok && type_ok && finaldeploy
 
     marker(x) = x ? "✔" : "✘"
     @info """Deployment criteria:
-    - $(marker(repo_ok)) ENV["TRAVIS_REPO_SLUG"]="$(travis_repo_slug)" occurs in repo="$(repo)"
-    - $(marker(pr_ok)) ENV["TRAVIS_PULL_REQUEST"]="$(travis_pull_request)" is "false"
-    - $(marker(tag_ok)) ENV["TRAVIS_TAG"]="$(travis_tag)" is (i) empty or (ii) a valid VersionNumber
-    - $(marker(branch_ok)) ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch)" (if tag is empty)
+    - $(marker(repo_ok)) CI repo slug = "$(repo_slug)" occurs in repo = "$(repo)"
+    - $(marker(pr_ok)) CI pull request indicator = "$(pull_request)" is "false"
```
mortenpi: To be consistent with the `pr_ok = ` line above:
```suggestion
    - $(marker(pr_ok)) CI pull request indicator = "$(pull_request)" is "false" or "False"
```

Could also maybe `lowecase` `pull_request` and say it's case-insensitive.

mortenpi:
Goes a long way towards fixing #988. We're only missing BitBucket now I think.

Regarding GitLab -- the assumption is that you'll still be pushing the pages to GitHub though, since GitLab pages uses a different system?

src/Documenter.jl
```diff
@@ -524,6 +528,7 @@ function deploydocs(;
         devurl = "dev",
         versions = ["stable" => "v^", "v#.#", devurl => devurl],
         forcepush::Bool = false,
+        uploader = "TRAVIS"
```
fredrikekre: Probably better as an enum.

mortenpi: Would have to export all of them then? Or you'd have to do `uploader = Documenter.APPVEYOR`?

Making it a `Symbol` would be another option.

asinghvi17:
> Regarding GitLab -- the assumption is that you'll still be pushing the pages to GitHub though, since GitLab pages uses a different system?

Yes, for now - I'll add a statement in the docs about that.  It seems that with Gitlab CI you only need to push to a different folder, though, so that may not be too hard.  They claim you can use any static site generator to create the website.

src/Documenter.jl
asinghvi17: I think an enum might be better, since we don't want the user giving invalid input...thanks @fredrikekre!  Yeah, we'd probably want to export them - but while building docs I don't think it's going to be a problem.

asinghvi17:
Thanks for the review, @mortenpi and @fredrikekre!
I think this is now good to go from my side.

src/Writers/LaTeXWriter.jl
```diff
@@ -181,6 +201,28 @@ end
...
+                @info "Cirrus CI detected"
+                get(ENV, "CIRRUS_TAG",                "")
+            elseif haskey(ENV, "APPVEYOR")
+                @info "AppVeyor CI detected"
+                get(ENV, "APPVEYOR_REPO_TAG_NAME",    "")
+            end
+            @info "No tag detected..."
+            else
+                ""
+            end
```
mortenpi: The indentation is a bit broken here.

But in any case -- this actually duplicates the code above. Could we refactor the `if` into a function that we can re-use here and above?

I am also wondering if it might be better to make the `@info`s `@debug`s. Although, there might be value in being verbose about this automagic behavior here.

mortenpi:
Was there are a reason, by the way, why you wanted to mark this "experimental"?

src/Writers/LaTeXWriter.jl
asinghvi17: I can make it a function, will take a while though - dont have too much free time today, so maybe in the evening.

asinghvi17:
Just because I haven't tested on all of these systems.

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@
 
 * ![Enhancement][badge-enhancement] Videos can now be included in the HTML output using the image syntax (`![]()`) if the file extension matches a known format (`.webm`, `.mp4`, `.ogg`, `.ogm`, `.ogv`, `.avi`). ([#1034][github-1034])
 
+* ![Enhancement][badge-enhancement] `deploydocs` now reads the appropriate environment variables on Travis CI, Gitlab CI, Cirrus CI, Drone CI, and AppVeyor, as does `LaTeXWriter`. ([#1067][github-1067])
```
mortenpi: > Just because I haven't tested on all of these systems.

Ah, fair enough. I use the experimental badge to mark experimental APIs. I think the API here is solid -- if something doesn't work, it's a bug.

But it is a good idea to mention the lack of testing somewhere. I might suggest:

```suggestion
* ![Enhancement][badge-enhancement] `deploydocs` now reads the appropriate environment variables on Travis CI, Gitlab CI, Cirrus CI, Drone CI, and AppVeyor, as does `LaTeXWriter`. However, be aware that CI systems other than Travis CI have not received extensive testing. ([#1067][github-1067])
```

src/Documenter.jl
```diff
@@ -547,13 +694,22 @@ function deploydocs(;
...
 
+    local cibranch, pull_request, repo_slug, tag, event_type, finaldeploy
+
+    if haskey(ENV, "CI")
+        cibranch, pull_request, repo_slug, tag, event_type, finaldeploy = read_ci_env(true, uploader=uploader)
+    else
+        cibranch = devbranch
+        pull_request = false
+        repo_slug = repo
+        tag = isempty(tag) && read_ci_env()[4]
```
mortenpi: This will set `tag` to `false` when `tag` is specified. You'd want to use the ternary `? :` here instead I think?

Do we want to rely on the CI variables here though, since this is the `haskey(ENV, "CI") === false` branch?

src/Documenter.jl
```diff
@@ -547,13 +694,22 @@ function deploydocs(;
...
+        event_type = "manual"
+        finaldeploy = haskey(ENV, "DOCUMENTER_KEY")
```
mortenpi: When pushing locally, the user will probably have their normal SSH keys set up, so there's no need for `DOCUMENTER_KEY`. Also, `git_push` messes around with `~/.ssh/config`, which is fine on CI, but we might want to avoid that on user's computers.

src/Documenter.jl
```diff
@@ -547,13 +694,22 @@ function deploydocs(;
         return
     end
 
+    local cibranch, pull_request, repo_slug, tag, event_type, finaldeploy
+
+    if haskey(ENV, "CI")
```
mortenpi: Perhaps the logic should be swapped around here: if `tag` is set, then do a special local build, otherwise try to automagically figure things out from CI?

src/Documenter.jl
```diff
@@ -462,12 +598,22 @@ the generated html. The following entries are valied in the `versions` vector:
...
    (as in e.g. `"stable" => "v^"`).
 
+**uploader** determines _which_ CI service will upload the documentation.  Defaults to Travis
+(`TRAVIS`).  Other options are `GITLAB_CI`, `CIRRUS_CI`, `DRONE`, and `APPVEYOR`.
+
+**tag** provides a specific tag, if you want one to be set.  Providing an empty string (if
+building manually) will make the documentation build as if there was no tag; but, if run on
+CI, it will read the appropriate environment variable to determine if a tag exists or not;
+if there is a provided tag, then that will take precedence over the environment variable,
+which will take precedence over the empty string.
```
mortenpi: In what situation do you want to have `tag` be set when running on CI?

asinghvi17: In case I want to build a tag associated with a different package - say MakieGallery building documentation for Makie.  They're versioned differently.  It's the same, I imagine, for any package whose docs are hosted externally.

src/Documenter.jl
asinghvi17: I also wrote a bit about this in https://github.com/JuliaDocs/Documenter.jl/pull/1067#discussion_r303698669, but the idea was that you could version documentation separate from the package.  I did change that around, a bit, though.

src/Documenter.jl
asinghvi17: Done, it now checks for CI before overwriting - otherwise it just uses the user's config.

src/Documenter.jl
asinghvi17: True, I'll just not set tag here then.

src/Writers/LaTeXWriter.jl
```diff
@@ -181,6 +201,12 @@ end
 function writeheader(io::IO, doc::Documents.Document)
     custom = joinpath(doc.user.root, doc.user.source, "assets", "custom.sty")
     isfile(custom) ? cp(custom, "custom.sty"; force = true) : touch("custom.sty")
+
+    tag = if haskey(ENV, "CI")
+            read_ci_env(false)[4]
+          else
+              ""
+          end
```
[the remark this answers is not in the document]
asinghvi17: Done!

src/Writers/LaTeXWriter.jl
```diff
@@ -90,7 +90,27 @@ function render(doc::Documents.Document, settings::LaTeX=LaTeX())
...
+                        elseif haskey(ENV, "CIRRUS_CI")
+                            @info "Cirrus CI detected"
+                            get(ENV, "CIRRUS_TAG",                "")
+                        elseif haskey(ENV, "APPVEYOR")
+                            @info "AppVeyor CI detected"
+                            get(ENV, "APPVEYOR_REPO_TAG_NAME",    "")
+                        end
+                    else
+                        ""
+                    end
```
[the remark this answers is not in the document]
asinghvi17: Done

src/Documenter.jl
```diff
@@ -573,30 +729,43 @@ function deploydocs(;
...
-    @info """Deployment criteria:
-    - $(marker(repo_ok)) ENV["TRAVIS_REPO_SLUG"]="$(travis_repo_slug)" occurs in repo="$(repo)"
-    - $(marker(pr_ok)) ENV["TRAVIS_PULL_REQUEST"]="$(travis_pull_request)" is "false"
-    - $(marker(tag_ok)) ENV["TRAVIS_TAG"]="$(travis_tag)" is (i) empty or (ii) a valid VersionNumber
-    - $(marker(branch_ok)) ENV["TRAVIS_BRANCH"]="$(travis_branch)" matches devbranch="$(devbranch)" (if tag is empty)
-    - $(marker(key_ok)) ENV["DOCUMENTER_KEY"] exists
-    - $(marker(type_ok)) ENV["TRAVIS_EVENT_TYPE"]="$(travis_event_type)" is not "cron"
-    Deploying: $(marker(should_deploy))
-    """
+    if HASKEY(ENV, "CI")
```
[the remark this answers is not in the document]
asinghvi17: Done

src/Documenter.jl
```diff
@@ -524,6 +670,7 @@ function deploydocs(;
         devurl = "dev",
         versions = ["stable" => "v^", "v#.#", devurl => devurl],
         forcepush::Bool = false,
+        uploader::CI_SYSTEM = TRAVIS
```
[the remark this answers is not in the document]
asinghvi17: Done

src/Documenter.jl
```diff
@@ -371,6 +373,138 @@ makedocs(
...
+
+    APPVEYOR
+
+end
+
+"""
+    read_ci_env([returnfinaldeploy::Bool]; uploader::CI_SYSTEM = TRAVIS)
+
+Read the CI environment variables, and return (always) an Array where the first
+five elements are, in order:
```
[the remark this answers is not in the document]
asinghvi17: Done

src/Documenter.jl
```diff
@@ -371,6 +373,138 @@ makedocs(
...
+
+    GITLAB_CI
+
+    CIRRUS_CI
+
+    DRONE
+
+    APPVEYOR
+
+end
```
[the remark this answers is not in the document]
asinghvi17: Done

asinghvi17:
Thanks @mortenpi for the extensive review!  I've added comments in the reviews.

I can separate the local dev changes into another PR (if this one is OK with tests and all), but it will take me some time - I'm sending my laptop out for repairs, and would need to set up the loaner...

mortenpi:
Thanks, @asinghvi17! I also added a few more fixes, mainly to make sure the documentation build passes.

However, this PR still has some issues. E.g. if you build Documenter's docs locally, `deploydocs` fails with some `getindex` error.

I think it would indeed be best to split the `tag` change into a separate PR (sorry for putting more work on your plate). The `tag` change touches a lot of relatively fragile code and having the two changes entangled makes it quite a bit harder to review. So splitting would allow us to merge the env. var changes quickly and then look more closely at the `tag` change.

src/Writers/LaTeXWriter.jl
```diff
@@ -181,6 +182,8 @@ end
 function writeheader(io::IO, doc::Documents.Document)
     custom = joinpath(doc.user.root, doc.user.source, "assets", "custom.sty")
     isfile(custom) ? cp(custom, "custom.sty"; force = true) : touch("custom.sty")
+
+    tag = haskey(ENV, "CI") ? read_ci_env(false)[4] : ""
```
asinghvi17:
```suggestion
    tag = haskey(ENV, "CI") ? read_ci_env(false)[:tag] : ""

File: CHANGELOG.md [hunk 16]
```diff
@@ -26,6 +26,10 @@
 
 * ![Enhancement][badge-enhancement] Videos can now be included in the HTML output using the image syntax (`![]()`) if the file extension matches a known format (`.webm`, `.mp4`, `.ogg`, `.ogm`, `.ogv`, `.avi`). ([#1034][github-1034])
 
+* ![Enhancement][badge-enhancement] `deploydocs` now reads the appropriate environment variables on Travis CI, Gitlab CI, Cirrus CI, Drone CI, and AppVeyor, as does `LaTeXWriter`. However, be aware that CI systems other than Travis CI have not received extensive testing. ([#1067][github-1067])
+
+* ![Enhancement][badge-enhancement] `deploydocs` can now be switched to tag building mode with the `tag` keyword, which overrides the tag information from CI environment variables. ([#1067][github-1067])
```
asinghvi17:
```suggestion

File: CHANGELOG.md [hunk 17]
```diff
@@ -26,6 +26,9 @@
 
 * ![Enhancement][badge-enhancement] Videos can now be included in the HTML output using the image syntax (`![]()`) if the file extension matches a known format (`.webm`, `.mp4`, `.ogg`, `.ogm`, `.ogv`, `.avi`). ([#1034][github-1034])
 
+* ![Enhancement][badge-enhancement] `deploydocs` now reads the appropriate environment variables on Travis CI, Gitlab CI, Cirrus CI, Drone CI, and AppVeyor, as does `LaTeXWriter`. However, be aware that CI systems other than Travis CI have not received extensive testing. ([#1067][github-1067])
+
```
asinghvi17:
```suggestion
```

asinghvi17:
@mortenpi , sorry for the delay on this.  I've removed the manual deployment stuff (I think) and this should be final if tests work.

marekdedic:
Hi,
joining a bit late I suppose, but I wanted to make Documenter work on Circle CI and found this...

I see you are doing a lot of mingling of the environment variables - is that really necessary? Wouldn't a better solution be to leave it to the CI config, which already does a lot of that (although I am not sure it does everything)

For example, instead of checking `TRAVIS_PULL_REQUEST == false`, I think it would be better to just put this in your config:
```yaml
jobs:
  include:
    - stage: "Documentation"
      if: type != pull_request
      julia: 1.0
      os: linux
      script:
        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
                                               Pkg.instantiate()'
      after_success: skip
```
https://docs.travis-ci.com/user/conditional-builds-stages-jobs/

That way, the core of Documenter can be (maybe entirely) **CI-independent**. Of course, supplying example configurations for some of the more common CIs would be great for the users.

marekdedic:
Follow up on my previous comment - a list of conditions on Travis, corresponding to the current checks:

- `occursin(travis_repo_slug, repo)` - I am not really sure why this check is there, maybe because of forks? Anyway, can be replaced by `if: repo = owner/name`
- `travis_pull_request == "false"` can be replaced by `if: type != pull_request`
- `isempty(travis_tag) || occursin(Base.VERSION_REGEX, travis_tag)` can be replaced, but it's very ugly: `if: tag = "" OR tag =~ ^v?(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:(-)|([a-z][0-9a-z-]*(?:\.[0-9a-z-]+)*|-(?:[0-9a-z-]+\.)*[0-9a-z-]+)?(?:(\+)|(?:\+((?:[0-9a-z-]+\.)*[0-9a-z-]+))?))$`
- `!isempty(travis_tag) || travis_branch == devbranch` can be replaced if we know the value of `devbranch`: `if: tag != "" OR branch = devbranch`
- `travis_event_type != "cron"` can be replaced by `if: type != cron`

What probably cannot be very easily replaced is the check for the existence of `DOCUMENTER_KEY`. On the other hand, I think the key shouldn't be passed as an environment variable in the first place - as far as I understand it, Travis can set up the key to be used by ssh: https://docs.travis-ci.com/user/private-dependencies/#user-key

I haven't used Travis in several years, so maybe I got something wrong - in such a case I would be glad to be corrected.

asinghvi17:
I think the logic here is that you want to be able to _build_ documentation on every CI run, possibly inspect it, but not actually _deploy_ it.

marekdedic:
Hmm, I see what you mean...

On CircleCI I would then separate into 2 jobs the building of the docs from their deployment.... 

However, it seems to me now that I'm looking into the docs that in Travis, there is no easy way to share data between jobs :( So you'd need to have 2 jobs and build the docs in both and that is just bad design.

I still think the deployment should be independent from the CI you are using, but as far as I can tell Travis offers no easy way to solve the problem you're taking about.

marekdedic:
By the way, regarding https://github.com/JuliaDocs/Documenter.jl/pull/1067#issuecomment-511165249 - yes, on GitLab, you don't push anything, you just tell the CI "this folder has the content of the pages". So you don't even need the `deploydocs` function. I can provide an example config.

src/Writers/LaTeXWriter.jl
```diff
@@ -13,7 +13,7 @@ navigation menu. It goes into the `\\title` LaTeX command.
 
 """
 module LaTeXWriter
-import ...Documenter: Documenter
+import ...Documenter: Documenter, read_ci_env
```
asinghvi17: WARNING: could not import Documenter.read_ci_env into LaTeXWriter

mortenpi:
Sorry for the delay with reviewing this, will take another look at the code tonight.

@asinghvi17 is right in that the `make.jl` script should always run on CI even if no deployment occurs. E.g. to catch documentation build errors in PRs. Also, `deploydocs` should not deploy anything if you're running `make.jl` locally.

So I am not sure we want to decouple the deploy checks from `deploydocs`. It would also mean that there would be more logic that the user has to maintain. `deploydocs` is just a convenience function for the user -- you could always write your own deployment phase.

That is not to say that there aren't refactorings that couldn't be done, but I think these things go beyond this PR. Feel free to open up a separate issue though to discuss this further.

> On the other hand, I think the key shouldn't be passed as an environment variable in the first place - as far as I understand it, Travis can set up the key to be used by ssh: https://docs.travis-ci.com/user/private-dependencies/#user-key

That feature is only for private repositories, no?

> By the way, regarding #1067 (comment) - yes, on GitLab, you don't push anything, you just tell the CI "this folder has the content of the pages". So you don't even need the deploydocs function. I can provide an example config.

I'd say that you'd still want one, ideally. The idea of `deploydocs` is that the user would have to do as little as possible. I do believe you are right that on GitLab the `deploydocs` implementation would be much simpler.

src/CISystems.jl
```diff
@@ -0,0 +1,136 @@
...
+five elements are, in order:
+    - `cibranch`: which branch is CI running on?
+    - `pull_request`: is CI building a pull request?  Empty string if not.
+    - `repo_slug`: The URL-friendly repo name.
+    - `tag`: The tag being built bu CI.  Empty string if not.
+    - `event_type`: The event that triggered CI.
+Furthermore, if `returnfinaldeploy` is true, the function will check whether the
+current CI provider matches `uploader` (set by default to Travis).
+
+Returns a NamedTuple.
```
mortenpi: Maybe, for a more cleaner API here, it would be better if `read_ci_env` would just return `nothing` if the CI env can not be detected? That way the call sites do not have to do the `haskey(ENV, "CI")` check, which may actually be different on different CI systems (in general). So instead, they would check if this returned `nothing` or not, and decide whether we're in CI based on that.

src/Writers/LaTeXWriter.jl
```diff
@@ -188,6 +189,8 @@ end
 function writeheader(io::IO, doc::Documents.Document)
     custom = joinpath(doc.user.root, doc.user.source, "assets", "custom.sty")
     isfile(custom) ? cp(custom, "custom.sty"; force = true) : touch("custom.sty")
+
+    tag = haskey(ENV, "CI") ? read_ci_env(false)[:tag] : ""
```
mortenpi: You pass `false` here, but not above?

Not sure if the `finaldeploy` check should really be in the `read_ci_env` function like this in the first place.

marekdedic:
Hi,
well, that's what I see as the biggest weakness of this approach - stuffing CI logic into this package and the `deploydocs` function. In my use case, I have `make.jl` for building the docs and `deploy.jl` for deploying them. Then I configure the CI `make.jl` if I just want to build the docs (and save them as an artefact for example) and both if I want to deploy the docs as well...

However, it is probably too late to change that now and force users to have two files.

I still stand behind the idea that CI logic should be left to the CI system and not to the deploy function. However, I am now not so sure how to do that properly.

Regarding the GitLab pages issue, I agree, having the deploydocs function there and doing most of the work makes sense to me there.

mortenpi:
> I still stand behind the idea that CI logic should be left to the CI system and not to the deploy function. However, I am now not so sure how to do that properly.

You are not wrong in the sense that we are currently missing a way to just easily deploy to `gh-pages` (#210). You need to do everything by hand if you need that. But `deploydocs` is just a convenience function for Travis users (but with this PR, will hopefully also work on a few other CI systems).

We definitely want some API to just do deployment (which `deploydocs` would internally call, most likely). But it does need some thought and refactoring. One interface that I have been contemplating that would achieve that is to pass arguments to `make.jl`. I.e. you'd just have to call something like `julia docs/make.jl deploy` to deploy the docs. See #75, [this gist](https://gist.github.com/mortenpi/31f9d38fb6c7bc5d858ac5b86c4fb039) and [this](https://github.com/JuliaDocs/Documenter.jl/wiki/Documenter.jl-roadmap#new-api-for-the-user).

simonbyrne:
Any movement on this? I would love to add support for Azure Pipelines as well.

asinghvi17:
I'll create a more minimized PR in a couple of days, hopefully that simplifies matters - I seem to have touched too much code here.

mortenpi:
This is very stale, so I'll go ahead an close this. Also, I think `deploydocs` kinda works this way now with the `DeployConfig`s.

PR #1071, opened 2019-07-17
Allow DocStringExtensions 0.8, bump version to 0.22.6

iamed2:
Currently I'm getting bumped down to 0.22.3 because it's the last version which didn't upper bound DocStringExtensions to 0.7.0, which means I can't use 0.22.4's fix for strict=true. This resolves that.


mortenpi:
LGTM, will tag right away.

iamed2:
thanks!

PR #1089, opened 2019-08-09
Add option to hide sitename in sidebar

mortenpi:
We currently do this for the Julia manual with some [CSS overrides](https://github.com/JuliaLang/julia/blob/5dd7636b9033a48853edb5c6b41c06bc3975ecf1/doc/src/assets/julia-manual.css#L1-L3). But with this, you can actually get rid of the HTML tag altogether.

Builds on top of #1043 (`mp/bulma`).
- [x] Change base branch to `master` before merging.


src/Writers/HTMLWriter.jl
```diff
@@ -152,9 +156,11 @@ struct HTML <: Documenter.Writer
             assets        :: Vector{String} = String[],
             analytics     :: String = "",
             collapselevel :: Integer = 2,
+            hide_sidebar_sitename :: Bool = false,
```
fredrikekre:
```suggestion
            sidebar_sitename :: Bool = true,
```

PR #1093, opened 2019-08-14
Fix checkdocs docstring

mortenpi:
The default of `checkdocs` has [never been `:none`](https://github.com/JuliaDocs/Documenter.jl/blame/master/src/Documents.jl#L276).


rfourquet:
I don't know about when version v0.24 is planned to be released, but if it's in a "long time", it would be great to backport this commit to v0.23! (if you agree, tell me if you would like me to make a PR).

fredrikekre:
We planned to do it any day now, but if you really want this in 0.23.X we can do that.

rfourquet:
Ah cool, no it's fine for me, I was just thinking to save the little trouble I had with this to other users, but this will be fixed soon anyway with 0.24.

PR #1100, opened 2019-08-19
Update guide.md

logankilpatrick:
Adding this since it's part of what is required and actually shows up in the example [here](https://github.com/JuliaLang/Example.jl/blob/master/docs/make.jl).

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -67,7 +67,7 @@ Add the following to your `make.jl` file
 ```julia
 using Documenter, Example
 
-makedocs(sitename="My Documentation")
+makedocs(modules=[Example], sitename="My Documentation")
 ```
 
 This assumes you've installed Documenter as discussed in [Installation](@ref) and that your


mortenpi:
Hmm, I would actually leave that code example as it is -- the `modules` keyword is not necessary there. As per the docstring, it is only used when checking for missing docstrings:

> `modules` specifies a vector of modules that should be documented in source. If any inline docstrings from those modules are seen to be missing from the generated content then a warning will be printed during execution of makedocs. By default no modules are passed to modules and so no warnings will appear.

What we could maybe use is a section on the guide page about document checks though?

logankilpatrick:
Okay, that makes sense. I will update it.

logankilpatrick:
Sorry for the delay. Ready to update this. Can you elaborate a little more on at you mean by "document checks"?

mortenpi:
Documenter can do various checks for you, e.g. making sure that all docstrings are included, links do not 404, doctesting. Not sure whether I forgot any, but the [`makedocs`](https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.makedocs) keywords should cover all the ones we have. Having them explained a bit more in a separate section might be handy.

mortenpi:
Not about `makedocs` in general, but specifically about the various checks that Documenter does. One of them being the check for missing docstrings, which is what the `module` keyword is used for.

logankilpatrick:
Ah got it, I will update my PR.

mortenpi:
Let's close this. I think what was discussed here is already in the `makedocs` docstring.

PR #1105, opened 2019-08-20
Compat caret specifier

Nosferican:
Change documentation to use the default caret specifier à la semver rather than the tilde specifier indicating patch versions as breaking with 0 major.

Would it be possible for the dev process in Documenter to use minor releases as breaking while keeping patches for non-breaking while 0 major?

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -165,7 +165,7 @@ Markdown.parse("""
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
 [compat]
-Documenter = "~$(v.major).$(v.minor)"
+Documenter = "$(v.major).$(v.minor)"
 ```
 """)
 ````


mortenpi:
@fredrikekre Is there any reason to prefer one over the other?

> Would it be possible for the dev process in Documenter to use minor releases as breaking while keeping patches for non-breaking while 0 major?

That's the standard Julia interpretation of the 0.x semver and we have always done that.

Actually, for a long time, we made sure that even the 0.x minor releases do not break anything. And with patch releases we are also conservative -- they are usually just bugfixes and do not contain new features, to provide additional stability.

Nosferican:
For the dev process you described it seems the compat should follow the standard Julia semver (i.e., default caret) rather than the tilde then.

mortenpi:
As far as I understand, `~0.23` and `^0.23` / `0.23` define an equivalent version range?

Nosferican:
They are the same while major is zero.
```
PkgA = "0.2.3" # [0.2.3, 0.3.0)
PkgB = "^0.2.3" # [0.2.3, 0.3.0)
PkgC = "~0.2.3" # [0.2.3, 0.3.0)
```
After major > 0, they differ.
```
PkgA = "1.2"   # [1.2.0, 2.0.0)
PkgB = "^1.2"   # [1.2.0, 2.0.0)
PkgC = "~1.2"   # [1.2.0, 1.3.0)
```
Tilde means the project has minors with potential breaking code deviating from semver.
Hence, why I think it would be best to use the standard caret if the project follows the usual semver.

PR #1109, opened 2019-08-21
Don't cleanup temporary directories meant for the user to inspect.

fredrikekre:
See #32851


mortenpi:
Could we use a patch release for this, or is it fine if this ships with 0.24 in a week or two?

fredrikekre:
Yea thats fine, 1.3 will be released later than that anyway I think.

Issue #1110, opened 2019-08-22
Return page sub-sections in search results?


ianshmean:
As an example, if you search for "macros" in the current main docs, the Macros section of the docs isn't in the search results, because it's a sub-section of the Metaprogramming docs page.

"macros" search results: https://docs.julialang.org/en/v1/search/?q=macros
What I was hoping to find: https://docs.julialang.org/en/v1/manual/metaprogramming/index.html#man-macros-1

Having page sections in search results seems like it would be helpful

ianshmean:
I just noticed that the section result is there, just way down the list.
Perhaps the ranking should be higher for sections that match the string exactly?

mortenpi:
Dup of #631?

ianshmean:
Indeed!

PR #1112, opened 2019-08-24
Increase search result prominence for close matches in result titles

### LINKED ISSUE #631: Exact page title match should be ranked first in the search results
Initially reported as part of #629. In the example there, a search for "markdown" (http://docs.julialang.org/en/latest/search/?q=markdown) produced the first few results:

>- [Markdown syntax](https://docs.julialang.org/en/latest/manual/documentation/#Markdown-syntax-1)(section)
>- [Markdown](https://docs.julialang.org/en/latest/stdlib/Markdown/#)(page)
>- [Markdown](https://docs.julialang.org/en/latest/stdlib/Markdown/#Markdown-1)(section)
> ...

The [Markdown page](https://docs.julialang.org/en/latest/stdlib/Markdown) should have been the first entry in the results list.

I wonder if the ranking is intentional, e.g. because the page has considerably less content than the section placed first. If that's the case, then I'd suggest indicating that factor in the search results (e.g. as a word count, or some abstracted "percentage match" number).

ianshmean:
I tested this through Chrome's DevTools (breakpoint before the function runs, adding this change, then running the page so that the index is generated with this change)

It makes search queries that closely match the result title rank much higher.
For "Macros" and "Arrays" queries in searching the base docs it makes those sections 1st or 2nd,  whereas they're currently way down in the list.

Fixes #631 and #1110

### DIFF
--- a/assets/html/search.js
+++ b/assets/html/search.js
@@ -165,7 +165,7 @@ $(document).ready(function() {
 
   var index = lunr(function () {
     this.ref('location')
-    this.field('title')
+    this.field('title',{boost: 10})
     this.field('text')
     documenterSearchIndex['docs'].forEach(function(e) {
         this.add(e)


mortenpi:
LGTM. Thanks, @ianshmean! It will be a short while before this will actually show up on Julia's stable docs (either with 1.3 of 1.4).

ianshmean:
@mortenpi I know this is probably frowned upon, but would https://github.com/JuliaLang/docs.julialang.org/pull/9 be acceptable? Seems like this fix would be helpful retroactively

Obviously just a suggestion, as it would break the rules..

ianshmean:
Perhaps a sensible approach would be to see how it goes for new releases first before doing anything retroactive..

PR #1113, opened 2019-08-25
Improving search matching based on result title: Increase field boost to 100

ianshmean:
Previously, not `boost`ing the `title` field (compared to the `text` field) meant poor ranking performance for key search results related to terms like "arrays", "macros", "types", "methods".

A boost of 10 improved this, but some spurious rankings remained. Boost 100 seems to give something closer to expectation across a variety of terms tried

Ref: #1112


mortenpi:
Thanks!

PR #1116, opened 2019-08-28
Add Julia 1.3 on Travis, don't test on Julia 1.1.

fredrikekre:


mortenpi:
Could also do 1.3 on Appveyor? Or you reckon it's not worth it?

fredrikekre:
IMO not worth it, in particular since AV don't run in parallell.

Issue #979, opened 2019-03-06
Allow configuring the docker image of DocumenterLaTeX


findmyway:
Currently the docker image and tag is hard-coded in the source code:

https://github.com/JuliaDocs/Documenter.jl/blob/28de68c0a10c7f7e8f7e6de9dcf266f9ab5947e1/src/Writers/LaTeXWriter.jl#L134

https://github.com/JuliaDocs/Documenter.jl/blob/28de68c0a10c7f7e8f7e6de9dcf266f9ab5947e1/src/Writers/LaTeXWriter.jl#L159

It would be very convient to allow configuring them.

mortenpi:
As an argument to `LaTeX()` perhaps? @fredrikekre do you have any thoughts on this?

fredrikekre:
> As an argument to LaTeX() perhaps?

Sure, that makes sense.

PR #1119, opened 2019-08-29
Fix PDF output format with `julia` code block.

### LINKED ISSUE #1118: The julia code block contains character `%` can not generate PDF format output.
Markdown code like this:
    
    ```julia
    1 % 2
    ```

will failed to generate PDF format output.

June-6th:
- Don't failed to generate PDF format with `julia` code block that
contain `%`.(Fixes #1118)
- Don't eat the last line of `julia` code block.

Before this PR:

Screenshot_20190829_190510

After this PR:

Screenshot_20190829_190630


mortenpi:
Hmm, on my system at least this seems to break `⊻` rendering. This code block

````markdown
```julia
1 % 2
1 ⊻ 2
1 | 2
```
````

becomes

```tex
\begin{minted}[escapeinside=⊻⊻]{julia}
1 % 2
1 ⊻\unicodeveebar⊻ 2
1 | 2
\end{minted}
```

as it should. But that renders as

Screenshot from 2019-08-30 11-22-55

After having played around with this for a while, it seems that minted's escaping is pretty broken in general? Using `%` _should_ work in my opinion (like it does for `lstlisting`). `⊻` maybe does not work because it is non-ASCII?

This gist should reproduce it: https://gist.github.com/mortenpi/574735e33d0f2e9093681a45e37bfd80

---

As a general comment, I think, if we can, it would be better to have the same escaping logic for the `minted` and `lstlisting` blocks. I.e. if we change `escapeinside` for one, we should probably change it for the other. That should obviate the need for two `_print_code_escapes` functions.

June-6th:
I can't reproduce your problem. My LuaTeX version is 1.07.0, what about you?

mortenpi:
Right, I was wondering if it might be something like that. Mine is 1.0.4. But that's from the Ubuntu 18.04 repositories, so this will break things if we build with our current Docker image.

mortenpi:
What seems to work is to add this to `documenter.sty`:

```tex
\newfontfamily\unicodevbarfont{DejaVu Sans}[Scale=MatchLowercase]
\newcommand\unicodevbar{{\unicodevbarfont ^^7c}}
```

and then do the code blocks as follows:

```tex
\begin{lstlisting}[escapeinside=||]
1 % 2
1 |\unicodeveebar| 2
1 |\unicodevbar| 2
\end{lstlisting}

\begin{minted}[escapeinside=||]{julia}
1 % 2
1 |\unicodeveebar| 2
1 |\unicodevbar| 2
\end{minted}
```

June-6th:
It is perfect! It would be better  to let `⊻` and `|` use the same color as other operators.

mortenpi:
> It would be better to let `⊻` and `|` use the same color as other operators.

Hmm. Do you mean that minted isn't highlighting them properly at the moment because we're doing the LaTeX magic?

June-6th:
Yes, they are darker.

mortenpi:
> Yes, they are darker.

I guess that would need getting the correct font color, thicknes etc. for the minted internals and applying it manually?

June-6th:
This may be the easiest way to do it, because lstlisting do not have the problem.

findmyway:
Hey @mortenpi ,

Is there any blocking issue or concern to merge this PR? 

The building pipeline for the pdf version of the Julia Documentation in Chinese has been broken for months due to the problem described above. It would be great to see this PR get merged.

Ref: https://github.com/JuliaCN/JuliaZH.jl/issues/85

mortenpi:
I think it will still break the Docker image currently?

mortenpi:
Ok, it doesn't actually break the Docker image, but PDFs built with it do not render correctly at the moment: https://github.com/JuliaDocs/Documenter.jl/pull/1119#issuecomment-526438867. Swapping the implementation to https://github.com/JuliaDocs/Documenter.jl/pull/1119#issuecomment-526453896 should work around that I think?

findmyway:
> What seems to work is to add this to `documenter.sty`:
> 
> ```
> \newfontfamily\unicodevbarfont{DejaVu Sans Mono}[Scale=MatchLowercase]
> \newcommnd\unicodevbar{{\unicodevbarfont ^^7c}}
> ```
> 
> and then do the code blocks as follows:
> 
> ```
> \begin{lstlisting}[escapeinside=||]
> 1 % 2
> 1 |\unicodeveebar| 2
> 1 |\unicodevbar| 2
> \end{lstlisting}
> 
> \begin{minted}[escapeinside=||]{julia}
> 1 % 2
> 1 |\unicodeveebar| 2
> 1 |\unicodevbar| 2
> \end{minted}
> ```
> 
> Edit: should be DejaVu Sans Mono for this symbol.

Hi @June-6th , can you confirm this works and make the change correspondingly?

<hr>

> Right, I was wondering if it might be something like that. Mine is 1.0.4. But that's from the Ubuntu 18.04 repositories, so this will break things if we build with our current Docker image.

I think when https://github.com/JuliaDocs/Documenter.jl/issues/979 gets resolved, this will not be a big problem. (At least it's not a big problem for us right now. Because we are using a customized docker image [here](https://github.com/JuliaCN/JuliaZH.jl/blob/master/contrib/LaTeXWriter.jl#L6))

June-6th:
@findmyway  Yes, it work well.

PR #1123, opened 2019-09-08
Some documentation tweaks

aminya:
Some documentation tweaks

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -128,9 +128,11 @@ jobs:
       julia: 1.0
       os: linux
       script:
-        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
-                                               Pkg.instantiate()'
-        - julia --project=docs/ docs/make.jl
+         - julia -e 'using Pkg; Pkg.activate(); push!(LOAD_PATH, pwd());' 
+        - julia --project=docs/ -e 'using Pkg; Pkg.activate();  push!(LOAD_PATH,pwd());'
+
+        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate();'
+        - julia --project=docs/ --color=yes docs/make.jl
       after_success: skip
 ```
 
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -128,11 +128,9 @@ jobs:
       julia: 1.0
       os: linux
       script:
-         - julia -e 'using Pkg; Pkg.activate(); push!(LOAD_PATH, pwd());' 
-        - julia --project=docs/ -e 'using Pkg; Pkg.activate();  push!(LOAD_PATH,pwd());'
-
-        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate();'
-        - julia --project=docs/ --color=yes docs/make.jl
+        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
+                                               Pkg.instantiate()'
+        - julia --project=docs/ docs/make.jl
       after_success: skip
 ```
 
--- a/docs/src/man/hosting/walkthrough.md
+++ b/docs/src/man/hosting/walkthrough.md
@@ -136,6 +136,7 @@ First, we need to Base64 encode the private key. Open Julia, and run the command
 
 ```julia
 julia> using Base64
+
 julia> read("path/to/private/key", String) |> base64encode |> println
 ```
 
--- a/docs/src/man/hosting/walkthrough.md
+++ b/docs/src/man/hosting/walkthrough.md
@@ -149,7 +149,7 @@ to the "Environment Variables" section. It should look like this:
 Now, add a new environment variable called `DOCUMENTER_KEY`, and set its value to the output
 from the Julia command above (make sure to remove the surrounding quotes).
 
-Finally, check that the "Display value in build log" is switched off and then click "Add".
+Finally, leave the "Display value in build log" as switched off and then click "Add".
 Congratulations! You've added the private key to Travis.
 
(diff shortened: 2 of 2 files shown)


docs/src/man/hosting.md
```diff
@@ -128,9 +128,11 @@ jobs:
       julia: 1.0
       os: linux
       script:
-        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
-                                               Pkg.instantiate()'
-        - julia --project=docs/ docs/make.jl
+         - julia -e 'using Pkg; Pkg.activate(); push!(LOAD_PATH, pwd());' 
+        - julia --project=docs/ -e 'using Pkg; Pkg.activate();  push!(LOAD_PATH,pwd());'
```
fredrikekre: These two first lines does not do anything since they are undone the second Julia quits, and you shouldn't have to modify the LOAD_PATH anyway.

docs/src/man/hosting/walkthrough.md
```diff
@@ -135,6 +135,7 @@ read [Generating an SSH Key](@ref).
 First, we need to Base64 encode the private key. Open Julia, and run the command
 
 ```julia
+julia> using Base64
```
fredrikekre:
```suggestion
julia> using Base64

```

docs/src/man/hosting/walkthrough.md
```diff
@@ -148,7 +149,7 @@ to the "Environment Variables" section. It should look like this:
 Now, add a new environment variable called `DOCUMENTER_KEY`, and set its value to the output
 from the Julia command above (make sure to remove the surrounding quotes).
 
-Finally, check that the "Display value in build log" is switched off and then click "Add".
+Finally, leave the "Display value in build log" as switched off and then click "Add".
```
fredrikekre:
```suggestion
Finally, make sure that the "Display value in build log" is switched off and then click "Add".
```

mortenpi:
```suggestion
Finally, leave the "Display value in build log" switched off and then click "Add".
```

docs/src/man/hosting.md
```diff
@@ -128,9 +128,11 @@ jobs:
...
+
+        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate();'
+        - julia --project=docs/ --color=yes docs/make.jl
```
mortenpi: The `push!`es to `LOAD_PATH` have no effect and are unnecessary. What is your goal here?

docs/src/man/hosting/walkthrough.md
mortenpi:
```suggestion
Finally, make sure that the "Display value in build log" is left switched off and then click "Add".
```

docs/src/man/hosting/walkthrough.md
aminya: Is this just a new line after the first command?

docs/src/man/hosting.md
aminya: Travis gave me errors in the documentation stage when I used the original config:
https://travis-ci.com/juliamatlab/MatLang/jobs/232115453/config

But I fixed the issue by adding these two lines:
https://travis-ci.com/juliamatlab/MatLang/jobs/232117413/config

I think it was because my package is not registered. 

I got those errors when building the package offline in Windows using PowerShell.

docs/src/man/hosting.md
aminya: Travis gave me errors in the documentation stage when I used the original config:
https://travis-ci.com/juliamatlab/MatLang/jobs/232115453/config

But I fixed the issue by adding these two lines:
https://travis-ci.com/juliamatlab/MatLang/jobs/232117413/config

I think it was because my package is not registered. 

I got those errors when building the package offline in Windows using PowerShell.

docs/src/man/hosting/walkthrough.md
mortenpi: Yep. This brings it in line with how it would actually be displayed in the REPL.

docs/src/man/hosting.md
mortenpi: Hmm. I am wondering if this is a bug in nightly somehow. @fredrikekre Do you see why the `julia --project=docs/ docs/make.jl` call did not detect the docs project/manifest in the first log?

The `LOAD_PATH`s should definitely not be necessary though.

aminya: I didn't upload `manifest` for the docs folder just a `project.toml`. So I thought maybe I need to write the script manually to create that.

fredrikekre: You seem to have deactivated the docs environment you start julia with: https://travis-ci.com/juliamatlab/MatLang/jobs/232115453#L225

mortenpi: Indeed: https://github.com/juliamatlab/MatLang/blob/3b287f7db3cc2ef115441d1498594ab10d8342db/docs/make.jl#L9

aminya: https://github.com/juliamatlab/MatLang/blob/master/docs/make.jl
I don't think so. In a commit before these two, I had some script but I removed it.
These two are for the same `make.jl` file. I forced pushed my commit to remove history, however.

mortenpi: The commit for the original Travis job definitely had the `make.jl` stuff (https://github.com/juliamatlab/MatLang/blob/3b287f7db3cc2ef115441d1498594ab10d8342db/docs/make.jl#L7-L10) and that is what caused the problem, not anything in `.travis.yml`.

aminya: Let me see if I can recreate the error with the latest `make.jl` file but the old Travis code

aminya: So the original Travis file works. It was my fault that I thought I removed that code from the `make.jl` file! 👀 
https://travis-ci.com/juliamatlab/MatLang/jobs/232190010

PR #1125, opened 2019-09-08
Allow at-contents to take Depth as a UnitRange

NHDaly:
Allow ```` ```@contents ```` to take `Depth` as a UnitRange:
````
```@contents
Pages = ["index.md"]
Depth = 2:5
```
````

This is useful to allow skipping the _title_ of a page (header level 1) when generating a Table of Contents, since the title is often redundant.

Fixes https://github.com/JuliaDocs/Documenter.jl/issues/245

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1238,7 +1238,7 @@ function domify(ctx, navnode, contents::Documents.ContentsNode)
         header = anchor.object
         url = string(path, '#', anchor.id, '-', anchor.nth)
         node = a[:href=>url](mdconvert(header.text; droplinks=true))
-        level = Utilities.header_level(header)
+        level = Utilities.header_level(header) - contents_header_level_offset(contents)
         push!(lb, level, node)
     end
     domify(lb)
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -330,7 +330,7 @@ document, but this can be adjusted using `Pages` and `Depth` settings as in the 
 ````markdown
 ```@contents
 Pages = ["foo.md", "bar.md"]
-Depth = 3
+Depth = 1:3
 ```
 ````
 
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -291,16 +291,18 @@ expected by the user. Try to stick to array literals as much as possible.
 ## `@contents` block
 
 Generates a nested list of links to document sections. Valid settings are `Pages` and `Depth`.
+Note that `Depth` can be either an `Int` (parsed as `1:Depth`) or a `UnitRange{Int}`,
+specifying the starting and ending depth to include in the generated contents.
 
 ````markdown
 ```@contents
 Pages = ["foo.md"]
-Depth = 5
+Depth = 1:5
 ```
 ````
 
 As with `@index` if `Pages` is not provided then all pages are included. The default
-`Depth` value is `2`.
+`Depth` value is `1:2`.
 
 ## `@example` block
 
--- a/src/Documents.jl
+++ b/src/Documents.jl
@@ -113,18 +113,21 @@ end
 
 struct ContentsNode
     pages       :: Vector{String} # Which pages should be included in contents? Set by user.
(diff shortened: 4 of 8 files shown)

### CODE REVIEW

**Comment by mortenpi** [comment]
At least in the HTML output, Documenter still seems to write them out on the same level. E.g. with `Depth = 2:4`, the level 2 headings, that should now be on the highest level, still end up on the second level:

contents

I reckon we'd want them to end up on the top level. Not 100% sure at the moment where the issue is coming from though.

**Comment by NHDaly** [comment]
> I reckon we'd want them to end up on the top level. Not 100% sure at the moment where the issue is coming from though.

I wasn't sure if moving them up a level was desirable or not. But now that you point it out, yeah I think you're right!

After a bit of digging, it looks like the change needs to be made in the renderers, e.g. here:
https://github.com/JuliaDocs/Documenter.jl/blob/13ec67eb0111374f747004dc1063e3bf49989bc0/src/Writers/MarkdownWriter.jl#L130

I've just pushed up a commit that subtracts the starting depth from the level. Let me know if that's a clean way to do it, or if there's a different factoring of the code you'd prefer? :) Thanks!

(Also, i'm not really sure how to write unit tests for this... is that something we should do? Thanks)

File: src/Writers/HTMLWriter.jl [hunk 1]
```diff
@@ -1238,7 +1238,7 @@ function domify(ctx, navnode, contents::Documents.ContentsNode)
         header = anchor.object
         url = string(path, '#', anchor.id, '-', anchor.nth)
         node = a[:href=>url](mdconvert(header.text; droplinks=true))
-        level = Utilities.header_level(header)
+        level = Utilities.header_level(header) - contents_header_level_offset(contents)
```
mortenpi: This currently errors with

```
ERROR: LoadError: UndefVarError: contents_header_level_offset not defined
```

Just needs a `using` for it, as the definition is in the `Documenter.Documents` module.

mortenpi:
What happens when you have 

```
* Foo
  - Foo 1
  - Foo 2
* Bar
  - Bar 1
  - Bar 2
```

and you have e.g. `Depth = 2:3`? It would just become:

```
* Foo 1
* Foo 2
* Bar 1
* Bar 2
```
?

NHDaly:
Yeah, that's true...

Maybe what would be better would be to have an option to drop all _preceding_ headers? That is, to only start the table-of-contents from the current position?


So this:
`````markdown
# My Page

## Table of Contents
```@contents
Pages = ["index.md"]
OnlyIndexBelow = true
```

# Section 1
## Subheader
...
# Section 2
...
`````

Would become:
> <h1>My Page</h1>
> <h2>Table of Contents</h2>
>
> - Section 1
>     - Subheader
> - Section 2
> <h1>Section 1</h1>
> <h2>Subheader</h2>
> <h1>Section 2</h1>
> ...


What do you think about that? The option name needs a bit of work, `OnlyIndexBelow = true` could be `SkipPrevious = true`, `OnlyIndexFollowing = true`, `StartHere = true`, `NoPrevious = true`, `StartIndex = "Section 1"`, or something else?

**Comment by mortenpi** [comment]
In general, you should not have `h1` headings on the page anyway (other than the top one), so that behavior would not be an issue. We could treat the `Depth` option as being quite a low-level feature (i.e. it does a very specific thing and it is up to the user to make sure that it makes sense within their docs).

That is not to say that we couldn't have a fancier option as well for the use case that you have in mind. But you would still need to figure out how to deal with the `h1` headings -- they _shouldn't_ be there, but they _can_ be.

Just a few thoughts on the second option:

* I would maybe have it be a special mode of at-contents, which would would enable with something like

````markdown
```@contents
PageContents = true
```
````

or

````markdown
```@contents
mode = :PageContents
```
````

It would then automatically figure out which page it is on, so no need to `Pages`. I guess `Depth` you could still maybe use?

* To handle `h1`s: one possible behavior would be to, if there are any `h1` headings below the at-contents block, treat them as `h2` and emit a warning. So, in that respect, this mode of at-contents would be opinionated about how you organize your headings.

But all this is quite a bit more work to implement I think. So happy to just go with generalizing the `Depth` option if that works for your use case.

NHDaly:
closing this stale PR.

PR #1126, opened 2019-09-08
fix warning: remove bad import Documenter: asset

NHDaly:
Fixes: `WARNING: could not import Documenter.asset into HTMLWriter`, which currently shows up on `master`.

It looks like that import was accidentally erroneously added in 5ae06778adad15d1e174a8ab9c88d13914aec5d4


mortenpi:
Indeed. Thanks for the fix! I think I was shuffling this function between the two files quite a bit as I was working on #1108.

NHDaly:
😁👍 makes sense. Thanks!

PR #1137, opened 2019-09-25
Scale figures to textwidth in latex writer

pfitzseb:
This isn't an ideal solution since we're also going to *upscale* small images, but fixing that [is nontrivial](https://stackoverflow.com/questions/122348/scale-image-down-but-not-up-in-latex).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Documenter.jl changelog
 
+## Version `v0.24.0`
+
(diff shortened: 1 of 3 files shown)


mortenpi:
What about [`adjustbox`](https://tex.stackexchange.com/a/96468/127437)?

pfitzseb:
Yeah, that's also one of the solutions mentioned in the stackoverflow post I linked. Is just adding packages [here](https://github.com/JuliaDocs/Documenter.jl/blob/master/assets/latex/documenter.sty) ok?

mortenpi:
Yep, I'd put them in the `.sty`. The only thing that would be good to check is to make sure that it still builds fine with the Docker image, i.e. that we have `adjustbox` available in the image.

pfitzseb:
Alright, changed to the `adjustbox` setup and can confirm the docker build still works fine.

mortenpi:
LGTM, will merge shortly. Thanks @pfitzseb!

PR #1143, opened 2019-10-04
Allow failures on Julia nightly

fredrikekre:


mortenpi:
Should we add testing for 1.3 maybe?

PR #1154, opened 2019-10-09
Fix documentation build

mortenpi:
Update references to DocumenterTools functions and modules.  Also, fix its version in documentation build so that it would not go out of date again.


fredrikekre:
Thanks. I pushed almost the exact same diff yesterday but forgot to open a PR; https://github.com/JuliaDocs/Documenter.jl/tree/fe/genkeys-docs

PR #1161, opened 2019-10-15
Support for pushing with https instead of ssh

fredrikekre:
Support for pushing with https instead of ssh and enable it for the GitHubActions DeployConfig.


fredrikekre:
Successful run: https://github.com/fredrikekre/Literate.jl/runs/260539334

PR #1171, opened 2019-11-08
Add User and IdentitesOnly lines to the generated SSH config

fjarri:
I am trying to set up documentation generation on CircleCI, and it seems that these entries are required for the `git push` to work, otherwise I get an error about the key not having write privileges (in fact, as I can see from the last access info on GitHub, it wasn't even used).

I am not sure what is the difference between Circle and Travis (where these entries are not required), so it would be great if someone more knowledgeable took a look at it, so that this PR doesn't break anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 
   **Possible breakage:** Packages overriding the default Documenter CSS file, relying on some external CSS or relying on Documenter's CSS working in a particular way will not build correct-looking sites. Custom themes should now be developed as Sass files and compiled together with the Documenter and Bulma Sass dependencies (under `assets/html/scss`).
 
-* ![Feature][badge-feature] Deployment is now more customizable and thus not as tied to Travis CI as before. ([#1147][github-1147])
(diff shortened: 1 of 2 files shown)


mortenpi:
This should have no effect when running on Travis, so LGTM. Thanks, @fjarri!

mortenpi:
> I am trying to set up documentation generation on CircleCI, and it seems that these entries are required for the `git push` to work, otherwise I get an error about the key not having write privileges (in fact, as I can see from the last access info on GitHub, it wasn't even used).

I wonder if CircleCI has a default SSH key generated that gets precedence of ours?

fjarri:
It does have a read-only key it uses to check out the repo, so maybe that's the reason. Travis seems to only have a single read/write key, judging by "Deploy Keys" settings tab in my projects. I can't change the access preferences now, and I don't remember if I could select them when I connected CircleCI.

mortenpi:
> It does have a read-only key it uses to check out the repo, so maybe that's the reason.

That's probably it then. SSH manages to log in with that one, so it never gets around to trying the write key.

PR #1176, opened 2019-11-14
Add more information to warning about unresolved link

baggepinnen:
Makes it easier to understand which link is unresolved.


src/Writers/HTMLWriter.jl
```diff
@@ -1666,7 +1666,7 @@ function fixlinks!(ctx, navnode, link::Markdown.Link)
         # provided files or generated by code examples)
         path = relhref(get_url(ctx, navnode), path)
     else
-        @warn "invalid local link: unresolved path in $(Utilities.locrepr(navnode.page))" link = link.url
+        @warn "invalid local link: unresolved path in $(Utilities.locrepr(navnode.page))" link = (link.url, link.text)
```
mortenpi: Good idea! I wonder if it would be even more readable if the URL and text would be split up into separate items?

```suggestion
        @warn "invalid local link: unresolved path in $(Utilities.locrepr(navnode.page))" text = link.text url = link.url
```

fredrikekre: Alternatively:
```suggestion
        @warn "invalid local link: unresolved path in $(Utilities.locrepr(navnode.page))" link.text link.url
```

Issue #839, opened 2018-09-22
Filtering included docstrings


thofma:
What are the rules for filtering docstrings? Consider the following function
```julia
julia> struct A{T} <: Number end
julia> """Number""" foo(x::Number) = x; # 1
julia> """Real""" foo(x::Real) = x; # 2
julia> """A{T}""" foo(x::A{T}) where {T} = x; # 3
julia> """A{Int}""" foo(x::A{Int}) = x; # 4
```
It is clear that `@docs foo(::Real)` and `@docs foo(::Number)` should return `#1` and `#2` respectively (and they do). But what about the following?
1) Should `@docs foo(::A{T} where {T})` give `#3`?
2) Should `@docs foo(::A{Int})` give `#4`?
3) What should `@docs foo(::Int)` do?
4) What should `@docs foo(::A{Float64})` do?
Now here is what is currently happening:
1) Pulls in `#1` and `#3`.
2) Pulls in `#4`.
3) Pulls in `#1`, `#2`, `#3`.
4) Pulls in `#1` and `#3`.

mortenpi:
I am not sure what the intended logic currently is, but I think I would expect it to get all the docstrings that are more specific than the signature specified. So IMO

1. `foo(::A{T} where {T})` should give 3 and 4.
2. `foo(::A{Int})` should give 4
3. `foo(::Int)` shouldn't match anything.
4. `foo(::A{Float64})` shouldn't match anything either.

Additionally:

* `foo(x::Number)` should give all four.
* `foo(x::Real)` should give 2.

I also initially thought that it should be consistent with the REPL `help?>` prompt, but I am no longer sure. In the REPL, if you e.g. `help?> foo(::Float64)`, then you should probably see all the general ones that match, especially since there might not be a specific method even defined (matches "upwards", if you will).

Whereas in Documenter I think we can assume that the doc author knows exactly which methods they want to document, so it should match "downwards". And if the author specifies the signature, they probably want to filter out unrelated more general docstrings.

Thoughts on this would be very welcome though.

thofma:
Thanks for the feedback. Note that with https://github.com/JuliaLang/julia/pull/29102 the REPL behavior will change. In particular `help?> foo(::Float64)` will give you only the docstring for the method which will be applied. Once the PR over at julia lands, only the docstrings of methods which are applicable for a given signature will be shown (so matches are "downwards").

[...]

PR #1179, opened 2019-11-16
add comment about specific method docstrings

Datseris:


docs/src/man/syntax.md
```diff
@@ -40,6 +40,17 @@ which will cause any unlisted docstrings to raise warnings when [`makedocs`](@re
...
+Notice also that you can use `@docs` to display the documentation strings of only specific
+methods, by stating the dispatch types. For example
+````markdown
+```@docs
+f(::Type1, ::Type2)
+```
+````
+will only display the documentation string of `f` that is related to these types.
+This can be useful when your module extends a function and adds a documentation
+string to that extention.
```
mortenpi: Thanks, @Datseris! I'll just add a small note that the signature should match exactly (related issue: #839).

```suggestion
string to that new method. Note that the method signature has to match exactly -- Documenter will
not match methods based on abstract types and dispatch rules.
```

Datseris:
@mortenpi I have added your suggestion, but I think you need to come up with an illustrative example. The currently standing text is confusing, at least for me. The problem is, it is not clear to me what "exactly" actually means. E.g. how does one include the method `f(::Vector{T}) where {T<:MyTypes}`?

mortenpi:
Parametric methods are actually problematic currently (#839) and I might refrain from commenting about them at the moment. But what I meant is this: suppose you have a method

```julia
foo(::Integer) = x
```

that has a docstring. Then neither `foo(::Number)` nor `foo(::Int64)` will match it in an at-docs block. The only way you can get that docstring is by listing exactly `foo(::Integer)` in the at-docs block.

Datseris:
Okay, I see your point. Then can you please add a comment that says exactly what you just told me, for the non-parametric methids? (I can do it instead but you should take the commit credit :D )

mortenpi:
Done. Does it look good to you @Datseris?

I will mention that it's actually completely accurate. If you only have the `foo(::Integer)` method and you put `foo(::Number)` into at-docs, it will actually splice `foo(::Integer)` into the docs. But I consider that to be a bug, since if you add a docstring to `foo(::Number)`, it will only splice the `::Number` one and not `::Integer`. So we are now documenting what I would consider to be the correct behavior and need to fix Documenter at some point.

> I can do it instead but you should take the commit credit :D

Gonna squash anyway, so not particularly worried about it :smile:

Datseris:
Looks good to me! If you agree then you should merge! :)

Issue #1177, opened 2019-11-14
GitHub Pages build not triggered with GITHUB_TOKEN


maxmouchet:
According to https://github.com/JuliaDocs/Documenter.jl/pull/1165 `GITHUB_TOKEN`is sufficient for Documenter to push to `gh-pages`.

However it seems that this token is not allowed to trigger pages builds [1]:
> The `GITHUB_TOKEN` that is provided as a part of GitHub Actions doesn't have authorization to create any successive events, such as GitHub Pages builds. So while you can push to the `gh-pages` branch using the `GITHUB_TOKEN`, it won't spawn a GitHub Pages build. You'll need to create a personal access token and supply it to your GitHub Action as a secret. 

Hence, docs. are correctly pushed, but GitHub Pages are never updated.  
Is there any workaround ?

I tried to add a manual build trigger [2] in my [workflow](https://github.com/maxmouchet/HMMBase.jl/blob/dbeb087f4632a9a8013e5778ab37fa8b141368d2/.github/workflows/documentation.yml):
```yaml
- name: Build and deploy
  env:
    DOCUMENTER_DEBUG: true
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      run: |
        julia --project=docs/ docs/make.jl
        curl -u maxmouchet:${GITHUB_TOKEN} -X POST https://api.github.com/repositories/164427411/pages/builds -H "Accept: application/vnd.github.mister-fantastic-preview+json"
```
but the GitHub API [returns](https://github.com/maxmouchet/HMMBase.jl/commit/dbeb087f4632a9a8013e5778ab37fa8b141368d2/checks?check_suite_id=307586853#step:5:149) the following error:
```json
{
  "message": "Resource not accessible by integration",
  "documentation_url": "https://developer.github.com/v3/repos/pages/#request-a-page-build"
}
```

[1] https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/td-p/26869
[2] https://stackoverflow.com/a/54088985

maxmouchet:
Of course, one workaround is to "create a personal access token and supply it to your GitHub Action as a secret", but the documentation should at-least specify that :-)

fredrikekre:
Yea, I thought we already had an issue about this. I had hoped GH would have fixed this but I guess not.

fredrikekre:
You can use a SSH deploy key now: #1183

[...]

PR #1183, opened 2019-11-17
Support SSH deploy keys from GH Actions.

fredrikekre:
X-ref: #1177

### DIFF
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -189,11 +189,6 @@ Implementation of `DeployConfig` for deploying from GitHub Actions.
 The following environment variables influences the build
 when using the `GitHubActions` configuration:
 
- - `DOCUMENTER_KEY`: must contain the Base64-encoded SSH private key for the repository.
-   This variable should be set in the GitHub Actions configuration file using a repository
-   secret, see the documentation for
-   [secret environment variables](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables).
-
  - `GITHUB_EVENT_NAME`: must be set to `push`.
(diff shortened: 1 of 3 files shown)


mortenpi:
Question: I assume that with the `GITHUB_TOKEN` you can't push to a different repository (e.g. something like we're doing for the Julia docs)? So to support that use case with GitHub Actions, should we maybe keep the `DOCUMENTER_KEY` approach around even after the page triggering gets fixed?

src/deployconfig.jl
```diff
@@ -274,15 +283,19 @@ function deploy_folder(cfg::GitHubActions; repo, devbranch, push_preview, devurl
...
-    ## GITHUB_TOKEN should exist (just check here and extract the value later)
-    token_ok = haskey(ENV, "GITHUB_TOKEN")
-    all_ok &= token_ok
-    println(io, "- $(marker(token_ok)) ENV[\"GITHUB_TOKEN\"] exists")
+    # ## GITHUB_TOKEN should exist (just check here and extract the value later)
+    # token_ok = haskey(ENV, "GITHUB_TOKEN")
+    # all_ok &= token_ok
+    # println(io, "- $(marker(token_ok)) ENV[\"GITHUB_TOKEN\"] exists")
+    ## DOCUMENTER_KEY should exist (just check here and extract the value later)
+    key_ok = haskey(ENV, "DOCUMENTER_KEY") || haskey(ENV, "GITHUB_TOKEN")
```
mortenpi: Could we print a warning here if the user only sets `GITHUB_TOKEN` that says something like "GitHub Pages may not update correctly, see #1177"?

fredrikekre: Done

PR #1186, opened 2019-11-19
Post a GitHub status with a link to generated documentation.

fredrikekre:
This currently only works on GitHub Actions, but might be possible
to enable on Travis CI too if one adds a (user) access token.

### DIFF
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -1,3 +1,5 @@
+import JSON
+
 """
     DeployConfig
 
@@ -55,6 +57,9 @@ This method must be supported by configs that push with HTTPS, see
 [`Documenter.authentication_method`](@ref).
 """
 function authenticated_repo_url end
+
+post_status(cfg::DeployConfig; kwargs...) = nothing
+post_status(; kwargs...) = post_status(auto_detect_deploy_system(); kwargs...)
 
 marker(x) = x ? "✔" : "✘"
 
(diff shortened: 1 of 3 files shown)


src/deployconfig.jl
```diff
@@ -304,6 +309,78 @@ function authenticated_repo_url(cfg::GitHubActions)
...
+        return post_github_status(type, repo, sha, subfolder)
+    catch
+        @debug "Failed to post status"
+    end
+end
+
+function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder=nothing) where S <: String
+    try
+        Sys.which("curl") === nothing && return
+        ## Exctract owner and repository name
```
mortenpi:
```suggestion
        ## Extract owner and repository name
```

src/deployconfig.jl
```diff
@@ -304,6 +309,78 @@ function authenticated_repo_url(cfg::GitHubActions)
...
+        elseif type == "failure"
+            json["description"] = "Documentation build failed"
+        else
+            error("unsupported type: $type")
+        end
+        push!(cmd.exec, "-d", sprint(JSON.print, json))
+        push!(cmd.exec, "https://api.github.com/repos/$(owner)/$(repo)/statuses/$(sha)")
+        # Run the command (silently)
+        io = IOBuffer()
+        res = run(pipeline(cmd; stdout=io, stderr=devnull))
```
mortenpi: Since it's only for debugging, should maybe capture stderr too? Otherwise, we might be throwing away some valuable debug information.

fredrikekre: I'm scared of accidentally leak credentials and such

mortenpi: Ah, that is a fair point.

fredrikekre: But I guess that will only show up in the julia error message which should be suppressed by the try-catch...

fredrikekre: Because we will only show `res` if we managed to spawn `curl`, and I guess `curl` will not leak it.

PR #1194, opened 2019-11-24
Fix mktempdir bug on Julia > 1.3.

fredrikekre:


mortenpi:
LGTM. Isn't this a breaking change in Julia though..?

fredrikekre:
No, just me assuming that the `cleanup` kwarg was implemented for all `mktempdir` methods.

PR #1195, opened 2019-11-26
Actually display log message

fredrikekre:
Actually display log message, and add some warnings and tests for broken deploy configurations.

### DIFF
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -33,8 +33,14 @@ arguments from [`deploydocs`](@ref).
     compatibility with future Documenter releases which may pass additional
     keyword arguments.
 """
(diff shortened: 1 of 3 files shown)


src/deployconfig.jl
```diff
@@ -290,6 +296,7 @@ function deploy_folder(cfg::GitHubActions; repo, devbranch, push_preview, devurl
         println(io, "- $(marker(auth_ok)) ENV[\"DOCUMENTER_KEY\"] or ENV[\"GITHUB_TOKEN\"]  exists")
     end
     print(io, "Deploying: $(marker(all_ok))")
+    @info String(take!(io))
```
fredrikekre: :slightly_smiling_face:

PR #1222, opened 2020-01-05
Change away from red inline code color

### LINKED ISSUE #1212: Why is inlined code rendered in red?
I noticed that `Documenter.jl` renders inlined code in red (unless a hyperlink is attached, in which case it is rendered in blue). See [here](https://juliareinforcementlearning.org/ReinforcementLearning.jl/latest/components/environments/) for an example.

I find this choice questionable and I am wondering how it is motivated. Indeed, the color red is more conventionally used to attract one's attention on a malformed element (such as a broken hyperlink) or to make an element strongly  stand-out. To me, rendering inlined code in black or gray would be more readable and less confusing.

Does anyone have thoughts on this?

mortenpi:
Goes from

from

to

to

Close #1212.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Documenter.jl changelog
 
+## Version `v0.24.4`
+
+* ![Enhancement][badge-enhancement] Change the inline code to less distracting black color in the HTML light theme. ([#1212][github-1212], [#1222][github-1222])
+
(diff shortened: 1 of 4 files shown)


mortenpi:
I'm a bit on the fence about whether we should do a similar change to the dark theme too. I find that the red is not quite as distracting there:
dark

PR #1223, opened 2020-01-06
add support for other language in HTML tag

### LINKED ISSUE #977: [Feature Request] Custom Translation on Assets and Text
There are some text in English generated by rendering, it would nice if we could change these text to customized version. e.g

https://github.com/JuliaDocs/Documenter.jl/blob/28de68c0a10c7f7e8f7e6de9dcf266f9ab5947e1/src/Writers/HTMLWriter.jl#L592

And also it would nice to allow have custom html lang tag, e.g `zh-cn`

FYI:
https://github.com/JuliaCN/JuliaZH.jl/pull/49

We are using a overwritten version of the above functions as a workaround in JuliaZH currently.

Roger-luo:
This should partly address #977

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * ![Enhancement][badge-enhancement] Change the inline code to less distracting black color in the HTML light theme. ([#1212][github-1212], [#1222][github-1222])
 
+* ![Enhancement][badge-enhancement] Add the ability specify the `lang` attribute of the `html` tag in the HTML output, to better support documentation pages in other languages. By default Documenter now defaults to `lang="en"`. ([#1223][github-1223])
+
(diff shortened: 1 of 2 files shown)


Roger-luo:
oh sorry I was in vacation didn't see it. thanks tho!

nil-is-lin:
This is a very useful feature. However, I found I don't know how to use it. It will be more perfect if you could add examples to the documentation.

mortenpi:
@nil-is-lin It is documented in [the docstring of `HTML`](https://juliadocs.github.io/Documenter.jl/stable/lib/internals/writers/#Documenter.Writers.HTMLWriter.HTML). I updated it in #1407 to make it slightly more precise, but feel free to comment there or open a PR if you think it could be further clarified.

PR #1224, opened 2020-01-07
Simplify/improve GitHub actions YAML setting example

tkf:
This PR contains three independent patches:

* Simplify the workflow YAML by not using `strategy.matrix`.  IIUC you don't need this if you are only running one combination of `julia` and OS.  Ref: https://github.com/bcbi/CompatHelper.jl/pull/171
* Update to actions/checkout v2: https://github.com/actions/checkout/releases/tag/v2.0.0. I didn't find any breaking change: https://github.com/tkf/MyPlayground.jl/pull/21
* Change `on` setting so that `push_preview = true` works out-of-the-box.

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -177,7 +177,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1.0.0
+      - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
         with:
           version: 1.3
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -175,17 +175,12 @@ on: [push]
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
-    strategy:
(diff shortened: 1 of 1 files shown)


fredrikekre:
Thanks. I believe the matrix strategy was the only working one back in GitHub actions beta when I first tried this.

docs/src/man/hosting.md
```diff
@@ -171,21 +171,21 @@ workflow configuration file:
 ```yaml
 name: Documentation
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
+    tags: '*'
+  pull_request:
```
fredrikekre:
```suggestion
  pull_request: # only needed for preview builds (push_preview=true in deploydocs)
```

tkf: Isn't this also required for running documenter in PRs from forks? (My post in the OP was a bit misleading.)

I guess people familiar with Travis would find this setup natural.

fredrikekre: Ah, right.

PR #1232, opened 2020-01-29
Support REPL softscope for Julia 1.5.

fredrikekre:
See https://github.com/JuliaLang/julia/pull/33864 and https://github.com/JuliaLang/julia/issues/34567.

### DIFF
--- a/test/doctests/src/softscope.md
+++ b/test/doctests/src/softscope.md
@@ -16,7 +16,7 @@ julia> @isdefined(t) # global
 false
 ```
 
-```jldoctest; filter = r"Stacktrace:(\\n \\[[0-9]+\\].*)*"
+```jldoctest; filter = r"Stacktrace:(\n \[[0-9]+\].*)*"
 julia> code = """
        s = 0 # global
        for i = 1:10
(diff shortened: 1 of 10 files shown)


test/doctests/src/softscope.md
```diff
@@ -0,0 +1,52 @@
...
+    s = t # new local `s` with warning
+end
+s, # global
+@isdefined(t) # global
+
+# output
+
+┌ Warning: Assignment to `s` in soft scope is ambiguous because a global variable by the same name exists: `s` will be treated as a new local. Disambiguate by using `local s` to suppress this warning or `global s` to assign to the existing global variable.
+└ @ ~/dev/Documenter/test/doctests/doctests.jl:3
+ERROR: UndefVarError: s not defined
```
StefanKarpinski: I guess the question is what do you actually want to have happen here? Is this usually short hand for "run this in the REPL and here's the output you should see"? or is it actually short for "run this in a script and here's the output you should see"?

fredrikekre: The latter. For the latter there is the `julia>` style doctests.

StefanKarpinski: Ok, then carry on, this is perfect as it is 😸

PR #1239, opened 2020-02-16
Document GITHUB_TOKEN workaround

### LINKED ISSUE #1177: GitHub Pages build not triggered with GITHUB_TOKEN

bramtayl:
Closes #1177

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -81,7 +81,7 @@ The three lines in the `script:` section do the following:
     Travis CI used to use `matrix:` as the section to configure to build matrix in the config
     file. This now appears to be a deprecated alias for `jobs:`. If you use both `matrix:` and
     `jobs:` in your configuration, `matrix:` overrides the settings under `jobs:`.
-    
+
     If your `.travis.yml` file still uses `matrix:`, it should be replaced with a a single
     `jobs:` section.
 
@@ -216,9 +216,13 @@ see the previous section.
 
 ### Authentication: `GITHUB_TOKEN`
 
-When running from GitHub Actions it is possible to authenticate using
-[the GitHub Actions authentication token
-(`GITHUB_TOKEN`)](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token). This is done by adding
+When running from GitHub Actions it is possible to authenticate using a personal
+access token.
+
+First, create a new personal access token named "GITHUB_TOKEN" with
+the "repo" scope checked [here](https://github.com/settings/tokens/new).
+
+Then, add
 
 ```yaml
 GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


docs/src/man/hosting.md
```diff
@@ -227,15 +231,9 @@ GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
...
-    elsewhere you should instead use a SSH deploy key.
-
-!!! warning "GitHub Pages and GitHub Token"
-    Currently the GitHub Page build is not triggered when the GitHub provided
-    `GITHUB_TOKEN` is used for authentication. See
-    [issue #1177](https://github.com/JuliaDocs/Documenter.jl/issues/1177)
-    for more information.
+    While GitHub will automatically create a github token for you named "GITHUB_TOKEN",
+    it will not have push access. Therefore, you must override
+    GitHub by making a new personal access token with the same name.
```
aminya: It should specifically say what to do.

aminya:
I also think more adjustment to the documentation is needed based on what I said here:
 https://github.com/JuliaDocs/Documenter.jl/issues/1177#issuecomment-586662631

docs/src/man/hosting.md
bramtayl: I can add a note like `(see instructions above)` because I did add instructions to do this?

aminya: Yes, and a link to that part of the instructions

bramtayl:
Now that github actions is up and running maybe we should just ditch all the travis and ssh-key docs?

aminya:
> 
> 
> Now that github actions is up and running maybe we should just ditch all the travis and ssh-key docs?

I agree. Probably it is better to move them to a separate document called `Travis.md`.

fredrikekre:
IMO we should not Document this option and only recommend SSH keys (as ww aöready do).

bramtayl:
Why?

fredrikekre:
Because access tokens are not limited to one repo, and is tied to a user, so it would be like giving your github password to everyone with write access to the repo in question.

bramtayl:
Hmm I think I see what your saying. Still seems like a pain to make SSH keys for each individual repo. I wonder if there's another solution?

fredrikekre:
You can reuse it if you want

mortenpi:
I have to agree with @fredrikekre here: we shouldn't recommend the use of personal access tokens for this.

> You can reuse [the SSH deploy key] if you want

I don't think this is true. If you try to add the same deploy key again, GitHub will complain if I remember correctly.

> Now that github actions is up and running maybe we should just ditch all the travis and ssh-key docs?

As long as we don't have #1177, we can't ditch the deploy key stuff. We could have GitHub Actions as the recommended setup (even as we're waiting for #1177), but no reason to ditch Travis -- it can still be useful for some people.

bramtayl:
Close in favor of #1244

PR #1244, opened 2020-02-19
remove travis docs, point to OnlinePackage

bramtayl:
Thought this would simplify the manual a bit

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -2,8 +2,8 @@
 
 After going through the [Package Guide](@ref) and [Doctests](@ref) page you will need to
 host the generated documentation somewhere for potential users to read. This guide will
-describe how to setup automatic updates for your package docs using either the Travis CI
-build service or GitHub Actions together with and GitHub Pages for hosting the generated
+describe how to setup automatic updates for your package docs using
+GitHub Actions together with and GitHub Pages for hosting the generated
 HTML files. This is the same approach used by this package to host its own docs --
 the docs you're currently reading.
 
@@ -14,11 +14,11 @@ the docs you're currently reading.
     proceed with the steps outlined here once you have successfully managed to build your
     documentation locally with Documenter.
 
-    This guide assumes that you already have [GitHub](https://github.com/) and
-    [Travis](https://travis-ci.com/) accounts setup. If not then go set those up first and
+    This guide assumes that you already have a [GitHub](https://github.com/)
+    account setup. If not then go set those up first and
     then return here.
 
-    It is possible to deploy from other systems than Travis CI or GitHub Actions,
+    It is possible to deploy from other systems than GitHub Actions,
     see the section on [Deployment systems](@ref).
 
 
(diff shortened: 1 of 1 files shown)


docs/src/man/hosting.md
```diff
@@ -42,135 +42,13 @@ service as the one testing your package. If you don't explicitly select
...
-        about Travis environment variables in [Travis User Documentation](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings).
-
-!!! note
-
-    There are more explicit instructions for adding the keys to Travis in the
-    [SSH Deploy Keys Walkthrough](@ref) section of the manual.
-
+generated documentation from the builder to GitHub. The SSH keys can be generated and
+submitted with `OnlinePackage.jl`. See the instructions for using that package
+[here](https://bramtayl.github.io/OnlinePackage.jl/latest/).
```
KristofferC: I read through that "documentation" but it is just a bunch of docstrings? I have no idea how I am supposed to use that package. It talks about some `User` type (what is that), something about deleting a repository (how is that relevant to this?), something about creating a user profile from a toml file where I still should go in and manually generate keys and then create some file locally? And what do I do if my repo is already online?

So far, I am quite unconvinced that this is any clearer and understandable than the existing documentation here that gets deleted.

mortenpi:
:+1: to mentioning OnlinePackage although, as @KristofferC said, more detailed instructions are always good.

A tangential comment regarding OnlinePackage: I only very briefly glanced at the code, but it looks like it more or less just handles the keys now and is pretty much Documenter-specific? If that is the case, that functionality would also be a good fit for [DocumenterTools](https://github.com/JuliaDocs/DocumenterTools.jl), if you'd like to merge them.

But either way, we do want to keep the Travis and `GITHUB_TOKEN` instructions around. The former are still useful for people who are using Travis for CI, and the latter will ideally become the default at some point (possibly soon if https://github.com/JuliaDocs/Documenter.jl/issues/1177#issuecomment-588091584 is true). Plus, it's good to have the logic documented, even if there is a tool that can automate the process.

bramtayl:
Ok y'all I hear you. I'll work on writing a detailed tutorial for OnlinePackage. Yes it's pretty SSH key specific; happy to merge into DocumenterTools (I've been pushing for it to get merged into PkgTemplates, but that doesn't seem to be going anywhere). Once that's done I'll edit the PR to just mention OnlinePackage and not delete any docs

bramtayl:
Given this https://github.com/JuliaDocs/Documenter.jl/issues/1177#issuecomment-588525572 I think I'll deprecate OnlinePackage

fredrikekre:
The functionality to automatically upload the keys might be useful to include in DocumenterTools.

bramtayl:
Ok, I can keep it around as long as you'd like. Does anyone know the official process for deprecating packages? Last time I deleted packages it caused chaos...

fredrikekre:
You don't have to do anything.

bramtayl:
Ok

PR #1254, opened 2020-03-08
Link to repo in navbar

mortenpi:
Adds a way to link to the root of the repository in the navbar. The prominence of the edit link is reduced by having it just as an icon:

repolink

By default, HTMLWriter will try to figure this out based on the remote repository, which can be configured using the `repo` argument of `makedocs`:

* Documenter now exports the `Remotes` module, which contains types to configure the remote explicitly, e.g.: `makedocs(repo = Remotes.GitHub("JuliaDocs", "Documenter.jl")`
* If the `repo` keyword is not passed, Documenter tries to determine it by looking at the `origin` remote of the source (as before).
* The old URL template strings can also be passed, but in that case the repository link will be hidden, as we don't know what part of the URL is the root.

The `HTML` format type now also takes the `repolink` keyword which can be used to either override the repository URL determined from `repo` (by passing a `String`), or hide the repository link (by passing `nothing`).

Close #1377


tfiers:
There is animo for this! https://julialang.slack.com/archives/C67TK21LJ/p1640113944190400

Anything we can do to help?

fredrikekre (approved):

gaurav-arya:
It seems like this change has not been included in the latest release, although it was included in the changelog for v0.27.22? https://github.com/JuliaDocs/Documenter.jl/blob/v0.27.23/src/Writers/HTMLWriter.jl

fredrikekre:
It is under the 0.28 release, no?

sloede:
Out of curiosity (since I've been watching this PR as well): Is there a (tentative) ETA on v0.28?

mortenpi:
That's my bad -- it did end up in the [auto-generated changelog of the GitHub release](https://github.com/JuliaDocs/Documenter.jl/releases/tag/v0.27.22). The only changelog that matters is what is [in `CHANGELOG.md`](https://github.com/JuliaDocs/Documenter.jl/blob/v0.27.23/CHANGELOG.md#version-v02723). We need to invoke TagBot in a way that it doesn't generate its own release notes -- they're not correct if you have a multi-branch backport workflow.

As for 0.28: I wouldn't want to put a date to it. There are a couple of fixes that should go in first that I haven't had the chance to tackle yet. However, you can always add a Manifest and use the `master` branch of Documenter, if you want to take advantage of the latest features (e.g. like is done for this package: https://juliaatoms.org/AngularMomentumAlgebra.jl/stable/).

PR #1265, opened 2020-03-20
Add note about doctests in docstrings

helgee:
Suggest to add modules with doctests in docstrings to the `modules` argument. Otherwise you get the strange situation where the doctests are not run when running `makedocs` but do run and potentially fail when running `doctest`.

CC: @fredrikekre

### DIFF
--- a/docs/src/man/doctests.md
+++ b/docs/src/man/doctests.md
@@ -237,6 +237,10 @@ DocMeta.setdocmeta!(MyPackage, :DocTestSetup, :(using MyPackage); recursive=true
 makedocs(modules=[MyPackage], ...)
 ```
 
+!!! note
+    Please make sure to include all modules that contain docstrings with doctests in the
+    `modules` argument to [`makedocs`](@ref). Otherwise these doctests will not be run.
+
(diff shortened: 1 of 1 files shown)


mortenpi:
Just to clarify: this is for when you are building docs for multiple packages at once? It should not be necessary to list all _submodules_ of your package separately if you already have the top-level module there.

helgee:
In this specific case I was completely missing the `modules` argument and from the docs it was not really clear (to me) that it is required.

docs/src/man/doctests.md
```diff
@@ -237,6 +237,10 @@ DocMeta.setdocmeta!(MyPackage, :DocTestSetup, :(using MyPackage); recursive=true
 makedocs(modules=[MyPackage], ...)
 ```
 
+!!! note
+    Please make sure to include all modules that contain docstrings with doctests in the
```
mortenpi: Sounds good. What do you think about clarifying slightly as follows?

```suggestion
    Make sure to include all (top-level) modules that contain docstrings with doctests in the
```

PR #1269, opened 2020-03-22
Fix minted/lstlistings escaping

### LINKED ISSUE #1118: The julia code block contains character `%` can not generate PDF format output.

### LINKED ISSUE #1200: LaTeX rendering trickiness
Here is a code block that seems to fail the build (Edit: doesn't actually fail the build, just that the output is empty):
````
```julia
+=  -=  *=  /=  \=  ÷=  %=  ^=  &=  |=  ⊻=  >>>=  >>=  <<=
```
````
becomes
````
\begin{minted}[escapeinside=\%\%]{julia}
+=  -=  *=  /=  \=  ÷=  %\%%=  ^=  &=  |=  %\unicodeveebar%=  >>>=  >>=  <<=\end{minted}
````

and same with `julia-repl`. No language works though, i.e.
```
\begin{lstlisting}[escapeinside=\%\%]
+=  -=  *=  /=  \=  ÷=  %\%%=  ^=  &=  |=  %\unicodeveebar%=  >>>=  >>=  <<=\end{lstlisting}
```
so seems to be something with `minted`.

fredrikekre:
Fix  #1200, fix #1118 , close #1119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * ![Bugfix][badge-bugfix] Remove `only`, a new export from `Base` on Julia 1.4, from the JS search filter. ([#1264][github-1264])
 
-* ![Bugfix][badge-bugfix] Fix errors in LaTeX builds due to bad escaping of certain characters. ([#1200][github-1200], [#1119][github-1119], [#1269][github-1269])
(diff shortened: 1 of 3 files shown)


fredrikekre:
Built the pdf for Julia 1.4.0 with this patch.

PR #1271, opened 2020-03-24
Allow replacements in doctest filters

mortenpi:
This allows regex - substitution pairs to be passed as doctest filters, giving more control over filtering.

For example, the following filter can be used to only retain the first 8 digits after the decimal point of a floating point number: `r"([0-9]+\.[0-9]{8})[0-9]+" => s"\1***"`

This means that this will pass (`sqrt(2) == 1.4142135623730951`):

````
```jldoctest; filter = r"([0-9]+\.[0-9]{8})[0-9]+" => s"\1***"
julia> sqrt(2)
1.41421356000
```
````

But this will not:

````
```jldoctest; filter = r"([0-9]+\.[0-9]{8})[0-9]+" => s"\1***"
julia> sqrt(2)
1.4142999999999
```
````

### CODE REVIEW

File: test/doctests/src/broken.md [hunk 1]
```diff
@@ -1,6 +1,13 @@
-This source file contains a broken doctest:
+This source file contains a broken doctests:
```
KristofferC:
```suggestion
This source file contains broken doctests:
```

PR #1280, opened 2020-04-08
Pass non-standard admonition categories on in HTML output

### LINKED ISSUE #1279: Feature request: custom admonition styling
## The problem

Prior to 0.24 (I think), it was possible to add custom admonition styling. Eg, I could make `docs/assets/admonitions.css` containing

```
.admonition.practice > .admonition-title {
 background-color: rgba(152,78,163,0.8);
}

.admonition.practice {
 background-color: rgba(152,78,163,0.3);
}
```

and 
```
format=Documenter.HTML(; # ...
    assets=String["assets/admonitions.css"])
```

inside `makedocs()`, to get purple admonitions when using `!!! practice`. This doesn't seem to work anymore, since the admonition name doesn't seem to get passed into the html for non-defaults. Eg **Example1**:

```
!!! practice "test"
    This is a test
```

gets turned into 

```
<div class="admonition"><header class="admonition-header">test</header><div class="admonition-body"><p>This is a test</p></div></div>
```

Whereas **Example2**:

```
!!! compat "test"
    This is a test
```

gets turned into 

```
<div class="admonition is-compat"><header class="admonition-header">test</header><div class="admonition-body"><p>This is a test</p></div></div>
```

---

## Proposal

Add the admonition type to the class, so that example 1 above would get turned into 

```
<div class="admonition is-practice"><header class="admonition-header">test</header><div class="admonition-body"><p>This is a test</p></div></div>
```

Then, if I understand correctly, I could change my `admonitions.css` to

```
.admonition.is-practice > .admonition-header {
 background-color: rgba(152,78,163,0.8);
}

.admonition.is-practice > .admonition-body {
 background-color: rgba(152,78,163,0.3);
}
```

mortenpi:
Arguably fixing a regression, although it is an undocumented "feature". But it is a good idea to pass these on.

Compared to pre-0.24 behaviour, the class names are now prepended with ~`admonition-`~ `is-category-` (not using `is-`, since that can conflict with various Bulma classes, like `is-hidden`) and fully sanitized (the final class names match `[a-z0-9-]+`).

Fix #1279.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * ![Bugfix][badge-bugfix] `Deps.pip` is again a closure and gets executed during the `deploydocs` call, not before it. ([#1240][github-1240])
 
+## Version `v0.24.8`
+
+* ![Enhancement][badge-enhancement] Non-standard admonition categories are (again) applied to the admonition `<div>` elements in HTML output (as `admonition-$category`). ([#1279][github-1279], [#1280][github-1280])
+
 ## Version `v0.24.7`
 
 * ![Bugfix][badge-bugfix] Remove `only`, a new export from `Base` on Julia 1.4, from the JS search filter. ([#1264][github-1264])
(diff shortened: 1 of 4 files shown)


kescobo:
Thank you!

kescobo:
Works like a charm:

Screen Shot 2020-04-08 at 7 58 34 PM

From

```
.admonition.is-category-practice > .admonition-header {
 background-color: rgba(152,78,163,0.8);
}

.admonition.is-category-practice > .admonition-body {
 background-color: rgba(152,78,163,0.3);
}
```

Would you be interested in a PR to actually add this to the docs?

mortenpi:
I am slightly hesitant to have it documented as public API just yet, in case we still want to make some changes to the structure of CSS classes. But if you don't mind, we could put the example with a few words of explanation into the [example build](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/src/man/style.md#styling-of-lists) (among other things, the example build functions as sort of a gallery of unofficial Documenter hacks).

PR #1281, opened 2020-04-08
[release-0.24] Backports for 0.24.8

mortenpi:
X-ref: #1280, #1287


fredrikekre:
Backported https://github.com/JuliaDocs/Documenter.jl/pull/1287 but forgot about the changelog... It is a pretty obscure thing anyway so not sure a changelog is worth it.

PR #1283, opened 2020-04-09
Fix text/latex output in HTMLWriter

### LINKED ISSUE #1278: "text/latex" output not rendered
Example using SymPy:

in md:
```@example tutorial
using SymPy
(x) = SymPy.symbols("x", real=true)
[x 2x]
```
Output in compiled html:
[\left[ \begin{array}{rr}x&2 x\end{array}\right]]

Image of compiled html:

edit: I should mention that it works in jupyter notebooks. For example: https://nbviewer.jupyter.org/github/JuliaPy/SymPy.jl/blob/master/examples/tutorial.ipynb

mortenpi:
Aiming to fix #1278.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1646,7 +1646,11 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
     elseif haskey(d, MIME"image/jpeg"())
         out = Documents.RawHTML(string("<img src=\"data:image/jpeg;base64,", d[MIME"image/jpeg"()], "\" />"))
     elseif haskey(d, MIME"text/latex"())
-        out = Utilities.mdparse(d[MIME"text/latex"()]; mode = :single)
+        # If the show(io, ::MIME"text/latex", x) output is already wrapped in \[ ... \], we
+        # unwrap it first, since when we output Markdown.LaTeX objects we put the correct
+        # delimiters around it anyway.
+        m = match(r"\s*\\\[(.*)\\\]\s*", d[MIME"text/latex"()])
+        out = Markdown.LaTeX(m === nothing ? d[MIME"text/latex"()] : m[1])
     elseif haskey(d, MIME"text/markdown"())
         out = Markdown.parse(d[MIME"text/markdown"()])
     elseif haskey(d, MIME"text/plain"())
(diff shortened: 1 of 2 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -1646,7 +1646,11 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
     elseif haskey(d, MIME"image/jpeg"())
         out = Documents.RawHTML(string("<img src=\"data:image/jpeg;base64,", d[MIME"image/jpeg"()], "\" />"))
     elseif haskey(d, MIME"text/latex"())
-        out = Utilities.mdparse(d[MIME"text/latex"()]; mode = :single)
```
mortenpi, line 1649: As far as I can tell, this `mdparse` call for LaTeX was introduced in #938. @fredrikekre I know this was ages ago, but do you remember whether there was any particular reason for this? I don't think we should expect `text/latex` representations be formatted as Markdown?

src/Writers/HTMLWriter.jl
```diff
@@ -1646,7 +1646,11 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
...
+        # If the show(io, ::MIME"text/latex", x) output is already wrapped in \[ ... \], we
+        # unwrap it first, since when we output Markdown.LaTeX objects we put the correct
+        # delimiters around it anyway.
+        m = match(r"\s*\\\[(.*)\\\]\s*", d[MIME"text/latex"()])
```
mortenpi, line 1652: The stripping of `\[` and `\]` is directly motivated by SymPy. I am wondering if we should also strip `$ .. $`,`$$ .. $$`, `\( .. \)` etc.

src/Writers/HTMLWriter.jl
fredrikekre: Probably just a copy-paste error.

fredrikekre:
Thanks, I had a similar fix on some machine.

The tricky thing here is what should happen with something like
```
This is $x$.
```
where not everything is math.

mortenpi:
Fair point, this may need a little bit more thought. It looks like it might end up being slightly breaking though, so let's leave this for 0.25.0.

src/Writers/HTMLWriter.jl
MichaelHatherly: Seems reasonable to me to be consistent here and just strip them all.

PR #1284, opened 2020-04-09
Add example admonition css to example build

kescobo:
Pseudo-document custom CSS for admonitions in example build. Re: #1280 

I wasn't sure how to test this build locally, but I'm ~92% sure it should work :-)

### DIFF
--- a/test/examples/src/man/style.md
+++ b/test/examples/src/man/style.md
@@ -25,6 +25,21 @@ Also, custom admonition classes can be used:
 !!! myadmonition "My Admonition Class"
 
     In the HTML output, this admonition has `is-category-myadmonition` applied to it.
+    Its style can be changed by adding styles to `.admonition-header` and `.admonition-body`
+    in a custom css file and adding it to the build with, for example:
+
+    ```julia
+    makedocs(
+        # ...
+        format=Documenter.HTML(;
+            #...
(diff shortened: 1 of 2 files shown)


test/examples/src/man/style.md
```diff
@@ -25,6 +25,21 @@ Also, custom admonition classes can be used:
...
+        # ...
+        format=Documenter.HTML(;
+            #...
+            assets=["assets/custom.css"]
+        )
+    )
+    ```
+
+    [See here](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/src/assets/custom.css)
+    for an example `custom.css` file.
```
mortenpi, line 42:
```suggestion
    See [`test/examples/src/assets/custom.css`](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/src/assets/custom.css)
    for an example of a custom CSS file.
```

mortenpi:
LGTM, thanks a lot!

> I wasn't sure how to test this build locally, but I'm ~92% sure it should work :-)

I can report that the posterior probability here is 100% :slightly_smiling_face: 

These test examples can be built by just calling `julia --project test/examples/make.jl` (although the environment also needs the test dependencies). Alternatively, they also get built during testing.

PR #1285, opened 2020-04-09
Add option to specify username when deploying via ssh

chrisbrahms:
This deals with the case of a username being specified in the the repository url in `deploydocs`. That is, if one calls `deploydocs` like this:
```julia
deploydocs(
    repo = "user@host:/path/to/repo",
)
```
deploying now works, rather than trying to push to `git@user@host:/path/to/repo`.

This came up when hosting documentation on a webserver where the `git` user already exists but I can't get access to it.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -185,7 +185,7 @@ when a minor version changes (i.e. except in patch releases).
 
 **`checkdocs`** instructs [`makedocs`](@ref) to check whether all names within the modules
 defined in the `modules` keyword that have a docstring attached have the docstring also
-listed in the manual (e.g. there's a `@docs` blocks with that docstring). Possible values
+listed in the manual (e.g. there's a `@docs` block with that docstring). Possible values
 are `:all` (check all names; the default), `:exports` (check only exported names) and
 `:none` (no checks are performed). If `strict` is also enabled then the build will fail if
 any missing docstrings are encountered.
@@ -244,7 +244,7 @@ $(SIGNATURES)
 
 Allows a page to be hidden in the navigation menu. It will only show up if it happens to be
 the current page. The hidden page will still be present in the linear page list that can be
-accessed via the previous and next page links. The title of the hidden page can be overriden
+accessed via the previous and next page links. The title of the hidden page can be overridden
 using the `=>` operator as usual.
 
 # Usage
@@ -344,7 +344,7 @@ currently running environment. See the manual section about
 **`root`** has the same purpose as the `root` keyword for [`makedocs`](@ref).
 
 **`target`** is the directory, relative to `root`, where generated content that should be
-deployed to `gh-pages` is written to. written to. It should generally be the same as
+deployed to `gh-pages` is written to. It should generally be the same as
 [`makedocs`](@ref)'s `build` and defaults to `"build"`.
 
 **`branch`** is the branch where the generated documentation is pushed. If the branch does
@@ -360,10 +360,10 @@ It can be used e.g. for a Markdown build. The following example installed the `p
 deps = Deps.pip("pygments", "mkdocs")
 ```
 
-**`make`** is the function used to specify an additonal build phase. By default, nothing gets
+**`make`** is the function used to specify an additional build phase. By default, nothing gets
 executed.
 
-**`devbranch`** is the branch that "tracks" the in-development version of the  generated
+**`devbranch`** is the branch that "tracks" the in-development version of the generated
 documentation. By default this value is set to `"master"`.
 
 **`devurl`** the folder that in-development version of the docs will be deployed.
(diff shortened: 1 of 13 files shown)


src/Documenter.jl
```diff
@@ -474,7 +474,7 @@ The documentation are placed in the folder specified by `subfolder`.
 function git_push(
         root, temp, repo;
         branch="gh-pages", dirname="", target="site", sha="", devurl="dev",
-        versions, forcepush=false, deploy_config, subfolder,
+        versions, forcepush=false, deploy_config, subfolder, username="git"
```
mortenpi, line 477: `username` is not used anywhere?

src/Documenter.jl
```diff
@@ -555,11 +555,23 @@ function git_push(
...
+            # The upstream URL to which we push new content and the ssh decryption commands.
+            upstream = "$(replace(repo, "$host/" => "$host:"))"
+        else
+            # if no user given, default to "git"
+            user = "git"
+            # Extract host from repo as everything up to first ':' or '/' character
+            host = match(r"(.*?)[:\/]", repo)[1]
+            # The upstream URL to which we push new content and the ssh decryption commands.
+            upstream = "git@$(replace(repo, "$host/" => "$host:"))"
+        end
```
mortenpi, line 573: Just a suggestion: it feels like it should be possible to do this with just a single regex, e.g.:

```julia
match(r"(?:([^@]*)@)?(.*?)[:\/]", "user@foo.com/blah.git")
```

And if the first capture group is `nothing`, then use the default username.

mortenpi: Also, could we refactor this into a function (returning a `user, upstream` tuple?) and add some unit tests, just to make sure that the code does exactly what it is supposed to?

src/Documenter.jl
chrisbrahms: that was left over from an earlier attempt, thanks for catching it

src/Documenter.jl
chrisbrahms: Yes, that is much more elegant (my feeble regex skills failed me). Where would be a good place to put the unit tests?

chrisbrahms:
I've factored out the repo-disassembly into `user_host_upstream` and added some tests. These are currently in `test/deployconfig.jl`--is that the correct place?

Regarding the new function `user_host_upstream`: to make `strict` happy, should I add a `@docs` entry for it or just remove the docstring?

src/Documenter.jl
```diff
@@ -474,7 +474,7 @@ The documentation are placed in the folder specified by `subfolder`.
 function git_push(
         root, temp, repo;
         branch="gh-pages", dirname="", target="site", sha="", devurl="dev",
-        versions, forcepush=false, deploy_config, subfolder,
+        versions, forcepush=false, deploy_config, subfolder
```
mortenpi, line 477: We generally keep the trailing commas:

```suggestion
        versions, forcepush=false, deploy_config, subfolder,
```

test/deployconfig.jl
```diff
@@ -228,3 +228,19 @@ end end
         @test cfg === nothing
     end
 end
+##
+@testset "Remote repository paths" begin
+    uhu = Documenter.user_host_upstream("github.com/JuliaDocs/Documenter.jl.git")
+    @test uhu == ("git", "github.com", "git@github.com:JuliaDocs/Documenter.jl.git")
+
```
mortenpi, line 235:
```suggestion

    uhu = Documenter.user_host_upstream("github.com:JuliaDocs/Documenter.jl.git")
    @test uhu == ("git", "github.com", "git@github.com:JuliaDocs/Documenter.jl.git")

    uhu = Documenter.user_host_upstream("gitlab.com/JuliaDocs/Documenter.jl")
    @test uhu == ("git", "gitlab.com", "git@gitlab.com:JuliaDocs/Documenter.jl")

```

src/Documenter.jl
mortenpi: We can add it here: https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/lib/internals/documenter.md

src/Documenter.jl
```diff
@@ -624,6 +621,29 @@ function rm_and_add_symlink(target, link)
...
+    #= the regex has three parts:
+    (?:([^@]*)@)?  matches any number of characters up to the first "@", if present,
+        capturing only the characters before the "@" - this captures the username
+    (?:([^\/:]*)[\/:]){1}  matches exactly one instance of any number of characters
+        other than "/" or ":" up to the first "/" or ":" - this captures the hostname
+    [\/]?(.*)  matches the rest of the repo, except an initial "/" if present (e.g. if
+        repo is of the form usr@host:/path/to/repo) - this captures the path on the host
+    =#
+    m = match(r"(?:([^@]*)@)?(?:([^\/:]*)[\/:]){1}[\/]?(.*)", repo)
+    isnothing(m) && error("Invalid repo path $repo")
```
mortenpi, line 640: This is failing on 1.0 (we're not depending on Compat):

```suggestion
    (m === nothing) && error("Invalid repo path $repo")
```

mortenpi:
> Regarding the new function `user_host_upstream`: to make `strict` happy, should I add a `@docs` entry for it or just remove the docstring?

We can add it here: https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/lib/internals/documenter.md

mortenpi:
Good to go as far as I can tell. Thanks for the contribution @chrisbrahms!

PR #1289, opened 2020-04-13
Don't skip duplicate docstrings

asinghvi17:
Anchors should take care of deduplication.


mortenpi:
Are you planning on working on something here or can we close this?

Just as a comment, in general, instructing Documenter to include multiple copies of the same docstring should be considered an error (e.g. you moved some docstrings around, but forgot to delete it from the old at-docs block).

PR #1292, opened 2020-04-14
Omit the number for the first link fragment on a page

fredrikekre:
Omit the number for the first link fragment on a page;
before: #foobar-1, #foobar-2
after: #foobar, #foobar-2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * ![Bugfix][badge-bugfix] `Deps.pip` is again a closure and gets executed during the `deploydocs` call, not before it. ([#1240][github-1240])
 
+* ![Enhancement][badge-enhancement] The first link fragment on each page now omits the number; before the rendering resulted in: `#foobar-1`, `#foobar-2`, and now: `#foobar`, `#foobar-2`. For backwards compatibility the old fragments are also inserted such that old links will still point to the same location. ([#1292][github-1292])
+
 ## Version `v0.24.9`
 
 * ![Bugfix][badge-bugfix] Canonical URLs are now properly prettified (e.g. `/path/` instead of `/path/index.html`) when using `prettyurls=true`. ([#1293][github-1293])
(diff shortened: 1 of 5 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -1218,16 +1218,23 @@ end
...
-            a[".docs-heading-anchor", :href => "#$aid"](mdconvert(h.text, h)),
-            a[".docs-heading-anchor-permalink", :href => "#$aid", :title => "Permalink"]
+        DOM.Tag(Symbol("h$(Utilities.header_level(h))"))[:id => lstrip(frag, '#')](
+            a[".docs-heading-anchor", :href => frag](mdconvert(h.text, h)),
+            legacy...,
+            a[".docs-heading-anchor-permalink", :href => frag, :title => "Permalink"]
         )
     else
-        a[:id => aid, :href => "#$aid"](domify(ctx, navnode, anchor.object))
+        error("hmmm...")
```
fredrikekre, line 1236: This path seems to not be used in Documenter at all. Can you even attach `id` to things that are not headers? I guess we can keep it, but maybe no need to add the legacy link here.

CHANGELOG.md
```diff
@@ -544,7 +546,11 @@
 [github-1269]: https://github.com/JuliaDocs/Documenter.jl/pull/1269
 [github-1279]: https://github.com/JuliaDocs/Documenter.jl/issues/1279
 [github-1280]: https://github.com/JuliaDocs/Documenter.jl/pull/1280
+<<<<<<< 048d2fa4b307fe88d5db28d300038b6f0468cfbb
 [github-1293]: https://github.com/JuliaDocs/Documenter.jl/pull/1293
+=======
+[github-1292]: https://github.com/JuliaDocs/Documenter.jl/pull/1292
+>>>>>>> Omit the number for the first link fragment on a page;
```
mortenpi, line 553: Bad merge here.

PR #1298, opened 2020-04-19
Discard build from version number when deploying

mortenpi:
With this, tags like `v1.2.3+foo` still deploy into `v1.2.3/`. This allows documentation to be depoyed for a version even after that version has been merged into the registry (and hence the tag in the repo should not be changed anymore).

A use case, for example, is when there is a configuration issue that kills the documentation build (but otherwise the package is fine). You can then fix the issue, commit the fix and locally tag that new commit with e.g. `v1.2.3+doc1`.


mortenpi:
Test runs:

* GitHub Actions: https://github.com/mortenpi/Example.jl/runs/599002031?check_suite_focus=true
* Travis CI: https://travis-ci.org/github/mortenpi/Example.jl/builds/676800586

PR #1288, opened 2020-04-13
Add warning about TagBot permissions

mortenpi:
Fix #1277

PR #1299, opened 2020-04-19
PDF/LaTeX: Escape characters in admonition titles

mortenpi:
Should resolve the PDF build error in #1288.


src/Writers/LaTeXWriter.jl
```diff
@@ -681,7 +681,6 @@ const _latexescape_chars = Dict{Char, AbstractString}(
     '\\' => "{\\textbackslash}",
     '\'' => "{\\textquotesingle}",
     '"' => "{\\textquotedbl}",
-    '_' => "{\\_}",
```
mortenpi, line 684: Technically unrelated, but this duplicates the loop below.

PR #1301, opened 2020-04-20
Make the depwarn test dependent on --depwarn value

mortenpi:
The nightly tests have been failing since JuliaLang/julia#35362 was merged, since the test that checks whether deprecation warnings are captured correctly assumed that the deprecation warning gets printed even if `--depwarn=no` (which is now the default).

Should also be pushed to #1300 once merged.

### DIFF
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -268,6 +268,8 @@ end
             # warnings enabled or not. To figure out whether that is the case or not, we can
             # look at the .depwarn field of the undocumented Base.JLOptions object.
             @test isdefined(Base, :JLOptions)
+            # hasfield was added in Julia 1.2. This definition borrowed from Compat.jl (MIT)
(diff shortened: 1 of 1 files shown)


fredrikekre:
Maybe backport this unless 0.25 is close? Actually, maybe backport in any case.

mortenpi:
I got backported in #1300. Should actually remove the labels now :slightly_smiling_face:

fredrikekre:
Okay, so there is something else that is broken on 1.5 then.

mortenpi:
Oh, you're right. `master` is failing as well. I am wondering if we should set up a GitHub Actions for nightlies -- it's easy to never notice nightly failures on Travis as they're allowed failures.

fredrikekre:
Documenter master works for me though, so looks like it has been fixed.

PR #1307, opened 2020-05-02
Add support for pushing PR previews to a different branch and/or different repo

DilumAluthge:
Currently, the only way to have PR previews is to put them in the `previews/PRXX` subdirectory in the same branch of the same repository as your actual docs. For example, if your actual docs are pushed to the `gh-pages` branch of the `MyPackage.jl` repo, then the PR preview for PR number 123 is pushed to the `previews/PR123` subdirectory in the `gh-pages` branch of the `MyPackage.jl` repo.

But, for a variety of reasons, you may not want to keep your PR previews in the same branch and/or repo as your actual docs.

This PR adds the ability to customize the branch and/or repo to which PR previews are pushed. The options are:
- Same repo, same branch - Prior to this PR, this is the only option. In this PR, we keep it as the default behavior.
- Same repo, different branch
- Different repo, same branch name
- Different repo, different branch name

We also add support for the `DOCUMENTER_KEY_PREVIEWS` environment variable. If the `DOCUMENTER_KEY_PREVIEWS` environment variable is set, then that will be the SSH key that is used for pushing PR previews. If the `DOCUMENTER_KEY_PREVIEWS` environment variable, then the regular `DOCUMENTER_KEY` environment variable will be used.

--- 

### Example
- Pull request: https://github.com/JuliaHealth/HealthBase.jl/pull/7
- Preview: https://juliahealth.org/HealthBase.jl-previews/previews/PR7/
- Repo to which previews are pushed: https://github.com/JuliaHealth/HealthBase.jl-previews/tree/gh-pages/previews

--- 

### Prior art

This is quite easy with Franklin. See e.g. my setup here:
- Workflow file: https://github.com/JuliaHealth/juliahealth.github.io/blob/dev/.github/workflows/deploy.yml
- Repo to which previews are pushed: https://github.com/JuliaHealth/juliahealth.github.io-previews/tree/gh-pages/previews


DilumAluthge:
This is ready for review @mortenpi @fredrikekre

DilumAluthge:
Bump @mortenpi

mortenpi:
I haven't gone through the code in detail yet but glancing at it, it looks good! Will try to go through this within the next few days.

Could you also add a note into `CHANGELOG.md`?

DilumAluthge:
> Could you also add a note into `CHANGELOG.md`?

Done!

Let me know if the changelog entry is too verbose.

src/deployconfig.jl
```diff
@@ -20,6 +28,21 @@ function documenter_key(::DeployConfig)
...
+Return the Base64-encoded SSH private key for the repository.
+Uses the `DOCUMENTER_KEY_PREVIEWS` environment variable if it is defined,
+otherwise uses the `DOCUMENTER_KEY` environment variable.
+
+This method must be supported by configs that push with SSH, see
+[`Documenter.authentication_method`](@ref).
+"""
+function documenter_key_previews(cfg::DeployConfig)
+    return get(ENV, "DOCUMENTER_KEY_PREVIEWS", documenter_key(cfg))
+
```
mortenpi, line 43:
```suggestion
```

DilumAluthge:
> One tiny request: could we have a test in test/deployconfig.jl that makes sure that DOCUMENTER_KEY_PREVIEWS gets properly fetched from the environment too

This part is done.

> (and gets ignored when it is in the env. but it's not a preview build)

This will be hard. The logic for this part is in the `git_push` function, so to test it we'd need to have some kind of integration test harness.

mortenpi:
> > (and gets ignored when it is in the env. but it's not a preview build)
> 
> This will be hard. The logic for this part is in the `git_push` function, so to test it we'd need to have some kind of integration test harness.

Ah, that's right, I didn't look at it closely enough. In that case, I'd say it's good to go. Thanks for the contribution!

For planning purposes: would it be fine for you to use Documenter `master` for a while? I would definitely like to have this in a minor release, but there are a few other things I would like to get into 0.25, so it might be a few weeks before a release.

src/Documenter.jl
```diff
@@ -461,13 +484,14 @@ function deploydocs(;
                 @debug "running extra build steps."
                 make()
             end
-            @debug "pushing new documentation to remote: '$repo:$branch'."
+            @debug "pushing new documentation to remote: '$deploy_branch:$branch'."
```
fredrikekre, line 487: `deploy_repo`?

DilumAluthge: Yeah good call

DilumAluthge:
> For planning purposes: would it be fine for you to use Documenter `master` for a while? I would definitely like to have this in a minor release, but there are a few other things I would like to get into 0.25, so it might be a few weeks before a release.

Sounds good! If you remember, just ping me whenever you make the release, so that I can remember to switch from using Documenter master to using Documenter 0.25.

src/Documenter.jl
mortenpi: Also `:$deploy_branch`?

DilumAluthge: Yep! #1310

fredrikekre:
The docs for how this interface should be used also needs to be updated.

mortenpi:
The fallbacks [for `deploy_folder` here](https://github.com/JuliaDocs/Documenter.jl/blob/51c2b29e0a5f58939327b2a65b9c03be5454aba7/src/deployconfig.jl#L58-L65) should also return `DeployDecision` -- currently `deploydocs` throws an error.

@fredrikekre, I guess the related docstring is the one you mean that should get updated?

fredrikekre:
Yes.

DilumAluthge:
https://github.com/JuliaDocs/Documenter.jl/pull/1315

PR #1309, opened 2020-05-03
Update guide.md

Kunz-David:
Added setting to prettyurls, that also chcecks for Github Actions ENV.

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -159,7 +159,8 @@ build/
        ```julia
        makedocs(...,
            format = Documenter.HTML(
-               prettyurls = get(ENV, "CI", nothing) == "true"
+               prettyurls = (get(ENV, "CI", nothing) == "true" ||
+                             get(ENV, "GITHUB_ACTIONS", nothing) == "true")
            )
        )
        ```


mortenpi:
I don't think this is necessary. GitHub Actions always sets `CI = true`: https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

Kunz-David:
Ahh, I see. I just assumed that when I'm using only GitHub Actions, that using only `CI = true` (which I thought was short for `TRAVIS_CI` or something similar) wouldn't do the trick.
I thought so because of this:
>Alternatively, if your goal is to eventually set up automatic documentation deployment
   with Travis CI (see [Hosting Documentation](@ref)), you can also use their environment
   variables to determine Documenter's behavior in `make.jl` on the fly:

So maybe this should be changed to say something like:
"Alternatively, if your goal is to eventually set up automatic documentation deployment with Travis CI **or GitHub Actions** (see [Hosting Documentation](@ref)), you can also use their environment variables to determine Documenter's behavior in `make.jl` on the fly:"

What do you think? @mortenpi

mortenpi:
That sounds reasonable to me. If you'd like to update the PR with that change, I'd be happy to merge it.

Kunz-David:
OK, nice :)
Here is the [PR](https://github.com/JuliaDocs/Documenter.jl/pull/1316).

PR #1310, opened 2020-05-04
Fix a typo in a debug statement; also add missing PR links to changelog

DilumAluthge:
Fixes a typo introduced in #1307 

Also adds missing PR links that we forgot to add in #1307


DilumAluthge:
cc: @mortenpi 
cc: @fredrikekre

mortenpi:
I took the liberty of adding the missing PR references here as well which I noticed only after merging.

PR #1315, opened 2020-05-05
Fix the `deploy_folder` fallbacks and the docstring for `deploy_folder`

DilumAluthge:


DilumAluthge:
@mortenpi @fredrikekre I think this should fix the `deploy_docs` stuff and the docstrings that Fredrik mentioned.

DilumAluthge:
I tried to add a docstring for `DeployDecision`. But this broke LinkCheck. When I removed the docstring for `DeployDecision`, then LinkCheck passed.

mortenpi:
I re-added the docstring, let's see what happens. I don't see anything in it that should cause linkcheck to fail. However, Travis and AppVeyor tests were both failing too, without the docstring.

DilumAluthge:
All right I think I should've fixed the test now.

DilumAluthge:
Okay, CI is passing now.

DilumAluthge:
@mortenpi @fredrikekre Anything else I should add to this PR?

PR #1316, opened 2020-05-11
Clarify prettyurl ENV condition in guide.md

Kunz-David:
Added clarification of using the prettyurl ENV["CI"] condition in guide.md.

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -153,7 +153,7 @@ build/
        ```
 
        Alternatively, if your goal is to eventually set up automatic documentation deployment
-       with Travis CI or GitHub Actions (see [Hosting Documentation](@ref)), you can also use their environment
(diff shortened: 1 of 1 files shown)


docs/src/man/guide.md
```diff
@@ -153,7 +153,7 @@ build/
        ```
 
        Alternatively, if your goal is to eventually set up automatic documentation deployment
-       with Travis CI (see [Hosting Documentation](@ref)), you can also use their environment
+       with Travis CI or GitHub Actions (see [Hosting Documentation](@ref)), you can also use their environment
```
fredrikekre, line 156:
```suggestion
       with e.g. Travis CI or GitHub Actions (see [Hosting Documentation](@ref)), you can also use their environment
```

mortenpi:
Thanks!

PR #1326, opened 2020-05-22
Document url for hosted previews

timholy:
I find myself checking https://discourse.julialang.org/t/ann-documenter-0-24/31381 every time I want to access a doc preview, so it seems time to deposit this information here.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -394,13 +394,21 @@ the generated html. The following entries are valid in the `versions` vector:
    (as in e.g. `"stable" => "v^"`).
 
 **`push_preview`** a boolean that specifies if preview documentation should be
-deployed from pull requests or not.
+deployed from pull requests or not. If your published documentation is hosted
+at `"https://USER.github.io/PACKAGE.jl/stable`, by default the preview will be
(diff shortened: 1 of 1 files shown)


src/Documenter.jl
```diff
@@ -394,14 +394,22 @@ the generated html. The following entries are valid in the `versions` vector:
...
 It defaults to the value of `branch`.
 
 **`repo_previews`** is the remote repository to which pull request previews are
 deployed. It defaults to the value of `repo`.
 
+!!! note
+    Pull requests made from forks will not have previews.
+    Hosting previews requires access to the deploy key.
+    Therefore, previews are available only for pull requests that were
+    submitted directly to the main repository.
```
fredrikekre, line 411: Should it be
```suggestion
    submitted directly from the main repository.
```
?

fredrikekre:
> I find myself checking https://discourse.julialang.org/t/ann-documenter-0-24/31381 every time I want to access a doc preview

Note that if you build with GitHub Actions Documenter will post a link to the preview on the pull request.

src/Documenter.jl
timholy: It's a little murky, but I agree this is better. :+1:

PR #1331, opened 2020-06-07
Fix linerange for empty svec

tkf:
I have a package that fails to generate documentation because `Documenter.Utilities.linerange` receives an empty `svec()` as the first argument.  I can upload this setup somewhere if it's unreasonable that the first argument is an empty `svec()`.  But I think it's nice to always handle the empty case cleanly.  So maybe this patch makes sense as-is?

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -490,6 +490,12 @@ end
 linerange(doc) = linerange(doc.text, doc.data[:linenumber])
 
 function linerange(text, from)
+    # text is assumed to be a Core.SimpleVector (svec) from the .text field of a Docs.DocStr object.
+    # Hence, we need to be careful when summing over an empty svec below.
+    #
+    # Also, the isodd logic _appears_ to be there to handle variable interpolation into docstrings. In that case,
+    # the .text field seems to become longer than just 1 element and every even element is the interpolated object,
+    # and only the odd ones actually contain the docstring text as a string.
     lines = sum(Int[isodd(n) ? newlines(s) : 0 for (n, s) in enumerate(text)])
     return lines > 0 ? (from:(from + lines + 1)) : (from:from)
 end
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -490,7 +490,7 @@ end
 linerange(doc) = linerange(doc.text, doc.data[:linenumber])
 
 function linerange(text, from)
-    lines = sum([isodd(n) ? newlines(s) : 0 for (n, s) in enumerate(text)])
+    lines = sum(Int[isodd(n) ? newlines(s) : 0 for (n, s) in enumerate(text)])
     return lines > 0 ? (from:(from + lines + 1)) : (from:from)
 end
 
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -129,6 +129,8 @@ end
         @test Documenter.Utilities.format_line(100:9999, formatting) == "100:9999"
     end
 
+    @test Documenter.Utilities.linerange(Core.svec(), 0) === 0:0
+
     # URL building
     filepath = string(first(methods(Documenter.Utilities.url)).file)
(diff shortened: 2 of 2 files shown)


src/Utilities/Utilities.jl
```diff
@@ -490,7 +490,7 @@ end
 linerange(doc) = linerange(doc.text, doc.data[:linenumber])
 
 function linerange(text, from)
-    lines = sum([isodd(n) ? newlines(s) : 0 for (n, s) in enumerate(text)])
+    lines = sum(Int[isodd(n) ? newlines(s) : 0 for (n, s) in enumerate(text)])
```
mortenpi, line 493:
```suggestion
    # text is assumed to be a Core.SimpleVector (svec) from the .text field of a Docs.DocStr object.
    # Hence, we need to be careful when summing over an empty svec below.
    #
    # Also, the isodd logic _appears_ to be there to handle variable interpolation into docstrings. In that case,
    # the .text field seems to become longer than just 1 element and every even element is the interpolated object,
    # and only the odd ones actually contain the docstring text as a string.
    lines = sum(Int[isodd(n) ? newlines(s) : 0 for (n, s) in enumerate(text)])
```

mortenpi:
Patch LGTM. I am curious though as to how you actually ran into this edge case?

tkf:
Hi, thanks for the review!

> how you actually ran into this edge case?

OK, so I managed to reproduce the bug in https://github.com/JuliaFolds/SplittablesBase.jl/pull/45. It turned out that I need a weird re-exporting -based setting https://github.com/JuliaFolds/SplittablesBase.jl/pull/44 _and_ Julia 1.5 or 1.6-DEV to get this "bug".

mortenpi:
Ok, that's interesting. On 1.4.2

```julia-repl
julia> Documenter.Utilities.linerange(Core.svec(), 0)
```

does not crash, but if you break out the `sum()`, then it does throw the error. I guess it's the compiler special-casing the zero-length `svec` somehow pre-1.5..?

In any case, it looks like you can run into empty `svec`s if you do the [re-exporting that you mentioned](https://github.com/JuliaFolds/SplittablesTesting.jl/blob/6ca30f2dd72c42f45a7c28b79d300acf1f8978f4/src/SplittablesTesting.jl#L5-L9). I think it's because `@doc(docstr, obj)` only copies over the Markdown object, but not the `.text` field.

One thing to note though: due to the way you attach the docstrings in SplittablesTesting, you actually get identical docstrings attached to separate objects in both `SplittablesTesting` and `SplittablesBase.Testing`, and so Documenter will complain about "missing docstrings in `SplittablesBase.Testing`". So if `SplittablesTesting` is supposed to export the API for the functions in `SplittablesBase.Testing`, it might be better to have the docstrings live in there.. or perhaps stored as strings in a separate `Dict` in `SplittablesBase.Testing` and only get put into the docsystem in `SplittablesTesting`.

tkf:
Thanks for the extra analysis/followup!

> On 1.4.2
> 
> ```
> julia> Documenter.Utilities.linerange(Core.svec(), 0)
> ```
> 
> does not crash

That's interesting. An MWE seems to be

```julia
julia> f() = count(_undef_, _undef2_::AbstractString)

julia> @code_typed f()
CodeInfo(
1 ─      Core.typeassert(Main._undef2_, Main.AbstractString)::AbstractString
│   %2 = π (Main._undef2_, AbstractString)
│   %3 = Main.count(Main._undef_, %2)::Any
└──      return %3
) => Any
```

which is in Julia 1.6-DEV.  In 1.4, I have:

```julia
julia> @code_typed f()
CodeInfo(
1 ─      Core.typeassert(Main._undef2_, Main.AbstractString)::AbstractString
│   %2 = π (Main._undef2_, AbstractString)
│   %3 = Main.count(Main._undef_, %2)::Int64
└──      return %3
) => Int64
```

> or perhaps stored as strings in a separate `Dict` in `SplittablesBase.Testing` and only get put into the docsystem in `SplittablesTesting`.

Yeah, it sounds like a better approach. I was a bit unsure about what I'm doing.

PR #1334, opened 2020-06-08
Set version to 0.25.0

mortenpi:
Time to get the enhancements and bugfixes that have accumulated on master out, open PRs can be postponed to 0.26.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Documenter"
 uuid = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-version = "0.24.7+0.25-DEV"
+version = "0.25.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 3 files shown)


mortenpi:
@fredrikekre We have test failures because we need to install DocumenterTools when testing. But it is not compatible (in the `[compat]` sense) with Documenter 0.25. Any way we could make it work?

One way is to push to DocumenterTools master and add Manifests here, but that's a bit ugly. Another way would be to tag a version of DocumenterTools that already is declared to be compatible with 0.25.. but I don't know if that is allowed in General?

fredrikekre:
> Another way would be to tag a version of DocumenterTools that already is declared to be compatible with 0.25.. but I don't know if that is allowed in General?

Yea you can do that.

fredrikekre:
If there are no changes needed in DocumenterTools for Documenter 0.25 we could also update the compat bounds in the registry.

mortenpi:
> If there are no changes needed in DocumenterTools for Documenter 0.25 we could also update the compat bounds in the registry.

We'd need to update the `Project.toml` there sooner or later though, so I feel that tagging a new release is actually the cleaner solution here.

mortenpi:
I'll go ahead and merge and tag this (finally) as soon as the tests pass. We can have the 0.25 release be a bit less conservative than usual about new features/enhancements, and rapidly tag 0.25.x releases as the open PRs get merged.

Issue #686, opened 2018-03-23
Switch over to LibGit2


mortenpi:
Since LibGit2 is available in core Julia, we should probably use it in `deploydocs` etc., and not depend on the `git` binary. Should make using Documenter on Windows a bit easier and also relevant for container-based builds (e.g. on GitLab, #661).

tpapp:
It seems that even `makedocs` uses `git`, eg with a repo spec like
```julia
    repo = "https://gitlab.com/tkpapp/GitlabJuliaDemo.jl/blob/{commit}{path}#{line}"
```

mortenpi:
Superseded by #2348. LibGit2 is basically deprecated, so we don't want to build on that.

PR #1335, opened 2020-06-09
Clarify that fatal error is fine

goretkin:
Could be worth suppressing `stdout`/`stderr` or not using the git CLI to begin with, but this is a small message that I would have appreciated.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -554,6 +554,7 @@ function git_push(
         try
             run(`git checkout -b $branch upstream/$branch`)
         catch e
+            @info "fatal error regarding $branch is expected on first deployment"
             @debug "checking out $branch failed with error: $e"
             @debug "creating a new local $branch branch."
             run(`git checkout --orphan $branch`)
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -554,9 +554,13 @@ function git_push(
         try
             run(`git checkout -b $branch upstream/$branch`)
         catch e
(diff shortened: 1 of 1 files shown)


src/Documenter.jl
```diff
@@ -554,6 +554,7 @@ function git_push(
         try
             run(`git checkout -b $branch upstream/$branch`)
         catch e
+            @info "fatal error regarding $branch is expected on first deployment"
```
mortenpi, line 557: I guess it could be more detailed while we're at it. E.g. something like this?

```suggestion
            @info """
            Checking out $branch failed, creating a new orphaned branch.
            This usually happens when deploying to a repository for the first time and
            the $branch branch does not exist yet. The fatal error above is expected output
            from Git in this situation.
            """
```

src/Documenter.jl
```diff
@@ -554,6 +554,7 @@ function git_push(
...
             @debug "checking out $branch failed with error: $e"
             @debug "creating a new local $branch branch."
```
mortenpi, line 559: I think we can remove this debug message now.

```suggestion
```

mortenpi:
> Could be worth suppressing stdout/stderr

That was my first thought as well, but actually it is probably better to print the Git output, since it may contain useful information when things are going wrong, especially on CI.

> not using the git CLI to begin with

That's, in principle, on the roadmap: #686

mortenpi:
LGTM: https://travis-ci.org/github/mortenpi/Example.jl/jobs/697878545#L236-L241

PR #1336, opened 2020-06-09
Added link to main repo site

Arkoniak:
This is a small change which makes site name in navbar clickable. You can see how it looks like in [https://arkoniak.github.io/PCloud.jl/dev/](https://arkoniak.github.io/PCloud.jl/dev/)

I am missing this feature in current implementation of `Documenter.jl` and as far as I remember there was a discussion regarding it some time ago on discourse.

PR in current state can be considered as proof of concept: if everything is ok, I can add necessary docstrings and may be tests (if I would be able to figure out how to test such a thing). If this implementation is not good, I am open to suggestion how it can be made better.


mortenpi:
I don't feel that it's ideal that the sitename/logo would link away from the documentation pages. I'd rather just have #1254, which will add an explicit repository link into the top bar.

Issue #1338, opened 2020-06-12
FR: compile documentation to LaTeX only


cserteGT3:
Would be nice if I could skip the pdf compilation part and just saving the `.tex` file.
(The reason for this, that I'm bad with docker and couldn't setup a gitlab CI job, that compiles to pdf. But I think I could setup one, that compiles to latex and another that compiles latex to pdf.)

PR #1339, opened 2020-06-12
add new platform to compile only to latex

cserteGT3:
As described in #1338 I would like to compile the documentation to latex only. Here's my implmentation for it.
Please let me know what you think about it.

### DIFF
--- a/docs/src/man/other-formats.md
+++ b/docs/src/man/other-formats.md
@@ -198,9 +198,9 @@ services:
 
 to your `.travis.yml` file.
 
-### Compiling to latex only
+### Compiling to LaTeX only
 
-There's a possibility to save only the `.tex` file and skip the pdf compilation.
+There's a possibility to save only the `.tex` file and skip the PDF compilation.
 For this purpose use the `platform="none"` keyword:
 
 ```
--- a/docs/src/man/other-formats.md
+++ b/docs/src/man/other-formats.md
@@ -197,3 +197,16 @@ services:
 ```
 
 to your `.travis.yml` file.
+
+### Compiling to latex only
+
+There's a possibility to save only the `.tex` file and skip the pdf compilation.
+For this purpose use the `platform="none"` keyword:
+
+```
+using DocumenterLaTeX
+makedocs(
+    format = LaTeX(platform = "none"),
+    ...
+)
+```
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -35,13 +35,14 @@ The `makedocs` argument `authors` should also be specified, it will be used for 
 
 # Keyword arguments
 
-**`platform`** sets the platform where the tex-file is compiled, either `"native"` (default) or `"docker"`.
+**`platform`** sets the platform where the tex-file is compiled, either `"native"` (default), `"docker"`,
+or "none" which doesn't compile the tex.
 See [Other Output Formats](@ref) for more information.
 """
 struct LaTeX <: Documenter.Writer
     platform::String
     function LaTeX(; platform = "native")
-        platform ∈ ("native", "docker") || throw(ArgumentError("unknown platform: $platform"))
+        platform ∈ ("native", "docker", "none") || throw(ArgumentError("unknown platform: $platform"))
         return new(platform)
     end
 end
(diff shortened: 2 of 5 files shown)


src/Writers/LaTeXWriter.jl
```diff
@@ -35,13 +35,14 @@ The `makedocs` argument `authors` should also be specified, it will be used for
 
 # Keyword arguments
 
-**`platform`** sets the platform where the tex-file is compiled, either `"native"` (default) or `"docker"`.
+**`platform`** sets the platform where the tex-file is compiled, either `"native"` (default), `"docker"`,
+or "none" which doesn't compile the tex.
```
mortenpi, line 39:
```suggestion
or "none" which doesn't compile the tex.

```

docs/src/man/other-formats.md
```diff
@@ -197,3 +197,16 @@ services:
 ```
 
 to your `.travis.yml` file.
+
+### Compiling to latex only
```
mortenpi, line 201:
```suggestion
### Compiling to LaTeX only
```

docs/src/man/other-formats.md
```diff
@@ -197,3 +197,16 @@ services:
...
+
+There's a possibility to save only the `.tex` file and skip the pdf compilation.
```
mortenpi, line 203:
```suggestion
There's a possibility to save only the `.tex` file and skip the PDF compilation.
```

src/Writers/LaTeXWriter.jl
```diff
@@ -94,7 +95,7 @@ function mktempdir(args...; kwargs...)
 end
 
 function render(doc::Documents.Document, settings::LaTeX=LaTeX())
-    @info "LaTeXWriter: rendering PDF."
+    @info "LaTeXWriter: rendering tex file."
```
mortenpi, line 98:
```suggestion
    @info "LaTeXWriter: creating the LaTeX file."
```

cserteGT3:
> However, just the `.tex` file is not enough to build the PDF afterwards. The temporary `build/` directory will also contain all the images, the `.sty` file etc. We should probably copy the whole thing over to `build/`?

Agree, I forgot about the other files. Made the changes, to copy the whole source directory to the build directory. Also corrected the documentation, as you requested.

inkydragon:
> We should probably copy the whole thing over to build/?

If I build HTML first and then build LaTeX without PDF, Will I get a mix of HTML and LaTeX file in `build/` dir?

Maybe copy the whole dir to `build/latex/` ?

cserteGT3:
> > We should probably copy the whole thing over to build/?
> 
> If I build HTML first and then build LaTeX without PDF, Will I get a mix of HTML and LaTeX file in `build/` dir?

[`makedocs`](https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.makedocs) cleans the build directory by default.

> Maybe copy the whole dir to `build/latex/` ?

and `makedocs` has a `build="build"` keyword to specify the build directory. I think that is sufficient.

mortenpi:
> `makedocs` cleans the build directory by default.

That is true. However, I believe currently `platform="none"` will not actually honor `clean = false` currently? The `force=true` in the `cp` will erase the original `build/`?

cserteGT3:
> > `makedocs` cleans the build directory by default.
> 
> That is true. However, I believe currently `platform="none"` will not actually honor `clean = false` currently? The `force=true` in the `cp` will erase the original `build/`?

Yes, I think so. I'll test it later today, and report back.

cserteGT3:
I've tested the followings (on Windows and WSL Linux):

1. `makedocs(..., format = Documenter.HTML(), clean=true)`
1. `makedocs(..., LaTeX(platform = "none"), clean=false)`

The first step is the usual Documenter build, and after the second command nothing is deleted, and the result is a mix of HTML and LaTeX files. Though, I'm not sure why this happens.

> cp(src::AbstractString, dst::AbstractString; force::Bool=false, follow_symlinks::Bool=false)
> 
>  Copy the file, link, or directory from src to dst. force=true will first remove an existing dst.
> 
>  If follow_symlinks=false, and src is a symbolic link, dst will be created as a symbolic link. If follow_symlinks=true and src is a symbolic link, dst will be a copy of the file or directory src refers to. Return dst.

mortenpi:
> and after the second command nothing is deleted, and the result is a mix of HTML and LaTeX files.

I can confirm that, although interestingly when I run `cp("foo", "bar"; force=true)` on some directories, the contents of `bar/` is lost. But Documenter seems to work the way we want it so :man_shrugging:

PR #1340, opened 2020-06-15
added simple ssh deploy config

ExpandingMan:
When using Documenter.jl for projects at my job, I have to use our private company github and very often wind up having to do just the sort of thing that I did here.  I think this would be useful to anyone who may not have the ability to, or not want to use either of travis or github actions for deploying documentation for any reason.

Obviously I have added no testing here at all, and it's not at all clear to me how we'd test this.  I was hoping you guys would give me some guidance on what it might take to get this merged, thanks.

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -282,7 +282,6 @@ See GitHub's manual for
 [Creating and using encrypted secrets](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets)
 for more information.
 
-
 ## `docs/Project.toml`
 
 The doc-build environment `docs/Project.toml` includes Documenter and other doc-build
@@ -339,6 +338,18 @@ or `git@`.
 
 See the [`deploydocs`](@ref) function documentation for more details.
 
+
+## Using Simple SSH
+If you do not have access to either Travis or Github for automated documentation updates,
+or you are working in a private git hosting service where using these is impractical, it
+may be desirable to push objects more simply via SSH.  For such cases, simply call, e.g.
+```julia
+deploydocs(
+    repo = "github.com/USER_NAME/PACKAGE_NAME.jl.git",
+    deploy_config = Documenter.SimpleSSHConfig("/path/to/ssh-key")
+)
+```
+If no path to the SSH key is provided, the `DOCUMENTER_KEY` environment variable will be used.
 
 
 ## `.gitignore`
@@ -457,4 +468,5 @@ Documenter.documenter_key
 Documenter.documenter_key_previews
 Documenter.Travis
 Documenter.GitHubActions
+Documenter.SimpleSSHConfig
 ```
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -484,6 +484,35 @@ function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder=nothi
 end
 
 
+"""
+    SimpleSSHConfig <: DeployConfig
+
+Implementation of deploy config for deploying directly via SSH.  This is particularly useful in cases
+where there is a private repository and it's impractical or impossible to set up github actions or
+travis for fully automated deployment.
+
+The constructor optionally accepts a key location, so that it is possible to use multiple distinct keys.
+If the key location is not passed or is an empty string, the `DOCUMENTER_KEY` environment variable will
+be used.
+"""
+struct SimpleSSHConfig <: DeployConfig
+    ssh_key_location::String
+end
+
+SimpleSSHConfig() = SimpleSSHConfig("")
+
+deploy_folder(::SimpleSSHConfig; repo, devbranch, push_preview, devurl, kwargs...) = devurl
+authentication_method(::SimpleSSHConfig) = SSH
+function documenter_key(cfg.::SimpleSSHConfig)
+    k = cfg.ssh_key_location
+    if isempty(k)
+        ENV["DOCUMENTER_KEY"]
+    else
+        Base64.base64encode(open(read, k))
+    end
+end
+
+
 ##################
 # Auto-detection #
 ##################


src/deployconfig.jl
```diff
@@ -484,6 +484,35 @@ function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder=nothi
 end
 
 
+"""
+    SimpleSSHConfig <: DeployConfig
+
+Implementation of deploy config for deploying directly via SSH.  This is particularly useful in cases
+where there is a private repository and it's impractical or impossible to set up github actions or
+travis for fully automated deployment.
```
mortenpi, line 492:
```suggestion
where there is a private repository and it's impractical or impossible to set up GitHub Actions or
Travis for fully automated deployment.
```

src/deployconfig.jl
```diff
@@ -484,6 +484,35 @@ function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder=nothi
...
+end
+
+SimpleSSHConfig() = SimpleSSHConfig("")
+
+deploy_folder(::SimpleSSHConfig; repo, devbranch, push_preview, devurl, kwargs...) = devurl
+authentication_method(::SimpleSSHConfig) = SSH
+function documenter_key(cfg::SimpleSSHConfig)
+    k = cfg.ssh_key_location
+    if isempty(k)
+        ENV["DOCUMENTER_KEY"]
```
mortenpi, line 509: I wonder if it should default to `~/.ssh/id_rsa`?

src/deployconfig.jl
```diff
@@ -484,6 +484,35 @@ function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder=nothi
...
 
 
+"""
+    SimpleSSHConfig <: DeployConfig
+
+Implementation of deploy config for deploying directly via SSH.  This is particularly useful in cases
+where there is a private repository and it's impractical or impossible to set up github actions or
+travis for fully automated deployment.
+
+The constructor optionally accepts a key location, so that it is possible to use multiple distinct keys.
```
mortenpi, line 494:
```suggestion
The constructor optionally accepts a path to the SSH key, so that it is possible to use multiple distinct keys.
```

docs/src/man/hosting.md
```diff
@@ -340,6 +339,18 @@ or `git@`.
 See the [`deploydocs`](@ref) function documentation for more details.
 
 
+## Using Simple SSH
```
mortenpi, line 342:
```suggestion
## Using Simple SSH

```

docs/src/man/hosting.md
```diff
@@ -340,6 +339,18 @@ or `git@`.
...
+If you do not have access to either Travis or Github for automated documentation updates,
+or you are working in a private git hosting service where using these is impractical, it
+may be desirable to push objects more simply via SSH.  For such cases, simply call, e.g.
```
mortenpi, line 345:
```suggestion
may be desirable to push objects more simply via SSH.  For such cases, simply call, e.g.

```

docs/src/man/hosting.md
```diff
@@ -340,6 +339,18 @@ or `git@`.
...
+```julia
+deploydocs(
+    repo = "github.com/USER_NAME/PACKAGE_NAME.jl.git",
+    deploy_config = Documenter.SimpleSSHConfig("/path/to/ssh-key")
+)
+```
+If no path to the SSH key is provided, the `DOCUMENTER_KEY` environment variable will be used.
```
mortenpi, line 352:
```suggestion

If no path to the SSH key is provided, the `DOCUMENTER_KEY` environment variable will be used.
```

src/deployconfig.jl
```diff
@@ -484,6 +484,35 @@ function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder=nothi
...
+If the key location is not passed or is an empty string, the `DOCUMENTER_KEY` environment variable will
+be used.
+"""
+struct SimpleSSHConfig <: DeployConfig
+    ssh_key_location::String
+end
+
+SimpleSSHConfig() = SimpleSSHConfig("")
+
+deploy_folder(::SimpleSSHConfig; repo, devbranch, push_preview, devurl, kwargs...) = devurl
```
mortenpi, line 504: I think this needs to return a `DeployDecision`?

src/deployconfig.jl
ExpandingMan: I suppose we can do that if all else fails.  Usually in environments where I have RSA keys I tend to wind up with a lot of them.  I will implement it as a final fallback though, change incoming...

src/deployconfig.jl
ExpandingMan: This worked for me, but let me look into it...

ExpandingMan:
Ok, obviously a bunch of things I need to do here.  Primarily, I need to carefully work through the github actions and travis `deploy_folder` methods to understand what's happening so that I can return an appropriate `DeployConfig`.  I think when I started this I based it off a hack I had been using which came from a slightly older version of Documenter maybe.

Also, I want to make sure that it is possible for users to put the deploy key *location* in an environment variable, rather than the key itself in the environment variable.  That seems like a much more common and perhaps even more secure way of selecting a key.

ExpandingMan:
Ok, I've gotten a chance to look over this.

## `deploy_folder` return
I took a more careful look at what this function does for travis and github actions, and I just can't think of a good analog for the simple SSH deployment.  I therefore left it as is: it just always deploys.  

## Where to find keys
I had from the beginning given the config a place to store a path to an SSH private key, part of the thinking behind this is that it potentially allows users to have multiple different keys for multiple different repos (not an inconceivable situation if one is using, e.g. both public github and a company git).  Most of the time however, I'd assume one would want to set this via the environment.  It is common to be set up to use SSH with a key in a file somewhere, however typically you would not have all of your SSH keys themselves in environment variables.  I therefore added the `DOCUMENTER_KEY_PATH` environment variable which allows users to set the location of their key without having to actually dump it to the environment.  I was going to also add this for Travis and Github actions, but those use cases look rather specialized, so I left it alone.  Finally, largely for the sake of consistency, I have it check `DOCUMENTER_KEY` as for the other deployments.  It's hard for me to imagine very many users wanting to use that in this case, but it's nice to be consistent.

## Auto detection
Before falling back to `nothing`, I check `DOCUMENTER_KEY_PATH` for `auto_detect_deploy_system`.

## Naming
`SimpleSSHConfig` is a horrible name, not least because the others aren't called `Config` at all.  Feel free to suggest changes to this, don't really have any better ideas.

Ok, so as far as I know that should be it.  Let me know what else you think needs to be done to get this merged, thanks.

mortenpi:
> I took a more careful look at what this function does for travis and github actions, and I just can't think of a good analog for the simple SSH deployment. I therefore left it as is: it just always deploys.

Not at all sure if this is a good idea, but could we prompt to the user (by default)? Since you'd expect this to be used from the command line, right?

> `SimpleSSHConfig` is a horrible name, not least because the others aren't called `Config` at all.

I think `SimpleSSH` or maybe `LocalSSH` would also work. But I don't actually have strong feelings about the name either way.

ExpandingMan:
I really don't like user prompts as it wreaks havoc with automation.  I suppose I could add an option to the `SimpleSSHConfig` struct that makes it possible to do that, how does that sounds?

mortenpi:
> I really don't like user prompts as it wreaks havoc with automation. I suppose I could add an option to the SimpleSSHConfig struct that makes it possible to do that, how does that sounds?

That's fair enough. The prompt might make sense if interactive/CLI was the primary use case, but not if it's automated. No need to implement an option for it.

docs/src/man/hosting.md
```diff
@@ -340,6 +339,21 @@ or `git@`.
...
+
+If you do not have access to either Travis or Github for automated documentation updates,
+or you are working in a private git hosting service where using these is impractical, it
+may be desirable to push objects more simply via SSH.  For such cases, simply call, e.g.
+
+```julia
+deploydocs(
+    repo = "github.com/USER_NAME/PACKAGE_NAME.jl.git",
+    deploy_config = Documenter.SimpleSSHConfig("/path/to/ssh-key")
+)
```
mortenpi, line 352: Should we maybe recommend guarding the `deploydocs` call here, e.g. with something like `"deploy" in ARGS && deploydocs(...)`? It would always deploy deploy when you're building docs locally at the moment?

ExpandingMan: Sounds good, commit incoming.

src/deployconfig.jl
```diff
@@ -492,6 +538,8 @@ function auto_detect_deploy_system()
         return Travis()
     elseif haskey(ENV, "GITHUB_REPOSITORY")
         return GitHubActions()
+    elseif haskey(ENV, "DOCUMENTER_KEY_PATH")
+        return SimpleSSHConfig()
```
mortenpi, line 542: I like this. It allows you to do a manual deployment without changing the `make.jl` script. I wonder if we should still default to `~/.ssh/id_rsa` if `isempty(DOCUMENTER_KEY_PATH)`? Then you can just run like this locally:

```
DOCUMENTER_KEY_PATH= julia --project=docs/ docs/make.jl
```

mortenpi:
There's no support for pushing tags locally at the moment though?

ExpandingMan:
Yeah, to be honest I'd really have to go digging through the code to figure out the appropriate way of pushing the tags, there's a lot involved in that that I'm totally unfamiliar with.  I suggest you merge this and then someone can add it later, if you want we can comment out the documentation for `SimpleSSHConfig` until a way of dealing with the tags are available.

This is just based on the ultra-simple setup I usually use for doing this in which I typically just point to `dev`.

docs/src/man/hosting.md
```diff
@@ -340,6 +339,25 @@ or `git@`.
...
+```julia
+"deploy" ∈ ARGS && deploydocs(
+    repo = "github.com/USER_NAME/PACKAGE_NAME.jl.git",
+    deploy_config = Documenter.SimpleSSHConfig("/path/to/ssh-key")
+)
+```
+
+Note that calling `deploydocs` with the `SimpleSSHConfig` will always deploy, so if you
+are worried about pushing to the remote Git repository you may want to guard against it
+by, as in this example, checking if the argument `"deploy"` was passed on the command line.
```
mortenpi, line 357: Small rephrasing suggestion:

```suggestion
Note that calling `deploydocs` with the `SimpleSSHConfig` will always deploy. So, to avoid
unwanted accidental deployments, it is good practice to guard the `deplodocs` call, e.g. by
checking whether `"deploy"` was passed as a command line argument to the `make.jl` script
(as is done in the example above).
```

src/deployconfig.jl
```diff
@@ -484,6 +484,52 @@ function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder=nothi
...
+
+Implementation of deploy config for deploying directly via SSH.  This is particularly useful in cases
+in which there is a private repository and it's impractical or impossible to set up GitHub Actions or
+Travis for fully automated deployment.
+
+The constructor optionally accepts a path to the SSH key, so that it is possible to use multiple distinct keys.
+If the key location is not passed or is an empty string, it will check for the `DOCUMENTER_KEY_PATH` environment
+variable for a path to an SSH key.  If `DOCUMENTER_KEY_PATH` is not set, it will check for `DOCUMENTER_KEY`
+as in the Travis and GitHub Actions configs.  Failing this, it will check for a key at `\$HOME/.ssh/id_rsa`.
+"""
```
mortenpi, line 498: Happy to leave the tagging as a future enhancement, but I am slightly concerned by someone just doing `git checkout vX.Y.Z` and running `make.jl`, expecting it to deploy the tag. So I think it would be good to have a warning for this in the docs:

```suggestion

!!! warning "Tagged versions"

    `SimpleSSHConfig` does not currently support pushing documentation for tagged versions, and
    will always deploy the documentation into development version directory in the remote
    repository.
"""
```

mortenpi:
If the doc updates look good to you, it's good to go I'd say. Any thoughts on [checking for `isempty(DOCUMENTER_KEY_PATH)`](https://github.com/JuliaDocs/Documenter.jl/pull/1340#discussion_r445238652)?

src/deployconfig.jl
ExpandingMan: Actually, in light of what we were discussing before with this always pushing, that seems rather dangerous.  In fact, now I'm even wondering if it's wise to do even this, could result in some users pushing when they don't intend to.

ExpandingMan:
Ok, thanks for your help with the docs.

See my comment above about whether it's even safe for us to default to `SimpleSSHConfig` under any circumstances, since it might result in some users pushing unexpectedly.

src/deployconfig.jl
mortenpi: Hmm. You are probably right. However, I feel that it is a nice workflow. Maybe one option would be to use a separate variable to trigger the automatic deployment, e.g. `DOCUMENTER_DEPLOY=local` and have it match `local` exactly?

ExpandingMan:
Had forgotten about this, but found myself needing it again.  Is there any interest in this any more?  Should I try re-basing it?  Do we have the feature some other way now?

mortenpi:
I still think it would be a good idea to have this.

ExpandingMan:
Ok... I tried rebasing, but I find manual merges confusing (I avoid having to do it like the plague, so it's pretty unfamiliar to me).  Can you sanity check the result?

fingolfin:
This has many conflicts now which would need to be resolved (via `rebase` or `merge`, doesn't matter). 

Last activity here was 2021. Is there still any interest in this feature by @ExpandingMan or anyone else?

PR #1345, opened 2020-06-18
Don't warn on 302 redirects

### LINKED ISSUE #1344: Valid DOI redirected links trigger warnings.
I'm seeing lots of doc warnings regarding redirected DOI links. For example

```
┌ Warning: linkcheck 'https://doi.org/10.1016/0169-8095%2894%2900090-Z' status: 302, redirects to https://linkinghub.elsevier.com/retrieve/pii/016980959400090Z.
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/PLD7m/src/DocChecks.jl:223
```

Both the [first](https://doi.org/10.1016/0169-8095%2894%2900090-Z) and [second](https://linkinghub.elsevier.com/retrieve/pii/016980959400090Z) links work. It doesn't seem like this should be a warning.

simonbyrne:
302s are meant to be temporary redirects, so we shouldn't warn. Fixes #1344.


mortenpi:
The CI failures are unrelated I believe: it's the usual intermittent failure we see on Travis. We should probably disable the linkcheck test that actually tries to connect to an external server on Travis, and maybe run it as part of the Linkcheck GitHub Actions workflow instead.

LGTM, I'm fine with treating 302 failures as successes.

simonbyrne:
bump?

mortenpi:
I just noticed that 307 is also ["Temporary Redirect (since HTTP/1.1)"](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection). I guess we should also disable the warning for 307 redirects?

Issue #1343, opened 2020-06-17
support computed external links?


ThomasBreuer:
As far as I understand, Documenter.jl supports internal links via `@ref`, and external links via the `[text](URL)` syntax, where `URL` must be an explicit address.
I am trying to use Documenter.jl for a module that cross-references various entries in the documentation of some other computer system.
The URLs in question may be subject to changes, for example depending on the version number of that system; hard-coded URLs are therefore not a good solution.
At the time when the documentation of the module gets built, these URLs can be computed by a Julia function that belongs to the module.
Would it be possible to extend Documenter.jl such that this kind of computed external links gets supported, or do you have a better idea for this functionality?

Currently I am using the following hack.
Write `[text](url-encode)` in the documentation, let Documenter.jl process it, and add a post-processing step that replaces each `url-encode` by the computed URL.
This works in the final HTML version of the documentation, and looks acceptable in the interactive Julia help. However, Documenter.jl issues warnings about `invalid local link`.

fredrikekre:
> Currently I am using the following hack.
Write [text](url-encode) in the documentation, let Documenter.jl process it, and add a post-processing step that replaces each url-encode by the computed URL.
This works in the final HTML version of the documentation, and looks acceptable in the interactive Julia help. However, Documenter.jl issues warnings about invalid local link.

If you just replace it in a preprocess step it should not complain.

ThomasBreuer:
Thanks for your reply.
The point is that the links occur in docstrings inside the Julia code files. Programmatically rewriting the code files whenever the documentation gets processed sounds risky.

c42f:
It would be great if Documenter allowed the user to customize the logic for `@ref` lookup.

For JuliaHub we now build and serve most package documentation, and allowing `@ref` to link documentation between packages would provide a very cohesive user experience.

XRef: https://github.com/JuliaDocs/Documenter.jl/issues/319

mortenpi:
Some sort of a hook into link generation is definitely something we could do, but I feel that this would need to be hashed out a bit more before it's actionable. Another question is whether we want to overload at-ref -- perhaps a dedicated at-link would make more sense?

[...]

PR #1351, opened 2020-06-24
@url

### LINKED ISSUE #688: at-ref to functions in other packages
From #425 it is my impression that references like 
```julia
module Foo
"See also [`Bar.fun`](@ref)"
somefun() = 1
end
```
are unsupported. Is this correct?

However I still think that this is the best way to refer to other packages in docstrings, and I am wondering if at least I could silence the error messages along the lines of
```
!! Unable to get the binding for '[`Bar.fun`](@ref)'. [src/index.md]
```

aminya:
Adds `@url` which allows cross-linking to other parts of the web (which includes other packages).

Defined in meta:
```julia
URLs = Dict("AcuteML" => "https://aminya.github.io/AcuteML.jl/dev/SyntaxReference/")
```

Later in the docs, I can use the following for referencing to its different parts:
```markdown
[AcuteML.@aml](@url AcuteML)
```
which builds the full URL for me.

This is a simple feature which can indirectly fix #688


mortenpi:
Sorry for not looking at this earlier. As I said in #1352, I don't think this conflicts with that one. This is actually kind of a workaround for the fact that the Julia Markdown parser doesn't support link reference definitions. A few thoughts:

1. Just bikeshedding: could also call it `@link`? As in, it doesn't necessarily have to be a full URL? E.g. maybe in the future this can be extended to support #1343, with user-defined functions generating the links.
2. I would be conservative with the URL keys, i.e. force them to match something like `[A-Za-z0-9-]`.
3. Relatedly, I think at the moment it would break if your key is e.g. `foo/bar` and then then you have `[](@url foo/bar/baz)`?
4. Currently, I believe if you write `[](@url foo)`, it appends a `/` to the reference URL? Probably best if it didn't, so that you could also use this feature to get the exact reference URL.

mortenpi:
Another thing that would be nice to have is to be able to pass a document-global list of these URLs to `makedocs`.

aminya:
> Sorry for not looking at this earlier. As I said in #1352, I don't think this conflicts with that one. This is actually kind of a workaround for the fact that the Julia Markdown parser doesn't support link reference definitions. A few thoughts:
> 
> 1. Just bikeshedding: could also call it `@link`? As in, it doesn't necessarily have to be a full URL? E.g. maybe in the future this can be extended to support #1343, with user-defined functions generating the links.

Sure. We can change it to `@link`. 

> 2. I would be conservative with the URL keys, i.e. force them to match something like `[A-Za-z0-9-]`.

:thinking: that is a good idea! We can use [IsURL.jl](https://github.com/zdroid/IsURL.jl/blob/master/src/IsURL.jl).

> 3. Relatedly, I think at the moment it would break if your key is e.g. `foo/bar` and then then you have `[](@url foo/bar/baz)`?
> 4. Currently, I believe if you write `[](@url foo)`, it appends a `/` to the reference URL? Probably best if it didn't, so that you could also use this feature to get the exact reference URL.

I wanted to update that. The script should extract the main part of the URL no matter what the user writes (with or without ending `/`). 


> Another thing that would be nice to have is to be able to pass a document-global list of these URLs to `makedocs`.

I think it is better to not use meta altogether. makedocs is a better place in my opinion.

mortenpi:
> > 1. I would be conservative with the URL keys, i.e. force them to match something like `[A-Za-z0-9-]`.
> 
>  that is a good idea! We can use [IsURL.jl](https://github.com/zdroid/IsURL.jl/blob/master/src/IsURL.jl).

Hmm, I am not sure we mean the same thing. I meant the keys of the `BaseURLs` dictionary that you use to refer to the actual URLs.

FWIW though, we have [`isabsurl` in Documenter](https://github.com/JuliaDocs/Documenter.jl/blob/f9f0b68ad11b68039b424557bf73511f55fed02f/src/Utilities/Utilities.jl#L607-L613), so there would be no need for an extra dependency (we want to avoid unnecessary dependencies in Documenter).

> I think it is better to not use meta altogether. makedocs is a better place in my opinion.

Since you already implemented it, I would keep it. I think it's good that you can defined the URLs close to where you use them if you indeed only use it on one page. It would also be consistent with e.g. doctest filtering (which you can do either per-doctest, per-page or globally).

aminya:
I'll take a look at this in one or two weeks. I've been quite busy lately.

fredrikekre:
I think it would be much better with a general placeholder mechanism. No point in keeping this just for URL placeholders in references IMO.

goerz:
It's a little bit of a hack, but [`DocumenterInterLinks`](https://github.com/JuliaDocs/DocumenterInterLinks.jl) (still experimental) could be used for something like this: [Can I use this plugin for general external links?](http://juliadocs.org/DocumenterInterLinks.jl/stable/howtos/#howto-external-links)

fingolfin:
@aminya do you have any plans to revive this PR?

aminya:
Sorry, I don't use Julia anymore.

PR #1353, opened 2020-06-29
Add COSMO.jl to examples

migarstka:
It would be great if you could add our documentation to the list of examples :)


mortenpi:
Sure thing, thanks for the contribution!

PR #1355, opened 2020-07-02
If no pages specified, put index.md in front of other pages

mortenpi:
If `pages` is not passed to `makedocs`, we determine the page list automatically (search for `*.md` and sort). This PR makes sure that `index.md` pages end up before other pages in this situation.


src/Builder.jl
```diff
@@ -159,6 +159,19 @@ function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
...
+"""
+    lt_page(a::AbstractString, b::AbstractString)
+
+Checks if the page path `a` should come before `b` in a sorted list. Falls back to standard
+string sorting, except for prioritizing `index.md` (i.e. `index.md` always comes first).
+"""
+function lt_page(a, b)
+    # note: length("index.md") == 8
+    a = endswith(a, "index.md") ? a[1:end-8] : a
+    b = endswith(b, "index.md") ? b[1:end-8] : b
```
MichaelHatherly, line 171: Will probably get string indexing errors if `"index.md"` is preceded by a wide Unicode char?

Maybe use `chop(a; tail=8)` or similar instead?

mortenpi: Ooh, thanks for catching that! I forgot that indexing wasn't Unicode-aware. `chop(; tail=8)` looks indeed to be the correct way to do this.

mortenpi: Actually, I think _technically_ `a[1:end-8]` would work here, since we are always just chopping of `index.md`, which we know is 8 code units at the end. Nevertheless, I think `chop` is the more correct way of doing this, so change to that.

PR #1357, opened 2020-07-05
Add specific HTML class to example output

### LINKED ISSUE #1026: Highlight output code-blocks differently
Output code blocks are not very easy to differentiate from input blocks, the only difference is that input blocks usually are syntag highlighted though. See e.g. https://fredrikekre.github.io/Literate.jl/stable/generated/example/#Basic-syntax-1

tkf:
This PR tweaks DOM of the `@example` output so that it has a different class than the code blocks.  We can then apply a different CSS style to visually distinguish the input and the output.

Before:

> image image

After:

> image image

Fix #1026.

_Edit by @mortenpi: added issue reference._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * ![Enhancement][badge-enhancement] When automatically determining the page list (i.e. `pages` is not passed to `makedocs`), Documenter now lists `index.md` before other pages. ([#1355][github-1355])
 
+* ![Enhancement][badge-enhancement] The output text boxes of `@example` blocks are now style differently from the code blocks, to make it easier to visually distinguish between the input and output. ([#1026][github-1026], [#1357][github-1357])
+
 ## Version `v0.25.0`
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -593,6 +596,8 @@
 [github-1344]: https://github.com/JuliaDocs/Documenter.jl/issues/1344
 [github-1345]: https://github.com/JuliaDocs/Documenter.jl/pull/1345
 [github-1355]: https://github.com/JuliaDocs/Documenter.jl/pull/1355
+[github-1357]: https://github.com/JuliaDocs/Documenter.jl/pull/1357
+
```
mortenpi, line 600:
```suggestion
```

tkf:
Thanks for the review and also adding the changelog!

KristofferC:
Perhaps worth thinking about if we want to update the defaults to use this?

tkf:
Actually, the screenshots in the OP are with the default CSS.

PR #1360, opened 2020-07-09
Use consistent HTML classs for stdout/err in example

tkf:
This is a follow up of #1357.  While #1357 handled evaluated output, it didn't handle the HTML class of stdout/stderr.  This PR fixes it.


tkf:
Great! Thanks for the review and the clean-ups.

It looks like the CIs are happy with it now.

PR #1365, opened 2020-07-14
Add "Powered by ..." message to footer

pfitzseb:
image

This should help a bit with SEO and therefore Julia's ranking (both for Julia itself and packages). AFAIU it's not ideal to put this in the footer, but I don't know enough about how Google ranks websites to figure out something better.

Open questions:
- [ ] Is everyone happy with this?
- [x] Should there be an opt-out mechanism?
- [ ] Should this be an opt-in feature (would completely defeat the purpose though, I think)?
- [x] Is the styling ok?

cc @ViralBShah 

_Edit by @mortenpi: fixes #1184_

### DIFF
--- a/assets/html/scss/documenter/layout/_main.scss
+++ b/assets/html/scss/documenter/layout/_main.scss
@@ -135,6 +135,7 @@
       .footer-message {
         font-size: 0.7em;
         margin: 0.5em auto 0 auto;
+        text-align: center;
       }
     }
   }
--- a/assets/html/scss/documenter/layout/_main.scss
+++ b/assets/html/scss/documenter/layout/_main.scss
@@ -133,7 +133,7 @@
       }
 
       .footer-message {
-        font-size: 0.9em;
+        font-size: 0.7em;
         margin: 0.5em auto 0 auto;
       }
     }
--- a/assets/html/scss/documenter/layout/_main.scss
+++ b/assets/html/scss/documenter/layout/_main.scss
@@ -133,7 +133,7 @@
       }
 
       .footer-message {
-        font-size: 0.7em;
+        font-size: 0.8em;
         margin: 0.5em auto 0 auto;
         text-align: center;
       }
--- a/assets/html/scss/documenter/layout/_main.scss
+++ b/assets/html/scss/documenter/layout/_main.scss
@@ -108,6 +108,7 @@
 
     .docs-footer {
       display: flex;
+      flex-wrap: wrap;
       margin-left: 0;
       margin-right: 0;
       border-top: 1px solid $border;
@@ -124,6 +125,16 @@
       }
       .docs-footer-nextpage {
         text-align: right;
+      }
+
+      .flexbox-break {
+        flex-basis: 100%;
+        height: 0;
+      }
+
+      .footer-message {
+        font-size: 0.9em;
+        margin: 0.5em auto 0 auto;
       }
     }
   }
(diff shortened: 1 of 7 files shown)


KristofferC:
I think it looks nice. Adding "Julia" to the page also makes it more likely to be found when googling something related to the package + "julia". Right now, there are many pages without the term "julia" at all which might lead to worse discoverability.

johnnychen94:
Literate.jl also adds this to the footer, so an opt-out mechanism would be a must from my perspective; otherwise, users could get two credit footnote by default... See the `credit` keyword in the [Literate docs](https://fredrikekre.github.io/Literate.jl/v2/outputformats/#Literate.DEFAULT_CONFIGURATION).

ViralBShah:
I think we should have this as the default message but easy to change.

assets/html/scss/documenter/layout/_main.scss
```diff
@@ -125,6 +126,16 @@
...
         text-align: right;
       }
+
+      .flexbox-break {
+        flex-basis: 100%;
+        height: 0;
+      }
+
+      .footer-message {
+        font-size: 0.9em;
```
mortenpi, line 136: I would make the font just a tiny bit smaller, so it would be slightly less intrusive visually:

```suggestion
        font-size: 0.7em;
```

footer

assets/html/scss/documenter/layout/_main.scss
```diff
@@ -125,6 +126,16 @@
...
+        margin: 0.5em auto 0 auto;
```
mortenpi, line 137: Currently, if the footer goes over multiple lines it is left-aligned (e.g. on mobile). I'd make it centered:

```suggestion
        margin: 0.5em auto 0 auto;
        text-align: center;
```

pfitzseb:
Alright, added  a `footer` argument to HTML that defaults to `"Powered by [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) and the [Julia Programming Language](https://julialang.org/)."`. Can be set to the empty string to hide the footer completely (could also allow `nothing` here, I suppose).

The argument checking for that argument is very rudimentary at the moment, so in theory you could set a header or whatever. I can add an early check for that if necessary.

mortenpi:
> The argument checking for `footer` is very rudimentary at the moment, so in theory you could set a header or whatever. I can add an early check for that if necessary.

What might work is to see if it parses to a single `Markdown.Paragraph`?

pfitzseb:
Done :)

test/htmlwriter.jl
```diff
@@ -52,6 +52,9 @@ end
     @test Documenter.HTML() isa Documenter.HTML
     @test_throws ArgumentError Documenter.HTML(collapselevel=-200)
     @test_throws Exception Documenter.HTML(assets=["foo.js", 10])
+    @test_throws ArgumentError Documenter.HTML(footer="foo\n\nbar")
+    @test_throws ArgumentError Documenter.HTML(footer="# foo")
```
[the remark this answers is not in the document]
pfitzseb: Ah, that's neat :) Good catch.

pfitzseb:
Alright, pushed an update to the docs.

johnnychen94:
Customized footer affects globally. Using this with Literate.jl would either 1) inserting two credit footers for Literate-generated pages, or 2) add credit for Literate even in other pages.

mortenpi:
LGTM, thanks again @pfitzseb!

> Customized footer affects globally. Using this with Literate.jl would either 1) inserting two credit footers for Literate-generated pages, or 2) add credit for Literate even in other pages that are not Literate-generated.

Yeah, it might be nice to be able to override it also on a per-page basis: #1370.

PR #1367, opened 2020-07-15
MathJax3 support

### LINKED ISSUE #1362: Add MathJax3 Writer
It would be good to support MathJax 3 for LaTeX rendering as well. My main motivation behind this is that MathJax 3 officially supports the [physics package](http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/physics/physics.pdf) as a plugin. From quick testing, it also seems to render a lot faster than MathJax 2 and really not much slower than KaTeX. It's probably better to add alongside MathJax 2, since the config formats are not compatible, so this would otherwise be breaking. I'd like to take a stab at this, if that's ok.

simeonschaub:
The options for MathJax3 are currently ignored, but the default rendering already works.
Should close #1362


src/Writers/HTMLWriter.jl
```diff
@@ -464,6 +505,17 @@ module RD
...
         ))
     end
+    function mathengine!(r::RequireJS, engine::MathJax3)
+        push!(r, RemoteLibrary(
+            "mathjax3",
+            "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-svg.js",
+        ))
+        push!(r, Snippet(["mathjax3"], ["MathJax"],
+            """
+            window.MathJax = $(json_jsescape(engine.config, 2))
```
simeonschaub, line 515: This doesn't work. I would appreciate any hints from people more familiar with Node and `require` on how to correctly configure this.

mortenpi: I think that's expected, since RequireJS loads the library before `require` calls the configuration callback. It looks like they've done away with the `Config()` function, so I couldn't really quickly figure out what would be the correct way of doing it is. But this might be relevant: https://docs.mathjax.org/en/latest/web/configuration.html#configuring-mathjax-after-it-is-loaded

simeonschaub: Is there a reason, we require the callback to be called after the library is loaded? Couldn't we just configure mathjax before it is loaded?

mortenpi: Well, that's just how RequireJS works (afaiu). I don't know if it possible to somehow run some code before RequireJS fetches and runs the module. Maybe it's possible to declare a custom dependency module somehow that RequireJS would have to load before MathJax..?

An alternative would be to drop RequireJS altogether for this, but currently the whole `documenter.js` generation is built around RequireJS..

simeonschaub:
Everything should work now. Do you think that's the right approach?

simeonschaub:
I don't really know how the tests here work, so please tell me what still needs to be added in that department. I was thinking that maybe we should already rename `MathJax` to `MathJax2` and make `MathJax` just an alias, so we can make v3 the default eventually. Or do you think it would be better to do this at a later time?

mortenpi:
> I don't really know how the tests here work, so please tell me what still needs to be added in that department.

The front end is not really being systematically tested. What would be nice, though, is to make sure that one of the builds in `test/examples/make.jl` would actually use `MathJax3`. This should make sure that this code path does not throw an error at least.

> I was thinking that maybe we should already rename MathJax to MathJax2 and make MathJax just an alias, so we can make v3 the default eventually. Or do you think it would be better to do this at a later time?

No, this would be good to do, if you don't mind. `MathJax` should stay as an alias, yea, for now. Preferably one that prints a deprecation warning. I think it could just be a `function` prints the deprecation and then calls `MathJax2(args...; kwargs...)`.

src/Writers/HTMLWriter.jl
```diff
@@ -464,6 +500,20 @@ module RD
...
         ))
     end
+    function mathengine!(r::RequireJS, engine::MathJax3)
+        push!(r, Snippet([], [],
+            """
+            window.MathJax = $(json_jsescape(engine.config, 2));
+
+            (function () {
+                var script = document.createElement('script');
+                script.src = 'https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-svg.js';
```
mortenpi, line 510: I would prefer if we'd hard-code the exact version here, to avoid any potential breakage.

I would also prefer if we'd CDN this [from cdnjs](https://cdnjs.com/libraries/mathjax). We use that for the other assets (except for fonts) and it's good to minimize the number of potential points of failure.

test/examples/make.jl
```diff
@@ -187,13 +187,22 @@ examples_html_doc = if "html" in EXAMPLE_BUILDS
...
-                    :Macros => Dict(
-                        :ket => ["|#1\\rangle", 1],
-                        :bra => ["\\langle#1|", 1],
+                #mathengine = MathJax(Dict(:TeX => Dict(
+                #    :equationNumbers => Dict(:autoNumber => "AMS"),
+                #    :Macros => Dict(
+                #        :ket => ["|#1\\rangle", 1],
+                #        :bra => ["\\langle#1|", 1],
+                #    ),
+                #))),
```
mortenpi, line 196:
```suggestion
```

test/examples/src/man/tutorial.md
```diff
@@ -367,6 +367,7 @@ mathengine = Documenter.MathJax(Dict(:TeX => Dict(
 
 ```math
 \bra{x}\ket{y}
+\pdv[n]{f}{x}
```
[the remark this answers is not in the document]
simeonschaub: This is just to double check that the physics package was actually working.

src/Writers/HTMLWriter.jl
```diff
@@ -464,6 +500,20 @@ module RD
...
+                script.async = true;
+                document.head.appendChild(script);
+            })();
```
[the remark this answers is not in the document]
simeonschaub: The problem is that MathJax 3 needs to be configured before it is actually loaded and there doesn't seem to be a straightforward way to change the config after loading. (See also [here](https://docs.mathjax.org/en/latest/web/configuration.html#configuring-mathjax-after-it-is-loaded)) This snippet is actually from https://docs.mathjax.org/en/latest/web/configuration.html#configuring-and-loading-in-one-script. I looked into whether a `RemoteLibrary` could also depend on a `Snippet`, but it seems like this is not supported by RequireJS, so I think this is the best way to do this.

simeonschaub:
I have no clue what's going on with the test failures...

simeonschaub:
:tada: Well, that took a while too figure out. Still don't really understand what was going on her, but it works now.

simeonschaub:
Did you have time to take another look at this?

src/Writers/HTMLWriter.jl
```diff
@@ -188,29 +188,68 @@ setting your own `tex2jax` value will override the default). This can be overrid
...
+                "TeX/autoload-all.js"
+            ],
+            :TeX => Dict(:equationNumbers => Dict(:autoNumber => "AMS"))
+        )
+        new((config === nothing) ? default : override ? config : merge(default, config))
+    end
+end
+
+@deprecate MathJax(config::Union{Dict,Nothing} = nothing, override=false) MathJax2(config, override) false
+@doc "deprecated – Use MathJax2 instead" MathJax
```
mortenpi, line 220:
```suggestion
@doc "deprecated – Use [`MathJax2`](@ref) instead" MathJax
```

src/Writers/HTMLWriter.jl
```diff
@@ -188,29 +188,68 @@ setting your own `tex2jax` value will override the default). This can be overrid
...
+    MathJax3(config::Dict = <default>, override = false)
+
+An instance of the `MathJax3` type can be passed to [`HTML`](@ref) via the `mathengine`
+keyword to specify that the [MathJax v3 rendering engine](https://www.mathjax.org/) should be
+used in the HTML output to render mathematical expressions.
+
+A dictionary can be passed via the `config` argument to configure MathJax. It gets passed to
+[`Window.MathJax`](https://docs.mathjax.org/en/latest/options/) function. By default,
+Documenter sets a custom configuration only for `tex2jax` and the `base`, `ams` and `autoload`
+packages are imported.
```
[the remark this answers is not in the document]
simeonschaub: Yes, I think, I can formulate that a bit better.

simeonschaub:
Sorry for the delay! I went into a bit more detail about which options are set by default in the docstring. This should now be good to go from my side.

mortenpi:
LGTM, thanks a lot @simeonschaub! I will aim to tag a new release with this in the next few days.

PR #1369, opened 2020-07-16
Doctest output starting with #

mortenpi:
The following doctest will fail

```jldoctest
julia> f(x) = println("# output from f\n$x");

julia> f(42)
# output from f
42
```

because we consider lines starting `#` as comments and include in the source part of the doctest, kinda: https://github.com/JuliaDocs/Documenter.jl/blob/3e2090abf8064c55ac8828c060513de7f8e6959f/src/DocTests.jl#L454-L456

Arguably, we should only consider the `#` as part of the doctest source if it is preceded by 7 spaces (our standard heuristic for determining if something is part of the source or output part):

```jldoctest
julia> f(x) = println("# output from f\n$x");

julia> f(42)
       # actual comment
# output from f
42
```

X-ref: JuliaLang/julia#36696


fredrikekre:
Nice!

Issue #1332, opened 2020-06-07
Add a "Star on GitHub" button along with Edit on GitHub in the docs


logankilpatrick:
Hey all, @tlienart had the great suggestion of adding a "Star on GitHub" button to drive more stars (a typical metric to assess the use of a package across OSS) to our Julia packages. The example is given (which is very telling IMO) is that Documenter only has 294 stars but is very widely used across the ecosystem.

guilhermebodin:
I would be happy to help in this one. 

Should it be something in the direction of https://joshday.github.io/OnlineStats.jl/latest/ ?

logankilpatrick:
Yes, maybe not that blatant but it's the exact idea.

mortenpi:
I'd say it should probably go into either the header (with the "Edit" links etc) or the sidebar. [MkDocs' Material theme](https://squidfunk.github.io/mkdocs-material/) is a good example in my opinion.

guilhermebodin:
To make it like this I think we need to make a request as in https://github.com/squidfunk/mkdocs-material/blob/4848c7dd41996db053c93c108f5a5be589d0cda3/src/assets/javascripts/patches/source/github/index.ts

Is this possible on Documenter.jl?

mortenpi:
Yep, you can include arbitrary JS in Documenter sites. Specifically, this loop

https://github.com/JuliaDocs/Documenter.jl/blob/567f200319a842db162a4c99b8d9d2b795d469e4/src/Writers/HTMLWriter.jl#L577-L581

gathers up [the JS assets](https://github.com/JuliaDocs/Documenter.jl/tree/master/assets/html/js) and outputs them into the `documenter.js` file, which then gets loaded with RequireJS.

guilhermebodin:
Nice, I got this script working
```
// libraries: jquery
// arguments: $

function round(value) {
  if (value > 999) {
    const digits = +((value - 950) % 1000 > 99)
    return `${((value + 1) / 1000).toFixed(digits)}k`
  } else {
    return value.toString()
  }
}

$(document).ready(function(){
  $.getJSON( "https://api.github.com/repos/JuliaDocs/Documenter.jl", function( data ) {
      console.log(`${round(data.stargazers_count || 0)} Stars`)
      console.log(`${round(data.forks_count || 0)} Forks`)
  });
});
```

Now I am not so sure if this is a good solution because everytime the document is ready it tries to get the json. 
[...]

PR #1377, opened 2020-07-28
Add Go to repo button

guilhermebodin:
This is my first iteration of something that minimally works for #1332 . It still looks ugly because there is no css.

Captura de Tela 2020-07-27 às 23 43 07

I still have to:

- [ ] Add the link that points to repository. 
- [ ] Add the repository owner and repository name
- [ ] Write proper HTML to make a proper layout
- [ ] Apply some css

Potentially missing things
- [ ] Add a logo, just like they have on https://squidfunk.github.io/mkdocs-material/

### DIFF
--- a/assets/html/js/repoinfo.js
+++ b/assets/html/js/repoinfo.js
@@ -1,0 +1,21 @@
+// libraries: jquery
+// arguments: $
+
+function round(value) {
+  if (value > 999) {
+    const digits = +((value - 950) % 1000 > 99)
+    return `${((value + 1) / 1000).toFixed(digits)}k`
+  } else {
+    return value.toString()
+  }
+}
+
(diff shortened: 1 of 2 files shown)


mortenpi:
Thanks for the PR! I can't take a look at the code right now, but as an FYI, for

> Add the link that points to repository.

there is already work in #1254. You're right though, that these two are closely related changes. Most of the work in #1254 was figuring out a way to pass more general remote repository information through without breaking existing setups. For that reason, I would suggest keeping this PR just for the star widget, and perhaps even work on top of that PR.

cossio:
How about a simpler thing: Just a Github button (with the mascot logo) pointing to the repo?

I mean, one could add things like repository owner and repository name later on.

PR #1386, opened 2020-08-03
Simplify some SCSS styles

zdroid:


assets/html/scss/documenter/layout/_search.scss
```diff
@@ -17,8 +17,6 @@
       //   @include font-awesome;
       //   content: "";
       // }
```
mortenpi, line 19: While we're at it, could we also remove this commented-out portion?

zdroid: Yeah, those two confused me too. There are some more above in `#documenter-search-results`.

mortenpi:
LGTM, thanks a lot for these cleanups @zdroid!

PR #1389, opened 2020-08-07
Work around 1.5 & nightly test failures

### LINKED ISSUE #1337: Doctest fixing broken on Julia 1.5/nightly
```jldoctest
julia> 1 + 2

julia> 3 + 4
```

becomes

```jldoctest
julia> 1 + 2
3
3
3j3u3l3i3a3>3 333 3+3 343
```

Julia 1.4 and below are fine. This is causing the nightly failures.

mortenpi:
Fix #1337

### DIFF
--- a/src/DocTests.jl
+++ b/src/DocTests.jl
@@ -422,7 +422,14 @@ function fix_doctest(result::Result, str, doc::Documents.Document)
     newcode = code[1:last(inputidx)]
     isempty(result.output) && (newcode *= '\n') # issue #772
     # second part: the rest, with the old output replaced with the new one
-    newcode *= replace(code[nextind(code, last(inputidx)):end], result.output => str, count = 1)
+    if result.output == ""
+        # This works around a regression in Julia 1.5.0 (https://github.com/JuliaLang/julia/issues/36953)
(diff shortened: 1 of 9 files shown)


src/DocTests.jl
```diff
@@ -422,7 +422,14 @@ function fix_doctest(result::Result, str, doc::Documents.Document)
...
     # second part: the rest, with the old output replaced with the new one
-    newcode *= replace(code[nextind(code, last(inputidx)):end], result.output => str, count = 1)
+    if result.output == ""
+        # This works around a regression in Julia 1.5.0 (https://github.com/JuliaLang/julia/issues/36953)
+        # Technically, it is only necessary if VERSION >= v"1.5.0-DEV.826"
+        newcode *= str
+        newcode *= code[nextind(code, last(inputidx)):end]
+    else
+        newcode *= replace(code[nextind(code, last(inputidx)):end], result.output => str, count = 1)
+    end
```
mortenpi, line 432: Instead of disabling the tests, I figured we could just add a workaround here to make sure that Documenter actually works properly on 1.5.0. `result.output == ""` is simple enough to short-circuit.

PR #1392, opened 2020-08-08
Correct render code block with tag "```julia-repl XXX"

likanzhan:
To correct render the code block with the following formatted tag:

````
```Julia-repl XXXX

```
````

such as [```julia-repl sayhello2](https://github.com/JuliaLang/julia/edit/master/doc/src/manual/metaprogramming.md#L521) in the `metaprogramming` manual.

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -441,7 +441,7 @@ const LEXER = Set([
 function latex(io::IO, code::Markdown.Code)
     language = isempty(code.language) ? "none" : code.language
     # the julia-repl is called "jlcon" in Pygments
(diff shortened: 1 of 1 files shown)


mortenpi:
Interestingly, we accidentally handle this case in the HTML output, because [e.g. `code[".language-julia-repl XXXX"]`](https://github.com/JuliaDocs/Documenter.jl/blob/a9c3ab7d4cd51592f4dd34c8f05744462d75c5b9/src/Writers/HTMLWriter.jl#L1599) drops everything after the space.

LGTM. Allowing for non-language information after the language name is also hinted at [in the CommonMark spec](https://spec.commonmark.org/0.29/#example-111).

PR #1398, opened 2020-08-12
Fix the nightly tests

mortenpi:
Nightly tests for docstring fixing are failing because of the `Array{1/2}` -> `Vector/Matrix` printing change in 1.6. This works around the version-dependence by reshaping the arrays into a higher-dimensional array, which still prints the same way.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,7 @@
 environment:
   matrix:
   - julia_version: 1.0
-  - julia_version: 1.4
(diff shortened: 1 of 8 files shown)


mortenpi:
The HTMLWriter Markdown node conversion tests still fail, but that appears to be due to JuliaDocs/DocStringExtensions.jl#93 and should be fixed by JuliaDocs/DocStringExtensions.jl#96.

Issue #1397, opened 2020-08-11
Add code coverage from deployed docs?


briochemc:
I use GitHub actions to deploy documentation (using Literate) separately from `test/runtests.jl`. The code bits are run and the output shows nicely in the docs and the notebooks, which is great (thanks Documenter × Literate!). But I would also like to get code coverage data from the code in the docs (using codecov). Is this possible? If yes, how?

I was assuming this required adding these last three lines (see below) to my `docs.yml` file (the top part comes from [Documenter's GitHub-Actions docs](https://juliadocs.github.io/Documenter.jl/dev/man/hosting/#GitHub-Actions)):

```yaml
name: Documentation

on:
  push:
    branches:
      - master
    tags: '*'
  pull_request:

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: julia-actions/setup-julia@latest
        with:
          version: '1.4'
      - name: Install dependencies
        run: julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
      - name: Build and deploy
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
        run: julia --project=docs/ docs/make.jl
      # I added these 3 lines below
      - uses: julia-actions/julia-uploadcodecov@latest
        env:
          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```

but I don't think it works. Maybe there's something else to do? Or maybe something entirely different? Or maybe it can't be  done at all? Any pointers appreciated!

fredrikekre:
That should work. What you are missing is probably just to start Julia with `--code-coverage=user`.

briochemc:
Great! Thanks for your help @fredrikekre! I think this worked! @mortenpi do you think this should be added to Documenter's docs?

mortenpi:
Yeah, I think it could be a useful thing to mention.

PR #1399, opened 2020-08-12
Add example of code coverage to docs

briochemc:
Just a suggestion to add an example of how to cover the code run during the docs deployment to codecov, as per issue/question https://github.com/JuliaDocs/Documenter.jl/issues/1397#issuecomment-672277080. (This probably needs polishing and generalized to include coveralls as well?)

_Edit by @mortenpi: close #1397_

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -282,6 +282,22 @@ See GitHub's manual for
 [Creating and using encrypted secrets](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets)
 for more information.
 
+### Add code coverage from your deployed docs
+
+If you want code run during the documentation deployment to be covered by codecov,
+you can edit the end of the docs part of your workflow configuration file so that 
+`docs/make.jl` is run with the `--code-coverage=user` flag and the coverage reports
+are uploaded to codecov:
+
+```yaml
+      - run: julia --project=docs/ --code-coverage=user docs/make.jl
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }}
+      - uses: julia-actions/julia-uploadcodecov@latest
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }
+```
 
 ## `docs/Project.toml`
 
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -284,7 +284,7 @@ for more information.
 
 ### Add code coverage from documentation builds
 
-If you want code run during the documentation deployment to be covered by codecov,
+If you want code run during the documentation deployment to be covered by Codecov,
 you can edit the end of the docs part of your workflow configuration file so that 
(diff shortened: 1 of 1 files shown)


docs/src/man/hosting.md
```diff
@@ -282,6 +282,22 @@ See GitHub's manual for
...
+you can edit the end of the docs part of your workflow configuration file so that 
+`docs/make.jl` is run with the `--code-coverage=user` flag and the coverage reports
+are uploaded to codecov:
+
+```yaml
+      - run: julia --project=docs/ --code-coverage=user docs/make.jl
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }}
+      - uses: julia-actions/julia-uploadcodecov@latest
```
mortenpi, line 297: Two comments here:

1. It looks like [julia-uploadcodecov](https://github.com/julia-actions/julia-uploadcodecov) is deprecated in favour of [codecov/codecov-action](https://github.com/codecov/codecov-action)?
2. It might be better to recommend a specific (major) version, instead of `latest`?

briochemc: I made a new commit with these 2 points implemented (I think).

briochemc:
> Mentioning how to do it for Coveralls would also be a nice-to-have.

Agreed, but I don't know how to do it. Maybe add an issue once this works for you for codecov?

mortenpi:
The [julia-processcoverage example](https://github.com/julia-actions/julia-processcoverage#usage) also has the  `with: file: lcov.info` section. Is that necessary, or can that be omitted as you have done?

Also, just to see it in action, do you have a public package where this is being used and is successfully uploading coverage?

briochemc:
> The [julia-processcoverage example](https://github.com/julia-actions/julia-processcoverage#usage) also has the `with: file: lcov.info` section. Is that necessary, or can that be omitted as you have done?

I *think* it works without it, but maybe that file does something I don't get?

> Also, just to see it in action, do you have a public package where this is being used and is successfully uploading coverage?

Yes, it should be working with [AIBECS.jl](https://github.com/JuliaOcean/AIBECS.jl)

mortenpi:
@SaschaMann, I hope you don't mind me tagging you, but perhaps you could comment: is the 

```yaml
        with:
          file: lcov.info
```

part necessary when using the `julia-processcoverage` GitHub Action?

SaschaMann:
@davidanthoff wrote that action, he will know if/why it's required.

Based on the codecov bash uploader (which their action wraps), I'd guess it may not be necessary, but I haven't used/tested it: https://github.com/codecov/codecov-bash/blob/master/codecov#L1208

davidanthoff:
I don't remember, I'm afraid :)

mortenpi:
LGTM. [This line here](https://codecov.io/gh/mortenpi/DocumenterGitHubActionsExample.jl/src/03719b1ea2197f5d9504a5880e33ad1c5d383574/src/DocumenterGitHubActionsExample.jl#L7) only gets run [in the documentation build](https://github.com/mortenpi/DocumenterGitHubActionsExample.jl), and the workflow is set up without the `with` section.

PR #1401, opened 2020-08-13
render a latex environment input as a paragraph

likanzhan:
Sometimes, the output of a `@example` block is itself a latex environment, such as a [DataFrame](https://github.com/JuliaData/DataFrames.jl/edit/master/docs/src/man/getting_started.md#L36).

````
```@example
using DataFrames
df = DataFrame(i=1:4, y='A':'D', z=5:8);
df
```
````

In this case, some special characters, such as `&` and `\`, are not need escaped, except for the linebreak symbol `\\`.
This pr intends to do this.
This is related to issue #1346.

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -498,8 +498,15 @@ function _print_code_escapes_inline(io, s::AbstractString)
 end
 
 function latex(io::IO, md::Markdown.Paragraph)
-    for md in md.content
-        latexinline(io, md)
+    if occursin(r"^\h*?\\begin{\p{Xan}*?}", md.content[begin])
+        for md in md.content
+            md = Base.replace(md, r"\\$" => "\\\\")
+            Base.print(io.io, md)
+        end
+    else
+        for md in md.content
+            latexinline(io, md)
+        end
     end
     _println(io, "\n")
 end

### CODE REVIEW

File: src/Writers/LaTeXWriter.jl [hunk 1]
```diff
@@ -498,8 +498,15 @@ function _print_code_escapes_inline(io, s::AbstractString)
 end
 
 function latex(io::IO, md::Markdown.Paragraph)
-    for md in md.content
-        latexinline(io, md)
+    if occursin(r"^\h*?\\begin{\p{Xan}*?}", md.content[begin])
```
mortenpi, line 501: The `[begin]` syntax was added in 1.4, so this would have to be rewritten as e.g. `first(md.content)` to be compatible with 1.0.

src/Writers/LaTeXWriter.jl
```diff
@@ -499,8 +499,15 @@ function _print_code_escapes_inline(io, s::AbstractString)
 end
 
 function latex(io::IO, md::Markdown.Paragraph)
-    for md in md.content
-        latexinline(io, md)
+    if occursin(r"^\h*?\\begin{\p{Xan}*?}", first(md.content))
```
mortenpi, line 502: Elements of `md.content` are not guaranteed to be strings, so this may throw an error, which it shouldn't: https://travis-ci.org/github/JuliaDocs/Documenter.jl/jobs/717872752#L320-L329

test/examples/src.latex_simple/index.md
```diff
@@ -28,3 +28,26 @@ julia> 127 % Int8
...
 ```
+
+## Issue 1401
+
+1. The output is a latex tabular environment
+
+```@example
+using DataFrames
+DataFrame(i=1:3, y='A':'C')
+```
```
mortenpi, line 39: As I said in https://github.com/JuliaDocs/Documenter.jl/pull/1403#discussion_r470536887, this at-example block will throw an error at the moment because we don't have DataFrames as part of the test dependencies. As we'd like to keep it that way, we'd need a custom type that would have the corresponding `show` method defined.

The `show` method could actually be defined in the same at-example block. If you're not quite sure what I mean, you can look at the image tests we have for inspiration: https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/src/man/tutorial.md#including-images-with-mime

test/examples/src.latex_simple/index.md
```diff
@@ -28,3 +28,26 @@ julia> 127 % Int8
...
+\begin{tabular}{r|ccc}
+        & i & y & z\\
+        \hline
+        & Int64 & Char & Int64\\
+        \hline
+        1 & 1 & 'A' & 5 \\
+        2 & 2 & 'B' & 6 \\
+        3 & 3 & 'C' & 7 \\
+        4 & 4 & 'D' & 8 \\
+\end{tabular}
```
mortenpi, line 52: Coming back to https://github.com/JuliaDocs/Documenter.jl/pull/1403#discussion_r470538154: do you intend this to render as just text or do you want this to render as a table? Because it _should_ render as text.

likanzhan: Yes, I intend to render as a table.

test/examples/src.latex_simple/index.md
likanzhan: Thanks. I tried to define a `show` method as you suggested, but with no success.  As DataFrames has already defined several [show](https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/io.jl#L39-L58) methods, also see the [_show](https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/io.jl#L250-L330) method, is it possible to import the method from there or is it valuable to define a custom type (If it valuable, I can see what I can do to help) in Documenter? Thanks. As an end user, it would be better if the tabular data can be presented in a reader friendly format.

mortenpi: We wouldn't want DataFrames as a test dependency just for this. However, while I haven't tested it, something like this should work I think:

````
```@example
struct Table end
Base.show(io, ::MIME"text/latex", ::Table) = write(io, raw"""
begin{tabular}{r|ccc}
        & i & y & z\\
        \hline
        & Int64 & Char & Int64\\
        \hline
        1 & 1 & 'A' & 5 \\
        2 & 2 & 'B' & 6 \\
        3 & 3 & 'C' & 7 \\
        4 & 4 & 'D' & 8 \\
\end{tabular}
""")
Table()
```

test/examples/src.latex_simple/index.md
mortenpi: Ok, but it _should_ render as text, because this is just text in a Markdown file. To include raw TeX like this, one should use an at-raw block:

````
```@raw latex
\begin{tabular}{r|ccc}
        & i & y & z\\
        \hline
        & Int64 & Char & Int64\\
        \hline
        1 & 1 & 'A' & 5 \\
        2 & 2 & 'B' & 6 \\
        3 & 3 & 'C' & 7 \\
        4 & 4 & 'D' & 8 \\
\end{tabular}
```
````

likanzhan: Example 2 works as expected.
Example-02

test/examples/src.latex_simple/index.md
likanzhan: Example 1 does not work. 
Example-01

mortenpi:
I took the liberty of updating the tests a bit, so the last inline comment is now attached to the wrong place. However, in the examples:

1. The `Table()` does work, but does not render correctly. I think it should be [handled here](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Writers/LaTeXWriter.jl#L395-L396), but isn't for some reason. So this would be nice to have a fix for.
2. The at-raw block here should render as a table (and does render correctly).
3. The text example below _should not_ render as a table, because it is semantically just text. To include raw LaTeX, you should use an at-raw block. I am pretty sure that the changes to the `Markdown.Paragraph` method are unnecessary.

This is how the examples render with master (since this branch currently errors): [DocumenterLaTeXSimple.pdf](https://github.com/JuliaDocs/Documenter.jl/files/5121284/DocumenterLaTeXSimple.pdf)

PR #1403, opened 2020-08-14
Add test items for #1401

likanzhan:
Add test items for #1401.

### DIFF
--- a/test/examples/src.latex_simple/index.md
+++ b/test/examples/src.latex_simple/index.md
@@ -29,7 +29,7 @@ julia> 127 % Int8
 julia> function foo end;
 ```
 
-## Issue 1400
+## Issue 1401
 
 1. The output is a latex tabular environment
 
--- a/test/examples/src.latex_simple/index.md
+++ b/test/examples/src.latex_simple/index.md
@@ -28,3 +28,26 @@ julia> 127 % Int8
 ```julia-repl sayhello2
 julia> function foo end;
 ```
+
+## Issue 1400
+
+1. The output is a latex tabular environment
+
+```@example
+using DataFrames
+DataFrame(i=1:3, y='A':'C')
+```
+
+2. The paragraph itself is a latex tabular environment
+
+\begin{tabular}{r|ccc}
+        & i & y & z\\
+        \hline
+        & Int64 & Char & Int64\\
+        \hline
+        1 & 1 & 'A' & 5 \\
+        2 & 2 & 'B' & 6 \\
+        3 & 3 & 'C' & 7 \\
+        4 & 4 & 'D' & 8 \\
+\end{tabular}
+


test/examples/src.latex_simple/index.md
```diff
@@ -28,3 +28,26 @@ julia> 127 % Int8
...
 ```
+
+## Issue 1401
+
+1. The output is a latex tabular environment
+
+```@example
+using DataFrames
+DataFrame(i=1:3, y='A':'C')
+```
```
mortenpi, line 39: This code block will not run at the moment because we don't have DataFrames in our test dependencies. Ideally, this would use a custom type with a specific `show` method for the LaTeX MIME in [`test/examples/make.jl`](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/make.jl). It can be just a show method that prints the Tex code below.

test/examples/src.latex_simple/index.md
```diff
@@ -28,3 +28,26 @@ julia> 127 % Int8
...
+\begin{tabular}{r|ccc}
+        & i & y & z\\
+        \hline
+        & Int64 & Char & Int64\\
+        \hline
+        1 & 1 & 'A' & 5 \\
+        2 & 2 & 'B' & 6 \\
+        3 & 3 & 'C' & 7 \\
+        4 & 4 & 'D' & 8 \\
+\end{tabular}
```
mortenpi, line 52: Do you intend this to render as text or a table? Because it _should_ render as just plain text. If you indeed want to include raw LaTeX, you'd need to use [the at-raw block](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@raw-format-block).

test/examples/src.latex_simple/index.md
likanzhan: DataFrames indeed defines a show method for [LaTeX MIME](https://github.com/JuliaData/DataFrames.jl/blob/master/src/abstractdataframe/io.jl#L250-L330).

And the following Tex Code is the output of this method of DataFrames:
````
\begin{tabular}{r|ccc}
        & i & y & z\\
        \hline
        & Int64 & Char & Int64\\
        \hline
        1 & 1 & 'A' & 5 \\
        2 & 2 & 'B' & 6 \\
        3 & 3 & 'C' & 7 \\
        4 & 4 & 'D' & 8 \\
\end{tabular}
````

This should be rendered as a latex table in the output pdf file.

In Ducumenter, however, this is treated as as plain text, and in the final output pdf file, it becomes something like

````
\begin{tabular}{r|cc} & i & y\ \hline & Int64 & Char\ \hline 1 & 1 & 'A' \ 2 & 2 & 'B' \ 3 & 3 & 'C' \ \end{tabular}
````
PDF

This is not what it should be.

P.S. The HTML MIME version seems correct:

HTML

Issue #658, opened 2018-03-06
Any Plans to support i18n?


Roger-luo:
Hi JuliaDocs team, 

Our old Chinese document (v0.3) in [JuliaCN](https://github.com/JuliaCN/julia_zh_cn) is out of date, we are planning to have a new Chinese document for Julia. Since, currently, julia source is using Documenter.jl, I'm wondering if you have any suggestions on i18n with Documenter.jl? (the Julia-i18n approach may not work for Chinese users, since that web service could be hard to reach in China and it would be more convenient to let user access docs inside REPL).

the original English documents collects docs from each stdlib, so we have to edit julia source (the `doc` directory and `doc` directory in `stdlib`), it may mess the original julia fork. It might be cleaner to have separated files/dir for Chinese docs. Previously, since all the docs are in `doc` with readthedocs services, we simply copied it, but it seems not working anymore with Markdown in the latest version.

We would like to hear your suggestions.

Bests,
JuliaCN team

mortenpi:
Definitely no objections to having Documenter help with i18n as much as possible, although there's nothing specific on the roadmap at the moment. Depending on what the specific features might be that you need, it might make sense to implement them as a plugin through a separate package (DocumenterI18N?).

Some quick thoughts:

* For the Julia manual pages, setting up a separate repository with the translated Markdown pages should be quite easy. And then you can build the translated docs with Documenter from scratch. Some of the `.md` files of the manual live under `stdlib/` in the Julia repo, so you'd need to copy those over as well.

* Docstrings are a bit harder since they come from the binary. But you can always programmatically access and edit docstrings on the fly, so it shouldn't be hard to overwrite them before calling `makedocs`.

* The docstrings could probably be overwritten in a package as well, which would allow i18n of the REPL (e.g. the user would start a session with e.g. `using JuliaZH`, and that would load translated docstrings into Base). To facilitate this, some of this work might have to happen in Base (I am not _that_ familiar with the docsystem code in Base).

* I guess it would be nice if Documenter(I18N) would keep track of whether a page / docstring has gotten out of date though?

Roger-luo:
We currently using this [JuliaCN.jl](https://github.com/JuliaCN/JuliaCN.jl) (still work in progress) to overwrite docstring in binary so we could support i18n in REPL, and translate documents in docs and use current functions of Documenter.jl to generate Chinese documentations.

[...]

Issue #297, opened 2016-10-04
Translated and Other Version Docs


ChrisRackauckas:
While having two badges works, the ideal case would be to have the ability to switch between versions of documentation via a menu. By allowing a full menu, this would allow choices beyond `Stable` and `Latest`. This can be good for keeping older versions of docs for backwards compatibility, or having documentation translated in other languages. Since Julia is already very widely used in places where English is not the main language, it is and will continue to be very important that we support translated documentation.

Maybe @Ismael-VC might have some good ideas for how to implement this. My idea would just be to let the package developer use different Markdown files for the translated files, and have a way of specifying the different kinds of docs. One way to implement this would be to allow one higher level in "pages" where we can specify the Markdown files for building the different versions/translations. 

Here are screenshots of examples from the Julia website and the Julia documentation for how this is shown:

![julialangsite](https://cloud.githubusercontent.com/assets/1814174/19095766/b073e4d8-8a4c-11e6-8044-a411112afb11.PNG)

![juliadocs](https://cloud.githubusercontent.com/assets/1814174/19095767/b076e052-8a4c-11e6-9e77-acd6f9fcd4a4.PNG)

ChrisRackauckas:
I talked this over with @Ismael-VC. It seems it wouldn't be too difficult to integrate a translation site into Documenter.jl. You can see that this was done for the Julia site via placing the Javascript:

https://github.com/JuliaLang/julialang.github.com/pull/252/files

This can be implemented in Documenter by simply adding the API key to the `makedocs` function, and adding the Javascript to the page headers. 

Another place to look would be integrating Crowdin instead of Transifex Live.

As for adding to the dropbox a button for other versions, that would take a Javascript expert. Look at the discussion here: julialang.github.com#428. Maybe a Juno dev could help out? @pfitzseb?

MichaelHatherly:
Adding an optional API key sounds reasonable to me. Probably best to stick with Transifex Live so that translators don't need to learn a different system.

mortenpi:
Having a widget to swap between versions is something we definitely need (also mentioned in #212).

It would be nice to integrate the language and version widgets into a single one I think -- could that be easily done with Transifex's widget?

ChrisRackauckas:
Now that you have the widget for versions, could translations be put in there?

MichaelHatherly:
[...]

PR #1407, opened 2020-08-18
Clarify the docstring of lang keyword

mortenpi:
Using slightly more accurate language, ref: https://github.com/JuliaDocs/Documenter.jl/pull/1223#issuecomment-675223146


nil-is-lin:
Thank you very much for your reply. Is there any way to switch different web pages according to the language, which may be useful for some non-native English users. I want to write web pages in two languages together, which may not be a wise way, but it does work. Does this need to declare two kinds of Lang tag?

mortenpi:
Documenter doesn't have any built-in support for multi-language docs and nothing specific on the roadmap for this (see #297, #658 for some discussion on this).

The `lang` keyword is a very specific feature to be able to declare the language of the current document, by setting the [`lang` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang). But it doesn't help you have multiple languages.

nil-is-lin:
OK, I see. I'll try something else. Thank you

PR #1415, opened 2020-09-14
simple typo fix "rather then" to "rather than"

vlandau:
Fixed small typo -- "rather then" should be "rather than".


mortenpi:
Thanks!

Issue #1416, opened 2020-09-17
Level 2 headings become Chapters in pdf build if pages contains arrays of files


cserteGT3:
I have the following page structure (as an MWE):
```
pages = ["Home" => "index.md",
        "f1" => ["f1/file1.md", "f1/file2.md"],
        "f2F1" => "f2/file1.md"]
```
The problem is, that the Level 2 headings (except those that are in the files in the array) become chapters in the generated pdf.
I think that they shouldn't be chapters, but subsections.
For easier understanding and exploring, I've created an MWE repo here: https://github.com/cserteGT3/DocumenterHeadingIssue

cserteGT3:
To clarify, what happens:
- `"Home"` becomes Part, Level 1 heading of `"index.md"` becomes Part, and Level 2 heading of `"index.md"` becomes Chapter
- `"f1"` becomes Part, Level 1 heading of `"f1/file1.md"` and `"f1/file2.md"` become Chapters and Level 2 heading of `"f1/file1.md"` and `"f1/file2.md"` become Subsections
- `"f2F1"` becomes Part, Level 1 heading of `"f2/file1.md"` becomes Part and Level 2 heading of `"f2/file1.md"` becomes Chapter

I suppose that it's not consistent and the same heading levels should translate to the same level in the pdf as well.

cserteGT3:
So I tried to track where it goes wrong. It happens in `Documenter.Writers.LaTeXWriter.files()`:
https://github.com/JuliaDocs/Documenter.jl/blob/cee9c16fc8c8ceeaae32b5bae9aa1705de5947ec/src/Writers/LaTeXWriter.jl#L739

I have two examples:

1. No array:

```julia
pgs =  ["Home" => "index.md",
			"f2F1" => "f2/file1.md"]
makedocs(
    format = LaTeX(platform="docker"),
    sitename = "TestPage",
	build = "pdfbuild",
    pages = pgs
)
```
Output of the `files` function:

```
julia> Documenter.Writers.LaTeXWriter.files(pgs)
2-element Array{Tuple{String,String,Int64},1}:
 ("Home", "index.md", 1)
 ("f2F1", "f2/file1.md", 1)
```

This generates a correct output, every level in the markdown is translated consistently to LaTeX levels.

2. array:

```julia
pgs =  ["Home" => "index.md",
            "f1" => ["f1/file1.md", "f1/file2.md"],
			"f2F1" => "f2/file1.md"]
makedocs(
[...]

PR #1417, opened 2020-09-17
Workaround for headings (#1416)

cserteGT3:
A workaround for #1416

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -741,8 +741,12 @@ function files!(out, p::Pair{S, T}, depth) where {S <: AbstractString, T <: Abst
 end
 
 function files!(out, p::Pair{S, V}, depth) where {S <: AbstractString, V}
-    push!(out, (p.first, "", depth))
-    files!(out, p.second, depth)
+    if p.second isa String
+        files!(out, p.first=>String(p.second), depth)
+    else
+        push!(out, (p.first, "", depth))
+        files!(out, p.second, depth)
+    end
 end
 
 files(v::Vector) = files!(Tuple{String, String, Int}[], v, 0)


mortenpi:
Just to make sure I understand: the only thing this should change is that if a `##` heading would become a L2 heading, which currently would be converted into a `\chapter{}`, it will be "demoted" to a `\section{}` instead?

Is the [example PDF](https://github.com/cserteGT3/DocumenterHeadingIssue/blob/master/docs/pdfbuild/TestPage.pdf) generated with the patched Documenter? It is weird that it now splits "Home" and "Index page" into separate `\part`s.. but that may well be a another bug that existed before.

For testing, could you perhaps add another simple LaTeX build [here](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/make.jl#L338-L353) (tests go in [here](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests_latex.jl)). As an idea, what we could do is use the `LaTeX(platform="none")` option to generate just the TeX file and parse out all the `\part`, `\section` etc., parts and make sure that they appear in the correct order.

cserteGT3:
> Just to make sure I understand: the only thing this should change is that if a `##` heading would become a L2 heading, which currently would be converted into a `\chapter{}`, it will be "demoted" to a `\section{}` instead?

Not exactly. Every heading will be converted to the correct LaTeX heading. I've updated the MEW repo's readme with an example, that I copy here as well:

Suppose we have the follownig structure:

```julia
PAGES = ["Part1" => "index.md",
            "Part2" => ["f1/file1.md", "f1/file2.md"],
            "Part3" => "f2/file1.md"]
```

### Documenter.jl@0.25

The latest stable release of Documenter generates the following (headings parsed from the `.tex` file):

```latex
\part{Part1}
\part{Level 1 at Index page}
\chapter{Level 2 at Index page}
\section{Level 3 at Index page}
\subsection{Level 4 at Index page}
\subsubsection{Level 5 at Index page}
\paragraph{Level 6 at Index page}
\part{Part2}
\chapter{Level 1 at f1-File1}
\section{Level 2 at f1-File1}
\subsection{Level 3 at f1-File1}
\subsubsection{Level 4 at f1-File1}
\paragraph{Level 5 at f1-File1}
\subparagraph{Level 6 at f1-File1}
\chapter{Level 1 at f1-File2}
\section{Level 2 at f1-File2}
\subsection{Level 3 at f1-File2}
\subsubsection{Level 4 at f1-File2}
\paragraph{Level 5 at f1-File2}
\subparagraph{Level 6 at f1-File2}
\part{Part3}
\part{Level 1 at f2-File1}
\chapter{Level 2 at f2-File1}
\section{Level 3 at f2-File1}
\subsection{Level 4 at f2-File1}
\subsubsection{Level 5 at f2-File1}
\paragraph{Level 6 at f2-File1}
```

### Documenter.jl with #1417

With this patch applied we get the following:

```latex
\part{Part1}
\chapter{Level 1 at Index page}
\section{Level 2 at Index page}
\subsection{Level 3 at Index page}
\subsubsection{Level 4 at Index page}
\paragraph{Level 5 at Index page}
\subparagraph{Level 6 at Index page}
\part{Part2}
\chapter{Level 1 at f1-File1}
\section{Level 2 at f1-File1}
\subsection{Level 3 at f1-File1}
\subsubsection{Level 4 at f1-File1}
\paragraph{Level 5 at f1-File1}
\subparagraph{Level 6 at f1-File1}
\chapter{Level 1 at f1-File2}
\section{Level 2 at f1-File2}
\subsection{Level 3 at f1-File2}
\subsubsection{Level 4 at f1-File2}
\paragraph{Level 5 at f1-File2}
\subparagraph{Level 6 at f1-File2}
\part{Part3}
\chapter{Level 1 at f2-File1}
\section{Level 2 at f2-File1}
\subsection{Level 3 at f2-File1}
\subsubsection{Level 4 at f2-File1}
\paragraph{Level 5 at f2-File1}
\subparagraph{Level 6 at f2-File1}
```

> Is the [example PDF](https://github.com/cserteGT3/DocumenterHeadingIssue/blob/master/docs/pdfbuild/TestPage.pdf) generated with the patched Documenter? It is weird that it now splits "Home" and "Index page" into separate `\part`s.. but that may well be a another bug that existed before.

No, the linked pdf is built with Documenter@0.25

> For testing, could you perhaps add another simple LaTeX build [here](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/make.jl#L338-L353) (tests go in [here](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests_latex.jl)). As an idea, what we could do is use the `LaTeX(platform="none")` option to generate just the TeX file and parse out all the `\part`, `\section` etc., parts and make sure that they appear in the correct order.

Thanks for the idea, it helped me to easier describe the issue as well. I'll write those tests.

cserteGT3:
A wrote those tests and also needed to modify the Markdown tests, because of the new markdown files I added.

mortenpi:
Ah, I think I see the problem now. Thanks for the extra info! I.e. if we have the `"Foo" => "bar.md"` case in `pages`, we currently create two headings for that on the same level (e.g. `\part` for "Part 1" and "Level 1 at Index page")? 

The `"Foo" => "bar.md"` syntax in `pages` is there so that the user could provide a shorter heading for the page. In the HTML backend we use `p.first` in the sidebar, but on the page we still render the first `#` Markdown heading. If the user does not provide an override in `pages`, we just use the `#` heading for the sidebar.

Of course in LaTeX we don't have a way to use two different names for a heading. So I would make the argument that in this case we should drop the user-provided override in the `String => String` case, instead of creating two levels of headings for the same heading.

E.g. I think with this patch the following case

```julia
pages = ["Introduction" => "introduction.md"]
```
```
introduction.md:
# Introduction to MyPkg.jl
```

would end up like

```tex
\part{Introduction}
\chapter{Introduction to MyPkg.jl}
```

which seems redundant.

```tex
\part{Introduction to MyPkg.jl}
```

should be sufficient, no?

test/examples/make.jl
```diff
@@ -395,3 +395,22 @@ else
...
+    @info("Building mock package docs: LaTeXWriter/latex_headings")
+    @quietly makedocs(
+        format = Documenter.Writers.LaTeXWriter.LaTeX(platform = "none"),
+        sitename = "Documenter LaTeX Heading Tests",
+        root  = examples_root,
+        build = "builds/latex_headings",
+        pages = ["Part1" => "headings/f1.md",
+                "Part2" => ["headings/f2.md", "headings/f3.md"],
+                "Part3" => "headings/f4.md"
+                ],
```
mortenpi, line 409: We should also test the case where the user does not provide an override, i.e.:

```suggestion
                "Part3" => "headings/f4.md",
                "headings/f5.md",
                ],
```

test/examples/make.jl
```diff
@@ -395,3 +395,22 @@ else
...
     nothing
 end
+
+examples_latex_headings_doc = if "latex_headings" in EXAMPLE_BUILDS
+    @info("Building mock package docs: LaTeXWriter/latex_headings")
+    @quietly makedocs(
+        format = Documenter.Writers.LaTeXWriter.LaTeX(platform = "none"),
+        sitename = "Documenter LaTeX Heading Tests",
+        root  = examples_root,
+        build = "builds/latex_headings",
```
mortenpi, line 405: It would be nice to move the heading `.md` files into a separate source directory (e.g. `src.latex_headings) and then put a `source = "src.latex_headings"` here (like we have for `latex_simple`).

cserteGT3:
I'm opposed to your suggestion. Consider the following example:
```
pages = ["Introduction" => "introduction.md",
    "Part2" => ["page1.md", "page2.md"]]
```
and

```
introduction.md:
# Introduction to MyPkg.jl

page1.md:
# Level 1 on Page 1

page2.md:
# Level 1 on Page 2
```

With your suggestion (if I understand correctly, it would compile to):
```
\part{Introduction to MyPkg.jl}
\part{Part 2}
\chapter{Level 1 on Page1}
\chapter{Level 1 on Page2}
```

My issue with that is, that the same level markdown headings are not compiled to the same levels in LaTeX, which seems inconsistent to me. (I mean, a Level 1 in `introduction.md` compiles to `part`, while Level 1 in `page1.md` and `page2.md` compile to `chapter`.) I would rather have the redundancy, than the incosistency.

I suggest the following example, that includes all the different arguments that `pages` could be (I hope):

mortenpi:
One thing to consider is that full consistency is not really feasible. You can also have more nested levels (e.g. as used in the [Julia manual](https://docs.julialang.org/en/v1/devdocs/init/)):

```julia
pages = [
    "Introduction" => "page1.md",
    "Part2" => [
        "page2_1.md",
        "Sub-part" => [
            "page2_2_1.md",
            "Sub-sub-page" => "page2_2_2.md"
        ],
    ]
]
```

which I believe in your proposed scheme would become

```tex
\part{Introduction}
  \chapter{L1 in page1.md}
\part{Part}
  \chapter{L1 in page2_1.md}
  \chapter{Sub-part}
    \section{L1 in page2_2_1.md}
    \section{Sub-sub-page}
      \subsection{L1 in page_2_2_2.md}
```

That said, I do agree that having L2 headings of a Markdown page as `\chapter`s is strange. Another edge case to consider is what would happen if you simply have `pages = ["page1.md"]` (e.g. single `index.md` case where `pages` is omitted).

Could we just have a rule that we either forces L1 headings to be `\chapter` or lower? Or L2 headings be `\section`s or lower? In all cases I would still advocate for not treating the `"Foo" => "foo.md"` case as introducing two different levels though (since `"Foo"` there is really just an alias). So maybe something like:

```tex
\part{L1 in page1.md}
    \section{L2 in page1.md}
    \section{L2 in page1.md}
\part{Part}
  \chapter{L1 in page2_1.md}
    \section{L2 in page2_1.md}
    \section{L2 in page2_1.md}
  \chapter{Sub-part}
    \section{L1 in page2_2_1.md}
      \subsection{L2 in page2_2_1.md}
      \subsection{L2 in page2_2_1.md}
    \section{L1 in page_2_2_2.md}
      \subsection{L2 in page_2_2_2.md}
      \subsection{L2 in page_2_2_2.md}
```

For another datapoint: @MichaelHatherly, how do you handle the headings in PDFs with Publish?

MichaelHatherly:
Publish doesn't really specify anything with regards to headings. The only thing the `toc.md` file does (`pages` equivalent here) is determine page inclusion order, actual h1-6 headers always just map to the same as section, subsection, etc no matter where they appear. To get chapter and part is left to the user with raw `{=latex}` blocks. When reviewing and deciding that I felt that there's just too much difference between html and latex for us to be able to make the right decision automatically in every case so an easily understandable scheme was best.

PR #1425, opened 2020-10-02
Add a showcase example for multiple uses of the same symbol.

### LINKED ISSUE #1361: add example to @docs for type & multiple constructors
I am thinking of making a PR to the docs, adding an example documenting constructors & types, and I am asking to check first. In

```julia
"what a splendid type"
struct Foo{T,S} end

"0"
Foo() = Foo{Nothing,Nothing}()

"1"
Foo{T}() where T = Foo{T,Nothing}()
```

I think that

````
```@docs
Foo        # what a splendid type
Foo()      # 0
Foo{T}()   # 1
Foo{T,S}() # undocumented
```
````

is the recommended way to get everything, is this correct?

### PR DESCRIPTION
**Pull request by tpapp**
Fixes #1361.

### DIFF
--- a/docs/src/showcase.md
+++ b/docs/src/showcase.md
@@ -274,7 +274,7 @@ The [`@index` block](@ref) can be used to generate a list of all the docstrings 
 Pages = ["showcase.md"]
 ```
 
-## Multiple uses of the same symbol
+### Multiple uses of the same symbol
 
 Sometimes a symbol has multiple docstrings, for example a type definition, inner and outer constructors. The example
 below shows how to use specific ones in the documentation.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by fredrikekre — APPROVED** [review]

File: docs/src/showcase.md [hunk 1]
```diff
@@ -274,6 +274,17 @@ The [`@index` block](@ref) can be used to generate a list of all the docstrings
 Pages = ["showcase.md"]
 ```
 
+## Multiple uses of the same symbol
```
mortenpi, line 277:
```suggestion
### Multiple uses of the same symbol
```

mortenpi:
LGTM, just changed it to be a subsection of the "Docstrings" section. Thanks!

PR #1426, opened 2020-10-02
Strip out $$ as well

blegat:
Rebases https://github.com/JuliaDocs/Documenter.jl/pull/1283 and strips `$$` as well

_Edit by @mortenpi: close #1283, fix #1278_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * ![Enhancement][badge-enhancement] The URL to the MathJax JS module can now be customized by passing the `url` keyword argument to the constructors (`MathJax2`, `MathJax3`). ([#1428][github-1428], [#1430][github-1430])
 
 * ![Bugfix][badge-bugfix] `Documenter.doctest` now correctly accepts the `doctestfilters` keyword, similar to `Documenter.makedocs`. ([#1364][github-1364], [#1435][github-1435])
+
 ## Version `v0.25.2`
 
 * ![Deprecation][badge-deprecation] The `Documenter.MathJax` type, used to specify the mathematics rendering engine in the HTML output, is now deprecated in favor of `Documenter.MathJax2`. ([#1362][github-1362], [#1367][github-1367])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -661,14 +661,11 @@
 [github-1389]: https://github.com/JuliaDocs/Documenter.jl/pull/1389
 [github-1392]: https://github.com/JuliaDocs/Documenter.jl/pull/1392
 [github-1400]: https://github.com/JuliaDocs/Documenter.jl/pull/1400
-<<<<<<< HEAD
 [github-1428]: https://github.com/JuliaDocs/Documenter.jl/issues/1428
 [github-1430]: https://github.com/JuliaDocs/Documenter.jl/pull/1430
 [github-1364]: https://github.com/JuliaDocs/Documenter.jl/issues/1364
 [github-1435]: https://github.com/JuliaDocs/Documenter.jl/pull/1435
-=======
 [github-1426]: https://github.com/JuliaDocs/Documenter.jl/pull/1426
->>>>>>> 5eca1784... Add CHANGELOG
 
 [documenterlatex]: https://github.com/JuliaDocs/DocumenterLaTeX.jl
 [documentermarkdown]: https://github.com/JuliaDocs/DocumenterMarkdown.jl
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,14 @@
 # Documenter.jl changelog
 
+## Version `v0.26.0`
+
+* ![Enhancement][badge-enhancement] Objects that render as equations and whose `text/latex` representations are wrapped in display equation delimiters `\[ ... \]` or `$$ ... $$` are now handled correctly in the HTML output. ([#1278][github-1278], [#1283][github-1283], [#1426][github-1426])
+
 ## Version `v0.25.3`
 
 * ![Enhancement][badge-enhancement] The URL to the MathJax JS module can now be customized by passing the `url` keyword argument to the constructors (`MathJax2`, `MathJax3`). ([#1428][github-1428], [#1430][github-1430])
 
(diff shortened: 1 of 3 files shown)


mortenpi:
Thanks for adopting that PR :smile: A few things to think about before we merge:

1. Will we ever be in a situation where we'll have these delimiters nested? E.g.. `\[$$ ... $$\]`? I feel that this will always be a bug in the generator of LaTeX.

   So, with that in mind, I am wondering if we should just `if ... else` match for the patterns and return as soon one matches, instead of passing the output to the next regex?

2. What are your thoughts on also stripping the `$ ... $` and `\( ... \)`?
3. As was mentioned in https://github.com/JuliaDocs/Documenter.jl/pull/1283#issuecomment-612235040, it's not 100% clear what the correct thing to do is if there are no detectable delimiters. We could fall back to `Utilities.mdparse`, but that means that we require the `show` methods to print delimiters.. i.e. for example if `show` prints just `x^2` then that won't be rendered as an equation. But this might be fine, since the MIME _is_ `text/latex`, which is not specific to TeX's math mode.
4. It would be good to have examples of all the delimiters we handle in the example/test file as well.

blegat:
I agree with 1. and 4, these are addressed in the new version of the PR

About 2. and 3., I agree with 

> But this might be fine, since the MIME _is_ `text/latex`, which is not specific to TeX's math mode.

The new version of the PR is non-breaking for cases where it is not wrapped by brackets or `$$` so this case may be left for later in case there is anything to be done.

src/Writers/HTMLWriter.jl
```diff
@@ -1733,7 +1733,18 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
...
         out = Documents.RawHTML(string("<img src=\"data:image/jpeg;base64,", d[MIME"image/jpeg"()], "\" />"))
     elseif haskey(d, MIME"text/latex"())
-        out = Utilities.mdparse(d[MIME"text/latex"()]; mode = :single)
+        # If the show(io, ::MIME"text/latex", x) output is already wrapped in \[ ... \] or $$ ... $$, we
+        # unwrap it first, since when we output Markdown.LaTeX objects we put the correct
+        # delimiters around it anyway.
+        latex = d[MIME"text/latex"()]
+        equation = false
+        m_bracket = match(r"\s*\\\[(.*)\\\]\s*", latex)
+        m_dollars = match(r"\s*\\\[(.*)\\\]\s*", latex)
```
mortenpi, line 1742: Should be this I think?
```suggestion
        m_dollars = match(r"\s*\\$$(.*)\\$$\s*", latex)
```

test/examples/src/man/tutorial.md
```diff
@@ -403,3 +403,33 @@ Hello World!
...
+
+```@example showablelatex
+struct LaTeXEquation
+    code :: String
+end
+Base.show(io, ::MIME"text/latex", latex::LaTeXEquation) = write(io, latex.code)
+nothing # hide
+```
+
+In an `@example` or `@eval`block, it renders as an equation:
```
mortenpi, line 419:
```suggestion
In an `@example` or `@eval` block, it renders as an equation:
```

test/examples/src/man/tutorial.md
```diff
@@ -403,3 +403,33 @@ Hello World!
...
+
+```@example showablelatex
+LaTeXEquation("x^2")
+```
```
mortenpi, line 423: This example, as is, will not render render as an equation anymore I think? Because this is the fallback situation? Could maybe be replaced with something like:

```suggestion
```@example showablelatex
LaTeXEquation("Foo $x^2$ bar.")
```
```

**Review by mortenpi — COMMENTED** [review]
Implementation LGTM, just the test examples need to be fixes. I think we'll merge this after 0.25.3 is tagged.

File: CHANGELOG.md [hunk 4]
```diff
@@ -656,10 +661,14 @@
...
 [github-1392]: https://github.com/JuliaDocs/Documenter.jl/pull/1392
 [github-1400]: https://github.com/JuliaDocs/Documenter.jl/pull/1400
+<<<<<<< HEAD
 [github-1428]: https://github.com/JuliaDocs/Documenter.jl/issues/1428
 [github-1430]: https://github.com/JuliaDocs/Documenter.jl/pull/1430
 [github-1364]: https://github.com/JuliaDocs/Documenter.jl/issues/1364
 [github-1435]: https://github.com/JuliaDocs/Documenter.jl/pull/1435
+=======
+[github-1426]: https://github.com/JuliaDocs/Documenter.jl/pull/1426
+>>>>>>> 5eca1784... Add CHANGELOG
```
mortenpi, line 671:
```suggestion
[github-1428]: https://github.com/JuliaDocs/Documenter.jl/issues/1428
[github-1430]: https://github.com/JuliaDocs/Documenter.jl/pull/1430
[github-1364]: https://github.com/JuliaDocs/Documenter.jl/issues/1364
[github-1435]: https://github.com/JuliaDocs/Documenter.jl/pull/1435
[github-1426]: https://github.com/JuliaDocs/Documenter.jl/pull/1426
```

CHANGELOG.md
```diff
@@ -1,11 +1,14 @@
...
 
+## Version `v0.26.0`
+
+* ![Enhancement][badge-enhancement] Objects that render as equations and whose `text/latex` representations are wrapped in display equation delimiters `\[ ... \]` or `$$ ... $$` are now handled correctly in the HTML output. ([#1278][github-1278], [#1283][github-1283], [#1426][github-1426])
+
 ## Version `v0.25.3`
 
 * ![Enhancement][badge-enhancement] The URL to the MathJax JS module can now be customized by passing the `url` keyword argument to the constructors (`MathJax2`, `MathJax3`). ([#1428][github-1428], [#1430][github-1430])
 
 * ![Bugfix][badge-bugfix] `Documenter.doctest` now correctly accepts the `doctestfilters` keyword, similar to `Documenter.makedocs`. ([#1364][github-1364], [#1435][github-1435])
```
mortenpi, line 11:
```suggestion
* ![Bugfix][badge-bugfix] `Documenter.doctest` now correctly accepts the `doctestfilters` keyword, similar to `Documenter.makedocs`. ([#1364][github-1364], [#1435][github-1435])

```

mortenpi:
Thanks again for adopting and iterating on this @blegat!

Issue #1428, opened 2020-10-03
script.src for MathJax3


ianqsong:
The version is currently specified and fixed on 3.0.5 for MathJax3 ([source](https://github.com/JuliaDocs/Documenter.jl/blob/6313dcdb21409977bc1c789ccceaa2478ec736a6/src/Writers/HTMLWriter.jl#L530)), but the lastest is on 3.1.2.
It seems to me that `https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-svg.js` is better as recommended by [mathjaxdoc](http://docs.mathjax.org/en/latest/web/start.html).
So I was wondering if it be better to have the `script.src` as a user-defined argument.

mortenpi:
I do strongly prefer deploying an exact version of a JS dependency. Even though MathJax should be following SemVer, there's always a possibility that a minor version change breaks something for someone, and so all existing already deployed docs suddenly break.

That said:

* We could definitely make the version and/or the URL customizable -- PRs to that effect are most welcome.
* I need to bump the versions of JS deps anyway, which should happen in 0.26.

ianqsong:
Please refer to PR #1430.

PR #1429, opened 2020-10-04
add an argument for mathjax cdn

ianqsong:
add a field for MathJax CDN that user can input by calling `Documenter.HTML`  ( #1428)


ianqsong:
sorry about the mess.

PR #1430, opened 2020-10-04
Allow specifying URLs for LaTeX engines

ianqsong:
add a field for MathJax CDN that user can input by calling Documenter.HTML

_Edit by @mortenpi: fix #1428_

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -190,7 +190,8 @@ user-provided dictionary is used.
 """
 struct MathJax2 <: MathEngine
     config :: Dict{Symbol,Any}
-    function MathJax2(config::Union{Dict,Nothing} = nothing, override=false)
+    url :: String
+    function MathJax2(config::Union{Dict,Nothing} = nothing, override=false; url = "")
         default = Dict(
             :tex2jax => Dict(
                 "inlineMath" => [["\$","\$"], ["\\(","\\)"]],
@@ -212,7 +213,7 @@ struct MathJax2 <: MathEngine
             ],
             :TeX => Dict(:equationNumbers => Dict(:autoNumber => "AMS"))
         )
-        new((config === nothing) ? default : override ? config : merge(default, config))
+        new((config === nothing) ? default : override ? config : merge(default, config), url)
     end
 end
 
@@ -241,7 +242,8 @@ user-provided dictionary is used.
 """
 struct MathJax3 <: MathEngine
     config :: Dict{Symbol,Any}
-    function MathJax3(config::Union{Dict,Nothing} = nothing, override=false)
+    url :: String
+    function MathJax3(config::Union{Dict,Nothing} = nothing, override=false; url = "")
         default = Dict(
             :tex => Dict(
                 "inlineMath" => [["\$","\$"], ["\\(","\\)"]],
@@ -253,7 +255,7 @@ struct MathJax3 <: MathEngine
                 "processHtmlClass" => "tex2jax_process",
             )
         )
-        new((config === nothing) ? default : override ? config : merge(default, config))
+        new((config === nothing) ? default : override ? config : merge(default, config), url)
     end
 end
 
@@ -373,7 +375,6 @@ struct HTML <: Documenter.Writer
     sidebar_sitename :: Bool
     highlights    :: Vector{String}
     mathengine    :: Union{MathEngine,Nothing}
-    mathjaxcdn    :: String
     footer        :: Union{Markdown.MD, Nothing}
     lang          :: String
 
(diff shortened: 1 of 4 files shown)


mortenpi:
Thanks for the PR!

I think it would be more natural to handle all of it in the `MathJax2/3` types. I.e. if you want to configure the URL, you'd just set something like `HTML(mathengine = MathJax3(url="https://...."))`. No reason to add another global to the `HTML` struct. Beyond that:

* Docstrings would need to be updated to reflect the new arguments.
* It would be good to add a test [to the example builds](https://github.com/JuliaDocs/Documenter.jl/blob/3b1934262da03c1bbf461420c46c9e12b5b05ec5/test/examples/make.jl#L196-L232) for overriding the URL, together with the relevant test [in the tests file](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests.jl) that would make sure that `documenter.js` contains the correct URL.

ianqsong:
A new field `url` is added to `MathJax2/3`, of which the default value copies the original one, i.e., "cdnjs.cloudflare.com". So [the tests file](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests.jl) stays unchanged. 

As this is just very small enhancement, any maintainers could make it happen soon, right?

src/Writers/HTMLWriter.jl
```diff
@@ -521,13 +524,14 @@ module RD
...
     function mathengine!(r::RequireJS, engine::MathJax3)
+        url = isempty(engine.url) ? "https://cdnjs.cloudflare.com/ajax/libs/mathjax/3.0.5/es5/tex-svg.js" : engine.url
         push!(r, Snippet([], [],
             """
             window.MathJax = $(json_jsescape(engine.config, 2));
 
             (function () {
                 var script = document.createElement('script');
-                script.src = 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/3.0.5/es5/tex-svg.js';
+                script.src = $url;
```
mortenpi, line 534: We need quotes here:

```suggestion
                script.src = '$url';
```

There is a `@test_broken` for this currently in the tests, so that would have to be updated as well.

mortenpi (changes requested):
I took the liberty of modifying the tests a bit, to make sure we test all the cases. I would have two more requests before merging though:

1. Also add the same argument for `KaTeX` (for consistency).
2. Update the docstrings to reflect this argument

test/examples/make.jl
```diff
@@ -213,7 +215,26 @@ else
...
+                    "packages" => ["base", "ams", "autoload", "physics"],
+                ),
+            ),
+            url = "https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js",
+        ),
+    )
+else
+    @info "Skipping build: HTML/deploy MathJax v2 (custom URL)" EXAMPLE_BUILDS get(ENV, "DOCUMENTER_TEST_EXAMPLES", nothing)
+    nothing
+end
```
mortenpi: This build actually doesn't completely work (MathJax loads, but does not render the equation at the end of `man/tutorials.md`), but I am not sure why at the moment.

ianqsong: I don't have so much knowledge of MathJax2/3, but the config part in the example of MathJax2 above is the same as the one of MathJax3.
```
MathJax3(
    Dict(
        :loader => Dict("load" => ["[tex]/physics"]),
        :tex => Dict(
            "inlineMath" => [["\$","\$"], ["\\(","\\)"]],
            "tags" => "ams",
            "packages" => ["base", "ams", "autoload", "physics"],
        ),
    ),
 ),
```

ianqsong:
> 1. Also add the same argument for `KaTeX` (for consistency).
I don't have much knowledge of `KaTax`, but there are three urls 
```
.../katex.min.css
.../katex.min.js
.../auto-render.min.js
```
Would you confirm that this is also the case for other CDNs?

I was thinking it be more efficient if I hand over this PR to you and you could do this in a single commit -- what'd you say?

mortenpi:
> I was thinking it be more efficient if I hand over this PR to you and you could do this in a single commit -- what'd you say?

I'm afraid that might not really be the case -- it's much less time-consuming just to review, so it would probably be a while before I get around to this in that situation.

Regarding KaTeX: you are right, I had forgotten that it was a bit more complicated. Let's leave KaTeX as it is.

ianqsong:
docstrings is updated. Many thanks for your help with the test cases.

mortenpi (approved):
Thanks @ianqsong! I took the liberty of doing a few edits. Once the tests pass, this should be good to go.

mortenpi:
@ianqsong was the closing of the PR accidental?

ianqsong:
sorry, a fat-finger mistake

Issue #1438, opened 2020-10-17
Heavy slowdown in ExpandTemplates due to `subtypes` function


epatters:
The steps of the `ExpanderPipeline` are dispatched by iterating over the subtypes of this abstract type. The `subtypes` function is called for every Markdown element processed, meaning that it can easily be called hundreds or thousands of times during a run of Documenter.

However, under some circumstances (possibly involving generated functions or type metaprogramming, see #1261), the `subtypes` function is extremely slow:

```
julia> import Documenter

julia> @time subtypes(Documenter.Expanders.ExpanderPipeline);
  0.057625 seconds (92.08 k allocations: 5.367 MiB)

julia> @time subtypes(Documenter.Expanders.ExpanderPipeline);
  0.011200 seconds (2.44 k allocations: 724.125 KiB)

julia> import Catlab

julia> @time subtypes(Documenter.Expanders.ExpanderPipeline);
 40.238011 seconds (1.29 M allocations: 109.522 MiB, 0.18% gc time)

julia> @time subtypes(Documenter.Expanders.ExpanderPipeline);
  5.237196 seconds (29.11 k allocations: 4.189 MiB)

julia> @time subtypes(Documenter.Expanders.ExpanderPipeline);
  5.265813 seconds (29.11 k allocations: 4.189 MiB)
```

I figure this should be considered a bug in Julia itself and I'm going to file another issue upstream. 

That being said, since the issue is currently severely impacting the build times for the docs of Catlab.jl and its dependers, I was hoping that a workaround could be implemented here. A simple workaround would move the call `sort(subtypes(T); by = order)` from inside [`dispatch`](https://github.com/JuliaDocs/Documenter.jl/blob/3b1934262da03c1bbf461420c46c9e12b5b05ec5/src/Utilities/Selectors.jl#L164) to the body of the main function [`expand`](https://github.com/JuliaDocs/Documenter.jl/blob/3b1934262da03c1bbf461420c46c9e12b5b05ec5/src/Expanders.jl#L25). That would ensure that `subtypes` is only called once during the `ExpandTemplates` step.

MichaelHatherly:
Thanks for tracking it down @epatters, would you be able to prepare a PR with that workaround and we can try get it sorted asap.

epatters:
Sure, I can prepare a PR. Thanks.

PR #1440, opened 2020-10-19
Cache subtypes to speed up selector dispatch

epatters:
Addresses #1438 by caching the results of the `subtypes` function. This PR reduces the runtime of the Catlab docs build from ~1 hr to 10 min.

The approach taken here has the side effect that a selector cannot have more steps assigned to it after it has been dispatched at least once, but hopefully that is not a problem.

### DIFF
--- a/src/Utilities/Selectors.jl
+++ b/src/Utilities/Selectors.jl
@@ -162,7 +162,10 @@ Selectors.dispatch(MySelector, args...)
 ```
 """
 function dispatch(::Type{T}, x...) where T <: AbstractSelector
-    for t in (sort(subtypes(T); by = order))
+    types = get!(selector_subtypes, T) do
+        sort(subtypes(T); by = order)
+    end
+    for t in types
         if !disable(t) && matcher(t, x...)
             runner(t, x...)
             strict(t) && return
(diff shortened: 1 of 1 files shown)


KristofferC:
Unrelated to the PR, but this whole "Selector system" feels kinda odd to me. Using dynamic dispatch and reflection for what amounts to a "do this, then do this" construct seems quite complicated and slow.

MichaelHatherly:
> Using dynamic dispatch and reflection for what amounts to a "do this, then do this" construct seems quite complicated and slow.

Yeah, would be good to get rid of it. I guess it just depends on whether any users are *actually* making use of it for extension purposes.

MichaelHatherly:
> The approach taken here has the side effect that a selector cannot have more steps assigned to it after it has been dispatched at least once, but hopefully that is not a problem.

@epatters did you also try the initial approach you outlined in https://github.com/JuliaDocs/Documenter.jl/issues/1438#issue-723841582?

> A simple workaround would move the call sort(subtypes(T); by = order) from inside dispatch to the body of the main function expand.

epatters:
@MichaelHatherly, I was originally going to try that, but then I realized it would fragment the `dispatch` logic across the package. I thought it would be better to isolate the workaround where the problem occurs so that the rationale will be clear to readers of the code.

mortenpi:
Does this kind of undo #620?

> If anyone who is using it sees this please speak up.

[DocumenterCitations.jl](https://github.com/ali-ramadhan/DocumenterCitations.jl) is an example that uses the selectors API for extending Documenter. But that shouldn't block us from improving it -- it's not, strictly, even part of the public API.

MichaelHatherly:
> DocumenterCitations.jl is an example that uses the selectors API for extending Documenter.

Cool package. Looks pretty active, so probably best not to disrupt it too much if possible.

PR #1441, opened 2020-10-19
Add support for ANSI colors

kimikage:
This starts from a Discourse topic: https://discourse.julialang.org/t/how-do-i-enable-ansi-colored-text-output-within-documenter-jl/48398

This changes:
- ~add a new field / argument `ansicolor` to `User` / `makedocs()`~
  - ~for global setting~
- add support for a keyword argument `ansicolor` in the `@example` block
  - for per-block setting to overide the global setting
- use the `ansicolor` option for the `:color` I/O context property
- translate "text/plain" output with ANSI escape codes in `HTMLWriter`
- add styles for ANSI colors to the default themes

These changes allow you to colorize the lazy print objects in the `@example` output blocks.
ansi2

However, there are some issues to be resolved.

I've put an ANSI escape code translator for HTML in a new package [`AnsiColoredPrinters.jl`](https://github.com/kimikage/AnsiColoredPrinters.jl) to make this feature work. The package is experimental and under development, and there are things left to be discussed before it is registered, including the name of the package. I won't refuse to contribute it as part of Documenter.jl, but I prefer to modularize the functionality into minimal packages.

The other is that currently there is no way to specify the `:color` I/O context property for output to `stdout` / `stderr`. (**Edit:** solved by https://github.com/JuliaDocs/IOCapture.jl/pull/1)
~This means that even if the `@example` or `@repl` code potentially supports ANSI colors, we cannot get the colored result.~

~Also, although `ansicolor` is `false` by default and is the responsibility of the user,~ we need to strip the ANSI escape codes if the output format does not support ANSI colors.  (This will be implemented soon.)

When the solutions for the above are found, we will need to discuss the compatibility with `highlight.js`, the theme for system 16 colors, etc.

### DIFF
--- a/assets/html/scss/documenter/_ansicolors.scss
+++ b/assets/html/scss/documenter/_ansicolors.scss
@@ -1,0 +1,216 @@
+$ansi-black: $black-ter !default;
+$ansi-red: null !default;
+$ansi-green: null !default;
+$ansi-yellow: null !default;
+$ansi-blue: null !default;
+$ansi-magenta: null !default;
+$ansi-cyan: null !default;
+$ansi-white: $grey-lighter !default;
+
+$ansi-light-black: null !default;
+$ansi-light-red: null !default;
+$ansi-light-green: null !default;
+$ansi-light-yellow: null !default;
+$ansi-light-blue: null !default;
+$ansi-light-magenta: null !default;
+$ansi-light-cyan: null !default;
+$ansi-light-white: $white-ter !default;
+
+@if $documenter-is-dark-theme {
+  $ansi-red: $red !default;
+  $ansi-green: $green !default;
+  $ansi-yellow: $yellow !default;
+  $ansi-blue: $blue !default;
+  $ansi-magenta: $purple !default;
+  $ansi-cyan: $turquoise !default;
+
+  $ansi-light-black: $grey-light !default;
+
+  $ansi-light-red: lighten($ansi-red, 15) !default;
+  $ansi-light-green: lighten($ansi-green, 15) !default;
+  $ansi-light-yellow: lighten($ansi-yellow, 10) !default;
+  $ansi-light-blue: lighten($ansi-blue, 15) !default;
+  $ansi-light-magenta: lighten($ansi-magenta, 10) !default;
+  $ansi-light-cyan: lighten($ansi-cyan, 15) !default;
+}
+
+@else {
+  $ansi-light-red: $red !default;
+  $ansi-light-green: $green !default;
+  $ansi-light-yellow: $yellow !default;
+  $ansi-light-blue: $blue !default;
+  $ansi-light-magenta: $purple !default;
+  $ansi-light-cyan: $turquoise !default;
+
+  $ansi-light-black: $grey !default;
+
+  $ansi-red: darken($ansi-light-red, 10) !default;
+  $ansi-green: darken($ansi-light-green, 10) !default;
+  $ansi-yellow: darken($ansi-light-yellow, 18) !default;
+  $ansi-blue: darken($ansi-light-blue, 10) !default;
+  $ansi-magenta: darken($ansi-light-magenta, 15) !default;
+  $ansi-cyan: darken($ansi-light-cyan, 10) !default;
+}
+
+.ansi span {
+  &.sgr1 {
+    font-weight: bolder;
+  }
+
+  &.sgr2 {
+    font-weight: lighter;
+  }
+
+  &.sgr3 {
+    font-style: italic;
+  }
+
+  &.sgr4 {
+    text-decoration: underline;
+  }
+
+  &.sgr7 {
+    color: $body-background-color;
+    background-color: $text;
+  }
+
+  &.sgr8 {
+    color: transparent;
+
+    span {
+      color: transparent;
+    }
+  }
+
+  &.sgr9 {
+    text-decoration: line-through;
+  }
+
+  &.sgr30 {
+    color: $ansi-black;
+  }
+
+  &.sgr31 {
+    color: $ansi-red;
+  }
+
+  &.sgr32 {
+    color: $ansi-green;
+  }
+
+  &.sgr33 {
+    color: $ansi-yellow;
+  }
+
+  &.sgr34 {
+    color: $ansi-blue;
+  }
+
+  &.sgr35 {
+    color: $ansi-magenta;
+  }
+
+  &.sgr36 {
+    color: $ansi-cyan;
+  }
+
+  &.sgr37 {
+    color: $ansi-white;
+  }
+
+  &.sgr40 {
+    background-color: $ansi-black;
+  }
+
+  &.sgr41 {
+    background-color: $ansi-red;
+  }
+
+  &.sgr42 {
+    background-color: $ansi-green;
+  }
+
+  &.sgr43 {
+    background-color: $ansi-yellow;
+  }
+
+  &.sgr44 {
+    background-color: $ansi-blue;
+  }
+
+  &.sgr45 {
+    background-color: $ansi-magenta;
+  }
+
+  &.sgr46 {
+    background-color: $ansi-cyan;
+  }
+
+  &.sgr47 {
+    background-color: $ansi-white;
+  }
+
+  &.sgr90 {
+    color: $ansi-light-black;
+  }
+
+  &.sgr91 {
+    color: $ansi-light-red;
+  }
+
+  &.sgr92 {
+    color: $ansi-light-green;
+  }
+
+  &.sgr93 {
+    color: $ansi-light-yellow;
+  }
+
+  &.sgr94 {
+    color: $ansi-light-blue;
+  }
+
+  &.sgr95 {
+    color: $ansi-light-magenta;
+  }
+
+  &.sgr96 {
+    color: $ansi-light-cyan;
+  }
+
+  &.sgr97 {
+    color: $ansi-light-white;
+  }
+
+  &.sgr100 {
+    background-color: $ansi-light-black;
+  }
+
+  &.sgr101 {
+    background-color: $ansi-light-red;
+  }
+
+  &.sgr102 {
+    background-color: $ansi-light-green;
+  }
+
+  &.sgr103 {
+    background-color: $ansi-light-yellow;
+  }
+
+  &.sgr104 {
+    background-color: $ansi-light-blue;
+  }
+
+  &.sgr105 {
+    background-color: $ansi-light-magenta;
+  }
+
+  &.sgr106 {
+    background-color: $ansi-light-cyan;
+  }
+
+  &.sgr107 {
+    background-color: $ansi-light-white;
+  }
+}
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -529,15 +529,32 @@ end
 # @example
 # --------
 
+# Find if there is any format with color output
+function _any_color_fmt(doc)
+    idx = findfirst(x -> x isa Documenter.HTML, doc.user.format)
+    idx === nothing && return false
+    return doc.user.format[idx].ansicolor
+end
+
 function Selectors.runner(::Type{ExampleBlocks}, x, page, doc)
+    matched = match(r"^@example(?:\s+([^\s;]+))?\s*(;.*)?$", x.language)
+    matched === nothing && error("invalid '@example' syntax: $(x.language)")
+    name, kwargs = matched.captures
     # The sandboxed module -- either a new one or a cached one from this page.
-    name = match(r"^@example[ ]?(.*)$", first(split(x.language, ';', limit = 2)))[1]
     mod = Utilities.get_sandbox_module!(page.globals.meta, "atexample", name)
     sym = nameof(mod)
     lines = Utilities.find_block_in_file(x.code, page.source)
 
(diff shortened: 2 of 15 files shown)


fredrikekre:
Nice! Is there ever a reason not to have this by default? Perhaps we can just look at the command line flag instead.

kimikage:
One problem with `IOCapture` is that `redirect_{stdout|stderr}(::IOContext)` is only available for Julia v1.6 (or later). (cf. https://github.com/JuliaLang/julia/pull/36688) 
Possible measures include:
- giving up the color display of `stdout`/`stderr` on v1.5 or earlier
- emulating `PipeEndpoint` as a subclass of `LibuvStream`
- pirating `get(::PipeEndpoint, key, default)`

Do you have any other good ideas?

fredrikekre:
> giving up the color display of stdout/stderr on v1.5 or earlier

I think this is fine.

kimikage:
>> giving up the color display of `stdout`/`stderr` on v1.5 or earlier
>
> I think this is fine.

I also think it's the safer of the three. Even if we don't offer the colored `stdout`/`stderr` as a feature of these packages, perhaps defining `get(::PipeEndpoint, key, default)` in the user code (i.e. "docs/make.jl"), would not be too much of a problem.

kimikage:
There is still a lot of work to be done, but the `stdout`/`stderr` and `@repl` blocks are now colored in HTML.
Also, although I haven't fully tested it yet, the interference with highlight.js (v10.3.1) doesn't seem to be a problem. I'm looking forward to the [improved highlighting](https://fredrikekre.se/posts/highlight-julia/). :heart_eyes:

stdout
repl

As suggested in https://github.com/JuliaDocs/Documenter.jl/pull/1441#issuecomment-711983438, instead of adding the `ansicolor` keyword argument to `makedocs`, the `:color` property of `stdout`, which is determined by the `--color` command line flag, is now used as the default value.

kimikage:
Currently, I added the `ansicolor` keyword argument to the `@example`/`@repl` block, but it might be better to generalize it to `IOContext`. (cf. issue #1450)
However, we will face the problem of `PipeEndpoint` again. :confused:

kimikage:
The system 16 colors are now generated from variables such as `$red` by default.

So, I changed the default color scheme while maintaining as much of the current look as possible. (I'm going to separate the color scheme changes into another PR.)

16colors

src/Writers/HTMLWriter.jl
```diff
@@ -1608,8 +1613,17 @@ mdconvert(b::Markdown.Bold, parent; kwargs...) = Tag(:strong)(mdconvert(b.text,
...
-    language = isempty(language) ? "none" : language
-    pre(code[".language-$(language)"](c.code))
+    class = isempty(language) ? ".nohighlight" : ".language-$(language)"
+    if language == "julia-repl"
+        input = IOBuffer(c.code)
+        printer = AnsiColoredPrinters.HTMLPrinter(input)
+        p = repr(MIME"text/html"(), printer)
+        # TODO: add support for specifying root tag in AnsiColoredPrinters
+        out = replace(replace(p, r"^<pre>\n" => "<code class=\"language-julia-repl\">"), r"/pre>$" => "/code>")
+        return pre(Tag(Symbol("#RAW#"))(out))
```
mortenpi, line 1623: Currently you get a `String` out of AnsiColoredPrinters, right? Could we return a `DOM` tree instead? AnsiColoredPrinters would probably need some API for user-defined printers though?

mortenpi (commented):
> I've put an ANSI escape code translator for HTML in a new package AnsiColoredPrinters.jl to make this feature work. The package is experimental and under development, and there are things left to be discussed before it is registered, including the name of the package. I won't refuse to contribute it as part of Documenter.jl, but I prefer to modularize the functionality into minimal packages.

I think the functionality is quite well-defined and general, so it make sense to have it as a separate package. Would you mind having that under the JuliaDocs umbrella though?

I would love to have you as its primary maintainer, but it would be good if there is a wider set of people who would can, if need be, release quick fixes, since the Julia master branch will be depending on this code.

assets/html/scss/documenter/_ansicolors.scss
```diff
@@ -0,0 +1,181 @@
...
+$ansi-light-black:  #888 !default;
+$ansi-light-red:    #d57 !default;
+$ansi-light-green:  #2a5 !default;
+$ansi-light-yellow: #d94 !default;
+$ansi-light-blue:   #08d !default;
+$ansi-light-magenta:#b8d !default;
+$ansi-light-cyan:   #0bc !default;
+$ansi-light-white:  #eee !default;
+
+span {
```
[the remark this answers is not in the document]
kimikage: Adding `documenter-ansi` is fine, but what is its primary purpose? In general, the looser the style specifications, the easier they are to customize (i.e. override).

What I'm concerned about is the condition on sticking `documenter-ansi`. By the new default, `ansicolor` is `true` in most cases. Do you want to make sure that the escape codes are actually included?

src/Writers/HTMLWriter.jl
kimikage: Returning a DOM tree is technically possible, but as you've noticed, there are two problems.
One is that the basic concept of AnsiColoredPrinters is the lazy printing using 3-arg `show`. Therefore, the output can be binary data instead of `String`, but it must be handled by `print`/`write`. That means we need to create some quite new API for the `DOM`.
The other is that `DOM` is a local type of Documenter.jl. If we use the general and portable `DOM` notation instead of the local type, it should be exactly "text/html". :stuck_out_tongue_closed_eyes:

Implementing a user-defined printer isn't hard if you use the private methods of AsiColoredPrinters, but refining the APIs to make them public is a bit of a hassle. :sweat_smile:

If there is a specific benefit to the authors and readers of the documents by expressing it in a DOM tree, I'd like to consider supporting it.

kimikage:
> Would you mind having that under the JuliaDocs umbrella though?

After making some miscellaneous changes (I think they will take about a week), I'd like to transfer the package to JuliaDocs.

However, before we do that, we need to decide on the package name. Even if it is based on the current name, I'd like to change "Ansi" to "ANSI".

assets/html/scss/documenter/_ansicolors.scss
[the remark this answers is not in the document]
mortenpi: It's just that the names `sgr1` etc. are very short and generic, so it feels like it might be a good idea to namespace them, to make sure it doesn't conflict with something in the future (but it's good to keep them short).

Actually, something like `.ansi` or `.ansicolor` would be a better class name.

src/Writers/HTMLWriter.jl
mortenpi: It would not be strictly necessary, but from a maintenance perspective, but it's probably a good idea if all the HTML generation happens in one place (in the `DOM` module). Also, once we get to `LaTeXWriter`, I think it would be good if Documenter would have full control over what LaTeX gets generated, in case we need to change that.

I haven't looked at the package very closely, but would it be possible to have the printer be an iterator that iterates over sections of characters that share the same styling or something along those lines? That way generating the correct array of `DOM` `<span>`s should be a matter of a few lines of code in Documenter.

mortenpi (commented):
> However, before we do that, we need to decide on the package name. Even if it is based on the current name, I'd like to change "Ansi" to "ANSI".

I also think that the capitalized name would look better.

src/Writers/HTMLWriter.jl
kimikage: The ANSI escape codes are a bit cumbersome to manage the stack in a tree model like HTML, because the start and end codes aren't necessarily one-to-one. Also, an HTML-specific issue is that the colors except the system 16 colors should be assigned in the `style` attribute directly.

So I'd like to add a mechanism to call a callback method "just before" writing an HTML tag. If we call `String(take!(io))` in the callback method, we should be able to get the text from the previous tag to the current position.

assets/html/scss/documenter/_ansicolors.scss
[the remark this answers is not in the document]
kimikage: `.ansi` is added to the `<code>`'s class now. However, it is also attached in the case of `ansicolor=false` for now.:sweat_smile:

src/Writers/HTMLWriter.jl
kimikage: The `HTMLPrinter` has a new `callback` option.
https://kimikage.github.io/ANSIColoredPrinters.jl/dev/output-formats/#html

So, the "text/plain" output is now DOMified (in a somewhat tricky way).

kimikage:
`ANSIColoredPrinters`' documentation is still unfriendly, but its code coverage is now over 90%. So I am going to apply for the transfer.
https://github.com/kimikage/ANSIColoredPrinters.jl/issues/10

fredrikekre:
I tried to revive this. I get errors from highlightjs though
```
One of your code blocks includes unescaped HTML. This is a potentially serious security risk. highlight.min.js:259:58
https://github.com/highlightjs/highlight.js/issues/2886
```
is that new or did you also see those @kimikage ?

kimikage:
Sorry I haven't been able to update this PR. I will try to `rebase` it in a few days.

fredrikekre:
Thanks for the rebase. Something seems wrong with the dark theme:
Screenshot from 2021-07-04 00-13-47

fredrikekre:
It also looks like hljs overrules things like
```
<code class="language-julia-repl ansi">julia&gt; printstyled("hello, world\n"; color=:blue)
<span class="sgr34">hello, world</span></code>
```
turning it into
```
<code class="language-julia-repl ansi hljs"><span class="hljs-meta">julia&gt;</span><span class="language-julia"> printstyled(<span class="hljs-string">"hello, world\n"</span>; color=:blue)
</span>hello, world</code>
```

fredrikekre:
For
````
```@repl
println("hello")
```
````
perhaps we need to emit something like
```
<pre>
<code "language-julia-repl">
println("hello")
</code>
<code "nohighlight">
hello
</code>
</pre>
```
i.e. split into two blocks where only the first is used with hljs

**Comment by fredrikekre** [comment]
I have some fixes for my last two comments which I can push alter. I think that this PR can be merged once the dark theme css is fixed and I will make a follow up PR:

**Comment by fredrikekre** [comment]
OMG in the test package I tried this on I had a custom css file overriding the documenter one.....

**Comment by kimikage** [comment]
I was considering the approach of writing a highlight.js plugin, but the plugin API does not work well with DOM manipulation.
For this reason, I think the approach proposed above of separating the REPL input and output is more elegant.

In this PR, I will postpone the ansi color support for `@repl` (comment out some code) and write documentation and tests.

**Comment by fredrikekre** [comment]
Sounds good

File: src/Expanders.jl [hunk 3]
```diff
@@ -530,14 +530,24 @@ end
...
-    name = match(r"^@example[ ]?(.*)$", first(split(x.language, ';', limit = 2)))[1]
     mod = Utilities.get_sandbox_module!(page.globals.meta, "atexample", name)
     sym = nameof(mod)
     lines = Utilities.find_block_in_file(x.code, page.source)
 
-    # "parse" keyword arguments to example (we only need to look for continued = true)
-    continued = occursin(r"continued\s*=\s*true", x.language)
+    # "parse" keyword arguments to example
+    continued = false
+    ansicolor = get(stdout, :color, false)::Bool
```
fredrikekre, line 543: I wonder if this is a useful check, does `stdout` on e.g. CI always give `true` here? Perhaps  `true` can just be default and if you don't want it you explicitly say that with `ansicolor=false`?

fredrikekre (commented):
.

src/Expanders.jl
fredrikekre: Actually, lets check `Base.JLOptions().color`, and if it is `2` then disable (e.g. if you started with `--color=no`)

fredrikekre:
```suggestion
    ansicolor = Base.JLOptions().color != 2 # disable if --color=no
```

src/Expanders.jl
```diff
@@ -597,10 +607,21 @@ end
...
-    matched = match(r"^@repl[ ]?(.*)$", x.language)
+    matched = match(r"^@repl(?:\s+([^\s;]+))?\s*(;.*)?$", x.language)
     matched === nothing && error("invalid '@repl' syntax: $(x.language)")
-    name = matched[1]
+    name, kwargs = matched.captures
+    # The sandboxed module -- either a new one or a cached one from this page.
     mod = Utilities.get_sandbox_module!(page.globals.meta, "atexample", name)
+
+    # "parse" keyword arguments to repl
+    ansicolor = get(stdout, :color, false)::Bool
```
fredrikekre, line 617:
```suggestion
    ansicolor = Base.JLOptions().color != 2 # disable if --color=no
```

mortenpi:
Thank you guys for working on this! Fredrik and I briefly discussed this on Slack how to proceed with this in terms of versions:

* We'll make this opt-in 0.27.4, to get this out to the users faster (to be toggled by new `ansicolor` keyword to `HTML`).
* We'll enable this by default in 0.28.0. So we'll have a follow-up PR that changes the default value (either now, or closer to the release). This is to avoid a large behavior change in a patch release.

Otherwise, if Fredrik is happy with this, then from my side this just needs a CHANGELOG entry.

fredrikekre:
I pushed those changes and added docs + changelog.

Issue #1413, opened 2020-09-06
Document how to generate docs locally


aviks:
Apologies if I've missed something, but I could not find, either by reading the Documenter docs, or googling, how to generate the docs locally while developing. The correct incantation is present in most travis configs, and was pointed out to me on Slack. Unless I have missed something obvious, I think it would be good to document the commands needed. They are not obvious, at least to me. At the very least, this issue can serve as a reminder to myself, since I had to look it up multiple times over the last months as I developed new packages. 

This is what I ended up doing, running from a shell. This presumes that the current directory is the package root, and that there exists a docs specific Project.toml. This project file must contain Documenter and the package in question as depdendencies. The following incantations can add the depedencies required, and generate the manifest

```
julia --project=docs/ -e 'using Pkg; Pkg.add("Documenter")'
julia --project=docs/ -e 'using Pkg; Pkg.add("MyPkg")'
julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));Pkg.instantiate()'
```

After that, the following incantation runs the doc generation. 

```
julia --project=docs/  docs/make.jl
```

mortenpi:
Yeah, this would be good to have in the manual somewhere. What I usually do is the following, if you have the standard setup with `docs/Project.toml`:

```
$ julia --project=docs/
pkg> instantiate
pkg> dev .
julia> include("docs/make.jl")
```

You can even drop a `using Revise` in there so that you wouldn't have to reload Julia when updating the docstrings.

aviks:
> What I usually do is the following

Yeah, that sounds reasonable. All of this is, of course, obvious in retrospect, but I've struggle with figuring this out, and seen other's do that.

haampie:
This came up on Slack too; the Documenter docs are making people run `julia make.jl` all the time.

haampie:
It would be nice to write a guide targeted for package developers who interactively write docs and view the result in their browser. I wrote a bit about how to do it from scratch, how to use Revise etc, but unfortunately Revise does not pick up docs for new function definitions it seems :( so I'm not sure if a Revise-first approach is that great

MichaelHatherly:
> unfortunately Revise does not pick up docs for new function definitions it seems :( so I'm not sure if a Revise-first approach is that great

Docstring updates work for me on Revise `3.1.4`, with either tracked packages of just using `includet`. (Everything should be Revise-first :smiley:)

[...]

PR #1451, opened 2020-10-22
Simplify the guide and focus on iterative package / docs development

haampie:
Solves #1413

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -2,67 +2,67 @@
 
 Documenter is designed to do one thing -- combine markdown files and inline docstrings from
 Julia's docsystem into a single inter-linked document. What follows is a step-by-step guide
-to creating a simple document.
+intended for users who iteratively develop a package and write documentation for it.
 
+## Preliminaries
 
-## Installation
-
-Documenter can be installed using the Julia package manager.
-From the Julia REPL, type `]` to enter the Pkg REPL mode and run
+As a package developer you want to keep a single Julia REPL session opened from which you
+generate your docs. To make Julia aware of changes to your code and docstrings, you need to
+install [Revise.jl](https://github.com/timholy/Revise.jl) first. Enter the Pkg REPL by
+pressing `]` and run:
 
 ```
-pkg> add Documenter
+(@v1.5) pkg> add Revise
+
+julia> using Revise
 ```
 
 
-## Setting up the Folder Structure
+## Setting up the Package Structure
 
 !!! note
     The function [`DocumenterTools.generate`](@ref) from the `DocumenterTools` package
     can generate the basic structure that Documenter expects.
 
-Firstly, we need a Julia module to document. This could be a package generated via
-`PkgDev.generate` or a single `.jl` script accessible via Julia's `LOAD_PATH`. For this
-guide we'll be using a package called `Example.jl` that has the following directory layout:
+Go back to the Pkg REPL and generate a new package:
+
+```julia
+(@v1.5) pkg> generate Example
+ Generating  project Example:
+    Example/Project.toml
+    Example/src/Example.jl
+```
+
+By default the generated Example.jl file contains a `greet()` function. We're going to add a
+docstring to it:
+
+```julia
+module Example
+
+"""
+    greet()
+
+Prints "Hello world!"
+"""
+greet() = print("Hello World!")
+
+end # module
+```
+
+Now create a `docs` folder within the package with the following structure:
 
 ```
 Example/
-├── src/
-│   └── Example.jl
-...
+├── docs
+│   ├── make.jl
+│   └── src
+│       └── index.md
+├── Project.toml
+└── src
+    └── Example.jl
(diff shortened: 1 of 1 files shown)


haampie:
This new guide is a bit sneaky, in the sense that Revise.jl only seems to pick up new docstrings when the newly created Example package contains at least one function with a docstring from the start. So instead of starting with 0 docstrings like in the original guide, a first step is to add a docstring to the `greet()` function.

docs/src/man/guide.md
```diff
@@ -2,134 +2,123 @@
 
 Documenter is designed to do one thing -- combine markdown files and inline docstrings from
 Julia's docsystem into a single inter-linked document. What follows is a step-by-step guide
-to creating a simple document.
+intended for users who iteratively develop a package and write documentation for it.
 
+## Preliminaries
```
mortenpi, line 8: I would add a paragraph here about why you'd want to use this more interactive workflow. E.g. something along these lines (but edits definitely welcome):

```suggestion

While the standard Documenter setup allows package documentation to be built on the command line by simply executing the `make.jl` script with e.g. `julia --project=docs/ docs/make.jl`, this is can be slow due to all the re-compilation happening in every new Julia session.

```

docs/src/man/guide.md
```diff
@@ -2,134 +2,123 @@
...
+As a package developer you want to keep a single Julia REPL session opened from which you
+generate your docs. To make Julia aware of changes to your code and docstrings, you need to
+install [Revise.jl](https://github.com/timholy/Revise.jl) first. Enter the Pkg REPL by
+pressing `]` and run:
 
 ```
-pkg> add Documenter
+(@v1.5) pkg> add Revise
+
+julia> using Revise
```
mortenpi, line 17: Might be good to warn here to make sure Revise gets added to the _global_ environment (even though it's implied by `(@1.5)>`), and not to either as a package dependency or in `docs/Project.toml`.

docs/src/man/guide.md
```diff
@@ -2,134 +2,123 @@
...
+(@v1.5) pkg> activate Example/docs
+ Activating new environment at `~/Documents/projects/Example/docs/Project.toml`
 
-Leave the newly added file empty and then run the following command from the `docs/` directory
+(docs) pkg> add Documenter
 
-```sh
-$ julia make.jl
+(docs) pkg> dev Example
+[ Info: Resolving package identifier `Example` as a directory at `~/Documents/projects/Example`.
```
mortenpi, line 94: This recommendation could be problematic, because it could clone a new package into `~/.julia/dev/Example.jl`, if you don't happen to be in the right directory? Perhaps better to use a relative path (`pkg> dev .` if in package root, or `pkg> dev ..` if under `docs/`) ?

docs/src/man/guide.md
```diff
@@ -2,134 +2,123 @@
...
+## Setting up the Package Structure
 
 !!! note
     The function [`DocumenterTools.generate`](@ref) from the `DocumenterTools` package
     can generate the basic structure that Documenter expects.
 
-Firstly, we need a Julia module to document. This could be a package generated via
-`PkgDev.generate` or a single `.jl` script accessible via Julia's `LOAD_PATH`. For this
-guide we'll be using a package called `Example.jl` that has the following directory layout:
+Go back to the Pkg REPL and generate a new package:
```
mortenpi, line 27: I think it would be really good to advertise PkgTemplates here.

docs/src/man/guide.md
```diff
@@ -2,134 +2,123 @@
...
-    ```
+```@docs
+Example.greet()
+```
+````
 
-Now add an `index.md` file to the `src/` directory.
+Since Documenter is a development tool and not a true dependency of your package, create
+a fresh environment in the `docs/` folder, and add Documenter and the Example package
+itself as dependencies:
```
mortenpi, line 85: Perhaps also recommend adding a `[compat]` for Documenter into `docs/Project.toml`?

docs/src/man/guide.md
```diff
@@ -177,61 +166,6 @@ build/
...
-```@meta
-CurrentModule = Example
-```
-
-```@docs
-func(x)
-```
-````
-
 ### Filtering included docstrings
```
mortenpi, line 169: I think heading should be "promoted" now, since `## Adding Some Docstrings` was removed:
```suggestion
## Filtering included docstrings
```

mortenpi (commented):
Thanks a lot @haampie! This is great -- this page really needs some updating. And sorry for a delayed review.

I wonder if it would be good to have a really quickstart section for a user who needs to edit the documentation of an existing package (so omit the setup, just focus on calling Documenter)?

knuesel:
It would be nice to get this merged... It would also fix #1500. @haampie are you still interested in working on this PR? Or would you prefer for someone else to take it over?

mortenpi:
I'll close this since it's stale, but hopefully we can re-use bits from here as part of #2321.

PR #1452, opened 2020-10-24
Empty Selector cache before running build pipeline

### LINKED ISSUE #1438: Heavy slowdown in ExpandTemplates due to `subtypes` function

mortenpi:
Small follow-up to #1440, to make sure that we would retain the feature where you can add selector stages between `makedocs` calls.

Just to make it concrete, with the current master the `DummyBuildStage` warning does not get printed if you run the following (but does work pre-#1440):

```julia
using Documenter: Documenter, makedocs, Selectors
makedocs(sitename = "foo")
abstract type DummyBuildStage <: Documenter.Builder.DocumentPipeline end
Selectors.order(::Type{DummyBuildStage}) = 2.5
Selectors.runner(::Type{DummyBuildStage}, args...) = @warn "DummyBuildStage"
makedocs(sitename = "foo")
```

@epatters I believe this should not affect the Catlab build times?

Close #1438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * ![Bugfix][badge-bugfix] `Documenter.doctest` now correctly accepts the `doctestfilters` keyword, similar to `Documenter.makedocs`. ([#1364][github-1364], [#1435][github-1435])
 
+* ![Bugfix][badge-bugfix] The `Selectors.dispatch` function now uses a cache to avoid calling `subtypes` on selectors multiple times during a `makedocs` call to avoid slowdowns due to [`subtypes` being slow][julia-38079]. ([#1438][github-1438], [#1440][github-1440], [#1452][github-1452])
+
 ## Version `v0.25.2`
 
 * ![Deprecation][badge-deprecation] The `Documenter.MathJax` type, used to specify the mathematics rendering engine in the HTML output, is now deprecated in favor of `Documenter.MathJax2`. ([#1362][github-1362], [#1367][github-1367])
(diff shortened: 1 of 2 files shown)


fredrikekre (approved):
I had written a comment suggesting this already in the original PR, but deleted it since it was easier to just press the button :stuck_out_tongue:

mortenpi:
> but deleted it since it was easier to just press the button

Somehow I immediately thought of this  https://youtu.be/1OitCT1KT4k?t=158 (from 2:38) :laughing:

epatters:
Thanks, can confirm that this does not affect the Catlab build time.

PR #1453, opened 2020-10-25
Do not capture interrupts with iocapture()

mortenpi:
Fixes a behaviour change introduced in #1422.

### DIFF
--- a/src/DocTests.jl
+++ b/src/DocTests.jl
@@ -217,7 +217,7 @@ function eval_repl(block, sandbox, meta::Dict, doc::Documents.Document, page)
                 # see https://github.com/JuliaLang/julia/pull/33864
                 ex = REPL.softscope!(ex)
             end
-            c = IOCapture.iocapture(throwerrors=false) do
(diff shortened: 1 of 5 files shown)


src/DocTests.jl
```diff
@@ -217,7 +217,7 @@ function eval_repl(block, sandbox, meta::Dict, doc::Documents.Document, page)
                 # see https://github.com/JuliaLang/julia/pull/33864
                 ex = REPL.softscope!(ex)
             end
-            c = IOCapture.iocapture(throwerrors=false) do
+            c = IOCapture.iocapture(throwerrors = :interrupt) do
```
fredrikekre, line 220: Unrelated to this PR; but perhaps the interface would be nicer if you passed an exception type
```julia
IOCapture.iocapture(throwerrors = InterruptException) do
    # ...
end
```

mortenpi: I think it's worth considering. Could you open an issue at IOCapture?

Issue #1055, opened 2019-07-02
Feature Request: copy to clipboard


Roger-luo:
I'm wondering if it is possible to enable this automatically for all code blocks in documentation, which make it much easier to try out examples.

mortenpi:
If you mean having a one-click widget to copy a whole code block to the clipboard, then that's something I was considering implementing in the new front end (#1043).

Roger-luo:
Oh, I just searched the issue didn't find it, looking forward to that!

mortenpi:
No issue specific to this feature, so let's leave this open until it's actually implemented :slightly_smiling_face:

PR #1454, opened 2020-10-27
Add copy to clipboard button for code snippets

abhishalya:
Preview:

image

Closes https://github.com/JuliaDocs/Documenter.jl/issues/1055


fredrikekre:
Nice!

Is it possible to get actual newlines instead of the character `\n` somehow? This is what I get after copying a codeblock with
```
1 + 1
2 + 2
```
```
julia> 1 + 1\n2 + 2
```

abhishalya:
Yeah, I'm trying to figure out a way to do it. Is there anything we can do from Julia side? The other option would be to handle via javascript.

fredrikekre:
It would also be nice to strip `# output` from the text in e.g.
````
```jldoctest
1 + 1
2 + 2

# output

4
```
````

**Comment by mortenpi** [comment]
For the newlines, I think we should modify the HTML generation in `DOM` so that newlines in attributes are replaced by `&#10;`. That way we don't get any weird indentation problems.

> It would also be nice to strip `# output` from the text in e.g.

As long as we render it as a single `<pre>` block, I think I would find it surprising if the copy-to-clipboard doesn't copy the content exactly.

File: src/Writers/HTMLWriter.jl [hunk 1]
```diff
@@ -1606,10 +1620,14 @@ mdconvert(b::Markdown.BlockQuote, parent; kwargs...) = Tag(:blockquote)(mdconver
 mdconvert(b::Markdown.Bold, parent; kwargs...) = Tag(:strong)(mdconvert(b.text, parent; kwargs...))
 
 function mdconvert(c::Markdown.Code, parent::MDBlockContext; kwargs...)
-    @tags pre code
+    @tags pre code button
     language = Utilities.codelang(c.language)
     language = isempty(language) ? "none" : language
-    pre(code[".language-$(language)"](c.code))
+    code_block = [
+        button[".copy-button .button .fas .fa-copy", Symbol("data-clipboard-text")=>c.code],
```
mortenpi, line 1627: Might be nice to have a `title` attribute on each button as well that would probably say "Copy to clipboard".

fredrikekre:
> As long as we render it as a single <pre> block, I think I would find it surprising if the copy-to-clipboard doesn't copy the content exactly.

Valid point, but it kind of defeats the purpose of having the button if the result is not pasteable into a Julia REPL? But perhaps it would be nice to render as two blocks instead, the `#output` in the final html is not very nice.

mortenpi:
> But perhaps it would be nice to render as two blocks instead, the `#output` in the final html is not very nice.

I think that's worth considering (#1455), but let's leave it for a separate PR.

fredrikekre:
It would be nice to also show `Copied!` or similar when you click the button, see e.g. the copy-buttons on https://doc.rust-lang.org/book/ch01-02-hello-world.html

assets/html/scss/documenter/components/_clipboard.scss
```diff
@@ -0,0 +1,3 @@
+.copy-button {
+  float: right;
```
mortenpi, line 2: This should make sure that it's the same distance from top and right side of the box:

```suggestion
  float: right;
  top: -0.2rem;
```

assets/html/scss/documenter/components/_clipboard.scss
```diff
@@ -0,0 +1,3 @@
...
+}
```
mortenpi, line 3: Perhaps it would be good to make it a bit smaller?

```suggestion
  font-size: 0.9em;
}
```

I.e. this

smaller

instead of this

bigger

mortenpi (commented):
Just a bunch of bikeshedding thoughts:

1. I think an arrow-pointing-to-clipboard icon (e.g. like they use on the clipboard.js website) would be better, but I couldn't actually find one in FontAwesome, so we can probably stick with the current one.

2. We could make the button hide itself if you're not hovering over the `<pre>` box (like we do with the "Source" links on docstrings), so it would not distract when not necessary.

3. How hard do you think would it be to make the "Copied" text appear next to the button (e.g. in a bubble, like on the clipboard.js site), as opposed to replacing the icon inside the button? It's just the button changing size does not look the prettiest in my opinion.

mortenpi:
Oh, one more thing, would you mind adding an entry into the [CHANGELOG.md](https://github.com/JuliaDocs/Documenter.jl/blob/master/CHANGELOG.md) file as well about this PR?

src/Utilities/DOM.jl
```diff
@@ -292,6 +292,9 @@ function escapehtml(text::AbstractString)
             char === '"'  ? write(buffer, "&quot;") : write(buffer, char)
         end
         String(take!(buffer))
+    end
+    if occursin("\n", text)
+        replace(text, "\n"=>"&#10;")
```
mortenpi, line 297: This should go into the `if`/`for` above. Currently it's actually breaking the escaping (the top `if` never returns from the function).

abhishalya: Yeah, I realised that. The problem here though is that if I replace it there, the code block doesn't render properly. I believe highlight.js needs those "/n"

mortenpi: Right. Should be okay if we only escape `\n`-s in attributes? I think we could implement that by adding a keyword argument to `escapehtml` and toggling that wherever we're escaping the attribute contents?

abhishalya:
@mortenpi  Have a look at this once you're free, I've tried to address as many comments as I could. Do let me know what you think about this :)

src/Utilities/DOM.jl
```diff
@@ -248,7 +248,11 @@ function Base.show(io::IO, n::Node)
         print(io, '<', n.name)
         for (name, value) in n.attributes
             print(io, ' ', name)
-            isempty(value) || print(io, '=', repr(escapehtml(value)))
+            if name === Symbol("data-clipboard-text")
+                isempty(value) || print(io, '=', repr(escapehtml(value, escape_new_lines=true)))
+            else
+                isempty(value) || print(io, '=', repr(escapehtml(value)))
+            end
```
mortenpi, line 255: `DOM` is an independent generic module, so its behavior shouldn't really depend on arbitrary argument names. I think we can just always escape newlines in attributes? Based on my reading of the standard, having [character references in attribute values should always be fine](https://html.spec.whatwg.org/#attributes-2).

src/Utilities/DOM.jl
```diff
@@ -248,7 +248,11 @@ function Base.show(io::IO, n::Node)
         print(io, '<', n.name)
         for (name, value) in n.attributes
             print(io, ' ', name)
-            isempty(value) || print(io, '=', repr(escapehtml(value)))
+            if name === Symbol("data-clipboard-text")
+                isempty(value) || print(io, '=', repr(escapehtml(value, escape_new_lines=true)))
```
mortenpi, line 252:
```suggestion
                isempty(value) || print(io, '=', repr(escapehtml(value, escape_newlines=true)))
```

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * ![Enhancement][badge-enhancement] Objects that render as equations and whose `text/latex` representations are wrapped in display equation delimiters `\[ ... \]` or `$$ ... $$` are now handled correctly in the HTML output. ([#1278][github-1278], [#1283][github-1283], [#1426][github-1426])
 
+* ![Enhancement][badge-enhancement] Code blocks now support copy to clipboard button to quickly run the code snippet in Julia REPL. ([#1454](github-1454))
```
mortenpi, line 7: We usually add the original issue reference(s) as well:

```suggestion
* ![Enhancement][badge-enhancement] Code blocks now support copy to clipboard button to quickly run the code snippet in Julia REPL. ([#1055](github-1055), [#1454](github-1454))
```

Will also need a `[github-1055]: https://github.com/JuliaDocs/Documenter.jl/issues/1055` in the link definitions section.

mortenpi (commented):
This is looking really good!

Just one more CSS gripe: it looks like the "Copied" message box ends up being _inside_ the `<pre>` box, so it's (1) partly cut off, and (2) creates a scrollbar on the code block when it appears:

copied

mortenpi (commented):
@abhishalya I'm really sorry for not reviewing earlier -- I completely missed that you updated the PR.

Besides the escaping thing, the only other issue I can still see is the "Copied" message creating scrollbars, instead of being of above everything. I tried to find a way around it, but I could quite figure it out. It feels like the solution [in this post](https://css-tricks.com/popping-hidden-overflow/) might be applicable.

Issue #1320, opened 2020-05-13
Respect users' preference for light/dark mode


cormullion:
Coming soon (or even already here) to a browser near you, a way of choosing to display light or dark mode:

https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme

Is this something that can be added to Documenter?

mortenpi:
Documenter's theme switching is JS-based, but that shouldn't be a problem, since it seems that you can [access that media feature via JS](https://web.dev/prefers-color-scheme/#loading-strategy) too. Would probably be a rather simple modification to [`themeswap.js`](https://github.com/JuliaDocs/Documenter.jl/blob/master/assets/html/themeswap.js).

PR #1456, opened 2020-10-29
Respect user preference for color scheme

abhishalya:
Automatically switches to 'non-primary' theme if user prefers dark scheme.
If there is a theme specified in the local storage then that takes more preference over this.

Closes https://github.com/JuliaDocs/Documenter.jl/issues/1320

### DIFF
--- a/assets/html/themeswap.js
+++ b/assets/html/themeswap.js
@@ -7,11 +7,9 @@ function set_theme_from_local_storage() {
   var theme =  window.localStorage.getItem("documenter-theme");
   // Check if the browser supports user color preference
   var darkPreference = false;
-  if(window.matchMedia('(prefers-color-scheme: dark)').media !== 'not all') {
-    // Check if the users preference is for dark color scheme
-    if(window.matchMedia('(prefers-color-scheme: dark)').matches === true) {
-      darkPreference = true;
-    }
+  // Check if the users preference is for dark color scheme
+  if(window.matchMedia('(prefers-color-scheme: dark)').matches === true) {
+    darkPreference = true;
   }
   // Initialize a few variables for the loop:
   //
@@ -33,8 +31,10 @@ function set_theme_from_local_storage() {
     // To distinguish the default (primary) theme, it needs to have the data-theme-primary
     // attribute set.
     var isprimary = (ss.ownerNode.getAttribute("data-theme-primary") !== null);
-    // If the ss is not primary then use it if the user's preference is towards dark colors
-    if(!isprimary) darkTheme = themename;
+    // Check if the theme is primary dark theme
+    var isDarkTheme = (ss.ownerNode.getAttribute("dark-theme-primary") !== null);
+    // Use ss is for dark theme then set the value of darkTheme
+    if(isDarkTheme) darkTheme = themename;
     // If we find a matching theme (and it's not the default), we'll set active to non-null
     if(themename === theme) active = i;
     // Store the style sheets of inactive themes so that we could disable them
@@ -48,7 +48,7 @@ function set_theme_from_local_storage() {
       ss.disabled = true;
     });
   }
-  else if(darkTheme !== null) {
+  else if(isDarkTheme !== null && darkPreference === true) {
     // If we did find an active theme, we'll (1) add the theme--$(theme) class to <html>
     document.getElementsByTagName('html')[0].className = "theme--" + darkTheme;
     // and (2) disable all the other theme stylesheets
@@ -57,12 +57,6 @@ function set_theme_from_local_storage() {
         ss.disabled = true;
       }
     });
-    // Store the theme in localStorage
-    if(typeof(window.localStorage) !== "undefined") {
-      window.localStorage.setItem("documenter-theme", darkTheme);
-    } else {
-      console.error("Browser does not support window.localStorage");
-    }
   }
 }
 set_theme_from_local_storage();
--- a/assets/html/themeswap.js
+++ b/assets/html/themeswap.js
@@ -1,10 +1,14 @@
 // Small function to quickly swap out themes. Gets put into the <head> tag..
 function set_theme_from_local_storage() {
-  // Browser does not support Web Storage, bail early.
-  if(typeof(window.localStorage) === "undefined") return;
-  // Get the user-picked theme from localStorage. May be `null`, which means the default
-  // theme.
-  var theme =  window.localStorage.getItem("documenter-theme");
+  // Intialize the theme to null, which means default
+  var theme = null;
+  // If the browser supports the localstorage and is not disabled then try to get the
+  // documenter theme
+  if(typeof(window.localStorage) != null) {
+    // Get the user-picked theme from localStorage. May be `null`, which means the default
+    // theme.
+    theme =  window.localStorage.getItem("documenter-theme");
+  }
   // Check if the browser supports user color preference
   var darkPreference = false;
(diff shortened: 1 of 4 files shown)


assets/html/themeswap.js
```diff
@@ -5,6 +5,14 @@ function set_theme_from_local_storage() {
   // Get the user-picked theme from localStorage. May be `null`, which means the default
   // theme.
   var theme =  window.localStorage.getItem("documenter-theme");
+  // Check if the browser supports user color preference
+  var darkPreference = false;
+  if(window.matchMedia('(prefers-color-scheme: dark)').media !== 'not all') {
```
mortenpi, line 10: Is it actually necessary to check this? I assume it's here for browsers that do not support that query? It _looks_ like that even if you give `matchMedia` a rubbish query, it will still return a sane object with `matches === false`. So it might be sufficient to just read `window.matchMedia('(prefers-color-scheme: dark)').matches`, and if that's `false`, assume we should show a light theme by default. But I might be wrong -- I am not at all familiar with this API.

assets/html/themeswap.js
```diff
@@ -25,8 +33,10 @@ function set_theme_from_local_storage() {
     // To distinguish the default (primary) theme, it needs to have the data-theme-primary
     // attribute set.
     var isprimary = (ss.ownerNode.getAttribute("data-theme-primary") !== null);
+    // If the ss is not primary then use it if the user's preference is towards dark colors
+    if(!isprimary) darkTheme = themename;
```
mortenpi, line 37: While not a big issue currently, it is possible that you'd have multiple light and dark themes, so this could pick the wrong theme. Maybe we should attach another attribute to mark a primary dark theme or something like that?

assets/html/themeswap.js
```diff
@@ -38,5 +48,21 @@ function set_theme_from_local_storage() {
...
+      if (ss.ownerNode.getAttribute("data-theme-name") !== darkTheme) {
+        ss.disabled = true;
+      }
+    });
+    // Store the theme in localStorage
+    if(typeof(window.localStorage) !== "undefined") {
+      window.localStorage.setItem("documenter-theme", darkTheme);
+    } else {
+      console.error("Browser does not support window.localStorage");
+    }
```
mortenpi, line 65: If I understand this correctly, this will, if you prefer a dark theme, will then store that dark theme as the preferred theme overall? This would mean that if the user disables the dark-theme on the OS level, docs would still be rendered with the dark theme, right? I think we don't want to do this -- it would be better if the theme would change with the user's OS settings (unless the user has explicitly requested another theme).

mortenpi (commented):
Currently it looks like it always defaults to the dark theme now, even though
```
> window.matchMedia('(prefers-color-scheme: dark)')
MediaQueryList {media: "(prefers-color-scheme: dark)", matches: false, onchange: null}
```
So it feels like something is broken here.

abhishalya:
Tried to address all above comments, also confirmed now that it works as it is supposed to.

assets/html/themeswap.js
```diff
@@ -25,8 +31,12 @@ function set_theme_from_local_storage() {
     // To distinguish the default (primary) theme, it needs to have the data-theme-primary
     // attribute set.
     var isprimary = (ss.ownerNode.getAttribute("data-theme-primary") !== null);
+    // Check if the theme is primary dark theme
+    var isDarkTheme = (ss.ownerNode.getAttribute("dark-theme-primary") !== null);
+    // Use ss is for dark theme then set the value of darkTheme
+    if(isDarkTheme) darkTheme = themename;
     // If we find a matching theme (and it's not the default), we'll set active to non-null
-    if(!isprimary && themename === theme) active = i;
+    if(themename === theme) active = i;
```
mortenpi, line 39: I think keeping the `isprimary` check here makes sense, because we don't want to mess around with applying class names when it's the primary light theme. Unless I am missing something?

```suggestion
    if(!isprimary && themename === theme) active = i;
```

assets/html/themeswap.js
```diff
@@ -38,5 +48,15 @@ function set_theme_from_local_storage() {
       ss.disabled = true;
     });
   }
+  else if(isDarkTheme !== null && darkPreference === true) {
```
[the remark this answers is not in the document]
mortenpi: Did you mean to check `darkTheme` here by the way? Because `isDarkTheme` will never be `null`.

mortenpi (commented):
Awesome, seems to work nicely now! A few more things that popped up as I was playing around with this:

1. There is the `if(typeof(window.localStorage) === "undefined") return;` at the beginning of `set_theme_from_local_storage()`.

   This will mean that in browsers where `window.localStorage` is `undefined`, but that prefer the dark theme, the theme will not be changed. So I think this should now be removed and the code below updated to make sure we handle the case where there is no `localStorage`.

   I also discovered that `window.localStorage` can be `null` (e.g. if you [disable `localStorage`](https://stackoverflow.com/a/6600933/1601695)). So I also think we should check [with `window.localStorage == null`](https://stackoverflow.com/a/2647888/1601695) instead.

2. The [theme picker in the settings dialog](https://github.com/JuliaDocs/Documenter.jl/blob/master/assets/html/js/themes.js) currently still shows "documenter-light" if the dark theme is selected by the browser. So we should update that code as well.

assets/html/themeswap.js
abhishalya: The problem here is if I prefer dark color scheme but I want to keep light theme for the site then I basically set the theme manually in the local storage. Since the theme is primary we ignore this `if` condition and then it moves on and basically uses my preference of dark scheme. We somehow need an override over the color preference in case of primary theme.

assets/html/themeswap.js
```diff
@@ -1,10 +1,20 @@
...
-  // Browser does not support Web Storage, bail early.
-  if(typeof(window.localStorage) === "undefined") return;
-  // Get the user-picked theme from localStorage. May be `null`, which means the default
-  // theme.
-  var theme =  window.localStorage.getItem("documenter-theme");
+  // Intialize the theme to null, which means default
+  var theme = null;
+  // If the browser supports the localstorage and is not disabled then try to get the
+  // documenter theme
+  if(typeof(window.localStorage) != null) {
```
mortenpi, line 7: I'm pretty sure we need to check the value, not the type here (`typeof(null) == "object"`):

```suggestion
  if(window.localStorage != null) {
```

assets/html/themeswap.js
mortenpi: Ah, you're right, never mind.

mortenpi (commented):
It looks like I can push to this branch myself. Could you also bring the branch up to date with master and cherry-pick this CHANELOG commit: https://github.com/JuliaDocs/Documenter.jl/commit/2751b43bfc4435248ed995b0997968d25e46267d.

Other than this and the one code suggestion, I think this is good to go :slightly_smiling_face:

mortenpi:
Could you merge master into this again? Or tick the "allow maintainers to push to this branch"? I can't merge this at the moment, because `CHANGELOG.md` conflicts.

abhishalya:
>Or tick the "allow maintainers to push to this branch"?

Sorry, since this is created from the MLH fork, I wasn't able to do that. Although I've asked Cory to give you access to the MLH repos, so you should be able to push directly :)

I've rebased the branch with master for now.

mortenpi:
Thanks @abhishalya!

PR #1461, opened 2020-11-03
docs: cleaning up PR previews

ericphanson:
I thought it would be helpful to mention @simonbyrne's `DocCleanup` script somewhere. I'm not totally sure this is the best place, but `push_preview` is mostly documented in the `deploydocs` docstring, and that didn't seem like the right place.

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -381,11 +381,12 @@ You also need to make sure that you have "gh-pages branch" selected as
 settings](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site),
 so that GitHub would actually serve the contents as a website.
 
-Note that the `gh-pages` branch can become very large, especially when `push_preview` is enabled to
-build documentation for each pull request. To cleanup the branch and remove stale documentation
-previews, a GitHub Actions workflow like the following can be used.
+**Cleaning up `gh-pages`.**
+Note that the `gh-pages` branch can become very large, especially when `push_preview` is
+enabled to build documentation for each pull request. To clean up the branch and remove
+stale documentation previews, a GitHub Actions workflow like the following can be used.
 
-```
+```yaml
 name: Doc Preview Cleanup
 
 on:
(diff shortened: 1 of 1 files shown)


docs/src/man/hosting.md
```diff
@@ -381,6 +381,12 @@ You also need to make sure that you have "gh-pages branch" selected as
 settings](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site),
 so that GitHub would actually serve the contents as a website.
 
+Note that the `gh-pages` branch can become very large, especially when `push_preview` is enabled to
+build documentation for each pull request. To cleanup the branch and remove stale documentation
+previews, a GitHub Actions workflow like
+[DocCleanup.yml](https://github.com/CliMA/TimeMachine.jl/blob/4d951f814b5b25cd2d13fd7a9f9878e75d0089d1/.github/workflows/DocCleanup.yml)
```
johnnychen94, line 387: Good to know this. Curious to ask, is this applicable to all repos even if `push_preview` is disabled?

It might be better to explicitly copy the action yml code rather than a URL pointer. Also, it might be better to move to https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions.

ericphanson: the `gh-pages` branch grows unbounded, AFAIK, but in general that's for good reason, since you want to keep old versions of the docs (tagged appropriately). But with `push_preview`, the docs it creates are stale as soon as the PR is merged-- and there tends to be way more of them, since generally there are more PRs than minor version releases. 

The URL is a permalink, so it should not go stale (unless the git history is rewritten, I guess), and this way you can see if there are updates to the script (by looking forward in the history). It also gives credit appropriately. I could copy it over though if you think that's much preferred.

I thought about putting it under the GitHub Actions section, but it seems like that section is specifically for building docs via GitHub Actions. You can use this workflow though whether or not you use GitHub Actions to build the docs (as long as its hosted on GitHub using the `gh-pages` branch).

mortenpi: It might be useful when the `master` is very active, as a new version of the `dev/` docs gets deployed with each commit and they all get retained in the Git history.

I also think it would make it slightly easier for the user if the workflow is directly in the documentation, but I also agree that it's unlikely that the permalink will break any time soon.

It is definitely in the correct section though.

mortenpi (approved):

ericphanson:
Ok, I've moved the workflow inline. Hopefully the attribution is OK (cc @simonbyrne).

> It might be useful when the master is very active, as a new version of the dev/ docs gets deployed with each commit and they all get retained in the Git history.

Ah, that does sound useful. I think the current version of this workflow just removes the previews though?

mortenpi:
Made a few tiny edits, but LGTM now. Thanks @ericphanson and @simonbyrne! Simon, just ping me if you see any issues with the attribution.

ericphanson:
Not sure where the right place to note this is, but the script errors if the PR in question did not have a PR preview generated, for example https://github.com/beacon-biosignals/KeywordSearch.jl/actions/runs/425988124. At KeywordSearch, we use `paths` specifier to only build docs when something that affects the docs changes, and thus some PRs don't have previews.

mortenpi:
Small `if [ -d ... ]` around the last three `git` commands should be sufficient to get around that I think?

Issue #566, opened 2017-09-26
Multiple identical footnotes


eveydee:
Right now, Documenter errors when multiple bodies for a footnote is encountered. This is problematic, when, for instance, rendering docstrings, where each docstring may contain identical references to the same article with identically labelled footnotes. Manually relabelling these for Documenter to work fixes it, but it would be better if the footnotes were either collected at the bottom, or the anchor tag to each footnote was given a unique reference based on its section/function name.

This came up in: https://github.com/JuliaLang/julia/pull/23679

mortenpi:
I would lean towards making the footnotes in each docstring be a separate group, as opposed to collecting them into a single catalogue as is done now. It would mean that you couldn't refer to footnotes defined in other docstrings, nor refer to a non-docstring footnote in within a docstring anymore. As you say, the anchor would have to be updated to have a docstring-specific prefix.

Alternatively, we could also do some more sophisticated scoping with the footnotes, but I am not sure whether the complexity would be worth it.

eveydee:
Referring to a footnote in another docstring probably not good practice, unless they would automatically be pulled into the docstring. `help? >` would just show a reference to a footnote that the user would have no idea what was. Even then, Documenter would also fail if the referenced footnote did not appear on the same page in the rendered documentation, as far as I can tell.

PR #1464, opened 2020-11-07
Remove duplicate footnotes for HTML pages

kahaaga:
## Issue 

Identical footnote IDs appearing multiple times in a docstring, or in different docstrings that are collected on the same HTML page, are rendered multiple times in the footnote list at the bottom of the HTML page. 

### Example

In [this documentation page](https://kahaaga.github.io/Entropies.jl/dev/generalized_entropy/), the footnote `[^Rényi1960]` appears four times in different methods.  This is caught by the doc checks:

```julia
Warning: footnote 'Rényi1960' has 4 bodies in src/generalized_entropy.md
```

but in the footnote list in the rendered HTML, duplicates are not handled:

Skjermbilde 2020-11-07 kl  13 52 59

Multiple docstrings that reference the same footnote should be possible to render on the same page without filling the footnote list with duplicates.

## Solution
If there are any footnotes, this PR adds a simple check for duplicate footnote IDs. Any footnotes with duplicate IDs are removed. 

After this PR, the example above renders as follows.
Skjermbilde 2020-11-07 kl  13 52 55

When running the doc checks, warnings about duplicate footnotes are still given, but any duplicates are removed from the final HTML.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1158,6 +1158,17 @@ end
 
 # Article (page contents)
 # ------------------------------------------------------------------------------
+function unique_footnotes(footnotes)
+    unique_ids = Int[]
+    unique_footnotes = String[]
+    for (i, fn) in enumerate(footnotes)
+        if fn.id ∉ unique_footnotes
+            push!(unique_ids, i)
+            push!(unique_footnotes, fn.id)
+        end
+    end
+    return footnotes[unique_ids]
+end
 
 function render_article(ctx, navnode)
     @tags article section ul li hr span a div p
@@ -1167,7 +1178,8 @@ function render_article(ctx, navnode)
     art_body = article["#documenter-page.content"](domify(ctx, navnode))
     # Footnotes, if there are any
     if !isempty(ctx.footnotes)
-        fnotes = map(ctx.footnotes) do f
+        # remove duplicate footnotes (if there are any duplicates), then process them
+        fnotes = map(unique_footnotes(ctx.footnotes)) do f
             fid = "footnote-$(f.id)"
             citerefid = "citeref-$(f.id)"
             if length(f.text) == 1 && first(f.text) isa Markdown.Paragraph


mortenpi:
I am not entirely sure this is the right way to go -- the footnote bodies can have different content. So really this is a situation where the user should sort out the footnotes.

Although it may not be the user's fault entirely.. it looks like this issue arises when you have the same reference in multiple footnotes? I would advocate for the idea in https://github.com/JuliaDocs/Documenter.jl/issues/566#issuecomment-332439396 to resolve this. It does mean that we'd have to attach the footnote bodies to each docstring separately.

As long as we don't have that, we could also have a stopgap solution:

1. Merge the footnotes together _iff_ their bodies are identical.
2. Make sure that the footnotes that have identical labels (but different bodies) are listed one after the other at the bottom of the page.

kahaaga:
> Although it may not be the user's fault entirely.. it looks like this issue arises when you have the same reference in multiple footnotes?

Precisely. Below is a use-case that warrants repeated references to the same footnote. 

> I would advocate for the idea in #566 (comment) to resolve this. It does mean that we'd have to attach the footnote bodies to each docstring separately.

Having docstring-specific footnote bodies listed at the bottom of each docstring would be very helpful. However, even if footnote bodies are attached to docstrings separately:

- Several references to the same footnote might still appear within the same docstring. This is a valid user case, and should not produce duplicate footnotes listed, even if the listing appears within each docstring. 
- Footnotes might also appear on the same page as the docstrings are listed that are not in any of the docstrings. These should be rendered at the bottom of the page (as they are now). 

## Updated PR

I updated the PR according to your proposal: 

1. __Footnotes are merged _iff_ their bodies are identical__. For now, identical bodies means equal string representations of those bodies. This could probably be done more intelligently, but works for non-complicated cases (see example below).
2. __Footnotes with identical IDs but different bodies are listed after one after the other__. This takes care of duplicates due to repeated reference to the same footnote listing (which is a valid use case). But it is still an ugly solution, because it is not clear which of the footnote references in the docstrings/text refer to which of the listed footnote bodies (see example below). One solution would be to number them (a), (c), and so on, both in the docstrings/text and in the footnote listing at the bottom of the page. Another solution, and perhaps the better one, is just to disallow repeated footnote IDs entirely. 

This approach can also be used to handle duplicate method-specific footnotes if the idea in #566 is implemented at some point in the future.

## Example 

Let's say I have a file `some_functions.jl` in my package that looks like this:

```julia
""" 
    fancy_algorithmA(x::Int)

This algorithm is based on publication A[^Rényi1960].

[^Rényi1960]: A. Rényi, Proceedings of the fourth Berkeley Symposium on Mathematics, Statistics and Probability, pp 547 (1960)
"""
function fancy_algorithmA(x::Int) end

""" 
    fancy_algorithmB(x::Int)

This algorithm is based on publication B[^Shannon1948], but contains some stuff 
from publication A[^Rényi1960] too. Also, it might be necessary to refer to 
publication B again within the same docstring[^Shannon1948]. This should not 
produce duplicates in the footnote list, even if footnotes are bound to 
docstrings for specific methods.

[^Shannon1948]: C. E. Shannon, Bell Systems Technical Journal 27, pp 379 (1948)
[^Rényi1960]: A. Rényi, Proceedings of the fourth Berkeley Symposium on Mathematics, Statistics and Probability, pp 547 (1960)
"""
function fancy_algorithmB(x::Int) end

""" 
    fancy_algorithmC(x::Int)

This is some variant of [^Rényi1960] where the same reference is used, but the 
body is different. 

[^Rényi1960]: A specific note from Rényi (1960). 
"""
function fancy_algorithmC(x::Int) end

export fancy_algorithmA, fancy_algorithmB, fancy_algorithmC
```

Then I have a `test.md` page that looks as follows:

```text
    # Duplicate references test

    The following docstrings can come from different modules: 

    ```@docs
    fancy_algorithmA
    fancy_algorithmB
    ```

    The documentation pages may further elaborate on the algorithms, and for that, publication
    A[^Rényi1960] might be referred to again again. The same goes for publication B [^Shannon1948].
    Repeating these footnotes outside the docstrings should not produce duplicates in the 
    footnote list at the bottom of the page. If footnote IDs and bodies are identical, these
    are considered duplicates in the updated PR, so they are not repeated in the footnote list.

    ```@docs
    fancy_algorithmC
    ```

    Also, the documentation pages might have footnotes that are not in any 
    docstring[^SomeotherPerson1990]. These should still be listed at the bottom of the page,
    like they are now.

    [^SomeotherPerson1990]: Some reference.
    [^Shannon1948]: C. E. Shannon, Bell Systems Technical Journal 27, pp 379 (1948)
```

With the updated PR, this renders as 

<img width="477" alt="Skjermbilde 2020-11-08 kl  16 23 33" src="https://user-images.githubusercontent.com/5237566/98469280-cbc3df80-21de-11eb-9bda-97f12bf36bdd.png">

mortenpi:
I didn't go through this yet, but one small request which can help expedite this: could you add a page to the [example HTML build](https://github.com/JuliaDocs/Documenter.jl/tree/master/test/examples) which would showcase all the edge cases with the footnotes? A new `footnotes.md` page would be perfectly fine (in fact, preferred).

Wimmerer:
Could I bump this? We need to add citations for the SuiteSparse stdlib, which would involve a *lot* of duplicates appearing in official docs. @ViralBShah

fingolfin:
For the specific use case of citations, nowadays I would also point to [DocumenterCitations.jl](https://github.com/JuliaDocs/DocumenterCitations.jl)

PR #1468, opened 2020-11-10
Add page title to search results

ianshmean:
With multiple dispatch, often the context of a method name is lost in the search results. `rounding` is a good example, where 5 of the top 8 results will land you in the `Dates` section

Current behavior:
image

This PR:

image

I did wonder about whether the page name should preceded the result, or whether there was a nicer way to format it in its current form

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * ![Enhancement][badge-enhancement] Objects that render as equations and whose `text/latex` representations are wrapped in display equation delimiters `\[ ... \]` or `$$ ... $$` are now handled correctly in the HTML output. ([#1278][github-1278], [#1283][github-1283], [#1426][github-1426])
 
+* ![Enhancement][badge-enhancement] The search page in the HTML output now shows the page titles in the search results. ([#1468][github-1468])
+
 ## Version `v0.25.3`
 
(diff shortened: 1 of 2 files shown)


mortenpi:
This LGTM, thanks @ianshmean!

> I did wonder about whether the page name should preceded the result, or whether there was a nicer way to format it in its current form

I think the information could be organized in some better way in principle. Maybe even multiple lines per search result in a box or something, with page title in a smaller font either above or below? I don't have any specific ideas though how it should look like though.

This definitely improves the status quo though, so I am happy to merge this as is. But if you want to iterate on this, we can also wait.

ianshmean:
I think go with this and if I have an idea I'm happy to do another PR 👍

ianshmean:
Would it be possible to get this into the julia 1.6 docs? Has it been released yet?

mortenpi:
No, it's still sitting on Documenter#master. In principle, there's nothing stopping us from releasing 0.26 though. What's the 1.6 timeline?

ianshmean:
I think it will fork in the next day or so, so it's tight if we want it to be straightforward. But I guess it's a logical thing to want to be added to an rc. 

If there really isn't anything stopping a release, then it could be simple!

PR #1472, opened 2020-11-10
Stop warning for mailto: links

friggog:
Previously mailto: links got treated as a local link so triggered a warning

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 * ![Bugfix][badge-bugfix] Type aliases of `Union`s (e.g. `const MyAlias = Union{Foo,Bar}`) are now correctly listed as "Type" in docstrings. ([#1466][github-1466], [#1474][github-1474])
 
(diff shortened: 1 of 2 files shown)


mortenpi:
LGTM, although I am wondering if we should do this check already in `isabsurl`: https://github.com/JuliaDocs/Documenter.jl/blob/c024d21a2a2fa1ebc027ac34b95aba5ec106f624/src/Utilities/Utilities.jl#L547-L553

friggog:
I'm not certain I'd count a mailto link as an absolute url - not sure exactly what the definition is but to me it seems to complicate the semantics of the isabsurl function?

Easy change if you'd prefer to roll it in vs what is there now

mortenpi:
That's fair, I was similarly unsure. For what it's worth, I think _technically_ they are (e.g. [IsURL](https://github.com/zlatanvasovic/IsURL.jl) considers them to be absolute), but we can leave figuring that out for another day. Thanks for the PR!

PR #1475, opened 2020-11-11
Use SVG for brand icons and FA icon for source link

friggog:
Uses inline SVG icons for repo brands (GitHub, GitLab etc) which allows for ADO logo as well.
Remove FA brands font as a result.
image

Had to update the stylesheet - not sure I've done this in the right/best way, also recompiling causes complete changes to the CSS, not sure if there is a preferred way to do this?

Also changed the source link for each function to use the FA code icon which I think looks nicer?
image

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1070,28 +1070,38 @@ function render_navbar(ctx, navnode, edit_page_link::Bool)
     # Set the logo and name for the "Edit on.." button.
     if edit_page_link && (ctx.settings.edit_link !== nothing) && !ctx.settings.disable_git
         host_type = Utilities.repo_host_from_url(ctx.doc.user.repo)
-        if host_type == Utilities.RepoGitlab
+        if host_type === Utilities.RepoGitlab
             host = "GitLab"
+            # from FontAwesome Brands: https://github.com/FortAwesome/Font-Awesome/blob/master/svgs/brands/gitlab.svg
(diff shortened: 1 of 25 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -1072,19 +1071,29 @@ function render_navbar(ctx, navnode, edit_page_link::Bool)
         host_type = Utilities.repo_host_from_url(ctx.doc.user.repo)
         if host_type == Utilities.RepoGitlab
             host = "GitLab"
-            logo = "\uf296"
+            logo = svg[:viewBox => "0 0 512 512", :xlmns=>"http://www.w3.org/2000/svg"](
```
fredrikekre, line 1074: Where are these taken from? Would be nice to add a comment with a URL or something.

mortenpi:
For recompiling the CSS (also applies to #1476), please use [DocumenterTools](https://github.com/JuliaDocs/DocumenterTools.jl) (it's not obvious, we really need to document this) -- it makes sure that the correct code formatting is used for the resulting CSS etc.:

```julia
using DocumenterTools: Themes
Themes.compile_native_themes()
```

The only thing you need to make sure is that your local Documenter clone is a dev. dependency of the environment where you also installed DocumenterTools.

friggog:
Yeah might useful to add this to the contributing page in the docs

mortenpi:
Closing, since we decided that it's better to stick with the status quo (since we can't include the Azure logo anyway): https://github.com/JuliaDocs/Documenter.jl/pull/2657#discussion_r2013034203 (thank you @friggog and also @fingolfin for the work on this though, even though it's not getting in)

PR #1476, opened 2020-11-11
Scroll overflow for long equations

### LINKED ISSUE #1470: Long ```math ... ``` overflows the div
If you have a long equation in a \``` math ... ``` block then the result will overflow the parent div:

image

If you just use inline math \`` ... `` then it wraps OK, although doesn't look quite as nice generally:

image

I don't know what the preferable solution is here,  but I feel like overflow should be avoided

friggog:
Resolves #1470


mortenpi:
This works great for KaTeX but not for MathJax. I think it should be enough to apply a class on the `<div>` that surrounds the equation

https://github.com/JuliaDocs/Documenter.jl/blob/ef9a5623ad267b236bb40db44ce1f1c83ea6d8f9/src/Writers/HTMLWriter.jl#L1637

and then apply the `overflow-x` style to that `<div>`?

friggog:
@mortenpi  good point, I think this should now work for all packages. The margin fix is a bit hacky but it seems like the easiest way to get clean layouts

assets/html/scss/documenter/components/_docstring.scss
```diff
@@ -48,6 +48,10 @@
       @extend .is-primary;
       @extend .is-size-7;
     }
+
+    .math-container {
+      overflow-x: auto;
+    }
```
mortenpi, line 54: I don't think this should be in `_docstring.scss`, since we want this to apply in all circumstances (currently the example cases don't get scrollbars)? I think the right place would be in the `.docs-main > article` selector: https://github.com/JuliaDocs/Documenter.jl/blob/60f368c0feb0cd17dafec13af2c83fbbaf948499/assets/html/scss/documenter/layout/_main.scss#L5-L10

mortenpi (approved):
LGTM, thanks for iterating on this @friggog!

I'll just note here that it looks like MathJax doesn't actually break inline equations like KaTeX does. But I don't think there's much we can do about it here.

PR #1501, opened 2021-01-03
Smallfixes

laborg:
While working on an EPUB exporter, I saw opportunity for small improvements.
- `slugify`ed pieces should start with an letter
- only extract the number for referenced issues

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -66,7 +66,11 @@ srcpath(source, root, file) = normpath(joinpath(relpath(root, source), file))
 Slugify a string into a suitable URL.
 """
 function slugify(s::AbstractString)
-    s = lstrip(x->x ∉ 'a':'z' && x ∉ 'A':'Z',s)
+    stripped_s = lstrip(!isletter,s)
+    # in case there is no letter at all, take the original approach
+    if !isempty(stripped_s)
+        s = stripped_s
+    end
     s = replace(s, r"\s+" => "-")
     s = replace(s, r"^\d+" => "")
     s = replace(s, r"&" => "-and-")
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -66,6 +66,7 @@ srcpath(source, root, file) = normpath(joinpath(relpath(root, source), file))
 Slugify a string into a suitable URL.
 """
 function slugify(s::AbstractString)
+    s = lstrip(x->x ∉ 'a':'z' && x ∉ 'A':'Z',s)
     s = replace(s, r"\s+" => "-")
     s = replace(s, r"^\d+" => "")
     s = replace(s, r"&" => "-and-")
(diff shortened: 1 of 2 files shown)


src/CrossReferences.jl
```diff
@@ -53,11 +53,12 @@ function basicxref(link::Markdown.Link, meta, page, doc)
...
         text = strip(sprint(Markdown.plain, Markdown.Paragraph(link.text)))
-        if occursin(r"#[0-9]+", text)
-            issue_xref(link, lstrip(text, '#'), meta, page, doc)
-        else
+        refissue = match(r"#[0-9]+", text)
+        if refissue === nothing
             name = Utilities.slugify(text)
             namedxref(link, name, meta, page, doc)
+        else
+            issue_xref(link, refissue[1], meta, page, doc)
```
mortenpi, line 61: I don't think this is correct -- there is no capture group in regex:

```julia
julia> x = match(r"#[0-9]+", "#124")
RegexMatch("#124")

julia> x[1]
ERROR: BoundsError: attempt to access 0-element Array{Union{Nothing, SubString{String}},1} at index [1]
```

Either way, as far as I can tell, this change is strictly non-functional? In which case I would just remove it -- it's debatable whether this actually improves readability, so I would rather err on the side of not making unnecessary non-functional changes.

src/Utilities/Utilities.jl
```diff
@@ -66,6 +66,11 @@ srcpath(source, root, file) = normpath(joinpath(relpath(root, source), file))
 Slugify a string into a suitable URL.
 """
 function slugify(s::AbstractString)
+    stripped_s = lstrip(!isletter,s)
+    # in case there is no letter at all, take the original approach
+    if !isempty(stripped_s)
+        s = stripped_s
+    end
```
mortenpi, line 73: Is there a particular problematic case that you came across and this tries to resolve?

laborg: I stumbled across the heading "1) problematic heading " giving me a warning when creating an epub because technically the slugified version starting with a "("  isn't a URL.

src/CrossReferences.jl
laborg: I'm not sure where the grouping went in this commit. I agree that this change isn't important.

src/Utilities/Utilities.jl
mortenpi: Technically, parentheses should be valid in URL fragments though?

But we already strip numbers, so I do agree that we should strip various other non-letters. But we probably want to keep e.g. `@` and `_`, as they are valid variable names in Julia.

laborg: Yes, they are allowed, but in this context the slugified target must match the id-attribute and for those I followed the recommendation from: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

odow:
Closing because HTML5 is much more flexible, and we now explicitly support starting with numeric constants. 

See discussion in https://github.com/JuliaDocs/Documenter.jl/pull/2325.

Issue #1458, opened 2020-10-30
Change of highlighting themes


kimikage:
I'm currently working on ANSI color support for the `@example`/`@repl` blocks. (cf. PR #1441) I would like to support LaTex output in the future, but I'm targeting HTML first.

There are several ways to specify ANSI colors, of which the  terminal-dependent 16 colors are the most commonly used. Since those colors are not standardized, I'm going to style them based on the colors (variables) of the theme, rather than hard-coding the color codes. Consequently, I'm planning to change the default theme slightly.

By contrast, in terms of thematic consistency, I have two gripes with the syntax-highlighting with highlight.js. One is that the color scheme of the dark theme (a11y-dark) doesn't match the rest of the documenter-dark theme, and the other is that there is no consistency between the dark theme and the light theme (highlight.jl's default).

Additionally, @fredrikekre has greatly [improved Julia's syntax highlighting](https://fredrikekre.se/posts/highlight-julia/). So, I see it as an opportunity to reconsider the themes.

Although I haven't decided on a specific theme selection and design approach yet, here's a demo of the new highlight.js.
![highlight](https://user-images.githubusercontent.com/12679384/97758962-56537180-1b43-11eb-926c-500f339b8417.gif)

mortenpi:
The more colorful version is the new theme, right? If yes, then that's a massive improvement and it would be great to have! I'll just admit here that I picked the first one that kinda worked and didn't think about it since.

I think it would make sense if the highlighting theme and the theme for the ANSI colors match. That's what you're suggesting, right? Since we have to vendor in the hljs themes anyway, we could easily roll our own. Might be a little bit of maintenance work when upgrading the hljs dependency, but I think we can tolerate that.

If you're going to mess around with the themes and the highlighter anyway, it might be worth updating [highlight.js to the latest version](https://github.com/JuliaDocs/Documenter.jl/blob/d06fa07e047e84ed63b58290e0c5d0c0fb3b8944/src/Writers/HTMLWriter.jl#L468), since there might be slight differences between the versions (especially since we're using v9 and the latest is v10). I believe it would be necessary to update to v10 anyway to get Fredrik's updates once it lands in a release.

kimikage:
FYI, I have confirmed that Documenter works with v10.3.1. However, since v10 the module has been anonymized, we need to specify the export keyword argument.

```julia
        hljs_version = "10.3.1"
        push!(r, RemoteLibrary(
            "highlight",
[...]

PR #1503, opened 2021-01-06
Update highlightjs version

laborg:
The update to the newest version of highlightjs ~~pulls in the work of @fredrikekre~~ to improve `julia` and `julia-repl`. This won't work on IE11, but ¯\_(ツ)_/¯...

The CSS files haven't changed with this update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Documenter.jl changelog
 
+## Version `v0.27.0`
+
+* ![Enhancement][badge-enhancement] The JS dependencies have been updated to their respective latest versions.
+
+  - highlight.js has been updated to `v10.5.0`, which also brings various updates to the highlighting of Julia code. Due to the changes in highlight.js, code highlighting will not work on IE11. ([#1503][github-1503])
+
 ## Version `v0.26.1`
 
 * ![Bugfix][badge-bugfix] HTML assets that are copied directly from Documenters source to the build output now has correct file permissions. ([#1497][github-1497])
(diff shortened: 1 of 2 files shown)


mortenpi (approved):
LGTM, thanks @laborg!

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Documenter.jl changelog
 
+## Version `v0.27.0`
+
+* ![Enhancement][badge-enhancement] The JS dependencies have been updated to their respective latest versions.
+
+  - highlight.js has been updated to `v10.5.0`, which also brings various updates to the highlighting of Julia code. Due to the changes in highlight.js, code highlighting will not work on IE11. ([#1503][github-1503])
```
fredrikekre, line 7: (All of) my changes are not in the official release, so is this really true?

fredrikekre (approved):

CHANGELOG.md
laborg: From the changelog of highlight.js:
 - enh(julia) Update keyword lists for Julia 1.x (#2781) Fredrik Ekre

fredrikekre: Right but that is just https://github.com/highlightjs/highlight.js/pull/2781 and does not include any of the other stuff that apparently was not working on IE11.

laborg: I think the word "various" isn't correct, but the IE11 stuff has nothing todo with your work, but refers to: https://github.com/highlightjs/highlight.js/releases/tag/10.0.0 :
> If your users have very old browsers, they may no longer be supported (no more IE11, etc.). (We're using ES2015 code now.)

mortenpi: Could you suggest a better phrasing? I am not really familiar with the changes in highlight.js

fredrikekre: I see, this looks ok then. I was told some of the changes from my blogpost didn't work on some older browsers so I thought you were talking about that.

Looks like IE11 has just 1% of users nowadays so maybe not even worth mentioning?

mortenpi:
I'll go ahead and merge this, since it looks like the CHANGELOG is fine as is.

> Looks like IE11 has just 1% of users nowadays so maybe not even worth mentioning?

Meh. Might as well mention it, since it's already there. You're right, though, in that this should have no noticeable effect on Julia users.

kimikage:
As mentioned in https://github.com/JuliaDocs/Documenter.jl/issues/1458#issuecomment-719913542, highlight.js v10 warns the invalid class "language-none".

PR #1510, opened 2021-01-26
Fix crash in case of empty `# output`

knuesel:
Without this PR, a doctest such as

    ```jldoctest
    nothing
    # output
    ```

results in a BoundsError exception:

```
ERROR: LoadError: BoundsError: attempt to access 1-element Array{SubString{String},1} at index [2]
Stacktrace:
 [1] getindex at ./array.jl:809 [inlined]
 [2] indexed_iterate at ./tuple.jl:82 [inlined]
 [3] eval_script(::Documenter.DocTests.MutableMD2CodeBlock, ::Module, ::Dict{Symbol,Any}, ::Documenter.Documents.Document, ::String) at /home/j/.julia/packages/Documenter/b3ri2/src/DocTests.jl:242
...
```


fredrikekre:
```jldoctest
nothing
# output

```
Still works I believe, but no reason to not support it without the extra empty line.

knuesel:
Yes with an empty line it works already (but the doc says that "the amount of whitespace appearing above and below the # output line is not significant").

fredrikekre (approved):

src/DocTests.jl
```diff
@@ -239,7 +239,9 @@ function eval_script(block, sandbox, meta::Dict, doc::Documents.Document, page)
     #
     #
     #       to mark `input`/`output` separation.
-    input, output = split(block.code, "# output\n", limit = 2)
+
+    # Append "\n" before split in case of empty output
+    input, output = split(block.code*"\n", "# output\n", limit = 2)
```
fredrikekre, line 244: Perhaps it is a bit cleaner to split by a regex
```suggestion
    input, output = split(block.code, r"^# output$"m, limit = 2)
```

knuesel: Very nice!

src/DocTests.jl
```diff
@@ -239,7 +239,9 @@ function eval_script(block, sandbox, meta::Dict, doc::Documents.Document, page)
     #
     #
     #       to mark `input`/`output` separation.
-    input, output = split(block.code, "# output\n", limit = 2)
+
+    # Append "\n" before split in case of empty output
```
fredrikekre, line 243:
```suggestion
```

fredrikekre (approved):

mortenpi (approved):
LGTM, thanks!

PR #1511, opened 2021-01-27
Check for empty CI environment variables

knuesel:
This PR changes `deployconfig.jl` to check that `DOCUMENTER_KEY`, `GITHUB_TOKEN` and `GITHUB_ACTOR` are non-empty. Currently it only checks if they are [defined](https://github.com/JuliaDocs/Documenter.jl/blob/eea5e5c80084dbee66009d040b264d6e16a867ff/src/deployconfig.jl#L416).

Why the current behavior is a problem: Many (most?) Julia packages configure CI with the following in the docs job:

```
        env:
          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }}
```

(That's what's generated by PkgTemplates and suggested by the Documenter [manual](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions).)

This means that deploys don't work with the initial configuration: both variables are defined, but `DOCUMENTER_KEY` will be empty unless/until the user configures the secret. At deploy time, Documenter gives priority to `DOCUMENTER_KEY` so the deploy will fail.

With this PR, Documenter will fall back to `GITHUB_TOKEN` when `DOCUMENTER_KEY` is empty, so deploys work "out of the box" since the token is always defined.

The PR also includes small docs change:

* Mention that `push_review` doesn't work from forks
* Make it clearer that `GITHUB_TOKEN` and `DOCUMENTER_KEY` are not necessarily both needed


fredrikekre (approved):

knuesel:
I just force-pushed a small change: I was using a bad test for empty string (`!==`, not reliable).

mortenpi (approved):
Also LGTM. And thanks a lot for for the detailed rationale in the OP!

PR #1512, opened 2021-01-27
Clarify treatment of outputs

### LINKED ISSUE #1509: Clarify treatment of outputs (streams, # hide and semicolon)
The documentation is not clear on what is included in the result block. It's a collection of tiny issues that I have gathered here as bullet points:

Documentation for `@example` [says](https://juliadocs.github.io/Documenter.jl/dev/man/syntax/#@example-block):

> Evaluates the code block and inserts the result into the final document along with the original source code.

* This should probably say "result of the last expression" (see below)

> `stdout` and `stderr`: The Julia output streams are redirected to the results block when evaluating `@example` blocks in the same way as when running doctest code blocks.

* The [doctest doc](https://juliadocs.github.io/Documenter.jl/dev/man/doctests/) doesn't describe this behavior (it says nothing about output streams).
* The quote above implies (to me at least) that `stdout` is redirected for the whole block, but only the last line of code is redirected:

       ```@example
       println("a")
       "b"
       ```
    Here the "a" doesn't show up in the result block.
* This behavior is not the same as doctest blocks: a `jldoctest` block includes stdout from all lines (here including "a" in the result). The different treatment of stdout also applies to the last expression: `isnothing(println("a"))` on the last line of an `@example` block shows only `true` in the result. The same in a doctest block will show "a" and `true`.

Documentation for `@repl` [says](https://juliadocs.github.io/Documenter.jl/dev/man/syntax/#@repl-block):

> These are similar to `@example` blocks, but adds a `julia>` prompt before each toplevel expression. `;` and `# hide` syntax may be used in `@repl` blocks in the same way as in the Julia REPL and `@example` blocks.

* `@example` doc says nothing about `;`
* `;` actually doesn't work in `@example` blocks
* `# hide` has a particular behavior (bug?) in `@repl` blocks: it adds an empty line (except if the expression returns nothing):

       ```@repl
       1
       2 # hide
       3
       ```

   This gives a gap in the result between the "1" and "3" prompts.

Documentation for script doctests says:

> The text that appears after `# output` is the textual representation that would be shown in the Julia REPL if the script had been `include`d.

* It might be good to precise here that semicolons are ignored (same as for `include`d scripts).

knuesel:
Closes #1509

### DIFF
--- a/docs/src/man/doctests.md
+++ b/docs/src/man/doctests.md
@@ -2,7 +2,7 @@
 
 Documenter will, by default, run `jldoctest` code blocks that it finds and makes sure that
 the actual output matches what's in the doctest. This can help to avoid documentation
-examples from becoming outdated, incorrect, or misleading. It's recommended that as many of
+examples from becoming outdated, incorrect, or misleading. It is recommended that as many of
 a package's examples be runnable by Documenter's doctest.
 
 This section of the manual outlines how to go about enabling doctests for code blocks in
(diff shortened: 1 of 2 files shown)


mortenpi:
Thanks @knuesel!

PR #1517, opened 2021-02-03
Remove GitHub token build trigger warning

### LINKED ISSUE #1177: GitHub Pages build not triggered with GITHUB_TOKEN

knuesel:
This PR removes a warning from the documentation and from the CI output, saying that GitHub Page builds are not triggered when using a GitHub token for authentication.

As discussed in #1177, the build trigger seems to be fixed now. I tested the following scenarios:

* Direct push
* Merging a pull request
* Pushing a tag to an older commit

These tests were made without pushing a commit manually to the `gh-pages`. The Pages build was correctly triggered in all cases.

Note that there is still a warning in the documentation for tags created with GitHub tokens (e.g. by TagBot).

Closes #1177

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -259,12 +259,6 @@ to the configuration file, as showed in the [previous section](@ref GitHub-Actio
     of the deployment is the same as the current repository. In order to push
     elsewhere you should instead use a SSH deploy key.
 
-!!! warning "GitHub Pages and GitHub Token"
-    Currently the GitHub Page build is not triggered when the GitHub provided
-    `GITHUB_TOKEN` is used for authentication. See
-    [issue #1177](https://github.com/JuliaDocs/Documenter.jl/issues/1177)
-    for more information.
-
 ### Authentication: SSH Deploy Keys
 
 It is also possible to authenticate using a SSH deploy key, just as described in
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -414,17 +414,7 @@ function deploy_folder(cfg::GitHubActions;
     end
 end
 
-function authentication_method(::GitHubActions)
-    if env_nonempty("DOCUMENTER_KEY")
-        return SSH
-    else
-        @warn "Currently the GitHub Pages build is not triggered when " *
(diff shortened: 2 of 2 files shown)


mortenpi:
LGTM, I can also confirm that the pages builds seem to be triggered without any issues, so these warnings are no longer necessary. Thanks @knuesel!

Although I think it's still work emphasizing that the user probably needs to set up a `DOCUMENTER_KEY` because of the TagBot issue.

Issue #1387, opened 2020-08-06
Error: mode == :single requires the Markdown string to parse into a single block │


blegat:
I am getting the following error with the output of an `@example` block. Why are the `_` splitting the block into several paragraphs ?
```
┌ Error: mode == :single requires the Markdown string to parse into a single block
│   s = "\$\$ x^{2}y_{1}^{2} + x^{2}y_{2}^{2} - 2xy_{1}^{2} + 2xy_{1}y_{2} + 2y_{1}^{2} \$\$"
│   md.content =
│    2-element Array{Any,1}:
│     :$                                                                                                                                                                
│     Markdown.Paragraph(Any["x^{2}y", Italic(Any["{1}^{2} + x^{2}y"]), "{2}^{2} - 2xy", Italic(Any["{1}^{2} + 2xy"]), "{1}y", Italic(Any["{2} + 2y"]), "{1}^{2} ", :$])
└ 
```

fredrikekre:
Julia's markdown parser uses ``` `` ``` for inline math and ```` ```math ```` blocks for multiline math.

blegat:
This was generated by Literate.jl. I have a code block that returns a polynomials and the polynomials gets printed in LaTeX so the MIME must be `text/latex` (see https://github.com/JuliaAlgebra/MultivariatePolynomials.jl/blob/9a29d5e691e39794a182b8c99172ea1412d73273/src/show.jl#L3)

fredrikekre:
Right. We probably need to rewrite `$` and `$$` similar to #1283.

PR #1518, opened 2021-02-03
Fix multiline display equations in HTML

odow:
This came from the JuMP docs: https://github.com/jump-dev/JuMP.jl/pull/2447#discussion_r569597250. We have some multiline display equations with `\n` in them (e.g., to format arrays).

Follows: https://github.com/JuliaDocs/Documenter.jl/pull/1426

Seems like it also fixes https://github.com/JuliaDocs/Documenter.jl/issues/1387

Is there a better way of testing this? I just followed #1426.


odow:
Ideas on the failing `docs/pdf/make.jl` build?

mortenpi:
> Ideas on the failing docs/pdf/make.jl build?

I think you end up with literal `\` and `n` characters in the TeX code in the updated examples now, since they are `raw""` strings?

odow:
Oh, thanks. I though I recalled `raw"` being cleverer than that with `\n`.

odow:
Tests passing.

mortenpi (approved):
I expanded the tests a bit, but otherwise LGTM! Thanks @odow!

odow:
Thanks Morten.

Issue #1516, opened 2021-02-02
Get tests passing on GitHub Actions Windows runners


mortenpi:
There are some weird, probably newline-related issues that fail the tests when (in `test/doctests`) running them on the `windows-latest` runners of GitHub Actions. I am unable to reproduce them locally on a Windows machine and AppVeyor works fine.

So because of that in #1496 I stuck with AppVeyor for Windows testing. But ideally we'd like to get the tests working reliably also on GitHub Actions. If anyone wants to take a look at this, I have some commits on the [`mp/gha-actions`](https://github.com/JuliaDocs/Documenter.jl/tree/mp/gha-windows) branch where I tried out a few things with partial success.

fredrikekre:
Might be useful to try https://github.com/marketplace/actions/debugging-with-tmate to get ssh access to the machine.

odow:
This sounds familiar: https://github.com/odow/SDDP.jl/blob/95e4a7fa9ed2e7ec91280abc7ceb793c23fc4b36/docs/make.jl#L20-L27
Before calling `:fix`, I had to rewrite the line endings. I don't have a Windows machine, but Morten, if you can't reproduce you probably [autocrlf](https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings) enabled. Or GH actions has it set differently.

mortenpi:
> if you can't reproduce you probably autocrlf enabled

Yeah, I think you're right that this is the culprit. As far as I can tell:

* AppVeyor sets [`core.autocrlf = input`](https://ci.appveyor.com/project/JuliaDocs/documenter-jl/build/job/bef8l26c9l0jp2wg#L74), so the checked out files have LF, and so the tests were passing.
* GHA sets [`core.autocrlf = true`](https://github.com/JuliaDocs/Documenter.jl/runs/1834468899?check_suite_focus=true#step:2:11), so the checked out files have CRLF.
* On the Windows VM where I tried to repro it.. it appears that `core.autocrlf = true` globally, but when I manually checked the files, they have been checked out with LF (:man_shrugging:; I think I did `pkg> dev` so maybe because of that). So it makes sense that the tests would pass, like on AppVeyor.

PR #1519, opened 2021-02-03
Normalize line breaks before replacements

odow:
Part of #1516


odow:
@mortenpi here you go! Easier than I thought.

fredrikekre (approved):

mortenpi:
Cool, thanks @odow! Let's see if we can get rid of AppVeyor now.

Issue #1342, opened 2020-06-16
Couple of improvements for generated PDF


ViralBShah:
In the PDF of the Julia manual (and presumably all PDFs), it would be nice to have:

- [x] Slightly more space for the chapter and section numbers
- [ ] Restricting TOC depth to just Chapter and Section (it goes much deeper) #1259
- [x] A way to incorporate the project logo on the title page
- [x] Start the TOC on the next page (currently starts on the title page)

Related: https://github.com/JuliaLang/julia/issues/35495

![80426085-043f4400-88b3-11ea-9b7a-b9cbf5add40b](https://user-images.githubusercontent.com/744411/84813750-67f60b80-afde-11ea-87ad-70ea526f64f4.png)

ViralBShah:
The `memoir` documentclass also creates more of a printable book, but perhaps we should just use article. I am sure that almost nobody prints this PDF, and it is largely for screen reading.

inkydragon:
I've done some experimenting, see https://github.com/inkydragon/Julia-LaTeX-doc. 
Here is the output PDF [TheJuliaLanguage-en-[4 over 4]-1c28e22.pdf](https://drive.google.com/file/d/1XU1_phqFnThR7k408f8TX0FWoup_3m4y/view?usp=sharing).

Instead of submitting a pull request, I put them in a separate project.
Because I'm not sure if these style changes should be applied to `Documenter.jl` which will affect all packages.
Maybe we should have some separate `LaTeXWriter.jl` for the julia's document.

I also made a title page design, which imitates the https://julialang.org/.
Welcome to provide a better design scheme. A sketch with a little explanation is enough.

![image](https://user-images.githubusercontent.com/5158738/85841866-51676580-b7d1-11ea-9778-a5b7dd754e89.png)

ViralBShah:
Right now, it is all part of Documenter, and applies to all projects. So we would need to have some capabilities that extend Documenter to allow for this.

mortenpi:
That looks really good. A bunch of the (general) style fixes (e.g. the TOC numbers, not doing the left/right page thing) should be incorporated into Documenter I'd say. Stuff that is specific to the Julia manual I would put into the Julia repository.

LaTeXWriter would need to be extended a little bit to support the logo and background image on the title page. But it does not have to be super sophisticated / generic. Something like being able to override the raw LaTeX source of the title page in the `makedocs` call would already be sufficient.

odow:
[...]

PR #1527, opened 2021-02-12
A few PDF improvements

jebej:
Ref: #1342 

1. Use the `oneside` option of the memoir class, which is better for reading on a screen and anything that isn't a bound book.
2. For now hardcode ToC depth to section.
3. Clear page after the title.


jebej:
Is there a way to see the output from CI?

mortenpi (approved):
We don't push the PDF outputs anywhere from CI, so one has to build it locally. Here's the PDF for Documenter's own manual: 
[Documenter.jl.pdf](https://github.com/JuliaDocs/Documenter.jl/files/5978675/Documenter.jl.pdf)

The changes LGTM though, thanks @jebej!

PR #1528, opened 2021-02-15
Remove unused fallback method for post_status

mortenpi:
In `git_push` we only use the method that passes the `DeployConfig` as the first positional argument, so this one is unnecessary. Potentially using this method in Documenter in the future is probably a bad idea, since it calls `auto_detect_deploy_system()` again, potentially introducing environment-dependent bugs.

### DIFF
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -99,7 +99,6 @@ This method must be supported by configs that push with HTTPS, see
 function authenticated_repo_url end
 
 post_status(cfg::Union{DeployConfig,Nothing}; kwargs...) = nothing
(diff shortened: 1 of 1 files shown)


fredrikekre:
I use it e.g. here: https://github.com/fredrikekre/Literate.jl/blob/master/docs/make.jl#L7

mortenpi:
Arguably it would be better to change that to

```julia
Documenter.post_status(Documenter.post_status(; type="pending", repo="github.com/fredrikekre/Literate.jl.git")
; type="pending", repo="github.com/fredrikekre/Literate.jl.git")
```
? `auto_detect_deploy_system` is called on 3 separate occasions in that `make.jl`, two of which are implicit calls.

fredrikekre (approved):

PR #1533, opened 2021-02-19
Thrown an error if the repo path contains the protocol

### LINKED ISSUE #1531: Throw an error if repo contains a protocol
Currently Documenter completely misinterprets the `repo` argument to `deploydocs` if it contains a protocol (e.g. `repo = "https://github.com/Foo/Bar.jl"`).

The argument is interpreted in this function:

https://github.com/JuliaDocs/Documenter.jl/blob/4be8243d31a6ffdbc64d779e3f8c2fb7c61de075/src/Documenter.jl#L693-L714

Something like `occursin(r"^[a-z]+://", repo) && error("...")` might be enough.

giordano:
Fix #1531.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 * ![Enhancement][badge-enhancement] `deploydocs` now throws an error if something goes wrong with the Git invocations used to deploy to `gh-pages`. ([#1529][github-1529])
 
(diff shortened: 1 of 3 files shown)


fredrikekre:
Huh, why did this try to deploy a preview?

mortenpi:
LGTM, thanks @giordano!

> Huh, why did this try to deploy a preview?

I think because external PRs still define `GITHUB_TOKEN`. It's just that it doesn't have the right permissions. Hasn't shown up before because we switched to GHA only recently.

Issue #1537, opened 2021-02-24
Use URI-encoding for SVG images coming from at-example blocks


jkrumbiegel:
If a bitmap plot is returned, it's wrapped in an image tag. If an svg plot is returned, it's not wrapped, which means the svg often will just spill over the column layout, because svg just enforces its own dimensions ruthlessly. I recently made a PR in VSCode about the same issue, which we could use in Documenter as well https://github.com/julia-vscode/julia-vscode/blob/953304c5d00ae68eb33e337110509ca6d63c1152/src/interactive/plots.ts#L139-L162

It involves URI encoding the svg string and then using it as the src for an `img` tag. This could also fix issues such as Cairo-produced svg's invalidating each other such as described here https://discourse.julialang.org/t/cairomakie-and-fontconfig/55931/13?u=jules

mortenpi:
Hmm. In principle, for the layout part, it should be sufficient to make sure that we apply the same CSS to the `<svg>` as we do to the `<img>` one.

But I agree that the interaction between images could still be an issues, so URI-encoding is better. We already do it for the binary images anyway: https://github.com/JuliaDocs/Documenter.jl/blob/80a86d763dea8adde06989c4f2a945c99c668097/src/Writers/HTMLWriter.jl#L1750-L1759

jkrumbiegel:
> make sure that we apply the same CSS to the <svg> as we do to the <img> one

That sadly doesn't help, because svgs don't respect css width/height settings if they have a viewBox attribute. https://css-tricks.com/scale-svg/

PR #1538, opened 2021-02-26
put svg with xmlns in img src tag

jkrumbiegel:
This fixes https://github.com/JuliaDocs/Documenter.jl/issues/1537

A small test I made with Luxor, to show that inline svgs are correctly scaled down to fit the column width:

````md
```@example
using Luxor
@svg begin
    background("black")
end 300 300
```

```@example
using Luxor
@svg begin
    background("red")
end 800 800
```

```@example
using Luxor
@svg begin
    background("blue")
end 1500 1500
```
````

before:

grafik

after:

grafik

### CODE REVIEW

File: src/Writers/HTMLWriter.jl [hunk 1]
```diff
@@ -1748,7 +1749,17 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
...
         out = Documents.RawHTML(d[MIME"text/html"()])
     elseif haskey(d, MIME"image/svg+xml"())
-        out = Documents.RawHTML(d[MIME"image/svg+xml"()])
+        svg = d[MIME"image/svg+xml"()]
+        # the xmlns attribute has to be present for data:image/svg+xml
+        # to work (https://stackoverflow.com/questions/18467982)
+        # URIs.escapeuri is needed to replace all special characters from the SVG string
+        # which could break the HTML
+        out = if occursin("xmlns=", svg)
+            escaped_svg = URIs.escapeuri(svg)
```
mortenpi, line 1758: Would base-64 encoding work as well? This way we could avoid the dependency.

src/Writers/HTMLWriter.jl
```diff
@@ -1748,7 +1749,17 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
...
+            Documents.RawHTML(string("<img src=\"data:image/svg+xml,", escaped_svg, "\" />"))
+        else
+            Documents.RawHTML(d[MIME"image/svg+xml"()])
```
mortenpi, line 1761: I wonder if we could "fix" the `<svg>` tag in this situation, e.g. by replacing `<svg` with `<svg 
  xmlns="http://www.w3.org/2000/svg"`?

mortenpi:
Could you also include a few additional test cases here: https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/src/man/tutorial.md#including-images-with-mime

It would be good to have a small matrix of cases there, e.g. whether the `<svg>` has a `viewBox` or not, if it has `width/height` or not etc. Quickly reading through [the article you linked](https://css-tricks.com/scale-svg/), it sounded like even with `img` tags there might be edge cases where the SVG doesn't scale quite right?

src/Writers/HTMLWriter.jl
jkrumbiegel: I'm not sure, I can test it. Maybe a function doing the escaping could be copied as well, it can't be that complicated.

src/Writers/HTMLWriter.jl
jkrumbiegel: Maybe, as far as my reading went, it seemed that this tag doesn't really do anything, but it's needed for browsers to accept the svg in the img src string. Normally embedded svgs are fine. I don't suspect that you could make an svg invalid by adding it if it doesn't have it already.

jkrumbiegel:
> It would be good to have a small matrix of cases there

Would that be purely visual tests, checked manually? Because I don't think I can write a test that detects a browser's ability to show the embedded svg :)

src/Writers/HTMLWriter.jl
jkrumbiegel: at least this page https://css-tricks.com/probably-dont-base64-svg/ recommends not to use base64, because of 33% larger filesize

jkrumbiegel:
I've made the following changes now:

SVG stays utf-8, with minimal changes to keep it from corrupting the surrounding html. I think that can also be nice if something doesn't look quite as expected, one can still inspect the SVG string manually in the browser, which doesn't work with base64.

I splice in the xmlns tag if it's not there, I couldn't find reasons why I shouldn't.

I've added a couple test cases on the page you linked.

jkrumbiegel:
I'd like to bump this as Makie's documentation could really use a simpler way to return svg's from example blocks

mortenpi (commented):
Sorry @jkrumbiegel, this dropped off my radar. Let's get this wrapped up.

src/Writers/HTMLWriter.jl
```diff
@@ -1748,7 +1748,43 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
     if haskey(d, MIME"text/html"())
         out = Documents.RawHTML(d[MIME"text/html"()])
     elseif haskey(d, MIME"image/svg+xml"())
-        out = Documents.RawHTML(d[MIME"image/svg+xml"()])
+        svg = d[MIME"image/svg+xml"()]
+
+        # The xmlns attribute has to be present for data:image/svg+xml
+        # to work (https://stackoverflow.com/questions/18467982)
+        svg_tag = match(r"<svg[^>]*>", svg).match
```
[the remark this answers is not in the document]
jkrumbiegel: You're right, I didn't think about this. The svg will be invalid then, but I've added an additional check to avoid a crash anyway. I'm now just returning the svg as is if it doesn't have a tag.

mortenpi (approved):
LGTM. Do you mind using `Documente#master` for Makie for a while though? I would like to leave this change for 0.27, but I don't want to rush that. This can easily be done if you commit the `docs/Manifest.toml`.

src/Writers/HTMLWriter.jl
```diff
@@ -1748,7 +1748,52 @@ function mdconvert(d::Dict{MIME,Any}, parent; kwargs...)
     if haskey(d, MIME"text/html"())
         out = Documents.RawHTML(d[MIME"text/html"()])
     elseif haskey(d, MIME"image/svg+xml"())
-        out = Documents.RawHTML(d[MIME"image/svg+xml"()])
+        svg = d[MIME"image/svg+xml"()]
+
+
+        svg_tag_match = match(r"<svg[^>]*>", svg)
+
```
mortenpi, line 1755:
```suggestion
        svg_tag_match = match(r"<svg[^>]*>", svg)
```

jkrumbiegel:
No I don't mind using master, thanks for helping to push this over the finish line.

devmotion:
I experienced the same problem with Jupyter notebooks, so maybe it would be useful to add the same fix to IJulia?

xref: https://discourse.julialang.org/t/cairomakie-and-fontconfig/55931/17?u=devmotion

mortenpi:
Thanks for iterating on this @jkrumbiegel!

cormullion:
Did this PR fix the "text in SVG images gets corrupted" issue? Still seeing this problem...

jkrumbiegel:
Yes, but it's only on master for now

cormullion:
OK, thanks (and for the fix!). I'll put my doc updates aside for a while...

mortenpi:
> I'll put my doc updates aside for a while...

If you have some docs that need this now, I would suggest just using `#master` for now by committing a `docs/Manifest.toml` file.

PR #1540, opened 2021-02-27
use valid identifiers for sandbox module names

JeffBezanson:
This allows base julia to switch to `var""` printing for invalid identifiers (https://github.com/JuliaLang/julia/pull/39841) and still have Documenter work with multiple julia versions.

### DIFF
--- a/src/DocTests.jl
+++ b/src/DocTests.jl
@@ -125,9 +125,9 @@ function doctest(ctx::DocTestContext, block_immutable::Markdown2.CodeBlock)
         # Define new module or reuse an old one from this page if we have a named doctest.
         name = match(r"jldoctest[ ]?(.*)$", split(lang, ';', limit = 2)[1])[1]
         if isempty(name)
-            sym = Symbol("__doctest__", filter(!=('#'), string(gensym())))
+            sym = Symbol("__doctest__", lstrip(string(gensym()), '#'))
         else
-            sym = Symbol("__doctest__", name)
+            sym = Symbol("__doctest__named__", name)
         end
         sandbox = get!(() -> Expanders.get_new_sandbox(sym), ctx.meta, sym)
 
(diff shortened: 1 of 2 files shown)


src/DocTests.jl
```diff
@@ -124,7 +124,11 @@ function doctest(ctx::DocTestContext, block_immutable::Markdown2.CodeBlock)
     if startswith(lang, "jldoctest")
         # Define new module or reuse an old one from this page if we have a named doctest.
         name = match(r"jldoctest[ ]?(.*)$", split(lang, ';', limit = 2)[1])[1]
-        sym = isempty(name) ? gensym("doctest-") : Symbol("doctest-", name)
+        if isempty(name)
+            sym = Symbol("__doctest__", filter(!=('#'), string(gensym())))
```
mortenpi, line 128: It looks like `!=('#')` doesn't work on 1.0, but this should be fine (and, arguably, a bit more readable):

```suggestion
            sym = Symbol("__doctest__", lstrip(string(gensym()), '#'))
```

src/DocTests.jl
```diff
@@ -124,7 +124,11 @@ function doctest(ctx::DocTestContext, block_immutable::Markdown2.CodeBlock)
...
+        else
+            sym = Symbol("__doctest__", name)
```
mortenpi, line 130: There is a theoretical possibility of the generated symbol conflicting with a named doctest (e.g. if someone uses `260` or something like that as their doctest name). This should get around that:

```suggestion
            sym = Symbol("__doctest__named__", name)
```

mortenpi (commented):
I guess, technically, the `gensym()`d one is no longer guaranteed to be unique, since `gensym()` would just check for the existence of `##NNN` but not `__doctest__NNN`? I don't think this will be an issue in practice though, since that's how the named ones work anyway.

Basically, LGTM, other than the two suggestions. We can get a patch release out as soon as this is merged to update Documenter in the Julia repo.

PR #1541, opened 2021-02-28
Illusrtate submodules in `@autodocs` example

goretkin:
I briefly expected `@autodocs` to recurse through defined submodules, and this example would have corrected my misunderstanding.


mortenpi:
Could we also explicitly mention that it is not recursive in the text?

goretkin:
> Could we also explicitly mention that it is not recursive in the text?

How's that?

mortenpi:
Looks great, thanks @goretkin!

PR #1545, opened 2021-03-05
Update SSH deploy example

staticfloat:
I have found that the links are all screwed up if you provide the full `git` clone URL, you should just provide the repository name.


mortenpi:
LGTM, thanks @staticfloat!

PR #1551, opened 2021-03-15
Fix highlight.js warning due to invalid "language-none"

kimikage:
In highlight.js v10, only "nohighlight" and "no-highlight" are valid for ignoring highlighting.

```
WARN: Could not find the language 'none', did you forget to load/include a language module?
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * ![Enhancement][badge-enhancement] The JS dependencies have been updated to their respective latest versions.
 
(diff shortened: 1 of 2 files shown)


mortenpi (approved):
LGTM. Would you mind adding the PR reference to the corresponding entry in the CHANGELOG as well?

kimikage:
I figured this was an attachment to PR #1503, so I added a PR reference after that number.

FYI, highlight.js v10.6 has already been released, and the initialization API has been changed.
(cf. https://github.com/highlightjs/highlight.js/blob/master/CHANGES.md#version-1060)
However, I don't think it needs to be updated urgently.

mortenpi:
Thanks @kimikage!

PR #1553, opened 2021-03-16
Let sitename link back to document root just like the logo.

fredrikekre:


mortenpi (commented):
Good idea. Needs a small CSS update as well though: the sitename is blue now (in the light theme) because it is a link. I'd say we want to keep it black, and also make sure it doesn't do anything for hover.

fredrikekre:
I thought there were instructions somewhere about how to compile the theme but can not find it now. Feel free to push those changes to this branch.

mortenpi:
It should be to go from the SCSS to CSS (assuming a `Pkg.dev`ed Documenter):

```julia
julia> using DocumenterTools: Themes
julia> Themes.compile_native_themes()
```

PR #1569, opened 2021-04-24
Use JuliaMono, drop Google Fonts CDN

### LINKED ISSUE #618: Fonts and CSS for Unicode Art

### LINKED ISSUE #1568: Use JuliaMono as the default monospace font
Now available from cdnjs: https://cdnjs.com/libraries/juliamono

mortenpi:
@cormullion Do you think we could maintain the CSS file with the `@font-face`s on the JuliaMono side? I created one here at the moment, [based on the Lato one](https://github.com/betsol/lato-font/blob/master/css/lato-font.css), that tries to incorporate all the woff2 files you have available via CDNJS.

One thing in that CSS file I am not entirely certain about is the use of the local version of the font (via `local('JuliaMono')`) -- fonts on CDNs usually don't appear to use it. I am entirely not sure about the pros and cons here, but it's probably in order to make sure that it doesn't conflict with an old version of the font, or even with a different font that happens to have the same name.

Close #1568. Should also fix our Unicode woes, so also closes #618.


cormullion:
Cool! You mean add a .CSS file to https://github.com/cormullion/juliamono? no problem, I think.

I think the 'local' might be to avoid downloading a file that's already present? I don't think that works that well anyway, since browsers increasingly don't allow access to local fonts.... Probably safe to leave it out....

assets/html/scss/juliamono.scss
```diff
@@ -0,0 +1,44 @@
+$juliamono-base-url: "https://cdnjs.cloudflare.com/ajax/libs/juliamono/0.037";
+
+@font-face {
+    font-family: JuliaMono;
+    font-weight: 300;
+    font-style: normal;
+    src: local('JuliaMono'), url("#{$juliamono-base-url}/JuliaMono-Light.woff2") format("woff2");
```
fredrikekre, line 7: What exactly do these different versions do? For my own website I used the one called `Regular` but that one is not even included here.

cormullion: I've not yet found out how CSS font weights works. :) 

In the font file I have these definitions:

Light 300

Regular 400

Medium 500

Semibold 600

Bold 700

Extrabold 800

Black 900

- Regular is the one I'd use by default.

fredrikekre: Okay. What determines which one is used? Will all those file be loaded by the browser? And for the case of Documenter, are all of them used?

cormullion: In my Documenter docs CSS I use just this one:

```
@font-face {
	font-family: JuliaMono-Regular;
	font-display: fallback;	/* allow 100 milliseconds then display fallback */
	src: url("https://cdn.jsdelivr.net/gh/cormullion/juliamono/webfonts/JuliaMono-Regular.woff2");
}
```

Probably not a good idea to ask for them all :)

mortenpi: For what it's worth, it looks like Chrome at least is smart enough not to download the `.woff2` if you're not actually using it. Currently, with this setup, on the `-Medium` is actually loaded, because the code blocks have the (default) weight of `400`.

That said, I agree that, if we're gonna have the CSS file here, we should only have `-Regular` there. However, @cormullion, what is the actual difference between the `.woff2` files of the different weights? Why are you generating different ones in the first place?

cormullion: Each `woff2` file contains the glyph shapes and data for one specific weight, so you need a separate file for each weight. (I think variable fonts might come in one big file, but this isn’t one of those.)

cormullion:
The CDNJS has finally completed the maintenance/update backlog, so JuliaMono v0.39 is now available at https://cdnjs.com/libraries/juliamono.

mortenpi:
Let's get this into 0.27.1 as well.

cormullion:
That's awesome, thanks!

Do you think anyone will notice/complain/pine for Roboto? Presumably it's not hard to revert (for some strange reason), given that it wasn't difficult to switch fonts before...

If there's an update to the font (eg from version , does this require an update here as well? I know the Pluto folks prefer to update font versions manually - I'm not sure whether there are notifications from cdn.jsdelivr or from cdnjs...

mortenpi:
> Do you think anyone will notice/complain/pine for Roboto? Presumably it's not hard to revert (for some strange reason), given that it wasn't difficult to switch fonts before...

JuliaMono fixes our Unicode woes, so it is the superior choice for the default in my opinion. And, as you said, it is customizable with a little bit of effort.

> If there's an update to the font (eg from version , does this require an update here as well?

Yep, the version is hard-coded, so an update to the font version would require a change here and a new release. It literally just requires a tiny change here though:
https://github.com/JuliaDocs/Documenter.jl/blob/bfdb3f271149a2c768e1b424eefd94b4e8aaa005/src/Writers/HTMLWriter.jl#L449

Issue #1571, opened 2021-04-29
Add docs on how to create a "Note" section in blue


logankilpatrick:
I was looking through the documenter docs and could not figure out how to add those special boxes in the docs. There should perhaps be a section which goes over this (shown here: https://docs.julialang.org/en/v1/#man-introduction).

MichaelHatherly:
https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#Syntax links to https://docs.julialang.org/en/v1/stdlib/Markdown/ which is where it's all documented (admonitions are all at the very bottom of the page), but perhaps needs some additional linking to from some other places.

mortenpi:
The [showcase page](https://juliadocs.github.io/Documenter.jl/dev/showcase/) also shows off all the elements and syntax, so its should source also work as an example of how to use everything (but it might be easy to miss this page).

logankilpatrick:
If we gave them their own section heading, that would make them indexed in the search right?

logankilpatrick:
I am not sure if the PR helped at all, still hard to find this.

mortenpi:
They do show up in search now: 
![Screenshot from 2021-05-22 10-48-25](https://user-images.githubusercontent.com/147757/119205072-4f924900-baeb-11eb-9812-7ff51bc4a58f.png)

logankilpatrick:
Got it, wasn't in dev.

PR #1574, opened 2021-05-02
Add headers to the admonitions for indexing

logankilpatrick:


logankilpatrick:
Closes: https://github.com/JuliaDocs/Documenter.jl/issues/1571 by making these more searchable. Right now searching "note" yields no useful results near the top.

mortenpi:
Could we make them L6 headings (`######`), so that the headings themselves would not be as prominent? They will still show up in search -- e.g. you can currently search for "Heading level" and all the levels will be there.

logankilpatrick:
Done!

Issue #1449, opened 2020-10-21
Add "this version is outdated" notice to versions other than latest


christopher-dG:
A lot of people have noticed that search engines tend to serve old versions of Julia documentation as well as package docs. Oftentimes, people don't actually realize they're looking at an old version. Some other doc services will display a notice along the lines of "this version is outdated, click here to go to this page in the current version" when displaying old documentation. Could we do something like that?

Here's an example from the Rust ecosystem:
![image](https://user-images.githubusercontent.com/17228795/96753474-878aaf80-1395-11eb-81d5-e27dd78616a7.png)

fredrikekre:
I guess it has to be a script or something that you run on old docs?

mortenpi:
This wouldn't help with existing deployed documentation, but we could do this going forward at least. On page load, the version selector lists all the versions, so you can check if the current one is the latest one.

cdsousa:
I think something like the CMake warning banner is very useful:

![Screenshot_20201104_091733](https://user-images.githubusercontent.com/333519/98092551-b23d3380-1e7e-11eb-8459-e1af24f5f10b.png)

eirikbrandsaas:
I support this so much, and I like @cdsousa example. 

It's happening so often that I google something, end up in the two-year-old DataFrames documentation without realizing it, and then spend 30 minutes reading outdated docs that with commands that no longer work.

fredrikekre:
Prototype here: https://fredrikekre.github.io/Literate.jl/v2.6/

eirikbrandsaas:
> Prototype here: https://fredrikekre.github.io/Literate.jl/v2.6/

Big improvement. For visibility, I think it would be better to put it as a banner like CMake or potentially just below the search box.

In the spirit of minimalism: `` or select the appropriate version in the drop-down menu below.'' isn't really necessary: I assume very few people are purposefully using older documentation and a) that subset has used Julia and read the documentation before and b)  it's easy to understand where to find the other version.

christopher-dG:
Awesome! I would recommend using the Julia logo's "inner" red colour with lower opacity to make the text more readable.

Also: This documents an~~d~~ old version -> This documents **an** old version

fredrikekre:
> For visibility, I think it would be better to put it as a banner like CMake or potentially just below the search box.

Yea maybe, but that requires more CSS than I know so.... I also kinda liked that it is next to the version selector but perhaps it doesn't matter. 

[...]

PR #1577, opened 2021-05-18
Add warning for outdated docs

### LINKED ISSUE #1302: Automatic noindex tags on versioned docs?
Versioned docs get out of date fast, so it would be nice if we could make sure they aren't indexed to push users towards the stable docs.

pfitzseb:
Fixes https://github.com/JuliaDocs/Documenter.jl/issues/1449 (and https://github.com/JuliaDocs/Documenter.jl/issues/1302).

image

This obviously only works for newly generated docs. For already generated old documentation package authors can manually run a script like https://gist.github.com/pfitzseb/4b910449e21b51fed4b4a2a7c20912ff to add the necessary elements.

Oh, and because I was wondering about whether adding a `noindex` meta tag dynamically actually prevents Google from indexing the page: [Yes, apparently it does](https://www.searchcandy.uk/seo/can-you-use-javascript-gtm-to-add-a-noindex-meta-robots-tag/).

_Edit by @mortenpi: fix #1302 x-ref_


fredrikekre:
> Oh, and because I was wondering about whether adding a noindex meta tag dynamically actually prevents Google from indexing the page: Yes, apparently it does.

https://github.com/JuliaDocs/Documenter.jl/issues/1302#issuecomment-723910244 :D

src/Documents.jl
```diff
@@ -231,6 +231,7 @@ struct User
     authors :: String
     version :: String # version string used in the version selector by default
     highlightsig::Bool  # assume leading unlabeled code blocks in docstrings to be Julia.
+    warn_outdated::Bool # insert warning popup for old versions of the documentation
```
mortenpi, line 234: I think this option should be passed via the `HTML` object, since it's only relevant for the HTML backend.

src/Writers/HTMLWriter.jl
```diff
@@ -1286,13 +1289,21 @@ function expand_versions(dir, versions)
...
         println(buf, "var DOC_VERSIONS = [")
         for folder in entries
             println(buf, "  \"", folder, "\",")
         end
         println(buf, "];")
+
+        newest = first(entries)
+        for s in symlinks
+            if s[1] == newest
+                newest = s[2]
```
mortenpi, line 1303:
```suggestion
            if s.first == newest
                newest = s.second
```

Might be a bit more readable -- it took me a sec to realize that `symlinks` is an array of `Pair`s.

src/Writers/HTMLWriter.jl
```diff
@@ -1286,13 +1289,21 @@ function expand_versions(dir, versions)
...
-function generate_version_file(versionfile::AbstractString, entries)
+function generate_version_file(versionfile::AbstractString, entries, symlinks = [])
     open(versionfile, "w") do buf
         println(buf, "var DOC_VERSIONS = [")
         for folder in entries
             println(buf, "  \"", folder, "\",")
         end
         println(buf, "];")
+
+        newest = first(entries)
```
mortenpi, line 1300:
```suggestion
        # The first element in entries corresponds to the latest version, but is usually not the full version
        # number. So this essentially follows the symlinks that will be generated to figure out the full
        # version number (stored in DOCUMENTER_CURRENT_VERSION in siteinfo.js).
        newest = first(entries)
```

I believe this is accurate?

src/Writers/HTMLWriter.jl
```diff
@@ -650,6 +650,9 @@ function render(doc::Documents.Document, settings::HTML=HTML())
         for filename in readdir(joinpath(ASSETS, "js"))
             path = joinpath(ASSETS, "js", filename)
             endswith(filename, ".js") && isfile(path) || continue
+            if filename == "warner.js" && !doc.user.warn_outdated
+                continue
+            end
             push!(r, JSDependencies.parse_snippet(path))
```
mortenpi, line 656: Is the `.js` handled properly when it's loaded via RequireJS?

mortenpi (commented):
The approach LGTM, just a few small nitpicks.

mortenpi:
> For already generated old documentation package authors can manually run a script like https://gist.github.com/pfitzseb/4b910449e21b51fed4b4a2a7c20912ff to add the necessary elements.

You wouldn't be willing to turn this into a function and put it into DocumenterTools by any chance?

pfitzseb:
Ok, just pushed a commit that cleans up the JS/CSS side of this. The overlay is now rendered as an admonition and respects themes:
image
image

assets/html/scss/documenter/components/_admonition.scss
```diff
@@ -10,7 +10,7 @@ $admonition-body-color: () !default;
 $admonition-header-background: () !default;
 $admonition-header-color: () !default;
 @each $name, $color in $admonition-colors {
-  $bg: lighten-color($color, 4);
+  $bg: lighten-color($color, 5);
```
pfitzseb, line 13: This change makes sure the default theme's admonitions all render with a dark background color, instead of the current behaviour:
image
New:
image

Imho this improves legibility by a lot.

src/Writers/HTMLWriter.jl
pfitzseb: It is, yes. I'd kinda like to load it a bit earlier, but it shouldn't matter in practice.

src/Documents.jl
pfitzseb: Good call. Will fix.

pfitzseb:
Ok, this is good to go from my end. Review comments are addressed.

> You wouldn't be willing to turn this into a function and put it into DocumenterTools by any chance?

That makes a lot of sense :) I totally forgot about that package.

src/Writers/HTMLWriter.jl
```diff
@@ -901,6 +910,15 @@ analytics_script(tracking_id::AbstractString) =
...
 
+function warning_script(should_warn)
+    if should_warn
+        return Tag(:script)[Symbol(OUTDATED_VERSION_ATTR)](
+            read(joinpath(ASSETS, "warner.js"), String)
+        )
+    end
+    return Tag(Symbol("#RAW#"))("")
+end
+
```
mortenpi, line 921: Hmm. What pros and cons do you see between (1) pasting into the HTML, (2) loading it from a file via a `<script>` tag (like we do for `themeswap.js`), or (3) loading via RequireJS like we do for most deps?

I guess RequireJS adds a small lag, which means the page renders at first without the box and then it appears, which can be annoying? But I feel that loading it from a separate file would be cleaner and consistent with `themeswap.js`? Loading time shouldn't be an issue with a separate file, since it seems to be working fine for the theme.

pfitzseb: Makes sense, I'll switch this to directly loading from the file.
The reason I stopped using RequireJS is to get a tag with `OUTDATED_VERSION_ATTR` into the source code so the accompanying PR in DocumenterTools can check whether it should add a warning to older pages.

pfitzseb:
Ok, this should be good to go.

Note that I've referenced https://github.com/JuliaDocs/DocumenterTools.jl/pull/51 in the docs, so that needs to be merged and tagged first, I think.

mortenpi (approved):
Sorry for the delay, but LGTM. Thanks a lot for iterating on this @pfitzseb!

src/Writers/HTMLWriter.jl
```diff
@@ -60,6 +60,9 @@ using ...Utilities.MDFlatten
 
 export HTML
 
+"Data attribute for the script inserting a wraning for outdated docs."
+OUTDATED_VERSION_ATTR = "data-outdated-warner"
```
mortenpi, line 64:
```suggestion
const OUTDATED_VERSION_ATTR = "data-outdated-warner"
```

src/Writers/HTMLWriter.jl
```diff
@@ -1286,13 +1304,26 @@ function expand_versions(dir, versions)
...
         for folder in entries
             println(buf, "  \"", folder, "\",")
         end
         println(buf, "];")
+
+        # The first element in entries corresponds to the latest version, but is usually not the full version
+        # number. So this essentially follows the symlinks that will be generated to figure out the full
+        # version number (stored in DOCUMENTER_CURRENT_VERSION in siteinfo.js).
+        # Every symlink points to a directory, so this doesn't need to be recursive.
+        newest = first(entries)
```
fredrikekre, line 1319: Is this really what should be used for the link? Shouldn't it be e.g. the top entry of the user specified list or something?

src/Documenter.jl
```diff
@@ -223,6 +223,9 @@ determined from the source file path. E.g. for `src/foo.md` it is set to `build/
 
 Note that `workdir` does not affect doctests.
 
+**`old_warn`** inserts a warning if the current page is not the newest version of the
```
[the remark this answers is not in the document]
pfitzseb: This got fixed in https://github.com/JuliaDocs/Documenter.jl/pull/1577/commits/c5d4da6170b3fe65be487b8ee9e3f8a21ed4a3c9

src/Writers/HTMLWriter.jl
pfitzseb: I don't see how that would work, given that the user provided `versions` argument doesn't necessarily contain any actual version numbers at all.

It might make sense to just use 
https://github.com/JuliaDocs/Documenter.jl/blob/325087c852c120227ab3d41dea6fbc7a3ac1332c/src/Writers/HTMLWriter.jl#L1236

fredrikekre: > doesn't necessarily contain any actual version numbers at all.

Why does that matter? I would guess most packages would like the link to direct to e.g. `stable` instead of the actual version number

pfitzseb: `DOCUMENTER_NEWEST` needs to be an actual version so we can compare it against `DOCUMENTER_CURRENT_VERSION` though.

fredrikekre: Perhaps two variables then?

pfitzseb: Hm, I still don't quite see what problem that solves. Why is a direct link to the full version (e.g. "v1.1.1") worse than linking to `stable` or `v1.1`, which then redirects to the full version anyways?

fredrikekre: Because if everyone used canonical URLs (e.g. `stable` or `v1`) things would be easier. E.g. if you copy the URL and paste somewhere

pfitzseb: Fair enough. Using the first entry in `entries` now, which is `stable` for the defaults.

pfitzseb:
Alright, let's get this merged?

mortenpi:
Thanks again @pfitzseb and @fredrikekre! And sorry it took a while to get merged -- I am a bit short on time right now.

@pfitzseb: I think it would be good to ANN this on Discourse, since sooner the people switch over to 0.27 the better for the ecosystem. Would you be willing to post something by any chance once we get 0.27 out? This is the only really major thing on master (that has been merged), so I don't think we need to advertise any of the other stuff that much.

pfitzseb:
Sure, I can do that.

PR #1583, opened 2021-05-25
Add note about Plots.jl with GR backend

ranocha:
I needed to do this to remove warnings from the docs in https://github.com/ranocha/SummationByPartsOperators.jl/pull/89

### DIFF
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -447,7 +447,7 @@ worry about relative paths.
     qt.qpa.xcb: could not connect to display 
     qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
     ```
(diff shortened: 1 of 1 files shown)


mortenpi (approved):
This great, thank you! I do kind of feel that this is partially an issue with Plots/GR, but it's a good idea to document the workaround -- I only [recently learned about this myself](https://github.com/JuliaAtoms/AtomicMiscellany.jl/commit/4c10cd4631bbb321f4616e8a457f5ad2cdeadb7b).

fredrikekre:
> I do kind of feel that this is partially an issue with Plots/GR

Agreed (https://github.com/fredrikekre/Literate.jl/issues/73#issuecomment-522063820) but I guess Plots + GR is common enough in docs to warrant this note.

PR #1586, opened 2021-05-31
Change Julia v1.4 -> v1.6 in Travis CI/Github Actions workflows

navidcy:
This PR changes the Travis CI and Github Actions workflows for deploying the Docs to use Julia v1.6 (instead of v1.4).

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -51,7 +51,7 @@ file. Note that the snippet below will not work by itself and must be accompanie
 jobs:
   include:
     - stage: "Documentation"
-      julia: 1.4
+      julia: 1.6
       os: linux
       script:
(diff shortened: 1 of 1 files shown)


KristofferC:
Alternatively, I think the number `1` can be used to use the latest released version?

navidcy:
I’m not sure about that. These prt pf the docs are just an example really. I just thought it’s good the example to be not as outdated. But leaving the `.6` there demonstrates how one can modify it accordingly to run on a particular version they like.

mortenpi:
LGTM, thanks @navidcy!

> Alternatively, I think the number `1` can be used to use the latest released version?

Printing sometimes subtly changes between Julia minor versions, which can cause doctests to become outdated. So on those grounds I would advocate specifying a minor version, to avoid suddenly getting CI failures when a new Julia version is released.

If you do not use doctests (or you test them in a separate job) then it should be completely safe to set `julia: 1` though.

PR #1590, opened 2021-06-09
Set version to 0.27.0

mortenpi:


mortenpi:
This also updates all the JS dependencies to their latest versions.

PR #1593, opened 2021-06-12
restore the global seed after calling into IOCapture

KristofferC:
See the discussion in https://github.com/JuliaLang/julia/pull/41184#issuecomment-859384653.


fredrikekre:
https://github.com/JuliaDocs/IOCapture.jl/pull/11

KristofferC:
Personally I think it makes more sense to have it here since the setup code gets run here but doesn't really matter I guess.

KristofferC:
Or maybe not...

PR #1595, opened 2021-06-14
Adjust wording of outdated warning

pfitzseb:
[Ref](https://discourse.julialang.org/t/ann-documenter-0-27/62726/11?u=pfitzseb).

cc @fredrikekre

### DIFF
--- a/assets/html/warner.js
+++ b/assets/html/warner.js
@@ -37,7 +37,7 @@ function maybeAddWarning () {
         document.body.removeChild(div);
     });
     const href = window.documenterBaseURL + '/../' + window.DOCUMENTER_STABLE;
-    div.innerHTML = 'This is an old version of the documentation. <br> <a href="' + href + '">Go to the newest version</a>.';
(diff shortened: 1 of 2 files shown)


assets/html/warner.js
```diff
@@ -37,7 +37,7 @@ function maybeAddWarning () {
         document.body.removeChild(div);
     });
     const href = window.documenterBaseURL + '/../' + window.DOCUMENTER_STABLE;
-    div.innerHTML = 'This is an old version of the documentation. <br> <a href="' + href + '">Go to the newest version</a>.';
+    div.innerHTML = 'This documentation is not for the latest version. <br> <a href="' + href + '">Go to the newest version</a>.';
```
fredrikekre, line 40: There is not `latest` and in the next sentence `newest`, perhaps it doesn't matter.

pfitzseb:
```suggestion
    div.innerHTML = 'This documentation is not for the latest version. <br> <a href="' + href + '">Go to the latest documentation</a>.';
```
Dunno whether that reads better...

mortenpi: I think it does read a little bit better, so let's go with that.

Issue #1594, opened 2021-06-13
BoundsError in preview when no versions exist yet


devmotion:
I opened a PR to one of my repos that adds a documentation with Documenter (https://github.com/devmotion/InferenceDiagnostics.jl/pull/6). Obviously, no versions of the docs exist yet. The documentation seems to be built correctly but `deploydocs` errors with
```julia
┌ Error: Failed to push:
│   exception =
│    BoundsError: attempt to access 0-element Vector{String} at index [1]
│    Stacktrace:
│      [1] getindex
│        @ ./array.jl:801 [inlined]
│      [2] first
│        @ ./abstractarray.jl:368 [inlined]
│      [3] (::Documenter.Writers.HTMLWriter.var"#68#69"{Vector{String}, Vector{Pair{String, String}}})(buf::IOStream)
│        @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/Pkhwa/src/Writers/HTMLWriter.jl:1319
│      [4] open(::Documenter.Writers.HTMLWriter.var"#68#69"{Vector{String}, Vector{Pair{String, String}}}, ::String, ::Vararg{String, N} where N; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
│        @ Base ./io.jl:330
│      [5] open
│        @ ./io.jl:328 [inlined]
│      [6] generate_version_file
│        @ ~/.julia/packages/Documenter/Pkhwa/src/Writers/HTMLWriter.jl:1308 [inlined]
│      [7] (::Documenter.var"#git_commands#24"{String, String, SubString{String}, String, Vector{Any}, Bool, String, String})(sshconfig::String)
│        @ Documenter ~/.julia/packages/Documenter/Pkhwa/src/Documenter.jl:589
```
(https://github.com/devmotion/InferenceDiagnostics.jl/pull/6/checks?check_run_id=2815389405#step:4:210)
It seems it errors since `entries` in https://github.com/JuliaDocs/Documenter.jl/blob/cf9e4a23193b6500aa1c349cdce19d524f51b5ba/src/Writers/HTMLWriter.jl#L1319 is empty.

mortenpi:
cc @pfitzseb

devmotion:
FYI I merged the PR nevertheless and the documentation was successfully deployed on the main branch.

PR #1596, opened 2021-06-14
check for empty entries in generate_version_file

pfitzseb:
Should fix https://github.com/JuliaDocs/Documenter.jl/issues/1594.

Should be perfectly fine to omit writing `DOCUMENTER_NEWEST` and `DOCUMENTER_STABLE` if there are no previously generated docs, since the warning will not be shown in that case anyways.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1312,6 +1312,9 @@ function generate_version_file(versionfile::AbstractString, entries, symlinks = 
         end
         println(buf, "];")
 
+        # entries is empty if no versions have been built at all
+        isempty(entries) && return
+
         # The first element in entries corresponds to the latest version, but is usually not the full version
         # number. So this essentially follows the symlinks that will be generated to figure out the full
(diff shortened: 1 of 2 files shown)


src/Writers/HTMLWriter.jl
```diff
@@ -1312,19 +1312,22 @@ function generate_version_file(versionfile::AbstractString, entries, symlinks =
...
-        # number. So this essentially follows the symlinks that will be generated to figure out the full
-        # version number (stored in DOCUMENTER_CURRENT_VERSION in siteinfo.js).
-        # Every symlink points to a directory, so this doesn't need to be recursive.
-        newest = first(entries)
-        for s in symlinks
-            if s.first == newest
-                newest = s.second
-                break
+        # entries is empty if no versions have been built at all
+        if !isempty(entries)
```
fredrikekre, line 1316: perhaps easier to just add `isempty(entries) && return`?

fredrikekre (approved):

mortenpi:
LGTM, thanks a lot!

PR #1598, opened 2021-06-15
Document behavior of thrown exceptions in `@repl`

ericphanson:
https://github.com/JuliaDocs/Documenter.jl/pull/1447 might be a better solution, but this documents the current behavior at least.


mortenpi:
LGTM. I do think we want #1447 in the long term, but good to have the behavior documented in the meanwhile.

PR #1599, opened 2021-06-16
Set version to 0.27.1

mortenpi:


.github/workflows/CI.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - master
       - release-*
+    tags: '*'
```
mortenpi, line 7: @fredrikekre I think this is what we need to get the tags working.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

odow: Correct https://github.com/jump-dev/JuMP.jl/blob/9c4dbfd829487af5dac8c45f89e4de7c7496e2a8/.github/workflows/documentation.yml#L5

mortenpi:
Thanks @odow!

PR #1609, opened 2021-06-21
Display Documenter version in document build info.

fredrikekre:


fredrikekre:
https://juliadocs.github.io/Documenter.jl/previews/PR1609/ :tada:

mortenpi (approved):
I'm happy with this, but would there also be a way to detect if you're running a modified version of Documenter (e.g. a specific commit from `master` etc)? Even something that would just detect if it is a tagged version or not would be nice (e.g. outputting `v0.27.3 (untagged)`).

That said, if you can't think of an easy way to do this, I am more than happy to keep it as is.

fredrikekre:
That's a bit tricky. A user will generally not have the git repository of Documenter (even if you add by branch/commit) so you would have to try and find some more information in the users manifest which to me doesn't seems worth it.

src/Writers/HTMLWriter.jl
```diff
@@ -796,6 +796,13 @@ function render_html(ctx, navnode, head, sidebar, navbar, article, footer, scrip
     )
 end
 
+function documenter_version()
+    project = joinpath(dirname(dirname(pathof(Documenter))), "Project.toml")
```
KristofferC, line 800: I think it would be better to do this at compile time since at runtime we might not have access to this file (eg if we are running from a sysimage)

fredrikekre: https://github.com/JuliaDocs/Documenter.jl/pull/1611

PR #1611, opened 2021-06-23
Move definition of Documenter version to precompilation-time.

fredrikekre:


mortenpi:
This is a good idea. Should we maybe put the variable in the global `Documenter` module, so that we could also use it in other parts of Documenter?

KristofferC (approved):

PR #1616, opened 2021-06-26
Support deployment to the root instead of to a version subfolder

fredrikekre:


fredrikekre:
I have deployed this for https://julialogging.github.io/

src/Documenter.jl
```diff
@@ -734,12 +743,22 @@ filesystems that are case-insensitive (e.g. on OS X, Windows). Without doing a `
 first, `git add -A` will not detect case changes in filenames.
 """
 function gitrm_copy(src, dst)
-    # --ignore-unmatch so that we wouldn't get errors if dst does not exist
-    run(`git rm -rf --ignore-unmatch $(dst)`)
-    # git rm also removed parent directories
+    # Remove individual entries since with versions=nothing the root
+    # would be removed and we want to preserve previews
+    if isdir(dst)
+        for x in filter!(x -> !(x in (".git", "previews")), readdir(dst))
```
mortenpi, line 749: Perhaps a bit easier to read:

```suggestion
        for x in filter!(!in((".git", "previews")), readdir(dst))
```

mortenpi (commented):
LGTM! In case something does break here, we should be able to see it when Documenter's docs deploy.

Could just use docs and a CHANGELOG note.

fredrikekre:
Added docs etc.

PR #1624, opened 2021-07-04
Make sure version regex work with \r\n line endings.

fredrikekre:


mortenpi:
Hmm, this made me wonder if we should put a `try ... catch` around the whole thing,  to make sure that an error here would not cause Documenter to fail to load. E.g. something like

```julia
const DOCUMENTER_VERSION = try
    project = joinpath(dirname(dirname(pathof(Documenter))), "Project.toml")
    toml = read(project, String)
    m = match(r"^version\s*=\s\"(.*)\"$"m, toml)
    VersionNumber(m[1])
catch e
    @error "Unable to parse Documenter version from Project.toml. This is non-fatal." exception = (e, catch_backtrace())
    # If the parsing somehow broke, we fall back to a nonsensical, but valid version object
    v"0.0.0-version-parsing-broken"
end
```

On the other hand, maybe throwing an error is better, so that one would catch the failure early.

Issue #890, opened 2018-11-20
$ not rendering properly


morris25:
The advised way to add dollar signs to a docstring is to double escape them, but the example docstring,
```
"""
\\\$ some text \\\$ some more text
"""
dolladollabills()="\$"
```
renders to :
<img width="456" alt="screen shot 2018-11-20 at 2 12 47 pm" src="https://user-images.githubusercontent.com/17068371/48800093-724f0700-ecce-11e8-9473-80133ef468b9.png">

I have yet to find a nice way to get dollar signs to show up properly in both the repl and the rendered docs.

odow:
This continues to bite me. 

```julia
shell> cat src/index.md
The price is \$1.

The price is \$1 or \$2.

julia> using Documenter

julia> Documenter.makedocs(sitename = "Bug", pages = ["index.md"])
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
[ Info: HTMLWriter: rendering HTML pages.
```
yields 

<img width="199" alt="image" src="https://user-images.githubusercontent.com/8177701/124351685-e48e7300-dc4f-11eb-85d1-0e89b645aaab.png">

My motivation to fix it is now high. Any pointers on where I should go looking?

mortenpi:
If my original guess in https://github.com/JuliaDocs/Documenter.jl/issues/890#issuecomment-440417723 is correct, I think it might be enough to do the relevant escaping here https://github.com/JuliaDocs/Documenter.jl/blob/6c8679bac445ab1c656744075b13a57835081cc3/src/Writers/HTMLWriter.jl#L1649

odow:
So I tried the escaping thing. It was sufficient to go
```Julia
function mdconvert(text::AbstractString, parent; kwargs...)
    if text == "\$"
        return DOM.Node("&#x24;")
    end
    return DOM.Node(text) 
end
```
and I had to modify 
https://github.com/JuliaDocs/Documenter.jl/blob/6c8679bac445ab1c656744075b13a57835081cc3/src/Utilities/DOM.jl#L271-L290
to prevent it being escaped.  

But... it seems that the math engines treat `&#x24;` as `$` when it comes to delimiters... so this had no effect.

[...]

PR #1625, opened 2021-07-04
Fix escaping of dollar symbols in HTMLWriter

odow:
Previously, Javascript LaTeX engines got confused trying to
distinguish between inline delimiters and dollar symbols for actual
currency. This works around the problem by wrapping (currency) dollar
symbols in a span tag.

Closes https://github.com/JuliaDocs/Documenter.jl/issues/890

I could use come pointers on how to test this properly. (Is there a way to test the output of the rendered HTML?) I got confused with the structure of `/test`. Here's my more-involved local test:

```julia
shell> cat src/index.md
The price is \$1.

The price is \$1 or \$2.

Here is some $x^2$ math.

The pr\$ice is \$s for some $x^2$.

julia> Documenter.makedocs(sitename = "Bug", pages = ["index.md"])
```

image

### DIFF
--- a/test/htmlwriter.jl
+++ b/test/htmlwriter.jl
@@ -171,5 +171,10 @@ end
             end
         end
     end
+
+    @test "Dollar escapes" begin
+        @test string(HTMLWriter.mdconvert("\$1")) == "\$1"
+        @test string(HTMLWriter.mdconvert("\$")) == "<span>\$</span>"
+    end
 end
 end
--- a/test/htmlwriter.jl
+++ b/test/htmlwriter.jl
@@ -172,7 +172,7 @@ end
         end
     end
 
-    @test "Dollar escapes" begin
+    @testset "Dollar escapes" begin
         @test string(HTMLWriter.mdconvert("\$1")) == "\$1"
         @test string(HTMLWriter.mdconvert("\$")) == "<span>\$</span>"
     end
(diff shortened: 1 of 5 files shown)


test/htmlwriter.jl
```diff
@@ -171,5 +171,10 @@ end
             end
         end
     end
+
+    @test "Dollar escapes" begin
```
mortenpi, line 175:
```suggestion
    @testset "Dollar escapes" begin
```

odow: Oops. My mistake

mortenpi (approved):
LGTM, I think this is a clever workaround. And doesn't have the potentially breaking aspect of changing the delimiter configuration.

We don't really have good integration tests of the outputs. The best we do currently are the `occursin`s [in the example builds](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests.jl). Potentially you could add a few tests there -- I think having your snippet in the `test/examples` build would be good.

Just a passing thought, beyond the scope of this PR, but in the future we could use [Gumbo](https://github.com/JuliaWeb/Gumbo.jl) to parse the generated HTML and then do some sanity checks on the AST. It worked well for generating the outdated version warnings.

odow:
@mortenpi how's this?

mortenpi:
Looks good @odow, thanks!

PR #1627, opened 2021-07-05
Support prerendering of code blocks.

fredrikekre:


fredrikekre:
I marked this as experimental for now.

src/Writers/HTMLWriter.jl
```diff
@@ -1694,15 +1754,17 @@ mdconvert(b::Markdown.BlockQuote, parent; kwargs...) = Tag(:blockquote)(mdconver
 
 mdconvert(b::Markdown.Bold, parent; kwargs...) = Tag(:strong)(mdconvert(b.text, parent; kwargs...))
 
-function mdconvert(c::Markdown.Code, parent::MDBlockContext; kwargs...)
+function mdconvert(c::Markdown.Code, parent::MDBlockContext; settings::HTML=HTML(), kwargs...)
```
mortenpi, line 1757: Constructing a `HTML` object with default values in the keyword does not feel quite right. Maybe set to `nothing` and check for that?

mortenpi (commented):
LGTM, other than the one thing.

This file is getting way too complicated though. I was wondering if the pre-rendering could be somehow more self-contained in a separate file.. but I don't actually see how. In any case, I think we need to properly refactor this at some stage.

src/Writers/HTMLWriter.jl
```diff
@@ -409,12 +422,53 @@ struct HTML <: Documenter.Writer
             mathengine    :: Union{MathEngine,Nothing} = KaTeX(),
             footer        :: Union{String, Nothing} = "Powered by [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) and the [Julia Programming Language](https://julialang.org/).",
             ansicolor     :: Bool = false, # true in 0.28
+            prerender     :: Bool = false,
+            node          :: Union{Cmd,String,Nothing} = nothing,
+            highlightjs   :: Union{String,Nothing} = nothing,
+            lang          :: String = "en",
+            warn_outdated :: Bool = true,
```
mortenpi, line 429:
```suggestion
            lang          :: String = "en",
            warn_outdated :: Bool = true,
            # experimental keywords
            highlightjs   :: Union{String,Nothing} = nothing,
            prerender     :: Bool = false,
            node          :: Union{Cmd,String,Nothing} = nothing,
```

test/prerender/prerender.jl
```diff
@@ -0,0 +1,96 @@
...
+@test length(HLJSFILES) == 2
+
+# Some failure modes
+@test_throws Base.IOError makedocs(;
+    sitename = "Prerendering code blocks",
+    format = Documenter.HTML(
+        prerender = true,
+        node = "nope",
+    ),
+)
```
mortenpi, line 95: Could we wrap this whole thing in a testset so that we'd get "n of n" successful tests report on CI?

mortenpi (approved):

Issue #491, opened 2017-05-25
```@repl block inside !!! note doesn't show output


shashi:
`@repl` block inside an `!!! note` is neither stylized as the REPL nor is the output shown. Is this intended?

mortenpi:
Can confirm. Other special blocks seem to be ignored in admonitions as well. Thanks for the report!

mortenpi:
I'm rather sure that the issue comes from [`src/Expanders.jl#L25-L32`](https://github.com/JuliaDocs/Documenter.jl/blob/6445adc41b1c27aa00f0784ca64571aefca9d339/src/Expanders.jl#L25-L32) -- the expander functions are only called on top-level blocks, and not recursively through the AST.

nluetts:
I just stumbled upon the same problem---is it not resolved yet? Or is there a workaround?

KristofferC:
>  the expander functions are only called on top-level blocks, and not recursively through the AST.

Even if you call it recursively throughout the AST that is not enough. You will get an expanded block entered into `page.mapping` but that will not be used by the renderer. You probably need to propagate this mapping all the way down to `mdconvert`.

ettersi:
I'd really like to see this fixed, and I'm happy to give a hand doing so. However, since there have already been two attempts which seem to eventually have run out of steam, I thought I'd ask for some input first on what people think would be the best way to resolve this.

KristofferC:
Is there anything wrong with https://github.com/JuliaDocs/Documenter.jl/pull/1747?

mortenpi:
#1747 contains some unnecessary changes as far as I can tell. But I think the general approach there (create a new expander step that recurses for certain elements, but only calls as subset of expander steps then) is good. It needs to be updated to the new MarkdownAST-based tree though. @ettersi I'd say you're welcome that adopt that PR finish it up if you'd like.

ettersi:
What's the recommended way to "adopt" a PR? I believe I can't fork a PR. I can of course create a new PR, but that would erase @jmert from the history, which doesn't seem fair.

mortenpi:
You'd have to start a new PR on your fork, but you can use their latest commit as a starting point (although, it might be easier to just re-do it from scratch, since merging with master might be more work at this point). One way or another, I'll make sure that there is a `Co-Authored-By` in the commit message when we merge.

ettersi:
Done: https://github.com/JuliaDocs/Documenter.jl/pull/1970

PR #1637, opened 2021-07-07
WIP: enable `@repl`, `@example` etc in admonition, lists etc

KristofferC:
This is a WIP to show one possible way of solving https://github.com/JuliaDocs/Documenter.jl/issues/491. It does this by recursing through the AST expand non-root nodes. Then in the HTML writer, the mapping created is passed through down all the way to `mdconvert` in a keyword argument and we then look up if we have an expanded version to use.

For a markdown with the following content:

````md
- foo
  ```@repl
  1+1
  ```
  - bar
    ```@example
    1+1 == 2
    ```

!!! warning
    ```@repl
    1+1
    ```

```@repl
2+2
````

it generates a HTML page as:

bild

No tests or anything yet because it would be good to get some feedback if the proposed method makes sense.

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -39,11 +39,26 @@ function expand(doc::Documents.Document)
         @debug "Running ExpanderPipeline on $src"
         empty!(page.globals.meta)
         for element in page.elements
-            Selectors.dispatch(ExpanderPipeline, element, page, doc)
+            run_dispatch(element, page, doc)
         end
         pagecheck(page)
     end
 end
+
+run_dispatch(element, page, doc) = Selectors.dispatch(ExpanderPipeline, element, page, doc)
+function run_dispatch(element::Markdown.List, page, doc)
+    Selectors.dispatch(ExpanderPipeline, element, page, doc)
+    foreach(x -> run_dispatch(x, page, doc), element.items)
+end
+function run_dispatch(element::Markdown.Admonition, page, doc)
+    Selectors.dispatch(ExpanderPipeline, element, page, doc)
+    foreach(x -> run_dispatch(x, page, doc), element.content)
+end
+function run_dispatch(element::Vector, page, doc)
+    foreach(x -> run_dispatch(x, page, doc), element)
+end
+
+
 
 # run some checks after expanding the page
 function pagecheck(page)
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -1358,13 +1358,13 @@ function domify(ctx, navnode)
     map(page.elements) do elem
         rec = SearchRecord(ctx, navnode, elem)
         push!(ctx.search_index, rec)
-        domify(ctx, navnode, page.mapping[elem])
+        domify(ctx, navnode, page.mapping[elem], page.mapping)
     end
 end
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/Expanders.jl [hunk 1]
```diff
@@ -39,12 +39,27 @@ function expand(doc::Documents.Document)
...
+    Selectors.dispatch(ExpanderPipeline, element, page, doc)
+    foreach(x -> run_dispatch(x, page, doc), element.items)
+end
+function run_dispatch(element::Markdown.Admonition, page, doc)
+    Selectors.dispatch(ExpanderPipeline, element, page, doc)
+    foreach(x -> run_dispatch(x, page, doc), element.content)
+end
+function run_dispatch(element::Vector, page, doc)
+    foreach(x -> run_dispatch(x, page, doc), element)
+end
```
mortenpi, line 59: Why can't these be methods of `Selectors.dispatch`?

src/Writers/HTMLWriter.jl
```diff
@@ -1754,7 +1757,9 @@ function mdconvert(link::Markdown.Link, parent; droplinks=false, kwargs...)
     droplinks ? link_text : Tag(:a)[:href => link.url](link_text)
 end
 
-mdconvert(list::Markdown.List, parent; kwargs...) = (Markdown.isordered(list) ? Tag(:ol) : Tag(:ul))(map(Tag(:li), mdconvert(list.items, list; kwargs...)))
+function mdconvert(list::Markdown.List, parent; kwargs...)
+    (Markdown.isordered(list) ? Tag(:ol) : Tag(:ul))(map(Tag(:li), mdconvert(list.items, list; kwargs...)))
```
mortenpi, line 1761:
```suggestion
    Markdown.isordered(list) ? Tag(:ol) : Tag(:ul))(map(Tag(:li), mdconvert(list.items, list; kwargs...))
```

Although could also be split into a multiline `if ... else ... end` for readability.

src/Writers/HTMLWriter.jl
```diff
@@ -1686,7 +1686,10 @@ function mdconvert(text::AbstractString, parent; kwargs...)
     return DOM.Node(text)
 end
 
-mdconvert(vec::Vector, parent; kwargs...) = [mdconvert(x, parent; kwargs...) for x in vec]
+function mdconvert(vec::Vector, parent; kwargs...)
+    mapping = get(kwargs, :mapping, nothing)
```
mortenpi, line 1690:
```suggestion
function mdconvert(vec::Vector, parent; mapping = nothing, kwargs...)
```

More idiomatic pehaps?

mortenpi (commented):
As implemented currently, it would expand _all_ at-blocks? I think it would make sense to run the code and at-meta blocks. at-index and at-contents.. is probably fine as well -- my main concern is whether they get handled by the CSS correctly, but since they should just be lists..

But I don't think at-docs and at-autodocs should get expanded outside of the top level.

KristofferC:
> As implemented currently, it would expand all at-blocks?

Yes, I can reduce it to only apply to "cody"-blocks.

mortenpi:
> Yes, I can reduce it to only apply to "cody"-blocks.

Yep, that sounds reasonable.

PR #1641, opened 2021-07-09
Theme the `julia>` prompt in the `julia-repl` language

fredrikekre:
Theme the `julia>` prompt in the `julia-repl` language as green using the same color as the ANSI color green for respective theme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   ```
   _This feature is experimental and subject to change in future releases._ ([#1627][github-1627])
 
+* ![Enhancement][badge-enhancement] The `julia>` prompt is now colored in green in the `julia-repl` language highlighting. ([#1639][github-1639], [#1641][github-1641])
+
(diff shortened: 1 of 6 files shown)


fredrikekre:
Screenshot from 2021-07-09 12-32-31
Screenshot from 2021-07-09 12-32-50

kimikage:
We could use the variable `$ansi-green` by writing the patch in `documenter-{dark|light}.scss` instead of `documenter/_patches.scss`. Although I can't say that is generally a good practice, I personally prefer the style with fewer hard-coded constants.

fredrikekre:
Good idea, I moved it.

fredrikekre:
Prompts have their own class now: https://github.com/highlightjs/highlight.js/pull/3490

mortenpi:
> Prompts have their own class now: [highlightjs/highlight.js#3490](https://github.com/highlightjs/highlight.js/pull/3490)

I noticed that in #1846. Should we remove the override?

PR #1645, opened 2021-07-14
Add hljs class to all code blocks

fredrikekre:
This adds hljs to all code blocks for two reasons:

 - nohighlight blocks are left untouched by HighlightJS, meaning that
   the hljs class isn't attached at runtime to those blocks. Adding this
   explicitly makes sure that CSS rules for code blocks still apply
   such that the correct "regular text color" is used.
 - If JavaScript is disabled in the browser this makes sure that at
   least the intended "regular text color" is used for code blocks.


fredrikekre:
Before:
Screenshot from 2021-07-14 13-50-09

After:
Screenshot from 2021-07-14 13-49-29

(The green color is just to show what is highlighted with the code color).

Issue #1629, opened 2021-07-06
ANSI color tweaks


fredrikekre:
I think in dark mode things look great, but in light mode `yellow`, `light_gray` and `light_yellow` are very hard to see

![Screenshot from 2021-07-06 16-42-42](https://user-images.githubusercontent.com/11698744/124621481-9a3d0a00-de7a-11eb-8930-ac5e9c27fe0c.png)
![Screenshot from 2021-07-06 16-43-44](https://user-images.githubusercontent.com/11698744/124621485-9ad5a080-de7a-11eb-9a2b-811e79a84465.png)

fredrikekre:
cc @kimikage

kimikage:
I was initially thinking of adjusting the color scheme of Documenter itself. (Whether my proposal is good or bad, I feel that the caption of Tip Admonition has too low contrast.)
![admonition](https://user-images.githubusercontent.com/12679384/125014442-b0301200-e0a8-11eb-8075-4832976bc558.png)
However, I am skeptical about such a color scheme change should be made in the patch version.
So, I think we have roughly three options:
1. tweak the ANSI colors only
2. tweak the basic color scheme, make the admonition colors look like the current scheme, and tweak the ANSI colors
3. tweak the basic color scheme (including the admonition colors) and the ANSI colors

kimikage:
I am busy for a while and am not at the stage where I can propose a new color scheme for Documenter yet. However, I would like to make a preliminary proposal for the part about ANSI colors. (Hopefully this weekend.)

I'm not really interested in changing the appearance of Documenter's default theme, I'm more interested in updating the theme to make it easier to customize.

kimikage:
WIP: https://kimikage.github.io/july-morning-theme/previews/PR1/design-language/

mortenpi:
I think the color scheme you're proposing @kimikage is a nice improvement over the current one, especially in the light theme. On the other hand, it is a perhaps noticeable, but not a major overhaul of the colors, so I would not mind having it in a patch release (I think it's a strict improvement over the current color scheme).

> WIP: https://kimikage.github.io/july-morning-theme/previews/PR1/design-language/

This is also really cool. Do you reckon we could make it part of Documenter's documentation?

PR #1646, opened 2021-07-16
Set version to 0.27.4.

fredrikekre:


kimikage:
Should the issue #1629 be included in 0.27.4?

PR #1647, opened 2021-07-17
[RFC] Tweak color schemes

### LINKED ISSUE #1629: ANSI color tweaks

kimikage:
This will update the definition of basic color variables such as `$red`. This also tweaks the ANSI colors by deriving from those basic colors.
However, the colors for admonitions etc. are kept as they are, for compatibility.

light
dark

Closes #1629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Version `v0.27.4`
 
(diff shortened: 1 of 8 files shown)


mortenpi:
This LGTM, thanks @kimikage! Is it ready to go on your end as well or was there something else you wanted to tweak here?

kimikage:
> Is it ready to go on your end as well or was there something else you wanted to tweak here?

Perhaps I may want to make a few more tweaks in the future, but the changes are likely to be minor and can be addressed in a future version, not here. :smile:

My concern is that not a few people would prefer a vivid theme like Monokai.

Issue #1612, opened 2021-06-24
allow plugins to pass additonal css-style additions


LazyScholar:
It would be nice to allow plugins like `DocumenterCitations.jl` to pass css-style additions to the `Documenter.jl` html-writer assets.

This would allow to add own plugin specific css styles to enhance readability of elements which are added to the HTML documentation by a plugin (without starting to clutter the `Documenter.jl` default style).

This might also allow the creation of plugins with advanced `.js` capabilities (a option to limit additions of assets like `.css` or `.js` to the scope of the plugin (instead of each page) might be needed here).

Currently this could only be done by raw html insertion into the markdown (by the user or the plugin) or by element wise css modification by the plugin.

----

__Examples which could be improved by small plugin css additions:__

[`DocumenterCitations.jl`](https://github.com/ali-ramadhan/DocumenterCitations.jl)
````
```@raw html
<style>
dl {
display: grid;
grid-template-columns: max-content auto;
}
dt {
grid-column-start: 1;
}
dd {
grid-column-start: 2;
margin-bottom: 0.75em;
}
</style>
```

```@bibliography
```
````
pure  |  with css addition
:-------------------------:|:-------------------------:
![bib1](https://user-images.githubusercontent.com/75576197/123248614-695aed80-d4e8-11eb-85bc-592291b7e481.png)  |   ![bib2](https://user-images.githubusercontent.com/75576197/123248618-69f38400-d4e8-11eb-8e25-e4e75170caa8.png)

[`DocStringExtensions.jl`](https://github.com/JuliaDocs/DocStringExtensions.jl)
````
```@raw html
<style>
#horlist ul {
display: grid;
grid-template-columns: repeat(auto-fit, minmax(25ch, 1fr));
grid-gap: 4px;
list-style: none;
}
</style>
<div id="horlist">
```

```@docs
NURBStoolbox
```

```@raw html
</div>
```
````
pure  |  with css addition
:-------------------------:|:-------------------------:
![exp1](https://user-images.githubusercontent.com/75576197/123249077-e6866280-d4e8-11eb-9986-d6e55dae3295.png) | ![exp2](https://user-images.githubusercontent.com/75576197/123249078-e71ef900-d4e8-11eb-830b-64103cbd0d29.png)

[`DocStringExtensions.jl`](https://github.com/JuliaDocs/DocStringExtensions.jl)
````
```@raw html
<style>
#horfield li > p + p::before {
content: ": ";
}
[...]

Issue #1575, opened 2021-05-05
Matrix with three rows shows a vertical scrollbar


jaakkor2:
Input like this shows a vertical scrollbar.  Documenter v0.26.3.

```
```math
A = \begin{bmatrix}
a & b\\
c & d
\end{bmatrix}
``` .
```math
B = \begin{bmatrix}
a & b & c\\
d & e & f\\
g & h & i
\end{bmatrix}
``` .
```
![documenter](https://user-images.githubusercontent.com/7318249/117127987-7e978380-ada5-11eb-9221-ecb621ee2afe.png)

LazyScholar:
Add there
https://github.com/JuliaDocs/Documenter.jl/blob/c942201b745abbdf155c880e1231cb8240507b41/assets/html/scss/documenter/layout/_main.scss#L12-L14
```
overflow-y: hidden;
```

or
add here
https://github.com/JuliaDocs/Documenter.jl/blob/c942201b745abbdf155c880e1231cb8240507b41/assets/html/scss/documenter/_patches.scss#L70-L72
```
padding: 0.5em 0 !important;
```
(this will lead increase the overall y size of the page (more y space between formulas))
I would prefer the first solution as it does not change the overall y-spacing although i could not test it for other use cases.

Issue #1433, opened 2020-10-08
Location of JS/script assets in the HTML


mortenpi:
The user can include additional JS scripts in the HTML page by passing the URLs via the `assets` keyword to `HTML`. Currently, the generated `<script>` tags get placed after the built-in dependencies (including RequireJS, [which can cause issues](https://github.com/JuliaDocs/Documenter.jl/issues/1247#issuecomment-703312434)). Based on that discussion, it looks like we could handle it better:

- [ ] By default, place the user-provided `<script>` tags _before_ the RequireJS loading (works around the issues with libraries that do not play well with RequireJS if they are loaded separately: https://github.com/requirejs/requirejs/issues/947#issuecomment-28677913).
- [ ] Allow the user to customize the location with a keyword to `asset()` (e.g. before, after RequireJS, at the end of the HTML file).

baptiste:
Have you seen workarounds for three.js? I'm hoping to use it for 3D visuals in my docs, but unless I manually move the `<script>` above requirejs it doesn't seem to work. (strangely [it kind of works](http://nano-optics.ac.nz/CoupledDipole.jl/dev/conventions/) if I use an old version of three.js, but there are too many features missing)

ericphanson:
This would be useful for https://github.com/asciinema/asciinema-player too, to include asciinema gifs. Re- https://github.com/asciinema/asciinema-player/issues/142

PR #1648, opened 2021-07-17
API for plugin assets

LazyScholar:
First time melding here. Hope i did not break anything.

### DIFF
--- a/src/Writers/HTMLWriter.jl
+++ b/src/Writers/HTMLWriter.jl
@@ -76,6 +76,25 @@ const ASSETS_SASS = joinpath(ASSETS, "scss")
 "Directory for the compiled CSS files of the themes."
 const ASSETS_THEMES = joinpath(ASSETS, "themes")
 
+"""
+    HTMLAsset
+
+Structure to hold a HTML asset.
+
+# Fields:
+
+- `class::Symbol`: symbols indicating the asset type (`:ico`, `:js` or `:css`)
+  They become a `<script>`, `<link rel="stylesheet" type="text/css">` or
+  `<link rel="icon" type="image/x-icon">` elements in `<head>`, respectively.
+
+- `uri::String`: address indicating the source location on the web or in
+  relation to the main page
+
+- `islocal::Bool`: boolean value indicating if the asset is located locally
+  (`islocal=true`)
+
+- `attributes::Dict{Symbol, String}`
+"""
 struct HTMLAsset
     class :: Symbol
     uri :: String
@@ -668,11 +687,13 @@ mutable struct HTMLContext
     search_index :: Vector{SearchRecord}
     search_index_js :: String
     search_navnode :: Documents.NavNode
+    asset_list::Vector{HTMLAsset}
 end
 
 HTMLContext(doc, settings=nothing) = HTMLContext(
     doc, settings, [], "", "", "", "", [], "",
     Documents.NavNode("search", "Search", nothing),
+    HTMLAsset[]
 )
 
 struct DCtx
@@ -803,6 +824,9 @@ function render(doc::Documents.Document, settings::HTML=HTML())
         copy_asset("themes/$(theme).css", doc)
     end
 
+    # read plugin assets and copy the files to the build directory
+    plugin_assets!(ctx)
+
     for page in keys(doc.blueprint.pages)
         idx = findfirst(nn -> nn.page == page, doc.internal.navlist)
         nn = (idx === nothing) ? Documents.NavNode(page, nothing, nothing) : doc.internal.navlist[idx]
@@ -847,6 +871,70 @@ function copy_asset(file, doc)
     assetpath = normpath(joinpath("assets", file))
     # Replace any backslashes in links, if building the docs on Windows
     return replace(assetpath, '\\' => '/')
+end
+
+# Plugin
+# ------------------------------------------------------------------------------
+
+"""
+    plugin_html_assets(::Documenter.Plugin)::Tuple{String,Vector{HTMLAsset}}
+
+This function allows plugins to register additional assets to the `HTMLWriter`.
+It should return a tuple with a string and a vector of [`HTMLAsset`](@ref).
+
+The string has to denote the absolute path to the directory containing the
+plugin assets. This will be used to copy the directory to the build plugin
+directory. This string can be empty if no local assets need to be registered
+omiting the copy of a plugin directory.
+The paths of local assets in the [`HTMLAsset`](@ref) vector will be modifyed by
+prepending `plugin/` to their relative `uri`.
+"""
+function plugin_html_assets(::Documenter.Plugin)
+    return "", HTMLAsset[]
+end
+
+"""
+    plugin_assets!(ctx::HTMLContext)
+
+This function will call the registering function [`plugin_html_assets`](@ref)
+for all plugins. In case that they return a vector of [`HTMLAsset`](@ref) those
+will be added to the `asset_list` field of `ctx` ([`HTMLContext`](@ref)).
+In the case that they return a non empty string the denoted directory will be
+copied to the directory `plugins` of build location.
+"""
+function plugin_assets!(ctx::HTMLContext)
+    plugin_dir = joinpath(abspath(ctx.doc.user.build),"plugin")
+    for (type, plugin) in ctx.doc.plugins
+        path, assets = plugin_html_assets(type)
+        # copy provided directory to the build directory
+        if !isempty(path)
+            !ispath(plugin_dir) && mkdir(plugin_dir)
+            target_dir = joinpath(plugin_dir,basename(path))
+            !isabspath(path) && !ispath(path) && error("Source directory '$path' provided by the plugin '$type' is not a valid path!")
+            ispath(target_dir) && @warn "Directory '$target_dir' already exists (overwritten)!"
+            cp(path, target_dir; force=true)
+        end
+        # add the assets to the asset_list vector and modify the path in the case
+        # of local paths
+        if !isempty(assets)
+            for i in 1:length(assets)
+                if assets[i].islocal
+                    file_path = normpath(joinpath(path, "..", assets[i].uri))
+                    if !isfile(file_path)
+                        @warn "File '$file_path' does not exist!"
+                        continue
+                    end
+                    file_path = joinpath(plugin_dir, assets[i].uri)
+                    if !isfile(file_path)
+                        @warn "File '$file_path' not found in build directory!"
+                        continue
+                    end
+                    assets[i] = HTMLAsset(assets[i].class, joinpath("plugin", assets[i].uri), true, assets[i].attributes)
+                end
+            end
+            append!(ctx.asset_list, assets)
+        end
+    end
 end
 
 # Page
@@ -1004,6 +1092,8 @@ function render_head(ctx, navnode)
         script[:src => relhref(src, ctx.themeswap_js)],
         # Custom user-provided assets.
         asset_links(src, ctx.settings.assets),
+        # General list of html assets.
+        asset_links(src, ctx.asset_list),
     )
 end
 


fredrikekre:
Thanks! Can you make 2 PRs instead? The first commit seems like a bugfix and could be merged as is I think while the second commit might need some discussion?

LazyScholar:
I could do it. Although the first PR is a very minor fix. So it can wait (unless you want to keep PRs for issues strictly separated).

mortenpi:
> I could do it. Although the first PR is a very minor fix. So it can wait (unless you want to keep PRs for issues strictly separated).

Generally we like to keep unrelated changes in different PRs, so it would be awesome if you could open a separate one for #1575.

mortenpi:
Re #1612 PR: why are you looping over the builder pipeline when trying to check for assets? Wouldn't looping over the `plugins` field of `Document` would make more sense?

LazyScholar:
> > I could do it. Although the first PR is a very minor fix. So it can wait (unless you want to keep PRs for issues strictly separated).
> 
> Generally we like to keep unrelated changes in different PRs, so it would be awesome if you could open a separate one for #1575.

Done.

LazyScholar:
> Re #1612 PR: why are you looping over the builder pipeline when trying to check for assets? Wouldn't looping over the `plugins` field of `Document` would make more sense?

I used `DocumenterCitatitions.jl` as a base to try an add that issue. I was unsure if i it would break anything if i added the information into the struct
https://github.com/ali-ramadhan/DocumenterCitations.jl/blob/a5513b091b984aea0c4753eb2a984174989da2e6/src/DocumenterCitations.jl#L18-L20
or if it would be more fitting to register such type called functions
https://github.com/ali-ramadhan/DocumenterCitations.jl/blob/a5513b091b984aea0c4753eb2a984174989da2e6/src/citations.jl#L2-L9
?

The first way would be better for performance and the second would allow to have page specific assets.

As of now `Documenter.jl` seems to use the same assets on each page so doing the first one would be also ok.

Either way why does `Documenter.jl`not build the asset list once and adds it to each page instead of rebuilding it every time (ref. Issue 1433)?

LazyScholar:
Okay i found the posiotion where where the plugin structs are stored. So looping over provided data structs is possible.
I can change it in a way that assets and directory paths stored there will be used.

If done that way there should be a note in the docs that those structs either have to contain two fields `assets::Vector{HTMLAsset}` + `asset_path::String` or that there should not exist fields with the same names/symbols with incompatible values.

Shall i change it that way? If so can you point me to the location in the docs for such a note?

LazyScholar:
I like this more. Is that ok so?
Only think that is missing is where to note that the plugin struct fields `asset_path` & `asset_list` in
```julia
struct PluginStructure <: Documenter.Plugin #passed to makedocs(plugin_struct, )
    some_plugin_data::Any
    asset_path::String
    asset_list::Vector{HTMLAsset}
end
```
are reserved for HTML asset additions.

LazyScholar:
And this is looping over the plugin struct types.

LazyScholar:
I deleted some dabbling comments i made while experimenting with other solutions.
I settled with those two commits:
- f6d17a0 : asset registration by  `asset_path::String`, `asset_list::Vector{HTMLAsset}` fields passed by the plugin struct on call of `makedocs` -> needs some notes in the docs
- 77b7baf : asset addition by type registered functions called by looping over plugins from within `HTMLWriter.jl`

src/Writers/HTMLWriter.jl
```diff
@@ -757,10 +757,12 @@ function render(doc::Documents.Document, settings::HTML=HTML())
         idx = findfirst(nn -> nn.page == page, doc.internal.navlist)
         nn = (idx === nothing) ? Documents.NavNode(page, nothing, nothing) : doc.internal.navlist[idx]
         @debug "Rendering $(page) [$(repr(idx))]"
-        render_page(ctx, nn)
+        render_page(doc, ctx, nn)
```
mortenpi, line 760: You don't need to pass `doc`. You can always access it via `ctx.doc`.

mortenpi (commented):
Thanks for iterating on this! This exposes a completely new API for plugin writers, so it's worth taking some time to get this right.

* I think it would be better to rely on multiple dispatch, rather than forcing the plugin writer to implement particular fields/properties. How about a function, like `extra_html_assets(::Plugin)`, that each plugin could implement, and would return the list of `HTMLAssets`?
* Could you also add a new documentation page where we could tart documenting this API? I think it might be worth documenting it first to get the design down, and then focus on the implementation details.

src/Writers/HTMLWriter.jl
LazyScholar: Oh i did not knew that i will change that.

LazyScholar: > Thanks for iterating on this! This exposes a completely new API for plugin writers, so it's worth taking some time to get this right.

I agree.

>     * I think it would be better to rely on multiple dispatch, rather than forcing the plugin writer to implement particular fields/properties. How about a function, like `extra_html_assets(::Plugin)`, that each plugin could implement, and would return the list of `HTMLAssets`?

Like those functions? They should do that only the names should be changed or are they not what you have in mind?
https://github.com/JuliaDocs/Documenter.jl/blob/77b7baf5911b3155332565e11020106afca49a8a/src/Writers/HTMLWriter.jl#L964-L983
Although the plugin struct fields could be optional. The moment fields with the same names would be used then there would it be necessary to check collisions/false input.

>     * Could you also add a new documentation page where we could tart documenting this API? I think it might be worth documenting it first to get the design down, and then focus on the implementation details.

A extra "Plug-in API" page Or a "How to Plug-in" page? I could start a page but i originally just wanted to add this and maybe refactor the asset addition structure for issue 1433. So i wont polish it, sorry.

TODO: I should also do asset input checking like here:
https://github.com/JuliaDocs/Documenter.jl/blob/77b7baf5911b3155332565e11020106afca49a8a/src/Writers/HTMLWriter.jl#L440-L444

LazyScholar:
> Thanks for iterating on this! This exposes a completely new API for plugin writers, so it's worth taking some time to get this right.

I agree.

>     * I think it would be better to rely on multiple dispatch, rather than forcing the plugin writer to implement particular fields/properties. How about a function, like `extra_html_assets(::Plugin)`, that each plugin could implement, and would return the list of `HTMLAssets`?

Like those functions? They should do that only the names should be changed or are they not what you have in mind?
https://github.com/JuliaDocs/Documenter.jl/blob/77b7baf5911b3155332565e11020106afca49a8a/src/Writers/HTMLWriter.jl#L964-L983
Although the plugin struct fields could be optional. The moment fields with the same names would be used then there would it be necessary to check collisions/false input.

>     * Could you also add a new documentation page where we could tart documenting this API? I think it might be worth documenting it first to get the design down, and then focus on the implementation details.

A extra "Plug-in API" page Or a "How to Plug-in" page? I could start a page but i originally just wanted to add this and maybe refactor the asset addition structure for issue 1433. I will start it but i wont polish/finish it, sorry.

TODO: I should also do asset input checking like here:
https://github.com/JuliaDocs/Documenter.jl/blob/77b7baf5911b3155332565e11020106afca49a8a/src/Writers/HTMLWriter.jl#L440-L444

src/Writers/HTMLWriter.jl
LazyScholar: Changed it.

mortenpi:
> Like those functions?

Ah, yes. I think I was looking at an old commit the other day. But hmm. A couple thoughts:

* For full flexibility, I think it might be better if the user specifies explicitly each and every asset that gets copied, rather than just overlaying a whole directory onto `build/`. I reckon in most cases it will be like 1 JS file and 1 CSS file that the plugin wants to add anyway.
* I think it would be cleaner and easier for the plugin writer if this was a single function to implement, rather than multiple.

We could maybe expand `HTMLAsset` in a way that you can also specify a source and destination file.

```julia
struct HTMLAsset
    class :: Symbol
    uri :: String
    source :: Union{String, Nothing}
...
```

where `source` is a full path to a local asset files, and a `nothing` value would correspond to `local = false`. `HTMLAsset` could then have a new constructor:

```julia
HTMLAsset(class::Symbol, src::String, dst::String = joinpath("assets", basename(src)))
```

The most likely case for the plugin writer would be doing something like  `HTMLAsset(:js, "path/to/myfancy.js")`, which will end up copying the files to `assets/myfancy.js` (I addition to including remote assets). And all the information will be contained in a single array of `HTMLAsset`s.

Do you have any thoughts or see any issues with this design?

> A extra "Plug-in API" page Or a "How to Plug-in" page?

Could just have a page called "Plugins" before between "Showcase" and "Library". Just a brief example showing how to implement your own plugin would be enough.

LazyScholar:
> Do you have any thoughts or see any issues with this design?

I think that would be perfect. So registering the `HTMLAsset` will push them into a vector. Which can be looped over to gather all assets. And with that static vector one can build the asset links of each page much cleaner. 

Which leads me to the next points i would change. Adding a additional field like `index::AbstractFloat` (defaulting to `0.0`) allowing to sort the assets. And refactoring the `asset()` calls in order to prepare for issue #1433 and set indices in order to place them in a certain order on the HTML header/footer (`index >= 0` are on header and `index < 0` are on the footer).

> > A extra "Plug-in API" page Or a "How to Plug-in" page?
> 
> Could just have a page called "Plugins" before between "Showcase" and "Library". Just a brief example showing how to implement your own plugin would be enough.

Okay, so a "How to Plug-In" page. I will write something up on how far i understand it.

LazyScholar:
> Do you have any thoughts or see any issues with this design?

Okay i had a look on how to implement. I suggest to make a abstract super type `HTMLAsset` and subtypes `HTMLAssetLocal`/`HTMLAssetGlobal` or two differently named wrapper functions. As the constructors will get either
- `class::Symbol`
- `source::String`
- `dst::String` optional -> defaulting to a string
or
- `class::Symbol`
- `uri::String`
which might lead to collisions if called with two arguments?
Or for global assets we enforce a third argument either empty string or a `false`. The `Union{String, Nothing}` could be dropped in case of two different subtypes or implied by a empty string?

LazyScholar:
Changes on 4407a4c:
- moved the checks of `HTMLAsset` to `asset()` and added checks for new `uri`&`src` fields
- added documentation on `HTMLAsset` and `asset()`
- now the plugin assets get collected into `HTMLContext.asset_list`
- after collecting all plugin assets the files listed on that list will be copied over to the build directory
- the `HTMLContext.asset_list` is used to create the additional DOM entries the same way as the user provided `HTML.assets` list

LazyScholar:
Still pondering on how to write a short overview on "How to Plug-in" for the docs.
I am afraid it is no easy task. Looking at how `DocumenterCitations.jl` does the hook into `Documenter.jl`it is necessary to include quite a few sub modules and functions one wants to hook into?

mortenpi (commented):
Sorry for the delay. Heaps of comments, but I like where this is going.

I think we're subtly changing the behaviour of `asset`, so I need to give that a bit more thought.

> Adding a additional field like index::AbstractFloat (defaulting to 0.0) allowing to sort the assets.

I am not a huge fan of this idea -- it doesn't really reliably guarantee anything. In principle, the assets from different plugins should not interact with each other. Of course, we don't live in an ideal world.. and this approach would sometimes allow for a workaround.

Do you think it would be enough if we guarantee that the assets get inserted in the order in which the plugins are passed to `makedocs`?

> And refactoring the asset() calls in order to prepare for issue #1433 and set indices in order to place them in a certain order on the HTML header/footer (index >= 0 are on header and index < 0 are on the footer).

Let's leave that for a separate PR.

src/Writers/HTMLWriter.jl
```diff
@@ -74,39 +74,66 @@ const ASSETS_SASS = joinpath(ASSETS, "scss")
...
 struct HTMLAsset
     class :: Symbol
     uri :: String
     islocal :: Bool
-
-    function HTMLAsset(class::Symbol, uri::String, islocal::Bool)
-        if !islocal && match(r"^https?://", uri) === nothing
-            error("Remote asset URL must start with http:// or https://")
-        end
-        class in [:ico, :css, :js] || error("Unrecognised asset class $class for `$(uri)`")
```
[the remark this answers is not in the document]
LazyScholar: with `!occursin(r"^https?://", src)` i assume

src/Writers/HTMLWriter.jl
```diff
@@ -117,21 +144,60 @@ Documenter.HTML(assets = [
...
+    # no file or path checks are done within this function!
+    if islocal === nothing
+        islocal = match(r"^https?://", src) === nothing
+    elseif islocal::Bool
+        if !islocal && match(r"^https?://", src) === nothing
+            error("Remote asset URL must start with http:// or https:// , asset: $(src)")
+        end
+    else
+        error("Not A valid type for islocal for: $(src)")
+    end
```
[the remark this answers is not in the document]
LazyScholar: Changed it to `typeof(islocal)<:Bool`. I disagree with `islocal :: Union{Bool, Nothing}` as it is either or.

src/Writers/HTMLWriter.jl
```diff
@@ -955,9 +1026,25 @@ function render_head(ctx, navnode)
         script[:src => relhref(src, ctx.themeswap_js)],
         # Custom user-provided assets.
         asset_links(src, ctx.settings.assets),
+        asset_links(src, ctx.asset_list),
     )
 end
 
+"""
+    add_plugin_assets(::Type{T})::Vector{HTMLAsset} where {T<:Documenter.Plugin}
```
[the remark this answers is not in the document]
LazyScholar: I go with `plugin_html_assets`.

src/Writers/HTMLWriter.jl
```diff
@@ -955,9 +1026,25 @@ function render_head(ctx, navnode)
...
+
+This function allows plugins to register additional assets to the `HTMLWriter`.
+It should return a vector of [`HTMLAsset`](@ref) simillar to the example of
+[`asset`](@ref).
+
+!!! info "Note:"
+    Please note that contrary to the regular use of the [`asset`](@ref)
+    function `src` has to be provided as absolute path for local assets!
+"""
+function add_plugin_assets(::Type{T})::Vector{HTMLAsset} where {T<:Documenter.Plugin}
```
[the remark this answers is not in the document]
LazyScholar: I changed it to `function add_plugin_assets(::Documenter.Plugin)::Vector{HTMLAsset}` as it has to be `Vector{HTMLAsset}` in order to be compatible with `HTMLContext.asset_list`.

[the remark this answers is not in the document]
LazyScholar: The function of the plugin writer won't have the output restriction as it is a "new function" but i'd like to have that in the docs as it is the blueprint function ("oh so is has to be named `plugin_html_assets` and it should return `typeof(HTMLAsset[])`).

src/Writers/HTMLWriter.jl
```diff
@@ -117,21 +144,60 @@ Documenter.HTML(assets = [
...
+    asset("https://example.com/fonts", class=:css),
     # Same as above, but for a local asset
     asset("asset/foo.script", class=:js, islocal=true),
+    # Asset provided via source and destination path
+    asset("/abs/path/to/extra_style.css", dst="asset/extra_style.css")
 ])
 ```
 """
-function asset(uri; class = nothing, islocal=false)
+function asset(src::String; class = nothing, islocal = nothing, dst = "")
```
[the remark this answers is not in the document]
LazyScholar: I somehow can't befriend me with functions without type restrictions and default values of nothing...
Sure in some cases it is useful to not restrict the types but it leads to more work. As now one has to test for valid input types and valid arguments on top. Isn't that one of the reasons to use types to have the function refuse obviously wrong arguments?
In the case of the optional arguments here i can understand it as they might be derived of the `src<:AbstractString`.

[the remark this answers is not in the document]
mortenpi: Ah, but in this case, `nothing` indicates a missing `dst` value, i.e. that the destination path is not set, instead of a simple default path. To me, using an empty string to indicate a missing value is similar to using `-1` for that if you know that otherwise you're expecting a positive number --- feels ugly in an age where type unions are cheap. :slightly_smiling_face: In any case, thanks for changing it!

src/Writers/HTMLWriter.jl
```diff
@@ -958,9 +1038,25 @@ function render_head(ctx, navnode)
...
         asset_links(src, ctx.settings.assets),
+        asset_links(src, ctx.asset_list),
     )
 end
 
+"""
+    plugin_html_assets(::Documenter.Plugin)::Vector{HTMLAsset}
+
+This function allows plugins to register additional assets to the `HTMLWriter`.
+It should return a vector of [`HTMLAsset`](@ref) simillar to the example of
```
mortenpi, line 1049:
```suggestion
It should return a vector of [`HTMLAsset`](@ref) similar to the example of
```

src/Writers/HTMLWriter.jl
```diff
@@ -958,9 +1038,25 @@ function render_head(ctx, navnode)
...
+[`asset`](@ref).
+
+!!! info "Note:"
+    Please note that contrary to the regular use of the [`asset`](@ref)
+    function `src` has to be provided as absolute path for local assets!
+"""
+function plugin_html_assets(::Documenter.Plugin)::Vector{HTMLAsset}
+    return HTMLAsset[] # return vector of assets which will be added
+end
```
mortenpi, line 1058: We can simplify this down to a one-liner:

```suggestion
plugin_html_assets(::Documenter.Plugin) = HTMLAsset[]
```

Regarding the `::Vector{HTMLAsset}` return type part -- as you said, this only applies to this particular method, so it's superfluous here (as the function body is trivial). However, we do want to keep it in the docstring, since we expect the plugin writers to return `::Vector{HTMLAsset}` (although, in principle, anything iterable with `eltype()=HTMLAsset` would also be okay).

src/Writers/HTMLWriter.jl
```diff
@@ -977,6 +1073,41 @@ function asset_links(src::AbstractString, assets::Vector{HTMLAsset})
...
 
+function plugin_assets(ctx::HTMLContext)
+    for (type, plugin) in ctx.doc.plugins
+        assets = plugin_html_assets(type)
+        if !isempty(assets)
+            # no path checking is done here as the plugin authors have to
+            # take care themselves that valid paths are provided (as the path
+            # completion for the relative source files in the asset()
+            # function will only complete for the docs/src directory)!
+            # so @blugin_author: no fancy asset() autocompletion for you
```
mortenpi, line 1084:
```suggestion
            # completion for the relative source files in the asset()
            # function will only complete for the docs/src directory)!
            # so @plugin_author: no fancy asset() autocompletion for you
```

Also, is the autocompletion comment still valid?

src/Writers/HTMLWriter.jl
```diff
@@ -74,67 +74,142 @@ const ASSETS_SASS = joinpath(ASSETS, "scss")
...
+        (class == :unknown) && error("Not supported derived class for: $(src)")
+    else
+        class in [:ico, :css, :js] || error("Unrecognised asset class $class for: $(uri)")
+    end
+    # automatically derive if the source is local or or online and perform
+    # checks if the flag has been set to true or false
+    # no file or path checks are done within this function!
+    if islocal === nothing
+        islocal = !occursin(r"^https?://", src)
+    elseif typeof(islocal)<:Bool
```
mortenpi, line 185:
```suggestion
    elseif isa(islocal, Bool)
```

would be more idiomatic for this check.

However, actually, if we'd constrain `islocal` to `::Union{Bool, Nothing}` in the method signature of `asset`, we could just use the inner condition directly and and drop the `else`, since we know that it has to be bool if it comes to this `else` branch.

mortenpi (commented):
Cool, this looks pretty good! Sorry for delay with the review.

I added a test in #1663 which is now broken, because absolute paths are now okay. With that in mind, could you add tests that:

1. Test the absolute path case, including that `assetlink` has correct paths (in the vein of the new tests)?
2. A test for the case when `src` is a remote URL (`islocal = false`), but `dst` is set. I think it should error, but I am not sure if it does?

src/Writers/HTMLWriter.jl
LazyScholar: True `isa(islocal, Bool)` is cleaner (forgot about this one).

For the `asset()` function i can live with `islocal::Union{Bool, Nothing}` so i will change it that way.

src/Writers/HTMLWriter.jl
LazyScholar: I think it is still valid as the file paths paths provided by **plugin authors** should be absolute paths as there is no way to derive the source file position from a relative path afterwards.
The only cases where relative paths are allowed are by `Documenter.jl` users as we assume relative paths are pointing to files in `<dev_dir>/docs/src` (although there are no checks in place for that, or?).

src/Writers/HTMLWriter.jl
LazyScholar: I agree and changed it that way.

LazyScholar:
> I added a test in #1663 which is now broken, because absolute paths are now okay. With that in mind, could you add tests that:
> 
>     1. Test the absolute path case, including that `assetlink` has correct paths (in the vein of the new tests)?

Yes. `src` has to be a relative path for `Documenter.jl` users and a absolute path for `Documenter.jl` plugin authors.

- In case for `Documenter.jl` users `src` is only used to copy the `src` string into the HTML header without checks (no file operations are done in consequence of using `asset()` as far as i can see).
- In case for plugin authors the path is used to copy the source files over and the `src`+`dst` string is used to place the file in the build path and to generate the string for the HTML header.

That might indeed lead to problems for `Documenter.jl` users later on. As the paths copied into the HTML header might not be valid. It would be nice to have a way to detect i asset has been called by a user or a plugin. Or to introduce a `_asset()` function which gets wrapped by `asset()` and a new `plugin_asset()` function and give those functions dedicated checks?

>     2. A test for the case when `src` is a remote URL (`islocal = false`), but `dst` is set. I think it should error, but I am not sure if it does?

It does not. I will put some checks in places for _2._ (although it might be useful to gather online source files locally, which in turn is a security risk).

test/htmlwriter.jl
```diff
@@ -77,15 +77,17 @@ end
...
+    @test_throws Exception asset("example.com/foo.js", islocal = false)
     @test_throws Exception asset("foo.js", islocal = false)
     @test_throws Exception asset("https://example.com/foo.js?q=1")
-    @test_throws Exception asset("https://example.com/foo.js", class=:error)
+    @test_throws Exception asset("https://example.com/foo.js", class = :error)
+    @test_throws Exception asset("https://example.com/foo.js", dst = "extra_assets/foo.js")
+    @test_throws Exception asset("https://example.com/foo.js", islocal = true)
     # Edge cases that do not actually quite work correctly:
-    let asset = asset("https://example.com/foo.js", islocal = true)
+    let asset = asset("https://example.com/foo.js", islocal = false)
```
LazyScholar, line 88: Not quite sure how i should change this.
I added more checks which lead this to throw exceptions.
After changing this your tests of the links throw errors.
Shall i drop those checks?

LazyScholar: I deleted those lines.

LazyScholar:
@mortenpi is the current implementation okay so far?
Possible changes would be:
- place a absolute path check in `HTML` struct constructor
- split the function `asset()` function into a function for users and plugin authors
You mentioned the PR in a issue related to the location/ordering of script/html assets shall i add a indexing number similar to `Documenter.jl/src/Builder.jl` (mentioned above)?

mortenpi:
@LazyScholar: apologies, I was not active here for a while. I still think a feature like this would be good. Do you think you would have time to revisit this PR in the near future? If yes, I could do another review round now.

LazyScholar:
> @LazyScholar: apologies, I was not active here for a while. I still think a feature like this would be good. Do you think you would have time to revisit this PR in the near future? If yes, I could do another review round now.

I will try, although tight on time. Will rebase and start from scratch.

LazyScholar:
@mortenpi found some time. Have a look.

mortenpi:
Thank you @LazyScholar! And sorry for the delay here. I am actually having a little bit of trouble reviewing this, since I don't have a clear vision of how exactly this will be used.

1. Could we list a couple of brief, but specific use cases for this API? I.e. what problem exactly and how this will solve? We could then benchmark the current implementation against those use case descriptions, to see if we should improve on it or not.
2. In the long term, it would be handy to have a brief manual page explaining how plugin writers should use these APIs. I think this could help with hashing out the API too. But on the other hand this is quite a bit of extra writing work, so this can also come later, and maybe also something I can tackle.

LazyScholar:
No problem @mortenpi, it is not urgent.

I will describe what i aimed for and how i imagine it to work.
To demo it try it with https://github.com/ali-ramadhan/DocumenterCitations.jl/pull/52 .

The main aim is to pass HTML-assets directly from one plug-in. As `Documenter.jl` already provides a solution for users via [`asset` (create HTML-asset element)](https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.Writers.HTMLWriter.asset) and the [`HTML` argument `assets` (to register special assets)](https://juliadocs.github.io/Documenter.jl/stable/lib/internals/writers/#Documenter.Writers.HTMLWriter.HTML) i tried to reuse most of it.

To use it the plug-in author has to use
```julia
function HTMLWriter.plugin_html_assets(::Type{"<name of the plugin>"})::Tuple{String, Vector{HTMLWriter.HTMLAsset}}
  return normpath(abspath("<here path of directory containing the assets (only the assets)>")), [ asset("<relative path of the asset to be included (has to include the name of the asset directory), same rules as for user assets>", class=:css, islocal=true) ]
end
```
Example:
```julia
const asset_dir = normpath(abspath(joinpath(@__DIR__, "..", "citations_assets")))
const asset_list = [ asset(joinpath("citations_assets","citations.css"),class=:css,islocal=true) ]

function HTMLWriter.plugin_html_assets(::Type{CitationBibliography})::Tuple{String, Vector{HTMLWriter.HTMLAsset}}
  return DocumenterCitations.asset_dir, DocumenterCitations.asset_list
end
```

How it should work:
1. The new function `plugin_assets!(ctx::HTMLContext)` is called within `render(doc::Documents.Document, settings::HTML=HTML())`.
2. This function loops through the plug-in list and calls the new function `plugin_html_assets(::Documenter.Plugin)`.
3. If one plugin registered this function with its own name it can return the absolute path to a asset directory and a vector of HTML assets.
4. In the case that a directory path has been passed the directory will fully copied to the `<build directory>/plugin` directory.
5. The `islocal=true` asset paths will be prepended with `plugin` and added to the new attribute `asset_list::Vector{HTMLAsset}` of `struct HTMLContext`.
6. Every time `render_head(ctx, navnode)` is called this vector `asset_list::Vector{HTMLAsset}` will be passed to `asset_links(src, ctx.asset_list)` which modifies and adds those plug-in assets to the head.

I tried to hold it very simple as my first attempt seemed too complicated. Provided documentation to the functions created by me. I wont create a own documentation page. I did not add test cases (no good idea on how to).

Possible improvements (not doing this):
- Refactor and generalize the existing function which copies assets to the build directory (arguments: list of filetypes to copy, source dir, target dir, overwrite flag) in order to reuse it in the new function `plugin_assets!(ctx::HTMLContext)`.
- Selectively add plug-in assets only to headers of sites where those where used.
- Allow users and plug-in authors to somehow dictate the order in which those should be included to the headers (floating point numbers would be the easiest like https://github.com/JuliaDocs/Documenter.jl/blob/0d088e4b42a972bf710c158bed2d803500a6d31d/src/Builder.jl#L74-L80)
- check if and how complicated plug-in assets work if the files have different directory depths (simple paths are already amended)
- check for already existing plug-in directories rename them automatically and amend the corresponding asset paths - or throw errors

LazyScholar:
https://github.com/JuliaDocs/Documenter.jl/issues/1943#issuecomment-1262989188

Sure, feel free to add tests. The Structure for the plug-in should be roughly as follows:

```
.
├── docs
│   ├── make.jl
│   └── src
│       └── index.md
├── mermaid_assets
│   └── mermaid_init.js
├── Project.toml
└── src
    └── Mermaid.jl
```
With those contents:

`src/Mermaid.jl`:
```julia
module Mermaid

using Documenter
using Documenter.Expanders
using Documenter.Selectors
using Documenter.Writers.HTMLWriter

# register plugin
struct MermaidPlugin <: Documenter.Plugin
  some_val::Bool
end
export MermaidPlugin
function MermaidPlugin()
  return MermaidPlugin(true)
end

# add html assets
const asset_dir = normpath(abspath(joinpath(@__DIR__, "..", "mermaid_assets")))
const asset_list = [ asset("https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js", class=:js, islocal=false)
                     asset(joinpath("mermaid_assets","mermaid_init.js"), class=:js, islocal=true) ]
function HTMLWriter.plugin_html_assets(::Type{MermaidPlugin})::Tuple{String, Vector{HTMLWriter.HTMLAsset}}
  return Mermaid.asset_dir, Mermaid.asset_list
end

# register expander
abstract type MermaidExpander <: Expanders.ExpanderPipeline end
Selectors.order(::Type{MermaidExpander}) = 11.05
Selectors.matcher(::Type{MermaidExpander}, node, page, doc) = Expanders.iscode(node, r"^@mermaid")

# register transformation
function Selectors.runner(::Type{MermaidExpander}, x, page, doc)
  raw_html = """<div class="mermaid">$(x.t.code)</div>"""
  x.t = Documenter.Documents.RawNode(:html, raw_html)
end

end # module Mermaid
```

`mermaid_assets/mermaid_init.js`:
```js
mermaid.initialize({startOnLoad:true});
```

`docs/make.jl`:
```julia
# get path of docs folders
docdir = dirname(@__FILE__);

# check if run by CI
CIflag = get(ENV,"CI","") != "";

# add look up paths
push!(LOAD_PATH, joinpath(docdir, ".."));

using Documenter;
using Mermaid;

makedocs(MermaidPlugin(),
         sitename = "Mermaid",
         authors = "???",
         format = Documenter.HTML(edit_link = CIflag ? :commit : "main" ),
         doctest = false,
         clean = true,
         pages = ["Home" => "index.md"],
        );
```

`docs/src/index.md`:
``````markdown
# Mermaid Test

```@mermaid
graph TD;
    A-->B;
    A-->C;
    B-->D;
    C-->D;
```
``````

Which provides this:
Screenshot 2022-10-02 at 02-23-24 HomeMermaid

After deleting some JS lines to circumvent those current errors:
Screenshot from 2022-10-02 02-39-00

MilesCranmer:
Could this be merged soon-ish? Would be super useful!

MilesCranmer:
Pinging this again. Would be very useful to have mermaid.js enabled so I can generate flowcharts in my docs. What is left to be done before merging?

Issue #937, opened 2019-02-03
deploydocs should create and maintain redirect index.html?


mortenpi:
Most Documenter setups have an `index.html` file like this

```html
<meta http-equiv="refresh" content="0; url=http://juliadocs.github.io/Documenter.jl/stable"/>
```

in the root of the `gh-pages` branch to redirect to the latest stable version of the docs. Should `deploydocs` just create this file automatically? Possibly with an option to disable this if the user wants to maintain their own `index.html`.

The behavior could be as follows:

* If there are tagged versions, have `index.html` point to `stable/`.
* If there are not tagged version, have it point to `dev/`.

Related: #264

fredrikekre:
Makes sense. I think most packages don't have that file at all though?

fredrikekre:
We can have a Documenter specific header and only modify it if it was indeed generated by Documenter.

kellertuer:
It seems my instance does not (yet?) create the index.html automatically. Do I have to activate that? It would be handy because until now only the stable/ link works – well “only” – that one works perfectly fine and great :)

christopher-dG:
Bump, I would very much like this feature. I've been pasting an `index.html` in manually for a bunch of packages but I'd prefer to have it handled for me, especially if it can properly choose between `/dev` and `/stable`!

mortenpi:
Fixed by #1657.

PR #1657, opened 2021-07-22
 Add `index.html` generation for redirecting to `./stable` or `./dev`

hyrodium:
This PR solves https://github.com/JuliaDocs/Documenter.jl/issues/937.


mortenpi:
Thanks @hyrodium, this is great!

I am concerned that this will overwrite user-created files -- there might be some cases where the user wants to maintain their own `index.html`. So if we want this in a patch release, we should not touch an existing `index.html`. 

However, the URL might change (e.g. `dev/` -> `v1/` -> `v2/`). So it is probably good if Documenter updates the file when necessary. I see two options:

1. We could potentially just change the behavior in the next breaking release to always update `index.html`.
2. Check for an HTML comment in `index.html` that indicates that the current version is automatically generated, and so can be overwritten.

I am kinda partial two (2), even if it's a little more complicated, since that still allows the user to also manually maintain an `index.html` file.

hyrodium:
Thanks for the review!
I've updated the function to check the existence of a comment `"<!--This file is automatically generated by Documenter.jl-->"` in `index.html`.

src/Writers/HTMLWriter.jl
```diff
@@ -1405,6 +1405,30 @@ function generate_version_file(versionfile::AbstractString, entries, symlinks =
...
+    isfile(redirectfile) && !startswith(read(redirectfile, String), comment) && return
+
+    version = devurl
+    stable_index = findfirst(v -> last(v)=="v^", versions)
+    if !isnothing(stable_index)
+        stable_version = first(versions[stable_index])
+        if stable_version in entries
+            version = stable_version
+        end
+    end
```
hyrodium, line 1424: In the current script, only `"stable"=>"v^" in versions` are supported. Should we take care of `"v#"`, `"v#.#"`and `"v#.#.#"`?

fredrikekre:
> Check for an HTML comment in index.html that indicates that the current version is automatically generated, and so can be overwritten.

This was what I was thinking too, and looks like you implemented that now in the PR so 👍

src/Writers/HTMLWriter.jl
```diff
@@ -1405,6 +1405,30 @@ function generate_version_file(versionfile::AbstractString, entries, symlinks =
...
+    isfile(redirectfile) && !startswith(read(redirectfile, String), comment) && return
+
+    version = devurl
+    stable_index = findfirst(v -> last(v)=="v^", versions)
+    if stable_index !== nothing
+        stable_version = first(versions[stable_index])
+        if stable_version in entries
+            version = stable_version
+        end
+    end
```
fredrikekre, line 1424: Can this reuse some functinality for the "outdated warning" code? Makes sense that those would point to the same version.

hyrodium: Are you suggesting to reuse the script of `generate_version_file`

https://github.com/JuliaDocs/Documenter.jl/blob/v0.27.4/src/Writers/HTMLWriter.jl#L1396-L1403

or `warner.js`?

https://github.com/JuliaDocs/Documenter.jl/blob/v0.27.4/assets/html/warner.js

I think the redirect page should be `./stable` or `./dev`, so we can't reuse the script here.
Sorry if I misunderstood something.

fredrikekre: I meant that `index.html` should redirect to the same thingthe outdated warning rediretion does.

hyrodium: Ah, I see. All I had to do was just picking the first element of `entries`.

https://github.com/JuliaDocs/Documenter.jl/blob/v0.27.4/src/Writers/HTMLWriter.jl#L1404
https://github.com/JuliaDocs/Documenter.jl/blob/v0.27.4/assets/html/warner.js#L40

hyrodium:
I've fixed the redirected destination, and I think this is ready to merge.

hyrodium:
I've updated the CHANGELOG.md. (btw, not NEWS.md:smile:)

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Documenter.jl changelog
 
+## Version `v0.27.6`
+* ![Feature][badge-feature] Add support for generating `index.html` to redirect to `dev` or `stable`. The redirected destination is the same as the outdated warning. If there's already user-generated `index.html`, Documenter will not overwrite the file. ([#937][github-937], [#1657][github-1657], [#1658][github-1658])
```
fredrikekre, line 4: Please also add links to the bottom of the file (`[github-1658]: ...` etc).

test/htmlwriter.jl
```diff
@@ -104,6 +111,8 @@ end
 
     mktempdir() do tmpdir
         versionfile = joinpath(tmpdir, "versions.js")
+        redirectfile = joinpath(tmpdir, "index.html")
+        devurl = "dev"
```
fredrikekre, line 115: Would be good with a test for for the case where no versions exist (to verify that `dev` is picked).

fredrikekre (commented):
.

CHANGELOG.md
hyrodium: Oh, I was careless.

test/htmlwriter.jl
hyrodium: I've added the tests in case6. (see comments)

CHANGELOG.md
```diff
@@ -686,6 +689,7 @@
 [github-929]: https://github.com/JuliaDocs/Documenter.jl/pull/929
 [github-934]: https://github.com/JuliaDocs/Documenter.jl/pull/934
 [github-935]: https://github.com/JuliaDocs/Documenter.jl/pull/935
+[github-937]: https://github.com/JuliaDocs/Documenter.jl/issue/937
```
fredrikekre, line 692:
```suggestion
[github-937]: https://github.com/JuliaDocs/Documenter.jl/issues/937
```

fredrikekre:
LGTM, thanks. I'll let @mortenpi look over it and merge.

mortenpi (approved):
LGTM, thanks @hyrodium and @fredrikekre!

CHANGELOG.md
```diff
@@ -893,6 +897,8 @@
 [github-1647]: https://github.com/JuliaDocs/Documenter.jl/pull/1647
 [github-1649]: https://github.com/JuliaDocs/Documenter.jl/pull/1649
 [github-1655]: https://github.com/JuliaDocs/Documenter.jl/issues/1655
+[github-1661]: https://github.com/JuliaDocs/Documenter.jl/pull/1657
+[github-1661]: https://github.com/JuliaDocs/Documenter.jl/pull/1658
```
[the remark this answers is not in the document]
hyrodium: Ahh, I was careless again. Thanks!

PR #1658, opened 2021-07-24
Remove `index.html`

hyrodium:
If #1657 is merged, `index.html` will be maintained by Documenter.jl.

We have the folllowing two options, and I choose the first one:
* Delete `index.html` and Documenter.jl will generate this file again.
  * :+1:  The author of commit will be `Documenter.jl`.
  * :-1:  We will lose the redirect page in short time.
* Modify `index.html` and add a comment `<!--This file is automatically generated by Documenter.jl-->` to be maintaned by Documenter.jl
  * :+1:  We don't lost the redirect page.
  * :-1:  The author of commit will be human.


fredrikekre:
I suggest we just add the documenter comment once the other pr is merged instead of deleting this.

hyrodium:
Thanks for the review. I've updated the file.

hyrodium:
I think we can merge this PR before #1657.

PR #1659, opened 2021-07-25
Fix example links

hyrodium:
https://evizero.github.io/Augmentor.jl/ doesn't seems the latest version of the package.


mortenpi:
Thanks! FWIW, they have committed to the root of `gh-pages` (instead of a version folder) at some point.

PR #1662, opened 2021-07-27
[LaTeXWriter] Avoid nested math environments

sbrisard:
The issue that is solved my this PR is discussed in [this thread](https://discourse.julialang.org/t/equation-numbering-in-documenter-jl/65349). The correction avoids incorrect nesting of explicit math environments into ``equation*/split`` blocks

```latex
\begin{equation*}
\begin{split}
...
\end{split}
\end{equation*}
```
=

### DIFF
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -574,7 +574,7 @@ end
 # equations in the Julia manual. However, if the equation is already wrapped in
 # align/align*, then there is no need to further wrap it (in fact, it will break).
 function latex(io::IO, math::Markdown.LaTeX)
-    if occursin(r"^\\begin\{align\*?\}", math.formula)
+    if occursin(r"^\\begin\{((equation)|(align)|(gather)|(flalign)|(multline)|(alignat)|(split))\*?\}", math.formula)
         _print(io, math.formula)
     else
         _print(io, "\\begin{equation*}\n\\begin{split}")


KristofferC:
Perhaps it would be a good idea to add a test for this?

sbrisard:
> 
> 
> Perhaps it would be a good idea to add a test for this?

Sure! I will do that. Thanks for the suggestion.

src/Writers/LaTeXWriter.jl
```diff
@@ -574,7 +574,7 @@ end
 # equations in the Julia manual. However, if the equation is already wrapped in
 # align/align*, then there is no need to further wrap it (in fact, it will break).
 function latex(io::IO, math::Markdown.LaTeX)
-    if occursin(r"^\\begin\{align\*?\}", math.formula)
+    if occursin(r"^\\begin\{((equation)|(align)|(gather)|(flalign)|(multline)|(alignat)|(split))\*?\}", math.formula)
```
mortenpi, line 577: I believe the brackets here are unnecessary:

```suggestion
    if occursin(r"^\\begin\{(equation|align|gather|flalign|multline|alignat|split)\*?\}", math.formula)
```

mortenpi (commented):
* A test would indeed be nice. It could be a simple `occursin` in here: https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests_latex.jl 
* Regarding `\eqref`: I don't think we want to special case that. In the HTML output, this would not produce anything meaningful.

sbrisard:
## Help needed!!!

I have been struggling with the tests for (too many) hours, now. I am currently trying to implement some simple tests, as requested. But the architecture of the tests is quite complex, and I find that `(@v1.6) pkg> test Documenter` skips the `LaTeX`-related tests... which is not what I want.

I have been fiddling with the `runtests.jl` file, and replaced the whole `@testset "Documenter" begin ... end` block with the following

```julia
@testset "Documenter" begin
    # Build the example docs
    @info "Building example/make.jl"
    include("examples/make.jl")
    include("examples/tests_latex.jl")
end
```

Also, in file `test/examples/make.jl`, I have replaced [lines 21 and 22](https://github.com/JuliaDocs/Documenter.jl/blob/cacf739b425b088e285af14a214cf4865bfb0019/test/examples/make.jl#L21) with

```julia
    ["latex_texonly"]
```

I would be extremely grateful for a cleaner fix (it could hardly be dirtier anyway!).

Thanks!

sbrisard:
>     * Regarding `\eqref`: I don't think we want to special case that. In the HTML output, this would not produce anything meaningful.

I don't quite agree. If you use the KaTeX engine, `\eqref` does not make any sense at all. However, with Mathjax, `\eqref{...}` *does* make sense, and does not need to be enclosed in `$...$` in the html file. In other words, the following markdown input

````md
```math
\begin{equation}
\label{eq20210803093550}
(a+b)^2 = a^2+2ab+b^2
\end{equation}
```

See this remarkable identity \eqref{eq20210803093550}.
````

Produces the following output

image

[note that the link to Eq. (1) is active and valid]. So for both LaTeX and HTML+Mathjax output, `\eqref{...}` works as expected, *without double backticks*. For other outputs, neither bare `\eqref{...}`, nor `\eqref{...}` surrounded with double backticks make sense. In the LaTeX output, `$\eqref{...}$` is weird. So I would really recommend that we accept bare `\eqref{...}` statements. But I agree, this is nit-picking.

**Comment by sbrisard** [comment]
The nested environments do occur in the HTML output as well. For example, the following input

````md
```math
\begin{equation}
(a+b)^2 = a^2+2ab+b^2
\end{equation}
```
```` 
that would display a *numbered* equation results in the following HTML+Mathjax output

```html
<p class="math-container">
\[\begin{equation}
(a+b)^2 = a^2+2ab+b^2
\end{equation}\]
</p>
```

This should really be corrected. This is not valid LaTeX, and it can't be excluded that in future versions of KaTeX/Mathjax, the LaTeX interpreter would be more restrictive and disallow such weird constructs.

PR #1664, opened 2021-08-03
Add documentation on how to easily test README.md

rikhuijzer:


docs/src/man/doctests.md
```diff
@@ -389,3 +389,28 @@ Alternatively, you can also pass the `doctest = :fix` keyword to [`makedocs`](@r
...
+## Testing README.md
+
+The easiest way to also run doctests on the README.md file is to add the readme to the
+docs of the module:
+
+```julia
+module MyModule
+
+function _update_module_doc()
+    path = joinpath(pkgdir(MyModule), "README.md")
```
KristofferC, line 402:
```suggestion
    path = joinpath(@__DIR__, "..", "README.md")
```

I think it is better to use a relative path here rather than `pkgdir` (which is much more complicated).

docs/src/man/doctests.md
```diff
@@ -389,3 +389,28 @@ Alternatively, you can also pass the `doctest = :fix` keyword to [`makedocs`](@r
...
+_update_module_doc()
+
+[...]
+
+end # module
+```
+
+Now, when users type `?MyModule` in the REPL, they'll see the contents of the readme
+file, and doctest blocks in the readme file will be tested. Be sure to call 
+`MyModule._update_module_doc()` before running the doctests.
```
KristofferC, line 415: I don't understand why you would need to call the `_update_module_doc()` manually. Isn't the docstring be attached to the module at precompile time?

rikhuijzer: It is for situations where the user doesn't close the REPL in between running doctests.

docs/src/man/doctests.md
rikhuijzer: Nice one :+1:

docs/src/man/doctests.md
rikhuijzer: To verify this, try to run `Pkg.test()` again after changing the README.md and not restarting the REPL. The tests will then use the old doctests (Julia 1.6).

rikhuijzer:
I guess the docs are not clear enough: https://discourse.julialang.org/t/validity-of-package-documentation-and-examples/66217/19.

PR #1665, opened 2021-08-04
Fix PR preview verification on 1.0

mortenpi:
X-ref: https://github.com/jump-dev/JuMP.jl/pull/2650#issuecomment-892321143


odow (approved):
Thanks @mortenpi

PR #1669, opened 2021-08-09
Improve the deploydocs docstring

### LINKED ISSUE #1668: Document what happens during deploydocs
In the [documentation for the `deploydocs` function](https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.deploydocs), I'm missing information on what exactly happens during the deployment to a `gh-pages` branch. My natural assumption would be that it simply copies the entire `docs/build` folder from the previous `makedocs` call to the appropriate `gh-pages` subfolder, e.g. `dev`. It seems like it might be skipping certain files (`.ipynb` files for example) in the process though, which is causing me [this problem](https://discourse.julialang.org/t/deploy-literate-non-html-files-with-documenter/66031) with deploying `Literate.jl`-rendered examples.

It would be great if the documentation could elaborate on the deployment process.

Also, I noticed that the documentation for `deploydocs` says

> Converts markdown files generated by makedocs to HTML and pushes them to repo.

Is that actually true, or ist that a "typo"? If true, it seems very confusing: doesn't `makedocs` do the conversion from markdown to HTML? Does `deploydocs` really perform any conversions of its own, or does it simply deploy the HTML files generated by `makedocs`?

goerz:
Closes #1668


goerz:
I was trying to view my changes to the documentation on my own branch locally, by running
~~~
julia --project=docs
~~~
from inside my topic branch.

Generating the docs runs through without errors. I then running a webserver in `./docs/build` so that I can view the result (to get around the "pretty" urls). For some reason, I'm not seeing my modifications. Any ideas? I'm fairly new to Julia development, so maybe I don't have the correct workflow down yet.

mortenpi (approved):
LGTM, thanks!

Regarding building locally: you also need to add the Documenter repository as a dev-dependency before running `make.jl` with `pkg> dev .`. It was likely using the latest tagged version, instead of the modified local version.

goerz:
> LGTM, thanks!
> 
> Regarding building locally: you also need to add the Documenter repository as a dev-dependency before running `make.jl` with `pkg> dev .`. It was likely using the latest tagged version, instead of the modified local version.

Yup, that did that trick! Thanks for the explanation (and the merge!)

PR #1673, opened 2021-08-13
WIP: Fix accesibility issues

pfitzseb:
The goal of this is to make sure Documenter-generated docs conform to [WCAG 2.0](https://www.w3.org/TR/WCAG20/).
I'm opening this early so I can document the changes as I work my way through them, which should hopefully make review a bit easier.

### Light theme `Version` contrast
**From**: Contrast ratio 3.93
image
**To**: Contrast ratio 5.26
image

### Dark theme code contrast
**From**: Contrast ratio 3.55
image
**To**: Contrast ratio 11.49
image

### DIFF
--- a/assets/html/scss/documenter-dark.scss
+++ b/assets/html/scss/documenter-dark.scss
@@ -47,7 +47,7 @@ $border-width: 1px;
 @import "documenter/utilities";
 @import "documenter/variables";
 
-$code: #e74c3c;
+$code: #ececec;
 $code-background: rgba(255, 255, 255, 0.05);
 
 $documenter-docstring-shadow: none;
--- a/assets/html/scss/documenter-light.scss
+++ b/assets/html/scss/documenter-light.scss
@@ -7,6 +7,9 @@ $warning: #ffdd57;
 $danger: #da0b00;
 $compat: #1db5c9;
 
+// slightly increase contrast over default l: 48%
+$grey: hsl(0, 0%, 42%);
+
 @import "documenter/utilities";
 @import "documenter/variables";
 


kimikage:
This is off-topic, but for consistency, I think it would be better to remove the dependency on the "darkly" theme, which also changes the font size of menus and admonitions.
I haven't checked the impact on plugins, but removing the darkly is not so hard.

mortenpi:
Sorry for getting around to this earlier, but these changes all LGTM. Thanks @pfitzseb!

I'd be happy to merge this in it's current form if you don't have additional fixes right now, since I think it's already an improvement.

> This is off-topic, but for consistency, I think it would be better to remove the dependency on the "darkly" theme, which also changes the font size of menus and admonitions. I haven't checked the impact on plugins, but removing the darkly is not so hard.

Darkly (kinda) is the current dark theme.. so it would have to be replaced with something. So when you say "remove", do you mean reducing the current dark theme down to something that would just override the colors basically? If yes, then I would support that completely --- just dropping Darkly in was kind of a shortcut to get a working dark theme in quickly, but it's definitely not ideal.

pfitzseb:
Yeah, I don't have any immediate plans to work on this further. There are some issues related to duplicate header ids, but I haven't yet have time to look into those.

sairus7:
I just paste here messages from Slack:

I can't get rid of the feeling that dark documenter theme is harder to read than light one. Font looks thicker and red code lines look too dark:
image

And for comparison of dark themes - from top to bottom:
1. Julia docs
2. Slack text
3. GitHub markdown

Looks like julia docs have more contrast, and bold text is blurry (open in new pane to watch in original ratio):

image

PR #1687, opened 2021-08-28
Pass file and line info to downstream loggers

ericphanson:
We can use `_file` and `_line` ([docs](https://docs.julialang.org/en/v1/stdlib/Logging/#Logging.@logmsg)) to cause the `@error` to act like it was emitted from where the doctest failed, instead of acting like it was emitted from `report` in Documenter. This is very useful because loggers can use the file and line info when trying to surface logs.

For instance, I want to use a GitHubActions.jl [GitHubActionLogger](https://github.com/julia-actions/GitHubActions.jl/blob/0f9a18e1421e7b350af9456ec18946410e6c8420/src/GitHubActions.jl#L209) which will emit GHA-friendly error messages, using the file and line info to create an inline-annotation in the diff showing where the error occurred:

Screenshot 2021-08-28 at 15 01 50

(That needs a small fix in GitHubActionLogger as well: https://github.com/julia-actions/GitHubActions.jl/pull/15).

We can only pass a single line, so I just chose the first line of the failing doctest. We could alternately choose the last line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * ![Bugfix][badge-bugfix] Checking whether a PR comes from the correct repository when deciding to deploy a preview on GitHub Actions now works on Julia 1.0 too. ([#1665](github-1665))
 
+* ![Bugfix][badge-bugfix] When a doctest fails, pass file and line information associated to the location of the doctest instead of the location of the testing code in Documenter to the logger. ([#1687](github-1687))
+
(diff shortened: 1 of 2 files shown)


fredrikekre (approved):
LGTM. Worth a small note in the changelog?

ericphanson:
Sure, I gave it a try

mortenpi:
LGTM as well, thanks!

Issue #1686, opened 2021-08-28
Fail docs build if there are doctest errors


ericphanson:
I think Documenter should (at least have the option to) fail `makedocs` if there are doctest errors, otherwise they need to either be tested elsewhere or they slip through. 

I don’t really think they should be tested as part of the usual tests since those are run on multiple Julia versions leading to annoying failures from printing changes, whereas the docs build is usually done on a single version of Julia.

I know `strict=true` does this but it also requires a lot of other things. In my opinion if you’re using doctests you generally want them to act like tests (ie red X in CI not green check).

I’m hoping to make seeing and fixing doctests more ergonomic (you’ll get a tiny PR to fix the logging statement so doctest `@error`s are more actionable downstream) but I think easily failing on them is an important step.

mortenpi:
Yep, I think more flexibility with the `strict` keyword would be perfectly fine. There is some earlier discussion here: https://github.com/JuliaDocs/Documenter.jl/issues/1504#issuecomment-762507285

ericphanson:
Ok great, I’ll try to make a PR soon. Would it be ok to just start with doctests or should such a PR cover other subsets of what strict does?

mortenpi:
I think starting with just doctests is fine. But the design should keep in mind that we probably want to add other options in the future.

PR #1689, opened 2021-08-29
Generalize `strict`

ericphanson:
Fixes https://github.com/JuliaDocs/Documenter.jl/issues/1686 by generalizing strict to accept a symbol or vector of symbols. I was going to do just `doctest` but looking at it, it seemed about as easy to do all of them.

I only found tests with `strict` for doctests though, but I modified those to try some of the new versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * ![Bugfix][badge-bugfix] Checking whether a PR comes from the correct repository when deciding to deploy a preview on GitHub Actions now works on Julia 1.0 too. ([#1665](github-1665))
 
+* ![Feature][badge-feature] The keyword argument `strict` in `makedocs` is more flexible: in addition to a boolean indicating whether or not any error should result in a failure, `strict` also accepts a `Symbol` or `Vector{Symbol}` indicating which error(s) should result in a build failure. ([][github-])
+
 ## Version `v0.27.5`
 
 * ![Bugfix][badge-bugfix] Fix an error introduced in version `v0.27.4` (PR[#1634][github-1634]) which was triggered by trailing comments in `@eval`/`@repl`/`@example` blocks. ([#1655](github-1655), [#1661](github-1661))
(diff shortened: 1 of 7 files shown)


ericphanson:
What this does:

* collects all the names we use for errors and puts them in `ERROR_NAMES`
* allows `strict` to be set to a Bool as now, or a `Symbol` from `ERROR_NAMES` or a `Vector{Symbol}` that is a subset of `ERROR_NAMES`. This allows any middle ground between `false` and `true`, by being strict about any subset of the possible errors. This is validated by `is_strict_kw`, a new utility function.
* Sets the `loglevel` when we log errors according to if we are being `strict` about that error or not, using `is_strict`, a new utility function
* Errors at the end of the build if we are strict about any errors that have occurred (or after doctest time if we are strict about `doctest` and any doctests have failed), using `is_strict`
* Adds some unit tests for `is_strict` and `is_strict_kw`, and runs some of the doctest tests with various equivalent (for the purpose fo doctests) choices of `strict`.

I am happy with it from my end and am ready for review when you get a chance.

ericphanson:
Bump

ericphanson:
Bump

src/Utilities/Utilities.jl
```diff
@@ -719,6 +720,45 @@ function get_sandbox_module!(meta, prefix, name = nothing)
     end
 end
 
+"""
+    is_strict(strict, val::Symbol) -> Bool
+
+Interanl function to checks if `strict` is strict about `val`, i.e.
```
albheim, line 726:
```suggestion
Internal function to check if `strict` is strict about `val`, i.e.
```

fredrikekre (commented):
LGTM but would be good to document somewhere what the different options for symbols are.

ericphanson:
The list is in the docstring (https://github.com/JuliaDocs/Documenter.jl/pull/1689/files#diff-2c620811ba00ffd5e7b818bdefbab47fe9d7e3e11d046e6fe18e5c01fdff3f0cR227) but maybe you’re saying the meaning of them all should be documented too? That would make sense to me, though I’m not sure where to put it.

ericphanson:
CI failures seem to be pkg server issues. (Especially considering everything passed just before and the only difference was typos in a docstring).

ericphanson:
Bump :). Would love to get this in soon-- I think it would massively increase the usability of doctests if there was an easy way to make them fail the built without also making everything else fatal.

ericphanson:
Bump

PR #1691, opened 2021-09-06
enable colored print for REPL output

aviatesk:
Won't be enabled for doctests (so I believe this should be safe).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 * ![Bugfix][badge-bugfix] When a doctest fails, pass file and line information associated to the location of the doctest instead of the location of the testing code in Documenter to the logger. ([#1687](github-1687))
 
+* ![Bugfix][badge-bugfix] Enabled colored printing for each output of `@repl`-blocks. ([#1691](github-1691))
+
 ## Version `v0.27.5`
 
 * ![Bugfix][badge-bugfix] Fix an error introduced in version `v0.27.4` (PR[#1634][github-1634]) which was triggered by trailing comments in `@eval`/`@repl`/`@example` blocks. ([#1655](github-1655), [#1661](github-1661))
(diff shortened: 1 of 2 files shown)


KristofferC:
Regarding doctests, how about producing colored output but stripping it before the comparison is made? I guess some people would not want that but it could be optional?

aviatesk:
> it could be optional?

So that people can do something like ?
```julia
julia> printstyled("red"; color=:red)
\e[31mred\e[39m
```

Honestly, I'm not super willing to support it. Maybe we can delay the support to the future until when someone finds it useful ?

KristofferC:
No, I meant so that people can still do:

````jl
```jldoctest
julia> @test 1==1
Test Passed
```
````

but have `Test Passed` written in green (like in the REPL).

**Comment by aviatesk** [comment]
Ah, I see. But it seems it requires much refactors in order to replace the contents of generated document with the actual outputs of doctests.

**Comment by KristofferC** [comment]
It was probably bad form of me to comment on this PR. It wasn't really meant as anything that should be done here, it was just a related thought I hade when I read the PR.

**Comment by fredrikekre** [comment]
Do you have an example of what this enables?

**Comment by aviatesk** [comment]
> It was probably bad form of me to comment on this PR. It wasn't really meant as anything that should be done here, it was just a related thought I hade when I read the PR description. I should just open a separate issue.

I see. Thanks for the clarification :)

> Do you have an example of what this enables?

An object may implement its own `show` method that may use color when enabled.
Probably the easiest repro would be something like:
```markdown
\`\`\`@example; ansicolor=true
struct Foo end
Base.show(io::IO, ::Foo) = printstyled(io, "Foo"; color = :blue)

Foo()
\`\`\`
```

aviatesk:
@KristofferC so can this get merged (and maybe can we tag a patch release) ?

fredrikekre:
Can you add a note in the changelog?

aviatesk:
Sure. Added.

PR #1694, opened 2021-09-09
fix example (linspace -> LinRange, broadcast)

gaelforget:
`linspace` has been removed, replaced with `LinRange`. Also I feel there should be a broadcast operation. Updated code worked as I would expect.

### DIFF
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -418,7 +418,7 @@ g(x) = cos(x) - x
 and then we plot `f` over the interval from ``-π`` to ``π``
 
 ```@example 1
-x = LinRange(-π, π, 16)
+x = range(-π, π; length=50)
 plot(x, f.(x), color = "red")
 savefig("f-plot.svg"); nothing # hide
 ```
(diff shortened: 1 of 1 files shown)


docs/src/man/syntax.md
```diff
@@ -418,8 +418,8 @@ g(x) = cos(x) - x
 and then we plot `f` over the interval from ``-π`` to ``π``
 
 ```@example 1
-x = linspace(-π, π)
-plot(x, f(x), color = "red")
+x = LinRange(-π, π, 16)
```
mcabbott, line 421: Would the canonical choice here just be `range`? LinRange is slightly weird leftover struct, which might sometimes be desirable for performance.

fredrikekre:
```suggestion
x = range(-π, π; length=50)
```

docs/src/man/syntax.md
```diff
@@ -653,8 +653,8 @@ final document.
 ```@eval
 using PyPlot
 
-x = linspace(-π, π)
-y = sin(x)
+x = LinRange(-π, π, 16)
```
fredrikekre, line 656:
```suggestion
x = range(-π, π; length=50)
```

PR #1696, opened 2021-09-15
fix `DocTestFilters = nothing` in at-meta

ranocha:
Using 
````
```@meta
DocTestFilters = nothing
```
````
as [suggested in the docs](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Filtering-Doctests), I get
```
ERROR: LoadError: MethodError: no method matching occursin(::Nothing, ::SubString{String})
Closest candidates are:
  occursin(::Union{AbstractChar, AbstractString}, ::AbstractString) at strings/search.jl:622
  occursin(::Regex, ::SubString; offset) at regex.jl:269
  occursin(::Regex, ::AbstractString; offset) at regex.jl:264
```
To me, it looks like this is the reason for this behavior.

### DIFF
--- a/src/DocTests.jl
+++ b/src/DocTests.jl
@@ -276,7 +276,7 @@ end
 function filter_doctests(strings::NTuple{2, AbstractString},
                          doc::Documents.Document, meta::Dict)
     meta_block_filters = get(meta, :DocTestFilters, [])
-    meta_block_filters == nothing && meta_block_filters == []
+    meta_block_filters === nothing && meta_block_filters = []
     doctest_local_filters = get(meta[:LocalDocTestArguments], :filter, [])
     for r in [doc.user.doctestfilters; meta_block_filters; doctest_local_filters]
         if all(occursin.((r,), strings))

### CODE REVIEW

File: src/DocTests.jl [hunk 1]
```diff
@@ -276,7 +276,9 @@ end
 function filter_doctests(strings::NTuple{2, AbstractString},
                          doc::Documents.Document, meta::Dict)
     meta_block_filters = get(meta, :DocTestFilters, [])
-    meta_block_filters == nothing && meta_block_filters == []
+    if meta_block_filters === nothing
```
KristofferC, line 279: ` meta_block_filters == nothing && (meta_block_filters = [])`

is probably what was intended

src/DocTests.jl
```diff
@@ -276,7 +276,9 @@ end
 function filter_doctests(strings::NTuple{2, AbstractString},
                          doc::Documents.Document, meta::Dict)
     meta_block_filters = get(meta, :DocTestFilters, [])
```
KristofferC, line 278: Completely unrelated to this PR but just saw this:

```suggestion
    meta_block_filters = get(Vector{Any}, meta, :DocTestFilters)
```

src/DocTests.jl
ranocha: Yes, I think so, too. However, i thought my version was a bit more readable (which seems to be fine given this typo)

src/DocTests.jl
ranocha: If you say so

src/DocTests.jl
KristofferC: I think this is the standard way of doing it as seen by the original code. It is just as easy to write `meta_block_filters == []` in your version.

ranocha: Changed in https://github.com/JuliaDocs/Documenter.jl/pull/1696/commits/f726d74c0255c47277cc7c2d308562a44c633f0d

Issue #1698, opened 2021-09-15
MethodError: no method matching `getindex(::Nothing, ::Int64)`


timholy:
https://github.com/timholy/SnoopCompile.jl/pull/267/checks?check_run_id=3610545554 shows a MethodError from `report`. I haven't dug into this package yet, though. Seems to happen on both Julia 1.6 and nightly.

fredrikekre:
Probably https://github.com/JuliaDocs/Documenter.jl/pull/1687 (cc @ericphanson ). [`lines` here](https://github.com/JuliaDocs/Documenter.jl/blob/50246d0f3dc9386254e5560184cac942077a826b/src/DocTests.jl#L404) is `nothing` if the source code can't be found (wonder why?).

timholy:
One confusing thing about SnoopCompile is that SnoopCompileCore is a subdir package. Maybe that's it?

PR #1699, opened 2021-09-15
Work around #1698

ericphanson:
I think this would revert to the old behavior if we can't find the line info. Not sure why that is though-- seems like we'd need to know that to know how to repro the error for a test.

fixes https://github.com/JuliaDocs/Documenter.jl/issues/1698

### DIFF
--- a/src/DocTests.jl
+++ b/src/DocTests.jl
@@ -402,7 +402,7 @@ function report(result::Result, str, doc::Documents.Document)
 
         $(result.output)
 
-        """, diff; _file=result.file, _line=line)
+        """, diff, _file=result.file, _line=line)
 end
 
 function fix_doctest(result::Result, str, doc::Documents.Document)
--- a/src/DocTests.jl
+++ b/src/DocTests.jl
@@ -402,7 +402,7 @@ function report(result::Result, str, doc::Documents.Document)
         $(result.output)
 
         """, diff; _file=result.file,
-          _line=lines === nothing ? (@__LINE__) : first(lines))
+          _line=lines === nothing ? nothing : first(lines))
 end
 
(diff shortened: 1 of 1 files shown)


src/DocTests.jl
```diff
@@ -401,7 +401,8 @@ function report(result::Result, str, doc::Documents.Document)
 
         $(result.output)
 
-        """, diff, _file=result.file, _line=lines[1])
+        """, diff; _file=result.file,
+          _line=lines === nothing ? (@__LINE__) : first(lines))
```
fredrikekre, line 405: This should also work, both this and your suggestion are equally useless I guess.
```suggestion
          _line=lines === nothing ? nothing : first(lines))
```

ericphanson: I thought maybe downstream loggers will expect `_line` will only be passed when it has an actual number, and I couldn't figure how to nicely conditionally pass it (could just write the whole macro twice but that seems less ideal), so I thought to pass a valid but useless line number instead of an invalid one.

In particular, the GitHubActionsLogger will do
```julia
julia> @warn "Hi" _line=nothing
::warning file=REPL[7],line=::Hi

julia> @warn "Hi"
::warning file=REPL[8],line=1::Hi
```
and the first one doesn't really look valid, though I suppose GitHubActions itself won't throw on seeing it lol. I can fix that in GitHubActionsLogger but it made me think that in general loggers might not expect `_line=nothing`.

ericphanson: I have no idea why CI is failing so I’ll try with `nothing`

timholy:
Belated thanks to @ericphanson and @fredrikekre!

Issue #1679, opened 2021-08-17
Doctest filter matches line by line by default


TorkelE:
I am trying to get the regex filter function to work for a simple doctest (with the goal of creating a filter which checks whenever there was an error or not).

Checking from here https://github.com/JuliaDocs/Documenter.jl/issues/452, it seems that `filter = r".*"` should create something which accepts any output. So I try:
(in all code blocks I will use ' to denote those things which create code in markdown, so to avoid messing up the formatting).
```julia
'''jldoctest; filter = r".*"
1
# output
1
'''
```
and 
```julia
'''jldoctest; filter = r".*"
1
# output
2
'''
```
which both works.

However, if I try anything a little bit complicated, it fails, e.g.
```julia
'''jldoctest; filter = r".*"
[1,2]
# output
2
'''
```

however, if it is a bit more similar, it does work:
```julia
'''jldoctest; filter = r".*"
[1,2]
# output
2-element Vector{Int64}:
 1
 3
'''
```
I also tried modifying the regex for multiple lines, e.g. `filter = r"\_.\{-}"`. Not sure whenever that is needed/correct though (in either case, didn't work)

mortenpi:
I think the issue is that [the `replace` call here](https://github.com/JuliaDocs/Documenter.jl/blob/68dbd53d4ff6b339e795a4a3328955ad5c689e0e/src/DocTests.jl#L283) is not greedy enough and keeps line endings:

```
julia> replace("""
       foo
       bar
       """, r".*" => "")
"\n\n"
```

So in the end the strings that get compared have a different number of newlines, and so don't match.

fredrikekre:
You can use the `s` modifier to let `.` match newlines (see docs for `@r_str`):
```
julia> replace("""
       foo
       bar
       """, r".*"s => "")
""
```

TorkelE:
Just confirming that 
```julia
'''jldoctest; filter = r".*"s
[1,2]

# output

3
'''
```
works.

Thanks a lot!

TorkelE:
Although maybe I was too fast. But the continuation problem might be separate, so created a new issue: https://github.com/JuliaDocs/Documenter.jl/issues/new

mortenpi:
I'll re-open this to leave it as a documentation issue -- we should mention in the docs that the filtering happens line by line.

PR #1720, opened 2021-10-24
Document that doctest filters match line-by-line

odow:
Closes https://github.com/JuliaDocs/Documenter.jl/issues/1679


docs/src/man/doctests.md
```diff
@@ -300,6 +300,12 @@ DocTestFilters = nothing
 The `DocTestFilters = nothing` is not strictly necessary, but good practice nonetheless to
 help avoid unintentional filtering in following doctest blocks.
 
+!!! info
+    Filters are applied line-by-line to the output. To match multiline strings,
+    use the `s` suffix of a regular expression to match newlines, e.g.,
+    `r"[0-9\.]+ seconds \(.*"s`. See the Julia documentation for `@r_str` for
+    more information on regular expression suffixes.
```
fredrikekre, line 307:
```suggestion
    The filter match is replaced with an empty string in both the expected and actual output using
    `replace`, e.g. `replace(str, filter => "")`. Note that this means that the same filter can match
    multiple times, and if you need the same filter to match multiple lines your regex need to account
    for that.
```

Issue #1168, opened 2019-10-29
Front end setting to make page full-width


mortenpi:
Sometimes the default max width of the main body column is too narrow, so it would be good if there was a way to make it wider, but without compromising readability in most cases.

* Have an option in the new settings dialog to disable the `max-width` on the main `<article>` tag. Implementation: apply a class e.g. `<html>` when that is toggled, similar to how we handle themes. Store the option in Web Storage.

* Also maybe have a keyboard shortcut for this, so that it would be easy to toggle (for people who know/want this). However, it's not really something you could rapidly toggle on-off I think since the reflow of the page will throw you to a different section of the page if you are somewhere in the middle.

Based on a suggestion by @KristofferC on Slack, writing it up so it would not get completely forgotten.

brianguenter:
This has been idle for a while. Is anything happening? I'd also like this capability.

mortenpi:
Nope, it's not really on the priority list, but happy to review a PR if someone wants to take a stab here. I'm okay with having some sort of an UI toggle for this.

PR #1724, opened 2021-10-26
Add responsive centered layout for widescreens

IlyaOrson:
This keeps the current desktop layout but centers both sidebar and content for widescreens.

Here is how the transition looks:
Captura de pantalla 2021-10-25 - 22 09 22
Captura de pantalla 2021-10-25 - 22 09 30
Captura de pantalla 2021-10-25 - 22 09 35

Compared to the current static left alignment in a widescreen:
Captura de pantalla 2021-10-25 - 22 14 48
 
I think a centered layout makes it look more natural. It is also popular in other modern doc generators.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,17 +34,9 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
           show-versioninfo: true
-      - uses: actions/cache@v1
-        env:
-          cache-name: cache-artifacts
-        with:
-          path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-${{ env.cache-name }}-
-            ${{ runner.os }}-test-
-            ${{ runner.os }}-
-      - uses: julia-actions/julia-runtest@latest
+      - uses: julia-actions/cache@v1
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v1
         with:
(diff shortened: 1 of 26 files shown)


IlyaOrson:
Bump.

It is possible to try it out with the Stylus add-on for the time being.
https://uso.kkx.one/style/221705

mortenpi:
It its current current state, I find the sidebar slightly jarring when it gets centered. Fredrik's suggestion to follow the Rust Book example, where the sidebar stays on the left, might be a better option.

However, more broadly, I am little hesitant to merge this kind of style change that fundamentally changes the UX. I would rather have such changes all bundled together into one big style update, rather than continuously messing the style. At the same time, I don't want to block contributions and improvements to the style either, so for that I think there are a couple of possible avenues:

* Better support for custom/user themes. Nothing wrong with making this available as an option that people can enable for their own packages.
* We could have some sort of an experimental style here, where we could more liberally experiment with styles.
* Somehow making it an option that the user can enable (see also #1168).

All that said: if you don't mind updating this, we could try the option Rust Book option. If it feels like a small enough change, then I might drop my objection to this particular change.

IlyaOrson:
I actually implemented that version first but got mostly negative reactions when I shared it in Slack, so I agree a toggle might be the best way to go with a general style change instead.
I thought this would be nicer for now because it is closer to the current version in some sense and it is barely noticeable in small screens.
It might be too much of personal taste thing to be a general change indeed, although I still feel the current style looks a bit outdated.

t-bltg:
Bump, this would be useful for `Plots` [auto generated tables](https://docs.juliaplots.org/dev/generated/attributes_series), which looks chopped.

Currently, we have to downsize `font-size` so that it fits.

AlexisRenchon:
The idea here is make it centered for widescreen? 
Would it be better to make it the default, but keep the option to change in settings?

Note, personally, I use a widescreen and would prefer centered. 
I am not sure what my preference is on a normal laptop so an option would be nice.

PR #1747, opened 2022-01-12
Enable recursive expansion of expanders within block-containing Markdown structures

### LINKED ISSUE #491: ```@repl block inside !!! note doesn't show output

jmert:
In some recent documentation I've been writing, I discovered that ` ```@repl ... ``` ` blocks would not expand if contained within a Markdown list. The first commit in this PR is (essentially) the definition I came up with to solve my immediate problem (i.e. as type/method definitions at the top of my `docs/make.jl` file).

I then discovered #1637 was a proposal for a very similar feature improvement, but it has a couple of unresolved review comments. The second commit takes comments from that other PR and updates this implementation to only recursively expand code-like expanders within nested contexts. It leverages an intermediate abstract type to partition the expanders which should be recursively expanded from those that should not.

I'd consider the third commit to be optional and can be dropped if the change is too disruptive. The idea is that the second commit requires defining "dummy" implementations of `Selectors.order` and `Selectors.matcher` for the intermediate abstract type. The third commit (IMHO) cleans up the abstraction by differentiating between concrete (singleton) subtypes which correspond to runnable actions and abstract types which can be used to build a type hierarchy.

- [x] Needs tests if/when the design approach and names are agreed upon.

Fixes #491

---

**Example**

````md
...multiple transformations are made...

- **Strings**

  As complex data themselves...

  - _Unicode_

    ... combining characters ...
    ```@repl
    acute_a = "\u00E1" # Latin Small Letter A with Acute
    a_combineaccute = "a\u0301" # "a" + Combining Acute Accent
    acute_a == a_combineaccute
    ```
````

**Before:**
> screenshot_before

**After:**
> screenshot_after

### DIFF
--- a/src/Builder.jl
+++ b/src/Builder.jl
@@ -38,38 +38,38 @@ abstract type DocumentPipeline <: Selectors.AbstractSelector end
 """
 Creates the correct directory layout within the `build` folder and parses markdown files.
 """
-abstract type SetupBuildDirectory <: DocumentPipeline end
+struct SetupBuildDirectory <: DocumentPipeline end
 
 """
 Runs all the doctests in all docstrings and Markdown files.
 """
-abstract type Doctest <: DocumentPipeline end
+struct Doctest <: DocumentPipeline end
 
 """
 Executes a sequence of actions on each node of the parsed markdown files in turn.
 """
-abstract type ExpandTemplates <: DocumentPipeline end
+struct ExpandTemplates <: DocumentPipeline end
 
 """
 Finds and sets URLs for each `@ref` link in the document to the correct destinations.
 """
-abstract type CrossReferences <: DocumentPipeline end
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/Builder.jl [hunk 1]
```diff
@@ -38,38 +38,38 @@ abstract type DocumentPipeline <: Selectors.AbstractSelector end
 """
 Creates the correct directory layout within the `build` folder and parses markdown files.
 """
-abstract type SetupBuildDirectory <: DocumentPipeline end
+struct SetupBuildDirectory <: DocumentPipeline end
```
fredrikekre, line 41: Why are these changes necessary? I believe there are some packages that extend Documenter by subtyping some of these.

PR #1748, opened 2022-01-12
Copy button for code blocks

pfitzseb:
image

Fixes https://github.com/JuliaDocs/Documenter.jl/issues/1055.

### DIFF
--- a/assets/html/scss/documenter/_elements.scss
+++ b/assets/html/scss/documenter/_elements.scss
@@ -34,7 +34,7 @@ h1, h2, h3, h4, h5, h6 {
   }
 }
 
-.code-container {
+pre {
   position: relative;
 
   .copy-button {
--- a/assets/html/scss/documenter/_elements.scss
+++ b/assets/html/scss/documenter/_elements.scss
@@ -33,3 +33,23 @@ h1, h2, h3, h4, h5, h6 {
     display: none !important;
   }
 }
+
+.code-container {
+  position: relative;
+
+  .copy-button {
+    opacity: 0;
+    transition: opacity 0.2s;
+    position: absolute;
+    right: 0.5em;
+    top: 0.5em;
+
+    &:focus {
+      opacity: 1;
+    }
+  }
+
+  &:hover .copy-button {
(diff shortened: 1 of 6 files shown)


fredrikekre:
I think it looks better with a copy icon like in https://github.com/JuliaDocs/Documenter.jl/pull/1454 instead of `Copy`

pfitzseb:
Fair.

image
image

fredrikekre:
Could the button be added only if javascript is enabled (i.e. at "runtime")? I did that here https://github.com/fredrikekre/fredrikekre.se/blob/master/_libs/clipboard.js for example. The way I did it there I don't need to do any unescaping, or include the code twice (once printed without colors as you do here IIUC). But maybe thats just because I used https://clipboardjs.com/ which handles that for me? Any reason not to use that?

pfitzseb:
Sure, we can do that as well.

fredrikekre:
Nice, add a changelog entry?

assets/html/scss/documenter/_elements.scss
```diff
@@ -33,3 +33,35 @@ h1, h2, h3, h4, h5, h6 {
     display: none !important;
   }
 }
+
+pre {
+  position: relative;
+
+  .copy-button {
+    opacity: 0;
```
fredrikekre, line 41: Perhaps make it visible without focus too? Something like
```suggestion
    opacity: 0.2;
```
looks pretty good IMO. Thoughts?

fredrikekre:
Something buggy with the icons using the dark theme? I see this:
Screenshot 2022-01-13 at 22 32 54
for example (firefox, macos).

fredrikekre:
Ah, looks like a CSS problem. In the dark theme `font-family` is overridden so it is missing `font-family: "Font Awesome 5 Free";`

pfitzseb:
image

assets/html/scss/documenter/_elements.scss
pfitzseb: It does, yeah :)

Issue #1753, opened 2022-01-17
Copy button does't working


ErikQQY:
Hello guys, the new copy button is really convenient and looks very good, but I think there is still some problem🤔

In the [Documenter docs](https://juliadocs.github.io/Documenter.jl/dev/man/guide/) the button is working fine, but the copy button looks like stuck in this docs: http://scifracx.org/FractionalCalculus.jl/dev/

![error](https://user-images.githubusercontent.com/52615090/149751249-6826684c-f44c-4589-b92a-055db8a98038.png)

fredrikekre:
@pfitzseb ^

pfitzseb:
#1754 will fix the issue, but you could also switch your hosting solution to https for now.

PR #1754, opened 2022-01-17
Improve copy button JS

pfitzseb:
Fixes https://github.com/JuliaDocs/Documenter.jl/issues/1753.


fredrikekre:
I'm okay with limiting this to https, but the patch doesn't look too complicated, so why not.

Issue #1554, opened 2021-03-16
Support workflow_dispatch events on GitHub Actions


mortenpi:
Currently Documenter won't deploy if you run a workflow with `workflow_dispatch`:

```
┌ Info: Deployment criteria for deploying devbranch build from GitHub Actions:
│ - :heavy_check_mark: ENV["GITHUB_REPOSITORY"]="..." occurs in repo="..."
│ - ✘ ENV["GITHUB_EVENT_NAME"]="workflow_dispatch" is "push"
│ - :heavy_check_mark: ENV["GITHUB_REF"] matches devbranch="master"
│ - :heavy_check_mark: ENV["GITHUB_ACTOR"] exists and is non-empty
│ - :heavy_check_mark: ENV["DOCUMENTER_KEY"] exists and is non-empty
└ Deploying: ✘
```

richardreeve:
I just had the same issue with running my documentation workflow with `release`. It does seem like this should be easy to fix if it's really just here to stop pull requests triggering deployment.

PR #1773, opened 2022-03-08
allow deploy docs on schedule

### LINKED ISSUE #1772: Allow docs build on `schedule`
Would it be possible to allow doc deploys on `schedule` jobs? Currently the only events that [trigger a docs build are `push` or `workflow_dispatch](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/deployconfig.jl#L346). 

My use case: I have some examples in my package docs that pull and plot recent data, I think it would be nice if the docs had up to date plots. One possible way to do this is to set up the docs job to run on a schedule (via a github action), however this doesn't trigger a docs build because `schedule` events don't trigger builds. 

Maybe a kwarg could be added to `deploydocs`? so the user could control which events they'd like to trigger builds.

Ianlmgoddard:
This PR adds the `deploy_on_schedule` kwarg to [`deploydocs`](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Documenter.jl#L481) and  [`deploy_folder`](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/deployconfig.jl#L304). This lets the user deploy documentation from scheduled github actions. 

I'm added some tests to check that the kwarg works as expected for `deploy_folder`. I couldn't see where the tests were for `deploydocs` but if tests exist for that function then I should probably add to them.

Also I think this a feature bump? unsure, any advice is appreciated.

closes #1772.

### DIFF
--- a/test/deployconfig.jl
+++ b/test/deployconfig.jl
@@ -381,6 +381,24 @@ end end
                                      devbranch="master", devurl="hello-world", push_preview=false)
         @test !d.all_ok
     end
+    # Build on `schedule` jobs
+    withenv("GITHUB_EVENT_NAME" => "schedule",
+            "GITHUB_REPOSITORY" => "JuliaDocs/Documenter.jl",
+            "GITHUB_REF" => "refs/tags/v1.2.3",
+            "GITHUB_ACTOR" => "github-actions",
+            "GITHUB_TOKEN" => "SGVsbG8sIHdvcmxkLg==",
+            "DOCUMENTER_KEY" => nothing,
+        ) do
+        cfg = Documenter.GitHubActions()
+        d = Documenter.deploy_folder(cfg; repo="github.com/JuliaDocs/Documenter.jl.git",
+                                     devbranch="master", devurl="dev", push_preview=true)
+        @test d.all_ok
+        @test d.subfolder == "v1.2.3"
+        @test d.repo == "github.com/JuliaDocs/Documenter.jl.git"
+        @test d.branch == "gh-pages"
+        @test Documenter.authentication_method(cfg) === Documenter.HTTPS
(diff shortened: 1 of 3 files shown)


Ianlmgoddard:
I didn't add the `kwarg` for the [`deploy_folder` dispatch](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/deployconfig.jl#L163) on `::Travis`. I don't think it's necessary?

test/deployconfig.jl
```diff
@@ -260,20 +269,54 @@ end end
...
+        "GITHUB_REF" => "refs/tags/v1.2.3",
+        "GITHUB_ACTOR" => "github-actions",
+        "GITHUB_TOKEN" => "SGVsbG8sIHdvcmxkLg==",
+        "DOCUMENTER_KEY" => nothing,
+        ) do
+        cfg = Documenter.GitHubActions()
+        d = Documenter.deploy_folder(cfg; repo="github.com/JuliaDocs/Documenter.jl.git",
+                                        devbranch="master", devurl="dev", push_preview=true,
+                                        deploy_on_schedule=false)
+        @test !d.all_ok
```
Ianlmgoddard, line 318: Might be better to explicitly test that this failed because the event name `schedule` isn't accepted when `deploy_on_schedule=false`.

fredrikekre:
I think it shouldn't be configurable in Documenter. If you don't want doc builds for scheduled workflows that should be configured in the workflow file instead IMO.

Ianlmgoddard:
> I think it shouldn't be configurable in Documenter.

Fair enough! I've redone the PR to be basically equivalent to #1554. Only real difference is the addition of a test to ensure all checks pass when the event is `schedule`.

fredrikekre:
Great, thank you!

PR #1771, opened 2022-03-02
Fix repl-code-block styling

pfitzseb:
Fixes https://github.com/JuliaDocs/Documenter.jl/issues/1770.

image

PR #1774, opened 2022-03-09
Force code padding

pfitzseb:


fredrikekre:
I suppose this makes it slightly more awkward to override these things yourself. Perhaps better to just modify my CSS file?

pfitzseb:
Yea, I think that'd be better.

pfitzseb:
Although maybe not? I've [removed the offending line in the original PR too](https://github.com/JuliaDocs/Documenter.jl/pull/1771/files#diff-c73999e36f0592994fec403b076e3f68418663eec065ebd604d2da123deed66e), so maybe it should be fixed in the Documenter CSS directly.

fredrikekre:
What was your conclusion here? :slightly_smiling_face:

pfitzseb:
I think this PR is fine -- you can still override the styling, but it'll just work with upstream themes.

PR #1776, opened 2022-03-11
docerror! to fix #1756

### LINKED ISSUE #1756: strict=true fails to change loglevel from Warning to Error
With `strict=true`, Documenter errors when encountering an issue, but it doesn't actually say "Error" - only "Warning" - on the issue itself, which makes it non-intuitive to figure out what's going on (e.g. search log for `error`).

**What happened:** At the example of [this docs build](https://github.com/JuliaGaussianProcesses/ApproximateGPs.jl/runs/4868224828?check_suite_focus=true):
```
┌ Warning: no docs found for 'LaplaceApproximation' in `@docs` block in src/api/index.md:7-11
│ ```@docs
│ SparseVariationalApproximation
│ LaplaceApproximation
│ posterior
│ ```
└ @ Documenter.Expanders ~/.julia/packages/Documenter/qdbx6/src/Expanders.jl:336
(...)
ERROR: LoadError: `makedocs` encountered an error. Terminating build
```

**What I would have expected** (also based on looking into the source, e.g. `loglevel` in DocChecks.jl):
The first line should have read
```
┌ Error: no docs found for 'LaplaceApproximation' in `@docs` block in src/api/index.md:7-11
```
instead.

st--:
Resolves #1756: With strict=true, Documenter errors when encountering any issue (based on checking doc.internal.errors in `Selectors.runner` in src/Builder.jl), but it doesn't actually say "Error" - only "Warning" - on the issue itself, which makes it non-intuitive to figure out what's going on (e.g. search log for error).

The reason for this is that on any issue, a tag is appended to the doc.internal.errors array, e.g.
```julia
push!(doc.internal.errors, :cross_references)
```
but in most places, it explicitly just prints a warning, e.g. in src/CrossReferences.jl:
```julia
@warn "'$slug' is not unique in $(Utilities.locrepr(page.source))."
```
In src/DocChecks.jl, there was already a `loglevel()` function that based on the strictness settings returns Logging.Error or Logging.Warn, so the log message of those checks was already correct.

In this PR, I combined the push! to doc.internal.errors & the appropriate logging message in a new function `docerror!`, and changed the other files to use this function as well.

### DIFF
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -530,7 +530,14 @@ end
         @test :meta_block ∈ doc.internal.errors
         @test_logs (:error, "doctest issue 123") (Documenter.Utilities.@docerror(doc, :doctest, "doctest issue $foo"))
         @test :doctest ∈ doc.internal.errors
-        @test_throws ArgumentError @macroexpand Documenter.Utilities.@docerror(doc, :foo, "invalid tag")
+        try
+            @macroexpand Documenter.Utilities.@docerror(doc, :foo, "invalid tag")
+            error("unexpected")
+        catch err
+            err isa LoadError && (err = err.error)
+            @test err isa ArgumentError
+            @test err.msg == "tag :foo is not a valid Documenter error"
+        end
     end
 end
 
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -530,6 +530,7 @@ end
         @test :meta_block ∈ doc.internal.errors
         @test_logs (:error, "doctest issue 123") (Documenter.Utilities.@docerror(doc, :doctest, "doctest issue $foo"))
         @test :doctest ∈ doc.internal.errors
+        @test_throws LoadError eval(:(Documenter.Utilities.@docerror(doc, :foo, "invalid tag")))
     end
 end
 
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -530,7 +530,7 @@ end
         @test :meta_block ∈ doc.internal.errors
         @test_logs (:error, "doctest issue 123") (Documenter.Utilities.@docerror(doc, :doctest, "doctest issue $foo"))
         @test :doctest ∈ doc.internal.errors
-        @test_throws LoadError eval(:(Documenter.Utilities.@docerror(doc, :foo, "invalid tag")))
+        @test_throws ArgumentError @macroexpand Documenter.Utilities.@docerror(doc, :foo, "invalid tag")
     end
 end
 
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -522,6 +522,15 @@ end
         @test_throws ArgumentError Documenter.Utilities.check_strict_kw(:a)
         @test_throws ArgumentError Documenter.Utilities.check_strict_kw([:a, :doctest])
     end
+
+    @testset "@docerror" begin
+        doc = (; internal = (; errors = Symbol[]), user = (; strict = [:doctest, :setup_block]))
+        foo = 123
+        @test_logs (:warn, "meta_block issue 123") (Documenter.Utilities.@docerror(doc, :meta_block, "meta_block issue $foo"))
+        @test :meta_block ∈ doc.internal.errors
+        @test_logs (:error, "doctest issue 123") (Documenter.Utilities.@docerror(doc, :doctest, "doctest issue $foo"))
+        @test :doctest ∈ doc.internal.errors
+    end
 end
 
 end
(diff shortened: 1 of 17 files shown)


fredrikekre:
A disadvantage with this strategy is that we lose source information (all log messages will look like they come from this new function).

st--:
> A disadvantage with this strategy is that we lose source information (all log messages will look like they come from this new function).

Ah, that's unfortunate. Could we fix it by turning it into a `@docerror` macro?

fredrikekre:
Yea, but maybe it is easier to keep the
```
@logmsg loglevel(...) "message"
```
pattern as was already there in some places?

st--:
I thought it was helpful to directly couple the push! to doc.internal.errors with the message, as this removes a lot of duplicated code

st--:
How about
```julia
macro docerror(doc, tag, msg, exs...)
    quote
        let
            push!($(doc).internal.errors, $(tag))
            if is_strict($(doc).user.strict, $(tag))
                @error $(msg) $(exs...)
            else
                @warn $(msg) $(exs...)
            end
        end
    end
end
```
and then instead of
```julia
push!(doc.internal.errors, :cross_references)
@warn "'$slug' is not unique in $(Utilities.locrepr(page.source))."
```
can write
```julia
@docerror doc :cross_references "'$slug' is not unique in $(Utilities.locrepr(page.source))."
```

st--:
I think it does (I started converting it over); could you turn on the workflow for me ?

st--:
it should all be ready for CI now

test/utilities.jl
```diff
@@ -522,6 +522,16 @@ end
...
     end
+
+    @testset "@docerror" begin
+        doc = (; internal = (; errors = Symbol[]), user = (; strict = [:doctest, :setup_block]))
+        foo = 123
+        @test_logs (:warn, "meta_block issue 123") (Documenter.Utilities.@docerror(doc, :meta_block, "meta_block issue $foo"))
+        @test :meta_block ∈ doc.internal.errors
+        @test_logs (:error, "doctest issue 123") (Documenter.Utilities.@docerror(doc, :doctest, "doctest issue $foo"))
+        @test :doctest ∈ doc.internal.errors
+        @test_throws LoadError eval(:(Documenter.Utilities.@docerror(doc, :foo, "invalid tag")))
```
fredrikekre, line 533: You can catch the error from `@macroexpand $docerror ...` instead of using `eval`.

st--: much better, thanks!

test/utilities.jl
```diff
@@ -522,6 +522,16 @@ end
...
     end
+
+    @testset "@docerror" begin
+        doc = (; internal = (; errors = Symbol[]), user = (; strict = [:doctest, :setup_block]))
+        foo = 123
+        @test_logs (:warn, "meta_block issue 123") (Documenter.Utilities.@docerror(doc, :meta_block, "meta_block issue $foo"))
+        @test :meta_block ∈ doc.internal.errors
+        @test_logs (:error, "doctest issue 123") (Documenter.Utilities.@docerror(doc, :doctest, "doctest issue $foo"))
+        @test :doctest ∈ doc.internal.errors
+        @test_throws ArgumentError @macroexpand Documenter.Utilities.@docerror(doc, :foo, "invalid tag")
```
fredrikekre, line 533: This throws `LoadError` on some older julia versions, so you have to do
```suggestion
        try
            @macroexpand Documenter.Utilities.@docerror(doc, :foo, "invalid tag")
            error("unexpected")
        catch err
            err isa LoadError && (err = err.error)
            @test err isa ArgumentError
            @test err.msg == "tag :foo is not a valid Documenter error"
        end
```

st--: thanks for figuring out how to fix it !

st--:
Phew, now all the tests are passing (except for codecov/patch, but I don't think there's an easy way of fixing that, so I hope it's acceptable as is) :)

fredrikekre:
Great, can you add a note in CHANGELOG.md? Just mention that the issue is fixed.

fredrikekre:
Code coverage dropped to 50% from 85%. Looks like every function that uses your new macro is marked as missed, so probably something is strange here...

st--:
> Code coverage dropped to 50% from 85%. Looks like every function that uses your new macro is marked as missed, so probably something is strange here...

That seems like a build issue, not an issue with my code - if you look at previous builds (e.g. [codecov/project](https://codecov.io/gh/JuliaDocs/Documenter.jl/compare/a31cddb882638d50fdbc9d087b941246993c00a4...e0dab4c5923d1aa8c13225a9c9f2dafa9b5d47f9) on e0dab4c it says 86.91% (+0.17%). Could you re-run the build ? GitHub has been having some issues as of yesterday.

fredrikekre:
Ok, that seems to have worked. Strange.

PR #1783, opened 2022-03-30
Added QuizQuestion to README

sylvaticus:
Added QuizQuestions.jl to "Other third-party packages that can be combined with Documenter".
QuizQuestions allows embedding simple interactive quizzes in the Markdown source page that is then converted to HTML by Documenter.


mortenpi:
LGTM, thanks!

Just an idea: it would be cool to see the widgets in action on an example page in the QuizQuestion docs :slightly_smiling_face:

Issue #1780, opened 2022-03-21
Monospace font "faux" bolding causing alignment issues in some cases 


mostol:
Hi! I was looking into some [alignment issues](https://github.com/FedeClaudi/Term.jl/issues/46) with the docs for the [Term.jl](https://github.com/FedeClaudi/Term.jl), package which uses Documenter.jl. From what I can tell, the source of the problem is that (in some browsers, at least,) when the default JuliaMono typeface is bolded, it uses the browser's "faux"-bolding technique, which can mess up the alignment on what *should* be a universally-monospace font. See [my comment on that issue](https://github.com/FedeClaudi/Term.jl/issues/46#issuecomment-1074442800) if you want a more detailed description of what I'm running into.

It looks like Documenter.jl is using a small `@font-family` ruleset as defined in the JuliaMono repo, but is *only* using `JuliaMono-Regular` (see #1569):
https://github.com/JuliaDocs/Documenter.jl/blob/bffc8665b9753ec7108b98b9788bef8a5cb4a482/src/Writers/HTMLWriter.jl#L525

Would it be a good idea to change this to use the [`juliamono.css`](https://cdnjs.cloudflare.com/ajax/libs/juliamono/0.044/juliamono.css) file instead of `juliamono-regular.css` to ensure *all* font weights are properly monospaced across browsers/devices?

cormullion:
Cool detective work! I wish I understood all this stuff better. One thing I don't get - how is Federico managing to get bold and italic monospaced markup into Documenter .md files? I'd quite like bold keywords ...  😃

fredrikekre:
> One thing I don't get - how is Federico managing to get bold and italic monospaced markup into Documenter .md files? I'd quite like bold keywords ...

Probably this is just a result of https://github.com/JuliaDocs/Documenter.jl/pull/1441. I think before that, (but after switching to JuliaMono) this wasn't a problem. See https://github.com/JuliaDocs/Documenter.jl/issues/618 which was closed by using JuliaMono as the default font.

mostol:
Yeah, it looks like `.ansi` CSS classes are applied to those code boxes where appropriate, so it just happens automatically it seems—which is really neat!

But on that topic, is there an easy way to apply custom CSS to a Documenter.jl-generated site?  Because for this particular issue, it seems like the fixes would either be:
1) Do nothing to Documenter.jl, and create some custom CSS for Term.jl so that it includes the necessary, broader font files.
[...]

PR #1784, opened 2022-04-01
Fix JuliaMono CSS to support all weights/variants

mostol:
Fixes https://github.com/JuliaDocs/Documenter.jl/issues/1780. Previous CSS file only supported JuliaMono Regular, which could create some spacings issues (see https://github.com/FedeClaudi/Term.jl/issues/46, for example). Now light/bold/italic font variations should use the proper font files, instead of faux-bolding/italicizing. Also updates JuliaMono source to v0.044 (up from v0.039). Tested on MacOS with Safari, and all font weights/variants worked correctly ☺️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


mortenpi:
Thanks, LGTM! Would you mind adding a note into `CHANGELOG.md` too?

mostol:
@mortenpi Like so?

mortenpi (approved):
Perfect, thank you!

PR #1785, opened 2022-04-03
Set TOC depth to chapter level

ViralBShah:


ViralBShah:
I guess that can be useful to have since the manual is so large. Maybe we just need to improve the spacing for sections in the manual. 

@inkydragon Would you know how to do that?

ViralBShah:
I think Julia manual is the biggest consumer of pdf manuals. Also a bit of extra spacing will not hurt other projects -so I think we should just do it universally. And we should revert my change to TOC depth here- and increase spacing for sections.

ViralBShah:
Mainly need to fix these two (which should be fixed in Documenter.jl):

Screen Shot 2022-04-04 at 12 17 34 PM

Screen Shot 2022-04-04 at 12 17 54 PM

ViralBShah:
Do we not need something like this to increase the spacing between the section number and section title?

```
\setlength{\cftsectionnumwidth}{2.5em \@plus \p@}
```

ViralBShah:
Thanks to your really nice comments in the code, I had indeed found that diagram, which is the most useful documentation!

inkydragon:
4.0em VSVS 3.0em

image

> Does this PR look good? 

LGTM.

> It would be nice to add the PDF building tips to Documenter's docs.

Yes, however, it looks like only julia's manual is slow to compile, so perhaps the PDF building tips should be added to julia's project documentation.

Documenter.jl might need to include a test that has multiple levels of chapters & sections, like julia's manual.

ViralBShah:
I installed `tectonic` with homebrew. It doesn't seem to have latex and such commands. Howevever, I have been hearing that people are rapidly adopting tectonic, and we may want to do that ourselves eventually. I had to use the docker method to build the pdf in julia base.

inkydragon:
> It doesn't seem to have latex and such commands.

```sh
tectonic -X compile --keep-logs -Z shell-escape  TheJuliaLanguage.tex
```

And if you want code highlighting, you need install `minted`: [installing - How to install minted in Ubuntu - TeX - LaTeX Stack Exchange](https://tex.stackexchange.com/questions/40083/how-to-install-minted-in-ubuntu)

`minted` needs to call an external program `pygments` which  is a python pkg.

ViralBShah:
With the tectonic command, I get:

```
error: ./documenter.sty:4: Package fontspec Error: The font "DejaVu Sans" cannot be found.

For immediate help type H <return>
Writing `TheJuliaLanguage.log` (12.21 KiB)

```

ViralBShah:
@mortenpi @fredrikekre Can I go ahead and tag a new release to get these pdf improvements into julia?

inkydragon:
> The font "DejaVu Sans" cannot be found.

Missing fonts
- `DejaVuSans.ttf`
- `DejaVuSansMono.ttf`

[Install dejavu-fonts on macOS with MacPorts](https://ports.macports.org/port/dejavu-fonts/)

ViralBShah:
Appreciate it. Now:

```
error: _minted-TheJuliaLanguage/681C970CC716FB99FA936E8A6E371DE25218153D23BFA92BF6039169C02165FC.pygtex:2: Undefined control sequence
Writing `TheJuliaLanguage.log` (58.65 KiB)
error: halted on potentially-recoverable error as specified
```

I have `pygments` installed.

inkydragon:
Deleting temporary files and folder `_minted-*`, then try again.

[pdftex - Minted not working - TeX - LaTeX Stack Exchange](https://tex.stackexchange.com/questions/237075/minted-not-working)

ViralBShah:
Thank you - that worked.

PR #1795, opened 2022-04-11
Add version option to LaTeX format

mortenpi:
@inkydragon This is what I had in mind as a stopgap solution for the version, until we support it for all formats in `makedocs`. Do the changes look good to you?


mortenpi:
Good catch @inkydragon!

PR #1796, opened 2022-04-11
Fix outdated GitHub Docs links

mortenpi:
They have been [throwing 301s](https://github.com/JuliaDocs/Documenter.jl/runs/5965277331?check_suite_focus=true). Should hopefully fix linkcheck.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * ![Enhancement][badge-enhancement] The preamble of the LaTeX source of the PDF build can now be customized by the user. ([#1746][github-1746], [#1788][github-1788])
 * ![Bugfix][badge-bugfix] Fix `strict` mode to properly print errors, not just a warnings. ([#1756][github-1756], [#1776][github-1776])
 * ![Bugfix][badge-bugfix] Disable git terminal prompt when detecting remote HEAD branch. ([#1797][github-1797])
+* ![Bugfix][badge-bugfix] When linkchecking HTTP and HTTPS URLs, Documenter now passes a realistic browser (Chrome) `User-Agent` header along with the request, in order to work around servers that try to use the `User-Agent` to block non-browser requests. ([#1796][github-1796])
 
 ## Version `v0.27.15`
 
@@ -1002,6 +1003,7 @@
 [github-1784]: https://github.com/JuliaDocs/Documenter.jl/pull/1784
 [github-1785]: https://github.com/JuliaDocs/Documenter.jl/pull/1785
 [github-1788]: https://github.com/JuliaDocs/Documenter.jl/pull/1788
+[github-1796]: https://github.com/JuliaDocs/Documenter.jl/pull/1796
 [github-1797]: https://github.com/JuliaDocs/Documenter.jl/pull/1797
 <!-- end of issue link definitions -->
 
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -26,8 +26,6 @@ makedocs(
     authors = "Michael Hatherly, Morten Piibeleht, and contributors.",
     linkcheck = "linkcheck" in ARGS,
     linkcheck_ignore = [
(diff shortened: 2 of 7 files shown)


mortenpi:
I believe the 403s that linkcheck complains about are an issue on GitHub's end, as I am also seeing them locally (intermittently) along with some 503s. I would not merge this PR until linkcheck actually passes.

mortenpi:
It actually starts looking like the 403s on `docs.github.com` are not an intermittent problem, but instead it looks like GitHub has decided to block non-browser requests based on the `User-Agent`. For example, simply running `curl -I` (which should be passing a `curl/7.68` UA):

```
$ curl -I https://docs.github.com/en
HTTP/2 403 
x-azure-ref: 0sadXYgAAAADZSTHMRxIAT5rY1kqwPFNHQUtMMzBFREdFMDMxOQA1OTZkNzhhMi1jYTVmLTQ3OWQtYmNkYy0wODM1ODMzMTc0YjI=
accept-ranges: bytes
date: Thu, 14 Apr 2022 04:48:49 GMT
via: 1.1 varnish
x-served-by: cache-akl10327-AKL
x-cache: MISS
x-cache-hits: 0
x-timer: S1649911729.432441,VS0,VE6
strict-transport-security: max-age=31557600
```

vs. when you spoof a browser UA:

```
$ curl -I --user-agent "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36" https://docs.github.com/en
HTTP/2 200 
cache-control: private, no-store
content-type: text/html; charset=utf-8
etag: "205a6-GRu78m11eWLYGT3vvB/H5ULI/5A"
set-cookie: _csrf=Yc_DxYKyMzpLByv2ZPRwPhMW; Path=/; HttpOnly; Secure; SameSite=Lax
access-control-allow-origin: *
content-security-policy: default-src 'none';prefetch-src 'self';connect-src 'self';font-src 'self' data: githubdocs.azureedge.net;img-src 'self' data: github.githubassets.com githubdocs.azureedge.net placehold.it *.githubusercontent.com github.com;object-src 'self';script-src 'self';frame-src https://graphql.github.com/ https://www.youtube-nocookie.com;style-src 'self' 'unsafe-inline';child-src 'self'
x-dns-prefetch-control: off
expect-ct: max-age=0
x-frame-options: SAMEORIGIN
x-download-options: noopen
x-content-type-options: nosniff
x-permitted-cross-domain-policies: none
referrer-policy: strict-origin-when-cross-origin
x-xss-protection: 0
x-powered-by: Next.js
x-azure-ref: 0qadXYgAAAACteYID9J/NRqwA1Kgj8QVXQUtMMzBFREdFMDMxMgA1OTZkNzhhMi1jYTVmLTQ3OWQtYmNkYy0wODM1ODMzMTc0YjI=
accept-ranges: bytes
date: Thu, 14 Apr 2022 04:48:42 GMT
via: 1.1 varnish
x-served-by: cache-akl10328-AKL
x-cache: CONFIG_NOCACHE, MISS
x-cache-hits: 0
x-timer: S1649911722.673083,VS0,VE339
vary: Accept-Encoding
strict-transport-security: max-age=31557600
content-length: 132518
```

To work around this, I changed linkcheck to spoof a realistic browser UA (somewhat arbitrarily picking the [example Chrome UA from Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#chrome_ua_string)), which makes linkcheck pass again with the `docs.github.com` links.

I considered having the spoofed UA be a fallback, if the initial linkcheck fails, but the complexity is probably not worth it. This does mean that theoretically it's possible that some linkchecks will now fail because the servers will consider this UA to be suspicious or something along those lines. But I think it should mostly have the opposite effect --- any server that cares about the UA on this level will hopefully consider this UA to be more trustworthy.

Issue #1074, opened 2019-07-25
Integrating Documenter with parsers for non-markdown documents


non-Jedi:
Related issue: #821.

I've started writing [a basic parser for org-mode
syntax](https://github.com/non-Jedi/Org.jl) and am wondering what the
best way of integrating it with Documenter would be. Looking through
the unfamiliar codebase, would my best bet be to implement
`Base.convert(Documenter.Utilities.Markdown2.MD, ::Org.Document)`? The
problem then is that I would still need to override the
`Documents.Page` constructor. I could keep going down this path, but
I'd basically just be trying to shoe-horn things into an architecture
that's not currently well-built as far as I can see for multiple
document parsers.

Any thoughts on the changes that would be necessary on the Documenter
end for this to work? Or an alternative approach for integration on my
end that wouldn't require Documenter changes?

mortenpi:
I have a branch that extends #552 which provides an interface for this kind of thing ([`mp/external-pages`](https://github.com/JuliaDocs/Documenter.jl/tree/mp/external-pages)). Essentially, you have to provide a parser function that goes `::String -> Markdown.MD`. You can see an example invocation [here](https://github.com/JuliaDocs/Documenter.jl/blob/50c28050db790886224741858ce1eb9fc4adc404/docs/make.jl#L19). That particular parser comes from [CMark.jl](https://github.com/mortenpi/CMark.jl).

I would encourage you to try out that branch and let me know if it works. I want to turn that into a proper feature in the near-ish future and feedback would definitely help. You just have to be able to convert your docs to the standard `Markdown.MD` trees.

CiaranOMara:
I too am interested in those external pages.

PR #1798, opened 2022-04-13
Added preprocess [default to  (page,path) -> page] option

sylvaticus:
The preprocess function is called after having read the String from
file and before it is parsed as a Markdown page.

### DIFF
--- a/src/Builder.jl
+++ b/src/Builder.jl
@@ -90,6 +90,7 @@ function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
     build  = doc.user.build
     source = doc.user.source
     workdir = doc.user.workdir
+    preprocess = doc.user.preprocess
 
 
     # The .user.source directory must exist.
@@ -128,7 +129,7 @@ function Selectors.runner(::Type{SetupBuildDirectory}, doc::Documents.Document)
 
             if endswith(file, ".md")
                 push!(mdpages, Utilities.srcpath(source, root, file))
-                Documents.addpage!(doc, src, dst, wd)
+                Documents.addpage!(doc, src, dst, wd, preprocess)
             else
                 cp(src, dst; force = true)
             end
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -221,6 +221,10 @@ ignored.
 **`linkcheck_timeout`** configures how long `curl` waits (in seconds) for a link request to
 return a response before giving up. The default is 10 seconds.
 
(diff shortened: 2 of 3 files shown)


sylvaticus:
ugh, from what I have understood the error doesn't seem to be related to this pull request...

mortenpi:
I appreciate the PR, but I am skeptical whether we want to add another option to `makedocs` for this. This can easily be achieved by writing a small loop in the `make.jl` before you run `makedocs`. But I am happy to leave this open for a while to get other opinions (@fredrikekre?).

PS: if your concern is modifying source files (which I agree would be bad practice), you can still have your source files in `src/`, but then have the preprocess loop write the processed files into e.g. `src-processed/` and then set the `source` keyword to `makedocs` correspondingly.

sylvaticus:
Hello, I find using a callback much neater. The motivations are exactly [those for the equivalent functionality in Literate.jl](https://fredrikekre.github.io/Literate.jl/v2/customprocessing/), just there it applies only to the `*.jl` -> `*.md` transformation, while here it would apply to any page whatever its origin is.
Yes I can cp to a temporary directory and preprocess/make documentation there but then the links "Edit on GitHub" would not work, so I need to cp to a temporary directory, work on the original directory and then cp back temporary directory -> original source directory, not something I like much... :-(
As I said, I just feel using a callback is the neatest approach, but again, that's just an opinion...

mortenpi:
Hmm. I concede the point about the edits links --- it would be nice if those would _just work_, without having mess around with the `EditURL`.

@fredrikekre: From the perspective of Literate UX, do you think it would be a good idea to have some way to hook Literate more directly into Documenter? Or do you think keeping them fully decoupled and using intermediate files like we do now is better?

But I am still a little hesitant to introduce this particular API. I do think though that some sort of user-provided parsing of files would be a good thing, but I think it should be generating the AST directly so that it would be more flexible (e.g. I have an ancient experimental branch that allows a different Markdown parser to be used: https://github.com/JuliaDocs/Documenter.jl/issues/1074#issuecomment-515477974). This type of pre-processing would then just be a special case. But such a feature is much more work.

sylvaticus:
Just a small comment: If you allow hack at the level of the AST then users would need  much more knowledge on the internals of Documenter/Markdown parser compared than just working on a string..

mortenpi:
That is true, but the user can fall back to using the standard parser and not actually work with the AST. Basically, rather than returning a string `s` from the callback, you would return e.g. `Markdown.parse(s)`.

PR #1800, opened 2022-04-19
Add a CI job to verify CHANGELOG

mortenpi:
Following up on #1787. This should catch various errors (missings links etc.) that get introduced into the CHANGELOG sometimes.

It depends on an [experimental branch of CommonMark.jl](https://github.com/mortenpi/CommonMark.jl/tree/mp/reference-links), as CommonMark does not fully support reference links right now (MichaelHatherly/CommonMark.jl#40). Once those changes get into CommonMark, the code here should be updated and then the `Manifest.toml` can be removed.

### DIFF
--- a/.github/workflows/repochecks.yml
+++ b/.github/workflows/repochecks.yml
@@ -20,13 +20,13 @@ jobs:
       - name: Test scripts/changelog.jl
         run: |
           # Run the unit tests of the changelog.jl script:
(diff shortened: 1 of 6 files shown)


mortenpi:
I'll go ahead an merge this as it does work and so that we could get the benefits of this CI job, even though the changes in CommonMark need some more work.

PR #1805, opened 2022-04-24
Update the warning for dev version docs

ViralBShah:


assets/html/warner.js
```diff
@@ -37,7 +37,7 @@ function maybeAddWarning () {
         document.body.removeChild(div);
     });
     const href = window.documenterBaseURL + '/../' + window.DOCUMENTER_STABLE;
-    div.innerHTML = 'This documentation is not for the latest version. <br> <a href="' + href + '">Go to the latest documentation</a>.';
+    div.innerHTML = 'This documentation is for the development version.<br> <a href="' + href + '">Go to the documentation for the latest stable release</a>.';
```
mortenpi, line 40: Hmm. This is not entirely correct, since the same message also gets displayed on old versions. Maybe just say "this is not for the current latest stable release" or something like that?

mortenpi: @ViralBShah What do you think of this?

```suggestion
    div.innerHTML = 'This documentation is not for the latest stable release, but for either the development version or an older outdated release..<br> <a href="' + href + '">Click here to go to the documentation for the latest stable release</a>.';
```

ViralBShah: I just deleted the word `outdated`, since older release is sufficient.

PR #1808, opened 2022-04-28
Refactor the repo/remote handling

mortenpi:
This factors out the refactoring part of #1254 to make the whole thing a little more manageable. In short, it introduces the `Remotes` module which allows for a more flexible way to define the remote repository. The `Remote` objects can be passed via the `repo` keyword to `makedocs`:

```julia
makedocs(repo = Remotes.GitHub("JuliaDocs", "Documenter.jl"), ...)
```

The user can also define a custom `Remotes.Remote` type and overload the `Remotes.repourl` and `Remotes.fileurl` functions to control the link generation.

We do quite a bit of special handling of GitLab, BitBucket and Azure Devops repositories, so it would probably be worth adding special `Remotes.Remote` types for each of those too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@
   **For upgrading:** To keep using the Markdown backend, refer to the [DocumenterMarkdown package][documentermarkdown]. That package might not immediately support the latest Documenter version, however.
 
 * ![Enhancement][badge-enhancement] The `ansicolor` keyword to `HTML()` now defaults to true, meaning that executed outputs from `@example`- and `@repl`-blocks are now by default colored (if they emit colored output). ([#1828][github-1828])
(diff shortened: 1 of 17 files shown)


mortenpi:
I think this in a mergeable shape now. I reduced it down to the bare minimum to introduce the `Remotes` module, without changing any behavior as far as I can tell (except one case, where I replaced one buggy behavior with another).

I do have a bunch of follow-up ideas, to properly fix up `getremote` etc., which will likely change some behaviors and also refactor parts of the code. And then there is the UI change in #1254. But to make it manageable to review and to make sure I am not accidentally changing or breaking anything I did not intend, I want to merge this minimal version first and do those things in follow-up PR(s).

@fredrikekre Would you have time to look through this? It's a little big, but I would love to get another set of eyes on this.

mortenpi:
Thanks @fredrikekre!

PR #1809, opened 2022-05-02
Set version to 0.27.17

mortenpi:


fredrikekre (approved):

mortenpi:
Let's get #1811 in as well.

PR #1816, opened 2022-05-10
improve error message to only print fatal errors

ericphanson:
Currently it prints all errors, which can be confusing (makes it look like `strict` options aren't working correctly)


mortenpi:
Thanks, yea, this makes more sense. The patch LGTM, so I am not sure where the test failures are coming from. Do you also get them locally? Also, could you add a small CHANGELOG note for this?

ericphanson:
I was able to repro locally. For some reason the test output is suppressed, but I was able to edit stuff to uncover an undefined variable exception- which was then obvious in retrospect because I use `c` later without defining it. Should be fixed, along with a changelog addition.

mortenpi:
Looks great, thank you!

PR #1818, opened 2022-05-11
Reduce padding in some container elements

### LINKED ISSUE #1814: Excessive padding in some cases (bikeshedding)
I feel that the padding used in some elements in the default style is a bit excessive and makes these element take up a lot of vertical space.

## Admonitions

gnome-shell-screenshot-62tlxj

Looking at e.g. the docs for Python https://docs.python.org/3/whatsnew/3.10.html and adopting that (7px) would give:

gnome-shell-screenshot-48ref4

## Code blocks

image

Going from 0.7 to 0.3 in the padding gives something like:

image

mortenpi:
Reduces and makes the padding more consistent between various container elements. For admonitions, it changes as follows:

padding-admonition-old
padding-admonition-new

It also makes the padding in the docstrings consistent with the admonitions, although I left the content vertical padding a little higher, since docstrings often start with a code block, and `0.5rem` looked a little too narrow then:

padding-docstrings-old
padding-docstrings-new

For code blocks, the vertical padding is slightly reduced, but left padding actually increases (to be consistent with admonitions and docstrings):

padding-code-old
padding-code-new

@KristofferC: what do you think? Not quite as narrow as you suggested, but perhaps a good compromise? Close #1814.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,6 +15,7 @@ jobs:
       matrix:
         version:
           - '1.0'
+          - '1.6'
           - '1'
           - 'nightly'
         os:
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
         with:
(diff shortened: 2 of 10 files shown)


KristofferC:
Yes, looks a lot better to my eyes. Thanks!

PR #1820, opened 2022-05-19
Fix `binding` for functors

### LINKED ISSUE #558: Documenter cannot handle new call synatx
Documenter cannot handle the new call syntax.
```julia
doc"""
dasd
"""julia
(::Foo)(::Int) = 1
```
In the REPL one can access the docstring as follows:
```julia
help?> (::Foo)(::Int)
  dasd
```
When I include it using
````julia
```@docs
(::Foo)(::Int)
```
````
I get:
```julia
ERROR: ErrorException("`binding` cannot understand expression `::Foo`.")

expression '(::Foo)(::Int)' in module 'Bla'
```

mgkurtz:
Together with JuliaLang/julia#45367 this should fix #558.

### DIFF
--- a/src/DocSystem.jl
+++ b/src/DocSystem.jl
@@ -61,8 +61,8 @@ end
 #
 binding(m::Module, x::Expr) =
     Meta.isexpr(x, :.) ? binding(getmod(m, x.args[1]), x.args[2].value) :
-    Meta.isexpr(x, [:call, :macrocall, :curly]) ? binding(m, x.args[1]) :
-    Meta.isexpr(x, :where) ? binding(m, x.args[1].args[1]) :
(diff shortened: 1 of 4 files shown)


mortenpi:
Is this fix backwards compatible with Julia versions that don't include the JuliaLang/julia/pull/45367 fix?

mgkurtz:
The change in handling `:where` is, by the way, not really related. It only removes a superfluous(?) shortcut (taking two recursion steps in one) which seemed to ask for being a bug. (And it _does_ produce an error, if one is mad enough to ask for the documentation of `Int where T` which is a valid Julia expression egal to `Int`.) Or did this special casing of `:where` had some other reason?

mortenpi:
I'd be happy to have the tests be version-guarded, although that would mean waiting for JuliaLang/julia#45367 to get merged.

PR #1824, opened 2022-05-24
Set version to 0.27.18

mortenpi:
I am planning on turning `master` into the dev branch for 0.28. I want to make some refactoring changes over the next few months (e.g. #1808, #1254; and also some upcoming). And while I don't expect them to be breaking per se, there might be unintended breakage and so I don't want to push these changes to people's CI automatically.

We can still release bugfixes and safe enhancements as 0.27.x patches via the usual backporting procedure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Documenter.jl changelog
 
-## Unreleased
+## Version `v0.27.18`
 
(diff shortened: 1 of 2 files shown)


mortenpi:
Okay, I forgot we had test dependencies on downstream packages. Let's not update the version then.

fredrikekre (approved):

PR #1819, opened 2022-05-15
Set `permissions: write-all` in `hosting.md`

rikhuijzer:
GitHub seems to have changed the permissions for the `secrets.GITHUB_TOKEN`. It does not allow writing by default anymore. I've now had two new repositories where I copy-pasted some workflow and got the following back
```
[docs cdb3219] build based on 2be61e3
 12 files changed, 16162 insertions(+)
 create mode 100644 dev/assets/documenter.js
 create mode 100644 dev/assets/search.js
 create mode 100644 dev/assets/themes/documenter-dark.css
 create mode 100644 dev/assets/themes/documenter-light.css
 create mode 100644 dev/assets/themeswap.js
 create mode 100644 dev/assets/warner.js
 create mode 100644 dev/index.html
 create mode 100644 dev/search/index.html
 create mode 100644 dev/search_index.js
 create mode 100644 dev/siteinfo.js
 create mode 100644 index.html
 create mode 100644 versions.js
remote: Write access to repository not granted.
fatal: unable to access 'https://github.com/rikhuijzer/MyProject.jl.git/': The requested URL returned error: 403
```
Explicitly setting
```
permissions: write-all
```
fixed it.

More information about the permissions for the `GITHUB_TOKEN` can be found at https://docs.github.com/en/actions/security-guides/automatic-token-authentication. I haven't yet found related release notes.

PR #1829, opened 2022-05-26
Set `permissions: content: write` in `hosting.md`

rikhuijzer:
Fix for the too permissive setting in https://github.com/JuliaDocs/Documenter.jl/pull/1819.

Thanks to @SaschaMann for noting that

> This is not a good practice. Instead, only the minimal permissions that are needed to make the workflow work should be given to the token.

See https://github.com/julia-actions/julia-docdeploy/pull/21 for details.

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -189,7 +189,8 @@ on:
 
 jobs:
   build:
-    permissions: write-all
+    permissions:
+      contents: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


mortenpi:
LGTM, thanks!

fredrikekre:
Does this allow publishing PR statuses?

SaschaMann:
No, those need either [`statuses`](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-statuses) or [`checks`](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-checks).

mortenpi:
So `deploydocs` calls out to these APIs:

https://github.com/JuliaDocs/Documenter.jl/blob/fdaa610e20c3a03c74dad53a67bf519d0baeebfa/src/deployconfig.jl#L499

https://github.com/JuliaDocs/Documenter.jl/blob/fdaa610e20c3a03c74dad53a67bf519d0baeebfa/src/deployconfig.jl#L521

Is `statuses: write` enough for these?

rikhuijzer:
> Is `statuses: write` enough for these?

Probably also `pull-requests: write` or what does that second API do? Someone here who can quickly try it out on one of their repositories? I'm not using this feature.

PR #1830, opened 2022-05-27
Automatically determine edit_link default

### LINKED ISSUE #1827: Default `edit_link` to default branch rather than "master"
Like is already done for `devbranch` https://github.com/JuliaDocs/Documenter.jl/pull/1751

Xref #1443, #1375

mortenpi:
Generalizes the solution to `devbranch` to also apply to `edit_link`. Fix #1827.

It also now prints a warning (for both `devbranch` and `edit_link`) if the `git remote` solution somehow fails. I think this is a good idea, because otherwise you might get some weird intermittent environment-dependent failures which are hard to debug. The warning should at least give you a hint where things might be going wrong.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -800,7 +800,14 @@ function git_remote_head_branch(varname, root; remotename = "origin", fallback =
     cmd = `git remote show $(remotename)`
     stderr_output = IOBuffer()
     git_remote_output = try
-        read(pipeline(setenv(cmd, env; dir=root); stderr=stderr_output), String)
+        # Older Julia versions don't support pipeline-ing into IOBuffer.
+        stderr_redirect = if VERSION >= v"1.6.0"
+            stderr_output
+        else
+            write(stderr_output, "(no output redirect on Julia $VERSION)")
+            devnull
+        end
+        read(pipeline(setenv(cmd, env; dir=root); stderr = stderr_redirect), String)
     catch e
         @warn """
         Unable to determine $(varname) from remote HEAD branch, defaulting to "$(fallback)".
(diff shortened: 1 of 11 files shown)


src/Utilities/Utilities.jl
```diff
@@ -785,6 +785,45 @@ function check_strict_kw(strict)
...
+it out automatically.
+"""
+function git_remote_head_branch(varname, root; remotename = "origin", fallback = "master")
+    env = copy(ENV)
+    env["GIT_TERMINAL_PROMPT"] = "0"
+    env["GIT_SSH_COMMAND"] = get(env, "GIT_SSH_COMMAND", "ssh -o \"BatchMode yes\"")
+    cmd = `git remote show $(remotename)`
+    stderr_output = IOBuffer()
+    git_remote_output = try
+        read(pipeline(setenv(cmd, env; dir=root); stderr=stderr_output), String)
```
mortenpi, line 800: @fredrikekre I removed the `ignorestatus` here, assuming that Git should return a non-zero status if `git remote show origin` is successful. Do you think this could break something?

fredrikekre: Probably fine since it is in `try-catch` since it will throw without `ignorestatus`, right?

mortenpi: I believe so. I assume that if it exits with a non-zero, there will not be anything meaningful to parse from the stdout anyway, so better to bail here, than in the parsing step.

fredrikekre:
I noticed in Literate that the call to `git remote show` here takes ~1s (https://github.com/fredrikekre/Literate.jl/commit/8054d2630b72bc190913b9b92cdb16e367b1ea51). It is not a huge problem here since you only call makedocs once, but a bit annoying when using LiveServer for example. (It wasn't a problem before since it was only used in deploydocs which doesn't run in that case).

mortenpi:
Should we also maybe do #1838 for this?

That said, something that I didn't think about before, I believe this could potentially block for up to 300s (TCP timeout) if you're having some weird networking problems.

PR #1831, opened 2022-05-30
Cache search objects in localStorage for faster search loading

IanButterworth:
I'm hoping to speed up the displaying of search results after a page reload. It's nice and fast when updating thee search box, but the building of the search dictionaries is really slow, especially on Base docs.

I see a few paths to do this:
1. Use a server-side search phrase -> results service - Can't be hosted by github pages etc. Won't work offline
2. Rework the docs pages to be within a long-lasting page and use frames to serve content so that the search dictionaries can be retained - Seems like a good idea but a lot of work. Will still be slow first load
3. Generate the dictionaries during mkdocs and deserialize them - Likely to be fast, but given users can provide custom search.js files (as Base does), it would likely be hard to make this work universally?
4. Store the `index` and `store` in browser local storage, which is per-origin, so that following page loads can use the cached versions

This PR is a demo of the last option.

Unfortunately, it's slower on first load and only slightly faster on following loads because of the need for compression due to the storage limit (on chrome at least) being 5MB, which the index of Base docs exceeds.

Perhaps the string size could be queried to see if compression is needed, but you'd need to know whether it was compressed at load time, so there would need to be an additional flag stored.

Anyway, I'm not sure this is a good idea, but I thought I'd share given it "works"

@mortenpi I recall we talked about some of this before

### DIFF
--- a/assets/html/search.js
+++ b/assets/html/search.js
@@ -1,5 +1,5 @@
-// libraries: jquery, lunr, lodash
-// arguments: $, lunr, _
+// libraries: jquery, lunr, lodash, lzstring
+// arguments: $, lunr, _, lzstring
 
 $(document).ready(function() {
   // parseUri 1.2.2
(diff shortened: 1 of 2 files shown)


IanButterworth:
Btw, one gripe I have is that I often instinctively press return after entering text in the search box, which, if you're on the search page already makes you needlessly hit the slow search load, given 1) the search results are likely already shown due to the text box update 2) the reload makes you hit the regeneration of the search dicts

Perhaps the return-key form entry could be disabled on the search page itself?

mortenpi:
I haven't looked at this PR in detail yet, but if we get some caching, that would be awesome.

> Perhaps the return-key form entry could be disabled on the search page itself?

Yes, I do that as well sometimes, so I think this would make a lot of sense. But separate PR might be good for this, as that's a no-brainer to merge.

assets/html/search.js
```diff
@@ -36,145 +36,168 @@ $(document).ready(function() {
   $("#search-form").submit(function(e) {
     e.preventDefault()
   })
+  var index = localStorage.getItem("index")
+  if (index === null) {
+    // If the index isn't retrievable, clear the store too just in case the regenerated index has changed
+    localStorage.clear()
```
fredrikekre, line 42: Isn't things like dark/light mode settings stored in here?

IanButterworth, line 42:
```suggestion
    localStorage.removeItem(store)
```

mortenpi (commented):
This seems to give about a 5x improvement (6s -> 1s for Julia manual, 500ms -> 100ms for Documenter's). I know we discussed generating the proper index on CI as well, but I think this here would also already be a pretty good improvement.

The only issues here that I see (other than the index being too big to store sometimes) is (1) that the cache is shared between different URLs and (2) it also doesn't get updated if the `search_index.js` file changes.

For (2) we could have some generated versioning string (could be the `sha1` of the index file?) in `siteinfo.js` that is checked, before the cache is used?

For (1) we could distinguish the storage based on the root URL of the page. But something I quickly tested is that it seems that if you're trying to `setItem` and the storage doesn't have enough space left, it fails (as opposed to e.g. removing old entries). This could be a problem if you browse the docs for multiple versions and then fill up your storage with the different versions. So ideally this would do some sort of memory management here. Or, alternatively, I guess we can just not support caching docs for different versions.

PR #1833, opened 2022-05-31
Disable enter key page refresh on search page

### LINKED ISSUE #1728: Feature request: Don't recompile search cache when [enter] is used in search field
If you search for something, then type in the term for another, the search is updated automatically, which is great.

However, if you naturally press return after the second search term, the page is refreshed and the search cache appears to be redone, introducing a delay.

It seems like [enter] could be turned into a no-op for any following searches?

IanButterworth:
As discussed https://github.com/JuliaDocs/Documenter.jl/pull/1831#issuecomment-1140581369

> One gripe I have is that I often instinctively press return after entering text in the search box, which, if you're on the search page already makes you needlessly hit the slow search load, given 1) the search results are likely already shown due to the text box update 2) the reload makes you hit the regeneration of the search dicts

> Perhaps the return-key form entry could be disabled on the search page itself?

I injected this js code locally and it worked, but I've not tried a full build. 
I guess there's a preview on the PR?

@mortenpi 

---

_Edit by @mortenpi: close #1728._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * ![Enhancement][badge-enhancement] The `ansicolor` keyword to `HTML()` now defaults to true, meaning that executed outputs from `@example`- and `@repl`-blocks are now by default colored (if they emit colored output). ([#1828][github-1828])
 
+## Version `v0.27.19`
+
+* ![Enhancement][badge-enhancement] On the HTML search page, pressing enter no longer causes the page to refresh (and therefore does not trigger the slow search index rebuild). ([#1728][github-1728], [#1833][github-1833])
+
 ## Version `v0.27.18`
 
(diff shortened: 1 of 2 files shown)


mortenpi (approved):
> I guess there's a preview on the PR?

It comes from an fork, so no. But I tested locally and it seems to work great! Thanks!

By the way, I think it also closes #1728.

IanButterworth:
Ha! Yes. Forgot I made that! 😁

IanButterworth:
Just tried this on my phone and it  meant the sidebar didn't hide on enter. Perhaps that needs a manual call now?

image

mortenpi:
Ah, good point. I agree that the sidebar should hide when you press "search". Would you have time to look at how to make that happen?

In fairness, I guess the whole UX here on mobile is a bit suboptimal (that the search box is in the sidebar), but that would be a much bigger change.

PR #1834, opened 2022-05-31
Hide sidebar on search

IanButterworth:
Hide sidebar on search, needed due to change in search behavior with #1833


mortenpi:
Awesome, LGTM. Thanks @IanButterworth!

PR #1836, opened 2022-06-02
Draft mode that skips (potentially) expensive parts of the build process

fredrikekre:
When passing `draft=true` to `makedocs` a draft version of the
documentation is built. Draft mode skips potentially expensive parts of
the building process and can be useful to get faster feedback when
writing documentation. Draft mode currently skips `@example`, `@repl`,
`@eval`, and `@setup` blocks. Draft mode can be disabled (or enabled) on
a per-page basis by setting `Draft = true` in an `@meta` block.

### DIFF
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -84,6 +84,7 @@ export Deps, makedocs, deploydocs, hide, doctest, DocMeta, asset,
         highlightsig = true,
         sitename = "",
         expandfirst = [],
+        draft = false,
     )
 
 Combines markdown files and inline docstrings into an interlinked document.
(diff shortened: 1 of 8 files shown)


mortenpi:
This is awesome!

Could we add a `test/examples` build that runs with `draft = true`, so that the code paths would get hit?

fredrikekre:
Added tests, and also disabled doctests in draft mode.

test/examples/make.jl
```diff
@@ -19,7 +19,7 @@ EXAMPLE_BUILDS = if haskey(ENV, "DOCUMENTER_TEST_EXAMPLES")
     split(ENV["DOCUMENTER_TEST_EXAMPLES"])
 else
     ["markdown", "html", "html-mathjax2-custom", "html-mathjax3", "html-mathjax3-custom",
-    "html-local"]
+    "html-local", "html-draft"]
```
fredrikekre, line 22: Do I need to update something more in order to have this run on CI?

src/Documenter.jl
```diff
@@ -188,6 +189,11 @@ is enabled by default.
 before the others. Documenter normally evaluates the files in the alphabetic order of their
 file paths relative to `src`, but `expandfirst` allows some pages to be prioritized.
 
+**`draft`** can be set to `true` to build a draft version of the document. In draft mode
+some potentially expensive steps are skipped (e.g. running `@example` blocks), which is
```
KristofferC, line 193:
```suggestion
some potentially time-consuming steps are skipped (e.g. running `@example` blocks), which is
```

is perhaps clearer?

PR #1838, opened 2022-06-03
Memoize result value in Utilities.getremote

fredrikekre:
Utilities.getremote is called when figuring out the URL to methods while
rendering the documentation. A profile of makedocs for Documenters docs
showed that almost a third of the collected samples was inside getremote,
and in particular in the spawned process to git. This patch memoizes the
result of getremote. This should give noticeable improvements, since
this function is often often called with the same argument.


mortenpi:
Looks reasonable to me, could just use a small CHANGELOG note.

KristofferC (approved):

PR #1840, opened 2022-06-03
Set version to 0.28.0-DEV.

fredrikekre:


mortenpi:
We can't really do this, since we have test dependencies on downstream packages. The DocumenterMarkdown one will be removed by #1826, but DocumenterTools will still be there.

PR #1841, opened 2022-06-05
Raise minimum required Julia version from 1.0 to 1.6

### LINKED ISSUE #1835: RFC: Drop support for Julia 1.0-1.5
So only support Julia 1.6+ from 0.28 onwards. Curious to know if there are any objections.

We don't really need <= 1.5 support for Julia docs themselves. For packages, you don't need to be able to run Documenter with every supported Julia version, and you can just build docs with Julia 1.6+. On the other hand, it would allow us to use JLLs etc. freely here (e.g. for Git, which is what prompted this).

If anyone really has a package that is only being run on Julia <= 1.5 (maybe privately), then they can still keep using the 0.27 branch.

fredrikekre:
fixes #1835


test/doctests/doctests.jl
```diff
@@ -228,19 +228,11 @@ rfile(filename) = joinpath(@__DIR__, "stdouts", filename)
...
-            @test success
-            @test is_same_as_file(output, rfile("41.stdout"))
-        end
-    else
-        # Old REPL scoping behaviour on older Julia version
-        run_makedocs(["hardscope.md"]) do result, success, backtrace, output
-            @test success
-            @test is_same_as_file(output, rfile("42.stdout"))
-        end
+    # Julia >1.5 REPL softscope,
```
KristofferC, line 231: Can remove?

mortenpi (approved):
LGTM, thanks @fredrikekre! I took the liberty of editing the CHANGELOG slightly.

PR #1849, opened 2022-06-16
Add folder selection for gh-pages

eirikbrandsaas:
Should probably make this clear so others dont make the mistake I did and assume it should be `docs/`.


fredrikekre:
Thanks

PR #1853, opened 2022-06-19
change CI badges from Travis/AV -> GHA

KristofferC:


mortenpi:
Good catch, thank you!

PR #1855, opened 2022-06-27
Use the Git JLL for git (via Git.jl; except on Macs)

mortenpi:
We still use the system `git` on Mac to work around the linking problems. Setting `GIT_TEMPLATE_DIR` (which Git.jl does) is necessary to make sure that our `git init` and `git clone` invocations do not pick up user-specific Git hooks etc (which can fail the tests).


KristofferC:
What's the size of this download?

mortenpi:
You mean the JLL artifacts? Is there a way to easily get that info? Git_jll appears the be on the order of 40-50MB (which is actually more than I thought it would be). But it also brings along a bunch of dependencies:

```
(gitjll) pkg> st -m
      Status `~/Julia/test/gitjll/Manifest.toml`
  [d7ba0133] Git v1.2.1
  [692b3bcd] JLLWrappers v1.4.1
  [21216c6a] Preferences v1.3.0
  [2e619515] Expat_jll v2.4.8+0
  [78b55507] Gettext_jll v0.21.0+0
  [f8c6e375] Git_jll v2.34.1+0
  [94ce4f54] Libiconv_jll v1.16.1+1
  [458c3c95] OpenSSL_jll v1.1.16+0
  [02c8fc9c] XML2_jll v2.9.14+0
```

Any particular circumstance where you'd be concerned about the size of the download?

fingolfin:
@mortenpi I believe this can be closed (you added `Git.jl` support in PR #2348)

Manifest.toml
asinghvi17, line 1: Looks like the Manifest got committed (by accident?)

mortenpi:
> you added `Git.jl` support in PR #2348

If at first you don't succeed, or something like that :smile: I probably forgot about this PR when I did the other one.

PR #1861, opened 2022-07-01
Fix missing doc check for methods from other packages

### LINKED ISSUE #1695: Docstring check confused with reexported function
At https://github.com/JuliaData/CategoricalArrays.jl/pull/367/checks?check_run_id=3575057492 I see this error:
```
┌ Error: 1 docstring not included in the manual:
│ 
│     Missings.levels :: Tuple{CategoricalArray}
│ 
│ 
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in @docs or @autodocs blocks.
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/ruzxx/src/DocChecks.jl:69
ERROR: LoadError: `makedocs` encountered an error. Terminating build
```
Yet the docstring is included in the manual, but it's `DataAPI.levels` (where the function is originally defined) rather than `Missings.levels`.

I could fix it by changing the definition of `levels` in Missings (see https://github.com/JuliaData/Missings.jl/pull/138) from
```julia
export levels
import DataAPI
const levels = DataAPI.levels
```
to:
```julia
export levels
using DataAPI: levels
```

That's not a big deal but I figured I would report it in case the fix would be easy.

### LINKED ISSUE #1857: Rebinding Module during DocChecks
Hi,

I ran into warnings while generating documentation for [FastLapackInterface](https://github.com/DynareJulia/FastLapackInterface.jl.git), due to the rebinding of certain functions from `LinearAlgebra.LAPACK`, to `FastLapackInterface` in this function:

https://github.com/JuliaDocs/Documenter.jl/blob/0d088e4b42a972bf710c158bed2d803500a6d31d/src/DocChecks.jl#L84

For example:
[LAPACK.getrf!](https://github.com/DynareJulia/FastLapackInterface.jl/blob/df66a83f61a7df457a8654e6d5a420a1e49cca49/src/lu.jl#L76) got rebound to `FastLapackInterface.getrf!`. 

This did not lead to any missing documentation, but it did cause warnings for all similar functions. I managed to fix it by explicitly importing them, but ideally I wouldn't have to.

Is this the desired behavior, if so, why?

Best,
Louis

mortenpi:
`allbindings` doesn't correctly construct the `Binding` keys when you have methods that extend functions from other packages. The constructed key is in the current module, rather than referring to the module where the function actually lives in. This works most of the time since usually the extended function will also be imported (e.g. with `using`) into the current module, but will fail if that is not the case (e.g. extending an unexported method by fully qualifying it with the module name).

In the same vein, whether a docstring is attached to an exported function or not is not always correctly determined if distinct functions with the same name are involved.

Fix #1857, fix #1695.

### DIFF
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -77,11 +77,12 @@ end
 
 function allbindings(checkdocs::Symbol, mod::Module, out = Dict{Utilities.Binding, Set{Type}}())
     for (obj, doc) in meta(mod)
-        isa(obj, IdDict{Any,Any}) && continue
+        isa(obj, IdDict{Any,Any}) && continue # is this ever true?!
+        @assert obj isa Docs.Binding
         name = nameof(obj)
         isexported = Base.isexported(mod, name)
         if checkdocs === :all || (isexported && checkdocs === :exports)
-            out[Utilities.Binding(mod, name)] = Set(sigs(doc))
+            out[obj] = Set(sigs(doc))
         end
     end
     out
--- a/src/DocChecks.jl
+++ b/src/DocChecks.jl
@@ -78,14 +78,19 @@ function allbindings(checkdocs::Symbol, mods)
 end
 
 function allbindings(checkdocs::Symbol, mod::Module, out = Dict{Utilities.Binding, Set{Type}}())
-    for (obj, doc) in meta(mod)
-        isa(obj, IdDict{Any,Any}) && continue # is this ever true?!
-        @assert obj isa Docs.Binding
-        name = nameof(obj)
-        # We only consider a name exported only if it is actually imported
-        isexported = (obj == Utilities.Binding(mod, name)) && Base.isexported(mod, name)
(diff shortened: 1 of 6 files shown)


src/DocChecks.jl
```diff
@@ -77,11 +79,13 @@ end
 
 function allbindings(checkdocs::Symbol, mod::Module, out = Dict{Utilities.Binding, Set{Type}}())
     for (obj, doc) in meta(mod)
-        isa(obj, IdDict{Any,Any}) && continue
+        isa(obj, IdDict{Any,Any}) && continue # is this ever true?!
+        @assert obj isa Docs.Binding
```
mortenpi, line 83: This bit actually confuses me.. are the keys to the `Docs.meta` dictionary ever not `Docs.Binding` objects? One option, just to be extra sure that it doesn't crash, could be to do `isa(obj, Docs.Binding) || continue` here.

PR #1869, opened 2022-07-07
Prevent `show` from mutating random state

LilithHafner:
To accommodate a [PR](https://github.com/JuliaLang/julia/pull/45222#discussion_r915524221) to base which utilizes global RNG in `show` in a hard to reproduce way.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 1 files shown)


fredrikekre:
Can you elaborate? This seem really weird to me and isn't obious from looking at the base PR.

fredrikekre:
Why doesn't sorting just use it's own rng?

LilithHafner:
https://github.com/JuliaLang/julia/pull/45222#discussion_r896907105

mortenpi:
I am wondering whether, even if there is a workaround for that PR, it might still be worth having this in Documenter, to make sure that doctests are more reproducible when the RNG is involved? Are there any cases where this might break any doctests? It would be nice to construct an for MWE though.

LilithHafner:
If `show` has consistent side effects, then repl doctests should call show after each evaluation just like the real REPL. This PR—eliminating RNG mutation side-effects—only makes sense if `show` has inconsistent RNG-mutation side effects while run in documenter. Here's a MWE for your consideration:

```julia
julia> using Random

julia> struct MyRange
           x::UnitRange{Int}
       end

julia> Base.show(io::IO, x::MyRange) = print(io, "A range containing ", unique(sort(rand(x.x, 10))), " and possibly other elements")

julia> Random.seed!(1729)
TaskLocalRNG()

julia> MyRange(1:10)
A range containing [2, 4, 5, 6, 8, 9, 10] and possibly other elements

julia> MyRange(1:10)
A range containing [1, 2, 3, 5, 6, 8, 9] and possibly other elements
```

The results I'm getting with this is that it is inconsistent in the REPL but consistent in `doctest(DocTestTest)`. I'm not sure why this is.

PR #1876, opened 2022-07-14
Use DocumenterTools as a dev dep in docs/

mortenpi:
This does a few things:

1. The source links to the DocumenterTools docstrings do not work right now because we Pkg.add it, but we need the source files to be located in a Git repository for gitremote to determine the remote repository. We can work around that by making it a Pkg.develop dependency instead.

2. If we add it as a dev dependency, we can use unreleased branches which declare support for newer Documenter versions, and so we can use the `-DEV` versions in the Documenter repository now. However, this requires checking out a particular [commit](https://github.com/JuliaDocs/DocumenterTools.jl/tree/documenter-0.28-dev) (or tag) of DocumenterTools. Pkg.develop can not be used for this since it does not know how to check out specific branches, and will fail as it tries to check out master (which will not be compatible with e.g. `0.28.0-DEV`).

3. Since adding it as a local dev dependency is a slightly non-trivial, this creates a separate `docs/instantiate.jl` script which does the necessary cloning and Pkg environment manipulations, just to make life easier locally and in the CI config.

Handling a minor release will still be a little tricky. But what I reckon we can do is: (1) prepare the `master` branch of DocumenterTools, which declares compat with Documenter 0.28, (2) update `docs/instantiate.jl` here to that branch and tag Documenter, and (3) tag the previously prepared commit of DocumenterTools. This way we never have to register a package that depends on a future release of Documenter.

Rather than depending on a commit from a side branch, I think it would be better to push custom tags to DocumenterTools for this. That would ensure that the commit will stay around, even if branches get deleted, and so the `git checkout` here should never fail.

This will also function as a test case for #1808, since this is an uncommon edge case that we actually handle (including docstrings from a third package in the manual with working source links).

### DIFF
--- a/docs/instantiate.jl
+++ b/docs/instantiate.jl
@@ -1,14 +1,22 @@
 # This script can be used to quickly instantiate the docs/Project.toml enviroment.
 using Pkg
-cd(@__DIR__) do
-    Pkg.activate(@__DIR__)
-    # Install the documenter branch of DocumenterTools
+documenter_directory = joinpath(@__DIR__, "..")
+project_directory = isempty(ARGS) ? @__DIR__() : joinpath(pwd(), ARGS[1])
+cd(project_directory) do
+    Pkg.activate(project_directory)
+    # Install a DocumenterTools version that declares compatibility with Documenter 0.28,
+    # but from an unreleased tag.
+    # https://github.com/JuliaDocs/DocumenterTools.jl/releases/tag/documenter-v0.1.14%2B0.28.0-DEV
     if isdir("dev/DocumenterTools")
-        @info "DocumenterTools already cloned to docs/dev/DocumenterTools"
+        @info "DocumenterTools already cloned to dev/DocumenterTools"
+        run(`git -C dev/DocumenterTools fetch origin`)
     else
         run(`git clone -n https://github.com/JuliaDocs/DocumenterTools.jl.git dev/DocumenterTools`)
     end
-    run(`git -C dev/DocumenterTools checkout --detach 336e27eeaf56852838b192b1fc7c0cce129d2d9f`)
-    Pkg.develop([PackageSpec(path = "dev/DocumenterTools"), PackageSpec(path = "..")])
+    run(`git -C dev/DocumenterTools checkout documenter-v0.1.14+0.28.0-DEV`)
+    Pkg.develop([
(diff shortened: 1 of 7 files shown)


docs/instantiate.jl
```diff
@@ -0,0 +1,15 @@
...
+    Pkg.activate(@__DIR__)
+    # Install the documenter branch of DocumenterTools
+    if isdir("dev/DocumenterTools")
+        @info "DocumenterTools already cloned to docs/dev/DocumenterTools"
+    else
+        run(`git clone -n https://github.com/JuliaDocs/DocumenterTools.jl.git dev/DocumenterTools`)
+    end
+    run(`git -C dev/DocumenterTools checkout --detach 336e27eeaf56852838b192b1fc7c0cce129d2d9f`)
+    Pkg.develop(path = "dev/DocumenterTools")
+    Pkg.develop(path = "..")
```
fredrikekre, line 13: Does this work?
```suggestion
    Pkg.develop([PackageSpec(path = "dev/DocumenterTools"), PackageSpec(path = "..")])
```

mortenpi: It does, and that's a good idea.

fredrikekre (approved):

docs/instantiate.jl
```diff
@@ -0,0 +1,22 @@
...
+    # Install a DocumenterTools version that declares compatibility with Documenter 0.28,
+    # but from an unreleased tag.
+    # https://github.com/JuliaDocs/DocumenterTools.jl/releases/tag/documenter-v0.1.14%2B0.28.0-DEV
+    if isdir("dev/DocumenterTools")
+        @info "DocumenterTools already cloned to dev/DocumenterTools"
+        run(`git -C dev/DocumenterTools fetch origin`)
+    else
+        run(`git clone -n https://github.com/JuliaDocs/DocumenterTools.jl.git dev/DocumenterTools`)
+    end
+    run(`git -C dev/DocumenterTools checkout documenter-v0.1.14+0.28.0-DEV`)
```
mortenpi, line 16: I also created the relevant tag in the DocumenterTools repo: https://github.com/JuliaDocs/DocumenterTools.jl/releases/tag/documenter-v0.1.14%2B0.28.0-DEV

PR #1880, opened 2022-07-19
add Woodpecker

uncomfyhalomacro:
Draft for now since I am going to add some documentation first :D

_Edit by @mortenpi: close #1878_


uncomfyhalomacro:
image
Currently works during my experiment on adding it

src/deployconfig.jl
```diff
@@ -21,11 +21,11 @@ Struct containing information about the decision to deploy or not deploy.
 - `subfolder::String` - The subfolder to which documentation should be pushed
 """
 Base.@kwdef struct DeployDecision
-    all_ok::Bool
-    branch::String = ""
-    is_preview::Bool = false
-    repo::String = ""
-    subfolder::String = ""
+  all_ok::Bool
```
fredrikekre, line 24: Please don't make whitespace changes.

uncomfyhalomacro:
I still have to figure out autodetection since `CI` is commonly used as an env var for all forges such as Github

uncomfyhalomacro:
@mortenpi @fredrikekre fixed a lot of typos now that i missed and i rebased it so and forced push with lease to make the history clean. Works fine now on my project that i tested this branch on.

fredrikekre (commented):
Please use 4 space indent. Can you also add a note about this in the CHANGELOG.md file?

src/deployconfig.jl
```diff
@@ -843,6 +843,167 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+      subfolder=subfolder)
+  else
+    return DeployDecision(; all_ok=false)
+  end
+end
+
+authentication_method(::Woodpecker) = Documenter.HTTPS
+function authenticated_repo_url(cfg::Woodpecker)
+  if !isnothing(get(ENV, "FORGE_URL", nothing))
+    return "https://$(ENV["CI_REPO_OWNER"]):$(ENV["DOCUMENTER_KEY"])@$(ENV["FORGE_URL"])/$(cfg.woodpecker_repo).git"
```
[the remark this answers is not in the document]
uncomfyhalomacro: Oh yeah. Thanks. I forgot. alright i will change that to something else ;) Would `PROJECT_ACCESS_TOKEN` good?

src/deployconfig.jl
```diff
@@ -855,7 +1016,19 @@ function auto_detect_deploy_system()
...
+      if ENV["CI"] in ["drone", "woodpecker"]
+        if ENV["CI"] == "drone"
+          @warn """Woodpecker is backward compatible to Drone 
+          but *there will be breaking changes in the future*"""
+        end
+        return Woodpecker()
+      else
+        @warn "May not be inside Woodpecker CI"
+        return nothing
+      end
```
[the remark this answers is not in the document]
uncomfyhalomacro: Will do!

uncomfyhalomacro:
This should be ready for another review now :smile:

fredrikekre:
Looks fine, can you add some tests to https://github.com/JuliaDocs/Documenter.jl/blob/master/test/deployconfig.jl?

src/deployconfig.jl
```diff
@@ -843,6 +843,201 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+            subfolder=subfolder
+        )
+    else
+        return DeployDecision(; all_ok=false)
+    end
+end
+
+authentication_method(::Woodpecker) = Documenter.HTTPS
+function authenticated_repo_url(cfg::Woodpecker)
+    if !isnothing(get(ENV, "FORGE_URL", nothing))
```
mortenpi, line 1026:
```suggestion
    if haskey(ENV, "FORGE_URL")
```

src/deployconfig.jl
```diff
@@ -843,6 +843,201 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+        return "https://$(ENV["CI_REPO_OWNER"]):$(ENV["PROJECT_ACCESS_TOKEN"])@$(ENV["FORGE_URL"])/$(cfg.woodpecker_repo).git"
+    else
+        # Regex will return the root of the URL e.g. https://github.com/JuliaDocs/Documenter.jl → github.com
+        # `CI_REPO_LINK` is an env var which returns a repository link e.g. https://github.com/JuliaDocs/Documenter.jl
+        m = match(r"https?:\/\/(?:.*\.)*(.+\..+?)\/", get(ENV, "CI_REPO_LINK", ""))
```
mortenpi, line 1031: The case of `m` being a `nothing` (if it doesn't match the regex, e.g. when it defaults to the empty string) should be handled here I think?

src/deployconfig.jl
```diff
@@ -843,6 +843,201 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+authentication_method(::Woodpecker) = Documenter.HTTPS
+function authenticated_repo_url(cfg::Woodpecker)
+    if !isnothing(get(ENV, "FORGE_URL", nothing))
+        return "https://$(ENV["CI_REPO_OWNER"]):$(ENV["PROJECT_ACCESS_TOKEN"])@$(ENV["FORGE_URL"])/$(cfg.woodpecker_repo).git"
+    else
+        # Regex will return the root of the URL e.g. https://github.com/JuliaDocs/Documenter.jl → github.com
+        # `CI_REPO_LINK` is an env var which returns a repository link e.g. https://github.com/JuliaDocs/Documenter.jl
+        m = match(r"https?:\/\/(?:.*\.)*(.+\..+?)\/", get(ENV, "CI_REPO_LINK", ""))      
+        FORGE_URL=m.captures[1]
+        if !isempty(FORGE_URL)
```
mortenpi, line 1033: You could require the first capture group to have non-zero length (with `+` instead of `*`), in which case you could remove this `if` (together with the `nothing` handling).

src/deployconfig.jl
```diff
@@ -843,6 +843,201 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+            return "https://$(ENV["CI_REPO_OWNER"]):$(ENV["PROJECT_ACCESS_TOKEN"])@$(FORGE_URL)/$(cfg.woodpecker_repo).git"
+        end
+    end
+  # Otherwise, raise an error
+    error("No repository URL or link found. Please check link or your `FORGE_URL` environment-variable again.")
```
mortenpi, line 1038: I don't think `authenticated_repo_url` should raise an error. If the deployment is broken, it should be caught by `deploy_folder`. I would suggest extracting the necessary variables in the `Woopecker` constructor, checking them in `deploy_folder`, and just constructing the string here.

src/deployconfig.jl
```diff
@@ -843,6 +843,201 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+    all_ok &= repo_ok
+    println(io, "- $(marker(repo_ok)) ENV[\"CI_REPO\"]=\"$(cfg.woodpecker_repo)\" occursin in re
+    po=\"$(repo)\"")
+
+    if build_type === :release
+        event_ok = in(cfg.woodpecker_event_name, ["push", "pull_request", "deployment", "tag"])
+        all_ok &= event_ok
+        println(io, "- $(marker(event_ok)) ENV[\"CI_BUILD_EVENT\"]=\"$(cfg.woodpecker_event_name)\" is \"push\", \"deployment\" or \"tag\"")
+        ## If a tag exist it should be a valid VersionNumber
+        m = match(r"^refs\/tags\/(.*)$", cfg.woodpecker_ref)
```
mortenpi, line 945: Should handle `m === nothing`. The GHA one does have the same issue, but it would be nice to crash with a more helpful error should `CI_COMMIT_REF` somehow contain nonsense.

src/deployconfig.jl
```diff
@@ -843,6 +843,201 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+        ## Deploy to deploydocs devurl kwarg
+        subfolder = devurl
+    else # build_type === :preview
+        m = match(r"refs\/pull\/(\d+)\/merge", cfg.woodpecker_ref)
+        pr_number = tryparse(Int, m === nothing ? "" : m.captures[1])
+        pr_ok = pr_number !== nothing
+        all_ok &= pr_ok
+        println(io, "- $(marker(pr_ok)) ENV[\"CI_COMMIT_REF\"] corresponds to a PR number")
+        if pr_ok
+            pr_origin_matches_repo = verify_github_pull_repository(cfg.woodpecker_repo, pr_number)
```
mortenpi, line 977: This calls out to GitHub API. Not sure that makes sense for Woodpecker?

mortenpi (commented):
Looks great for the most part, just some nitpicks.

* Do you have an example run on Codeberg/Woopecker that uses this branch?
* I don't think the PR is up to date with master? There should be some conflicts in the CHANGELOG.

uncomfyhalomacro:
> Looks great for the most part, just some nitpicks.
> 
>     * Do you have an example run on Codeberg/Woopecker that uses this branch?
> 
>     * I don't think the PR is up to date with master? There should be some conflicts in the CHANGELOG.



> 

https://codeberg.org/uncomfyhalomacro/SBOL3.jl/src/branch/main/.woodpecker/.doc-woodpecker-experiment.yml I experimented this on my soon-ish to start project of mine. Result will update my `pages` branch https://codeberg.org/uncomfyhalomacro/SBOL3.jl/src/branch/pages. -> the result would be this https://uncomfyhalomacro.codeberg.page/SBOL3.jl/dev/

I will rebase from upstream branch then :D and fix some conflicts

uncomfyhalomacro:
Just found an issue on the rendered pages where if I click the edit button, it gives the wrong link even with correct root URL. I have to fix and resolve results from the code review later.

mortenpi:
If the edit link behavior was something that _just_ appeared, then it might also be something from #1808.

uncomfyhalomacro:
> If the edit link behavior was something that _just_ appeared, then it might also be something from #1808.

nah i found out it was something else haha. anyway. it is ready for review now (again) :D

src/deployconfig.jl
uncomfyhalomacro: I wish there was a way to verify properly but it seems woodpecker does not have an env var that checks what hosting you use. Although that is fine. For now, I am just going to check if the pr number extracted from the commit ref from the built-in env var `CI_COMMIT_REF` is a valid number and also is the same as the pr number from the built-in env var `CI_COMMIT_PULL_REQUEST`

src/deployconfig.jl
uncomfyhalomacro: a `@warn` should be fine i guess

mortenpi: I don't think so. What would `authenticated_repo_url` return in that case? It should always return a sane value the `deplodocs` can use.

mortenpi:
> it is ready for review now (again) :D

Did you forget to push? I see you marked the comments resolved, but I believe the code hasn't changed?

uncomfyhalomacro:
> > it is ready for review now (again) :D
> 
> Did you forget to push? I see you marked the comments resolved, but I believe the code hasn't changed?

No, it has changed. I just need to push one more to resolve `authenticated_repo_url`

src/deployconfig.jl
```diff
@@ -843,6 +843,200 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+    if haskey(ENV, "FORGE_URL")
+        return "https://$(ENV["CI_REPO_OWNER"]):$(ENV["PROJECT_ACCESS_TOKEN"])@$(ENV["FORGE_URL"])/$(cfg.woodpecker_repo).git"
+    else
+        # Regex will return the root of the URL e.g. https://github.com/JuliaDocs/Documenter.jl → github.com
+        # `CI_REPO_LINK` is a built-in env var which returns a repository link e.g. https://github.com/JuliaDocs/Documenter.jl
+        # The variable is passed through the pipeline during the runs.
+        # See https://woodpecker-ci.org/docs/usage/environment to check the environment-variable 
+        m = match(r"https?:\/\/(?:.+\.)*(.+\..+?)\/", get(ENV, "CI_REPO_LINK", "")) 
+        forge_url = m !== nothing ? m.captures[1] : ""  # if `nothing`, just give an empty string so deploydocs can use it
+                                                        # even though we know that errors
```
mortenpi, line 1035: An empty string is not really a sane value here though. This failure, should it occur, should be caught in `deploy_folder` and `all_ok` set to `false` there. I think you could add a `woodpecker_forge_url` to the struct, move this whole logic into the constructor, and check for the empty string in `deploy_folder`.

src/deployconfig.jl
```diff
@@ -843,6 +843,200 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+"""
+    Woodpecker <: DeployConfig
+
+Implementation of `DeployConfig` for deploying from Woodpecker CI.
+
+The following environmental variables are built-in from the Woodpecker pipeline
+influences how `Documenter` works:
+ - `CI_REPO`: must match the full name of the repository <owner>/<name> e.g. `JuliaDocs/Documenter.jl`
+ - `CI_BUILD_EVENT`: must be set to `push`, `tag`, `pull_request`, and `deployment`
+ - `CI_COMMIT_REF`: must match the `devbranch` keyword to [`deploydocs`](@ref), alternatively correspond to a git tag.
```
mortenpi, line 859: `CI_COMMIT_PULL_REQUEST`, `CI_REPO_LINK` and `CI_REPO_OWNER` should also probably be here? And while you do mention `FORGE_URL` and `PROJECT_ACCESS_TOKEN` below, it might be good to have them in this list too, for completeness.

src/deployconfig.jl
```diff
@@ -843,6 +843,200 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+
+More about pipeline syntax is documented here: https://woodpecker-ci.org/docs/usage/pipeline-syntax
+
+Lastly, another environment-variable used for authentication is
+the `PROJECT_ACCESS_TOKEN` which is an access token you defined by 
+the forge you use e.g. GitHub, GitLab, Codeberg, and other gitea 
+instances. Check their documentation on how to create an access token. 
+This access token should be then added as a secret as documented in 
+https://woodpecker-ci.org/docs/usage/secrets. 
+
```
mortenpi, line 892:
```suggestion
```

src/deployconfig.jl
```diff
@@ -843,6 +843,200 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+```
+
+More about pipeline syntax is documented here: https://woodpecker-ci.org/docs/usage/pipeline-syntax
+
+Lastly, another environment-variable used for authentication is
+the `PROJECT_ACCESS_TOKEN` which is an access token you defined by 
+the forge you use e.g. GitHub, GitLab, Codeberg, and other gitea 
+instances. Check their documentation on how to create an access token. 
+This access token should be then added as a secret as documented in 
+https://woodpecker-ci.org/docs/usage/secrets.
```
mortenpi, line 891: The markdown parser doesn't pick up links automatically.

```suggestion
More about pipeline syntax is documented here: <https://woodpecker-ci.org/docs/usage/pipeline-syntax>

Lastly, another environment-variable used for authentication is
the `PROJECT_ACCESS_TOKEN` which is an access token you defined by 
the forge you use e.g. GitHub, GitLab, Codeberg, and other gitea 
instances. Check their documentation on how to create an access token. 
This access token should be then added as a secret as documented in 
<https://woodpecker-ci.org/docs/usage/secrets>.
```

src/deployconfig.jl
```diff
@@ -843,6 +843,200 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+    woodpecker_repo::String
+    woodpecker_event_name::String
+    woodpecker_ref::String
+end
+
+"""
+    Woodpecker()
+
+Initialize woodpecker environment-variables. Further info of
+environment-variables used are in https://woodpecker-ci.org/docs/usage/environment
```
mortenpi, line 904:
```suggestion
environment-variables used are in <https://woodpecker-ci.org/docs/usage/environment>
```

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
 
 * ![Bugfix][badge-bugfix] Documenter now generates the correct source URLs for docstrings from other packages when the `repo` argument to `makedocs` is set (note: the source links to such docstrings only work if the external package is cloned from GitHub and added as a dev-dependency). However, this change **breaks** the case where the `repo` argument is used to override the main package/repository URL, assuming the repository is cloned from GitHub. ([#1808][github-1808])
 
+* ![Enhancement][badge-enhancement] Added Woodpecker CI as another CI solution for deploying docs. ([#1880][github-1880])
+
```
mortenpi, line 21: I created some conflicts again by merging a few PRs. Could you also move this up the list a little, so that it would be the last of the enhancements?

src/deployconfig.jl
```diff
@@ -843,6 +843,200 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+        deploy_branch = branch_previews
+        deploy_repo = repo_previews
+        is_preview = true
+        ## deploydocs to previews/PR
+        subfolder = "previews/PR$(something(pr_number1, 0))"
+    end
+
+    owner_ok = env_nonempty("CI_REPO_OWNER")
+    token_ok = env_nonempty("PROJECT_ACCESS_TOKEN")
+    auth_ok = owner_ok | token_ok
```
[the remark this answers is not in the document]
uncomfyhalomacro: Removed `owner_ok`. Going to stick with the `PROJECT_ACCESS_TOKEN`. This is a user-defined secret using the value from a generated access token from the user's forge of choice. This is to check if it exists. I borrowed some logic from `GitHubActions` `deployfolder` where it checks if `GITHUB_TOKEN` exists.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 * ![Bugfix][badge-bugfix] Documenter now generates the correct source URLs for docstrings from other packages when the `repo` argument to `makedocs` is set (note: the source links to such docstrings only work if the external package is cloned from GitHub and added as a dev-dependency). However, this change **breaks** the case where the `repo` argument is used to override the main package/repository URL, assuming the repository is cloned from GitHub. ([#1808][github-1808])
 * ![Bugfix][badge-bugfix] Documenter no longer uses the `TRAVIS_REPO_SLUG` environment variable to determine the Git remote of non-main repositories (when inferring it from the Git repository configuration has failed), which could previously lead to bad source links. ([#1881][github-1881])
 
+* ![Enhancement][badge-enhancement] Added Woodpecker CI as another CI solution for deploying docs. ([#1880][github-1880])
+
```
mortenpi, line 23:
```suggestion
* ![Enhancement][badge-enhancement] Woodpecker CI is now automatically supported for documentation deployment. ([#1880][github-1880])
* ![Bugfix][badge-bugfix] Documenter now generates the correct source URLs for docstrings from other packages when the `repo` argument to `makedocs` is set (note: the source links to such docstrings only work if the external package is cloned from GitHub and added as a dev-dependency). However, this change **breaks** the case where the `repo` argument is used to override the main package/repository URL, assuming the repository is cloned from GitHub. ([#1808][github-1808])
* ![Bugfix][badge-bugfix] Documenter no longer uses the `TRAVIS_REPO_SLUG` environment variable to determine the Git remote of non-main repositories (when inferring it from the Git repository configuration has failed), which could previously lead to bad source links. ([#1881][github-1881])

```

The suggestion is messy, but it should just slightly rephrase the new entry and move it up a bit.

src/deployconfig.jl
```diff
@@ -843,6 +843,216 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+    #
+    repo_link_ok = !isempty(cfg.woodpecker_repo_link)  # if repo link is an empty string then it is not valid
+    all_ok &= repo_link_ok
+    forge_url_ok = !isempty(cfg.woodpecker_forge_url)  # if the forge url is an empty string, it is not a valid url
+    all_ok &= forge_url_ok
+
+    repo_ok = occursin(cfg.woodpecker_repo, repo)
+    all_ok &= repo_ok
+    println(io, "- $(marker(repo_ok)) ENV[\"CI_REPO\"]=\"$(cfg.woodpecker_repo)\" occursin in re
+    po=\"$(repo)\"")
```
mortenpi, line 960:
```suggestion
    println(io, "- $(marker(repo_ok)) ENV[\"CI_REPO\"]=\"$(cfg.woodpecker_repo)\" occursin in repo=\"$(repo)\"")
```

mortenpi (approved):
Cool! I think this LGTM now, other than the two minor things.

Do you links to a run on Codeberg with the latest version of this here by any chance, including the CI logs?

uncomfyhalomacro:
> Cool! I think this LGTM now, other than the two minor things.
> 
> 
> 
> Do you links to a run on Codeberg with the latest version of this here by any chance, including the CI logs?

https://ci.codeberg.org/uncomfyhalomacro/SBOL3.jl/build/12/1 here 😄

mortenpi:
Hmm. In the configuration, unless I am mistaken, but:

* `julia -e 'using Pkg; Pkg.develop(url="https://codeberg.org/uncomfyhalomacro/SBOL3.jl");'` will just add SBOL3 into the global environment, which seems unnecessary?
* The `rm docs/Project.toml` seems unnecessary -- you can just modify the existing environment. In fact, usually you'd want to just add current package as dev dep and then `Pkg.instantiate` it, especially if you have a bunch of other packages there.
* `Pkg.add(url="https://codeberg.org/uncomfyhalomacro/SBOL3.jl")` will make the documentation build use your `main` version no matter what. You should do `Pkg.develop(path=pwd())` instead, to make sure you're using SBOL3 from the checked out repository.

CHANGELOG.md
```diff
@@ -16,9 +16,9 @@
...
 
+* ![Enhancement][badge-enhancement] Woodpecker CI is now automatically supported for documentation deployment. ([#1880][github-1880])
 * ![Bugfix][badge-bugfix] Documenter now generates the correct source URLs for docstrings from other packages when the `repo` argument to `makedocs` is set (note: the source links to such docstrings only work if the external package is cloned from GitHub and added as a dev-dependency). However, this change **breaks** the case where the `repo` argument is used to override the main package/repository URL, assuming the repository is cloned from GitHub. ([#1808][github-1808])
 * ![Bugfix][badge-bugfix] Documenter no longer uses the `TRAVIS_REPO_SLUG` environment variable to determine the Git remote of non-main repositories (when inferring it from the Git repository configuration has failed), which could previously lead to bad source links. ([#1881][github-1881])
```
mortenpi, line 21: Whoops, this was my bad.

```suggestion
* ![Bugfix][badge-bugfix] Documenter no longer uses the `TRAVIS_REPO_SLUG` environment variable to determine the Git remote of non-main repositories (when inferring it from the Git repository configuration has failed), which could previously lead to bad source links. ([#1881][github-1881])

```

uncomfyhalomacro:
> Hmm. In the configuration, unless I am mistaken, but:
> 
>     * `julia -e 'using Pkg; Pkg.develop(url="https://codeberg.org/uncomfyhalomacro/SBOL3.jl");'` will just add SBOL3 into the global environment, which seems unnecessary?
> 
>     * The `rm docs/Project.toml` seems unnecessary -- you can just modify the existing environment. In fact, usually you'd want to just add current package as dev dep and then `Pkg.instantiate` it, especially if you have a bunch of other packages there.
> 
>     * `Pkg.add(url="https://codeberg.org/uncomfyhalomacro/SBOL3.jl")` will make the documentation build use your `main` version no matter what. You should do `Pkg.develop(path=pwd())` instead, to make sure you're using SBOL3 from the checked out repository.

For the first one. Yeah I can remove it because that is just an experimentation for finding some issues from Pkg. It was because I have many errors caused by `Pkg.jl`. This is because the package is not yet registered so even with `julia --project=docs/ using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()` that will error. So that is my workaround for now

uncomfyhalomacro:
It is a separate issue. Maybe I should have open an issue in Pkg.jl hmm

uncomfyhalomacro:
> > Logs for why i use `Pkg.add` instead of the usual `Pkg.develop(PackageSpec(path=pwd()))`: https://ci.codeberg.org/uncomfyhalomacro/SBOL3.jl/build/2/9
> 
> If you put the `add` after the `develop` it should be fine I believe. The standard setup definitely works for unregistered packages.

Going to try that

uncomfyhalomacro:
@mortenpi well that worked. thanks! https://ci.codeberg.org/SynBioJulia/SBOL3.jl/build/1/3 anyway just ignore the part where Documenter fails. Forgot to update my repository link. I moved my project to an org.

docs/src/man/hosting.md
```diff
@@ -417,6 +417,37 @@ jobs:
 
 _This workflow was taken from [CliMA/ClimaTimeSteppers.jl](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/0660ace688b4f4b8a86d3c459ab62ccf01d7ef31/.github/workflows/DocCleanup.yml) (Apache License 2.0)._
 
+## Woodpecker CI
+
+To run a documentation build from Woodpecker CI, one should create an access token
+from their forge of choice: GitHub, GitLab, or Codeberg (or any Gitea instance). 
+This access token should be added to Woodpecker CI as a secret named as
+`project_access_token` - all lowercase. Next, create a new pipeline configuration file
```
fredrikekre, line 425: Is this a convention or requirement in Woodpecker? You use `PROJECT_ACCESS_TOKEN` later, so is Woodpecker capitalizing the name? In that case, does it matter?

src/deployconfig.jl
```diff
@@ -843,6 +843,215 @@ authentication_method(::Buildkite) = Documenter.SSH
...
+end
+
+authentication_method(::Woodpecker) = Documenter.HTTPS
+function authenticated_repo_url(cfg::Woodpecker)
+    # `cfg.woodpecker_forge_url` should be just the root of the URL e.g. github.com, gitlab.com, codeberg.org
+    # otherwise, it will be equal to `cfg.woodpecker_repo_link`
+    # If so, we just split the `http(s)://` from the string we want 
+    # e.g. `https://github.com/JuliaDocs/Documenter.jl` to `github.com/JuliaDocs/Documenter.jl`.
+    if haskey(ENV, "FORGE_URL")
+        return "https://$(ENV["CI_REPO_OWNER"]):$(ENV["PROJECT_ACCESS_TOKEN"])@$(ENV["FORGE_URL"])/$(cfg.woodpecker_repo).git"
```
fredrikekre, line 1045: I wonder if Documenter might leak this URL to the logs. For GitHub actions, which also uses an authenticated URL like this it isn't much of a problem since there you use a temporary token.

fredrikekre (approved):

docs/src/man/hosting.md
uncomfyhalomacro: Resolved in https://github.com/JuliaDocs/Documenter.jl/pull/1880/commits/f94f3a552b6c879c92a2163e468b29e1989e09c9

src/deployconfig.jl
uncomfyhalomacro: The token will be shown as `****` in the logs so the token isn't leaked. I am not sure what you mean leaking of the URL. you mean that Documenter may show `https://JuliaDocs:myprojectaccesstoken@github.com/JuliaDocs/Documenter.jl`? It will only show as `https://****:****@github.com/****/Documenter.jl`.

uncomfyhalomacro: Although, unless they `echo`-ed their secrets to a file, then I'm going to question their life choices

uncomfyhalomacro: You can see a successful log here https://ci.codeberg.org/SynBioJulia/SBOL3.jl/build/6/3. There is nothing much you can't see, even the url passed from the `authenticated_repo_url` function is not showing.

docs/src/man/hosting.md
mortenpi: If it gets passed uppercase anyway, maybe better to recommend setting declaring it uppercase as well, just to avoid this confusion?

src/deployconfig.jl
mortenpi: I wonder if it would be possible to use [`git-credential-store`](https://git-scm.com/docs/git-credential-store) for this? I don't have experience with it, but it looks like it would then be similar to how the handle the SSH keys.

uncomfyhalomacro: I don't think having to use the git credential store is necessary. git credential store just does what with the URL we have to authenticate stored as a text file. We are just adding an extra step that I don't think is needed.

uncomfyhalomacro: Although I have an idea. We can use https://github.com/str4d/rage or https://github.com/FiloSottile/age for this if you want to push with the git credential store. it is an extra step. you juts need to encrypt the credential store file with the corresponding URLs, add it to your project repo and then deencrypt it inside the container. It is many steps but I think that is what you want to avoid leaking of information?

uncomfyhalomacro: Anyway seems we can use SSH. I guess I am just overthinking a lot. See https://github.com/JuliaDocs/Documenter.jl/pull/1880/commits/c7c5bf1d0ba549a4d05ab32b77465daa568a2bba and https://ci.codeberg.org/SynBioJulia/SBOL3.jl/build/12 that now uses `DOCUMENTER_KEY`. I will add the documentation for creating the keys.

uncomfyhalomacro:
One more documentation on how to generate SSH keys and converting the private key to a base64 key and then another review so this could be merged

uncomfyhalomacro:
@fredrikekre @mortenpi This is ready to be merged. You can review again I guess.

fredrikekre (approved):

src/deployconfig.jl
mortenpi: Hmm the logs for that build say

```
│ - ✔ ENV["PROJECT_ACCESS_TOKEN"] exists and is non-empty
```

? Did it actually use the SSH key?

mortenpi: Also, just for future reference, here is where Documenter would happily leak the secret if it were to run into an error

https://github.com/JuliaDocs/Documenter.jl/blob/396ee24391c081792e2f3cc32bb47699d7ddc3dd/src/Documenter.jl#L631-L642

While the censoring by the CI would generally catch this, I think ideally we should avoid this altogether. But that can be a subject of a follow-up PR (#1889). Just for the record though, I assume we'd have Documenter write the credentials file on every `deplodocs` run, just like we do with the SSH key.

mortenpi (approved):
@uncomfyhalomacro Would you be okay using the master branch of Documenter for a while? It's going to be a while before we tag 0.28.0, but I am not sure I want to go through the hassle of backporting this unless it's really necessary.

uncomfyhalomacro:
It should be fine! No worries 😄

mortenpi:
Cool. And thanks a lot for iterating on the PR!

Issue #245, opened 2016-08-28
Use UnitRange for Depth in @contents block


mortenpi:
From #226:

> BTW Is there any way to only show the contents from depth 2 for a page, i.e. excluding the page title as redundant?

and since [we decided that it's probably a good idea](https://github.com/JuliaDocs/Documenter.jl/issues/226#issuecomment-241528818):

> Yes, a minimum depth sounds reasonable. Perhaps using a `UnitRange` instead of an `Int` for the `Depth` value. So something like
> 
> ``` @contents
>    Pages = [
>        "algorithms.md",
>        "graph-types.md",
>        "graphs-builtin.md",
>        "interface.md",
>    ]
>    Depth = 2:4
> ```
> 
> to display only levels `2` to `4`, while internally `Depth = N` is rewritten as `Depth = 1:N`.

Opening this so that this thought wouldn't get lost.

PR #1890, opened 2022-07-27
Support UnitRange for Depth configuration in `at-contents` block

fredrikekre:
This patch adds support for `MinDepth` in `at-contents` for specifying
the minimum header level depth that should be shown, similar to `Depth`.
This is useful mostly for single-page content blocks, since you can then
filter out the superfluous level 1 header.

_Edit by @mortenpi: close #245_

### DIFF
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -322,7 +322,7 @@ Depth = 5
 ````
 
 As with `@index` if `Pages` is not provided then all pages are included. The default
-`Depth` value is `2`, i.e. header levels 1 and 2 are included. `Depth` also accept
+`Depth` value is `2`, i.e. header levels 1 and 2 are included. `Depth` also accepts
 `UnitRange`s, to make it possible to configure also the minimum header level to be shown.
 `Depth = 2:3` can be used to include only headers with levels 2-3, for example.
 
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -5,6 +5,7 @@ For supported Markdown syntax, see the [documentation for the Markdown standard 
 
 ```@contents
 Pages = ["syntax.md"]
+Depth = 2:2
 ```
 
 ## `@docs` block
@@ -321,7 +322,9 @@ Depth = 5
 ````
 
 As with `@index` if `Pages` is not provided then all pages are included. The default
-`Depth` value is `2`.
+`Depth` value is `2`, i.e. header levels 1 and 2 are included. `Depth` also accept
+`UnitRange`s, to make it possible to configure also the minimum header level to be shown.
+`Depth = 2:3` can be used to include only headers with levels 2-3, for example.
 
 ## `@example` block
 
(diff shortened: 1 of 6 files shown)


docs/src/man/syntax.md
```diff
@@ -5,6 +5,7 @@ For supported Markdown syntax, see the [documentation for the Markdown standard
 
 ```@contents
 Pages = ["syntax.md"]
+MinDepth = 2
```
fredrikekre, line 8: Maybe it even makes sense to default `MinDepth = 2` when `lenght(Pages) == 1`?

mortenpi: I am a little unsure about having one argument's default depending on the value of the other, since it could be surprising that changing `Pages` would change the depth behavior. On the other hand, I do see the utility. It would also remove any `h1`s further down the page though I believe?

Maybe a note in the documentation would be enough to push users to set `MinDepth = 2`?

mortenpi (commented):
Not opposed to a new option either, but what do you think about allowing `UnitRange`s to `Depth` instead for this? Refs: #245 and #1125.

fredrikekre:
> Not opposed to a new option either, but what do you think about allowing UnitRanges to Depth instead for this? 

Ah, forgot about that. Yea that should be fine (although you will have to look up the default max depth even when you only want to change the default min).

docs/src/man/syntax.md
fredrikekre: Yea, I was thinking that if you only have one entry in `Pages` your probably have it at the top of that same page and thus the H1 header is not needed. And multiple H1 headers on the same page is not so common. If you don't want this behavior you can of course override it. But perhaps we can do this by default at a later stage if it seems useful.

src/Documents.jl
```diff
@@ -113,19 +113,23 @@ end
...
+            Depth  = 1:2,
             build  = error("missing value for `build` in `ContentsNode`."),
             source = error("missing value for `source` in `ContentsNode`."),
             others...
         )
-        new(Pages, Depth, build, source, [])
+        if Depth isa Integer
+            Depth = 1:Depth
+        end
+        new(Pages, Depth.start, Depth.stop, build, source, [])
```
mortenpi, line 132: I don't think the `.start` and `.stop` fields are technically documented?

```suggestion
        new(Pages, first(Depth), last(Depth), build, source, [])
```

mortenpi (approved):
One minor comment, but LGTM!

docs/src/man/syntax.md
```diff
@@ -321,7 +322,9 @@ Depth = 5
 ````
 
 As with `@index` if `Pages` is not provided then all pages are included. The default
-`Depth` value is `2`.
+`Depth` value is `2`, i.e. header levels 1 and 2 are included. `Depth` also accept
```
mortenpi, line 325:
```suggestion
`Depth` value is `2`, i.e. header levels 1 and 2 are included. `Depth` also accepts
```

fredrikekre:
cc @NHDaly (since you started on this feature in #1125 you might be interested)

PR #1903, opened 2022-08-15
Add aria-label to copy code button for accessibility

pfitzseb:


fredrikekre (approved):

mortenpi:
LGTM, but I wonder if it should also set a `title` attribute, for mouse-over tooltip?

mortenpi:
Thanks @pfitzseb!

PR #1906, opened 2022-08-18
Normalize line endings in MD source before parsing

mortenpi:
Line endings in Markdown source files are now normalized to `LF` before parsing, to work around [a bug in the Julia Markdown parser][julia-29344] where parsing is sensitive to line endings, and can therefore cause platform-dependent behavior.

I wonder if this might cause problems with doctest fixing though..

[julia-29344]: https://github.com/JuliaLang/julia/issues/29344


mortenpi:
@fredrikekre Do you happen to have any thoughts on this, especially regarding the doctest fixing?

PR #1908, opened 2022-08-19
Various bugfixes and enhancements to LaTeXWriter

mortenpi:
I think `-interaction=batchmode` is actually more appropriate than `nonstopmode`, since the latter includes "diagnostic messages" that are basically just noise. `-halt-on-error` should help in the cases where LaTeX doesn't clearly state that something was an error (like when you have missing fonts).

The full logs for the recent Julia 1.8.0 failure should just be

```
Latexmk: This is Latexmk, John Collins, 1 January 2015, version: 4.41.
Rule 'pdflatex': Rules & subrules not known to be previously run:
   pdflatex
Rule 'pdflatex': The following rules & subrules became out-of-date:
      'pdflatex'
------------
Run number 1 of rule 'pdflatex'
------------
Latexmk: applying rule 'pdflatex'...
------------
Running 'lualatex  -interaction=batchmode -halt-on-error -shell-escape -recorder  "TheJuliaLanguage.tex"'
------------
This is LuaTeX, Version 1.0.4 (TeX Live 2017/Debian)
 system commands enabled.
tput: No value for $TERM and no -T specified

luaotfload | main : initialization completed in 0.091 seconds
luaotfload | db : Reload initiated (formats: otf,ttf,ttc); reason: "Font phvb8t not found.".
luaotfload | resolve : sequence of 3 lookups yielded nothing appropriate.Failure to make 'TheJuliaLanguage.pdf'
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 256
Latexmk: Errors, in force_mode: so I tried finishing targets
```

which is far better than the 2MB of output we're getting now. And it fails only after a few seconds, rather than running for several minutes.

- https://tex.stackexchange.com/questions/258814/what-is-the-difference-between-interaction-nonstopmode-and-halt-on-error
- http://latexref.xyz/Command-line-options.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
 * ![Bugfix][badge-bugfix] Documenter no longer uses the `TRAVIS_REPO_SLUG` environment variable to determine the Git remote of non-main repositories (when inferring it from the Git repository configuration has failed), which could previously lead to bad source links. ([#1881][github-1881])
 * ![Bugfix][badge-bugfix] Line endings in Markdown source files are now normalized to `LF` before parsing, to work around [a bug in the Julia Markdown parser][julia-29344] where parsing is sensitive to line endings, and can therefore cause platform-dependent behavior. ([#1906][github-1906])
 
+## Version `v0.27.23`
+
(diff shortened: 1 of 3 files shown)


mortenpi:
It also appears that we do not handle `Markdown.LineBreak`s at all in LaTeXWriter, leading a `MethodError`.

PR #1909, opened 2022-08-20
attempt to fix broken PDF links

matthias314:
This is supposed to address several issues regarding the PDF version of the Julia documentation, namely JuliaLang/julia#38054 and probably also the second comment in JuliaLang/julia#43652. This PR certainly needs improvements, but hopefully it gets the ball rolling. I find it annoying that the Julia documentation has so many broken links, and with this PR they all seem to disappear.

As far as I can tell, the main problem is that the field `anchor.nth` is used when creating a hypertarget in https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/Writers/LaTeXWriter.jl#L275-L279 but it is not used by hyperlinks. I don't know if this problem only happens for `anchor.nth == 1` or in general. In the case of the Julia documentation, the value seems to be always `1`, so I have omitted this field completely. I have noticed that in `HTMLWriter.jl` the case `anchor.nth == 1` receives special treatment.

A secondary problem is that the hypertarget is inserted before the text it refers to. Many targets are chapter or section headings, so there is a good chance that there will be a page break between the target and the text. On the other hand, if the target is moved behind the text, then the text is not visible when users follow the link. I've addressed this by using the LaTeX macro `\label` to create the target. Such labels then need to be referenced by `\hyperref` instead of `\hyperlink`. Since I don't know how to tell from within Julia which kind of target a link refers to, I have added a macro `\hyperlinkref` to `documenter.sty` which inserts either `\hyperref` or `\hyperlink`, depending on whether a corresponding label exists or not.

Other changes: In https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/Writers/LaTeXWriter.jl#L290-L294 a link is added to a target at the same spot. I have removed the link because it's unclear to me what the purpose of a link is that links to itself. (Theoretically, there could again be a page break between the target and the link.)

For the Julia documentation, the methods `latex(io::IO, index::Documents.IndexNode, page, doc)` and `latex(io::IO, contents::Documents.ContentsNode, page, doc)` are apparently never called, so changes in this code haven't been tested. In the latter function I have kept the `anchor.nth` when generating the link id, and I've moved the assignment `id =` to where it belongs.

I've also noticed that you use hashes of the anchor ids to get the target labels. Would it be a problem to simply use the ids themselves? For example. can they contain any characters that cause problems?

### DIFF
--- a/assets/latex/documenter.sty
+++ b/assets/latex/documenter.sty
@@ -85,3 +85,9 @@
 \usepackage{graphicx}
 \usepackage[export]{adjustbox}
 %
+
+% Some internal link targets are implemented with \label, some with \hypertarget,
+% but they require different links. This inserts a \hyperref if a corresponding label exists,
+% and \hyperlink if it doesn't.
+\def\hyperlinkref#1#2{\@ifundefined{r@#1}{\hyperlink{#1}{#2}}{\hyperref[#1]{#2}}}
+%
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -85,6 +85,10 @@ Context(io) = Context{typeof(io)}(io, false, Dict(), 1, "")
 _print(c::Context, args...) = Base.print(c.io, args...)
 _println(c::Context, args...) = Base.println(c.io, args...)
 
+# Labels in the TeX file are hashes of plain text labels.
+# To keep the plain text label (for debugging), say _hash(x) = x
+_hash(x) = string(hash(x))
+
 
 const STYLE = joinpath(dirname(@__FILE__), "..", "..", "assets", "latex", "documenter.sty")
 const DEFAULT_PREAMBLE_PATH = joinpath(dirname(@__FILE__), "..", "..", "assets", "latex", "preamble.tex")
@@ -273,9 +277,9 @@ function latex(io::IO, vec::Vector, page, doc)
 end
 
 function latex(io::IO, anchor::Anchors.Anchor, page, doc)
-    id = string(hash(string(anchor.id, "-", anchor.nth)))
-    _println(io, "\n\\hypertarget{", id, "}{}\n")
+    id = _hash(Anchors.label(anchor))
     latex(io, anchor.object, page, doc)
+    _println(io, "\n\\label{", id, "}{}\n")
 end
 
 
@@ -288,10 +292,9 @@ function latex(io::IO, node::Documents.DocsNodes, page, doc)
 end
 
 function latex(io::IO, node::Documents.DocsNode, page, doc)
-    id = string(hash(string(node.anchor.id)))
+    id = _hash(Anchors.label(node.anchor))
     # Docstring header based on the name of the binding and it's category.
-    _println(io, "\\hypertarget{", id, "}{} ")
-    _print(io, "\\hyperlink{", id, "}{\\texttt{")
+    _print(io, "\\hypertarget{", id, "}{\\texttt{")
     latexesc(io, string(node.object.binding))
     _print(io, "}} ")
     _println(io, " -- {", Utilities.doccat(node.object), ".}\n")
@@ -329,9 +332,9 @@ function latex(io::IO, index::Documents.IndexNode, page, doc)
 
     _println(io, "\\begin{itemize}")
     for (object, _, page, mod, cat) in index.elements
-        id = string(hash(string(Utilities.slugify(object))))
+        id = _hash(string(Utilities.slugify(object)))
         text = string(object.binding)
-        _print(io, "\\item \\hyperlink{")
+        _print(io, "\\item \\hyperlinkref{")
         _print(io, id, "}{\\texttt{")
         latexesc(io, text)
         _println(io, "}}")
@@ -352,7 +355,6 @@ function latex(io::IO, contents::Documents.ContentsNode, page, doc)
         # Filter out header levels smaller than the requested mindepth
         level = level - contents.mindepth + 1
         level < 1 && continue
-        id = string(hash(string(anchor.id, "-", anchor.nth)))
         # If we're changing depth, we need to make sure we always print the
         # correct number of \begin{itemize} and \end{itemize} statements.
         if level > depth
@@ -370,7 +372,8 @@ function latex(io::IO, contents::Documents.ContentsNode, page, doc)
             end
         end
         # Print the corresponding \item statement
-        _print(io, "\\item \\hyperlink{", id, "}{")
+        id = _hash(Anchors.label(anchor))
+        _print(io, "\\item \\hyperlinkref{", id, "}{")
         latexinline(io, header.text)
         _println(io, "}")
     end
@@ -676,8 +679,8 @@ function latexinline(io::IO, md::Markdown.Link)
     else
         if occursin(".md#", md.url)
             file, target = split(md.url, ".md#"; limit = 2)
-            id = string(hash(target))
-            wrapinline(io, "hyperlink") do
+            id = _hash(target)
+            wrapinline(io, "hyperlinkref") do
                 _print(io, id)
             end
         else
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,12 @@
 * ![Bugfix][badge-bugfix] Documenter now generates the correct source URLs for docstrings from other packages when the `repo` argument to `makedocs` is set (note: the source links to such docstrings only work if the external package is cloned from GitHub and added as a dev-dependency). However, this change **breaks** the case where the `repo` argument is used to override the main package/repository URL, assuming the repository is cloned from GitHub. ([#1808][github-1808])
 * ![Bugfix][badge-bugfix] Documenter no longer uses the `TRAVIS_REPO_SLUG` environment variable to determine the Git remote of non-main repositories (when inferring it from the Git repository configuration has failed), which could previously lead to bad source links. ([#1881][github-1881])
 * ![Bugfix][badge-bugfix] Line endings in Markdown source files are now normalized to `LF` before parsing, to work around [a bug in the Julia Markdown parser][julia-29344] where parsing is sensitive to line endings, and can therefore cause platform-dependent behavior. ([#1906][github-1906])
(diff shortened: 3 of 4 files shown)


src/Writers/LaTeXWriter.jl
```diff
@@ -273,9 +273,11 @@ function latex(io::IO, vec::Vector, page, doc)
 end
 
 function latex(io::IO, anchor::Anchors.Anchor, page, doc)
-    id = string(hash(string(anchor.id, "-", anchor.nth)))
-    _println(io, "\n\\hypertarget{", id, "}{}\n")
+    # TODO: currently ignoring anchor.nth
+    # id = string(hash(string(anchor.id, "-", anchor.nth)))
+    id = string(hash(anchor.id))
```
mortenpi, line 278: I believe we need to keep using `.nth` here. The [CI artifacts here](https://github.com/JuliaDocs/Documenter.jl/actions/runs/2895291989) contain the PDFs from the `test/examples` builds that do use at-index and at-contents blocks, and it looks like at-contents links are currently still broken, which I think is due to this.

mortenpi:
Thank you for the PR! I'm planning a backport release with some TeX fixes, and this fits in there perfectly.

> I don't know if this problem only happens for anchor.nth == 1 or in general. In the case of the Julia documentation, the value seems to be always 1, so I have omitted this field completely. I have noticed that in HTMLWriter.jl the case anchor.nth == 1 receives special treatment.

`anchor.nth` is important when two links happen to have the same `.id` (e.g. the same section heading in multiple places). The special casing in HTML is just because _usually_ you don't need to disambiguate, and so we don't want to have a trailing `-1` in all URLs (which are semi-visible to the readers).

So I do think we need to consistently use the `.nth` field when generating the labels. However, since we hash them into random strings anyway, there is no need to special case the `1`.

> Would it be a problem to simply use the ids themselves? For example. can they contain any characters that cause problems?

It's a design decision that predates me, but yes, I'm quite confident it's there to make sure we don't accidentally break the TeX by due to some weird characters. The docstrings `.id`s in particular can be problematic I think.

> I have removed the link because it's unclear to me what the purpose of a link is that links to itself.

I presume it was there to be consistent with the original Markdown writer. But I agree that it doesn't serve any purpose in the PDF.

> I've addressed this by using the LaTeX macro \label to create the target.

I am fine with the macro as well, but I wonder if we could use `\label` for both? Although I guess you can only attach `\label` to certain special LaTeX commands?

mortenpi:
Also, while we're doing this, could you factor out the `id` generation into a separate function? The logic is currently duplicated in multiple places. I think it would require two methods (or even functions): one for `Anchors.Anchor` and the other one for a generic object (i.e. for docstring links).

src/Writers/LaTeXWriter.jl
matthias314: If the anchors have `.nth` (at least if `> 1`), then it must also appear in the links. Is this currently the case? I couldn't see anything like this in `LaTeXWriter.jl`. (My understanding of the whole package is very limited at present.)

mortenpi: In short, yes, it should be used when generating the labels. But in the PDF, there is no point in special casing the `.nth == 1` case, since the link labels are not user-readable anyway.

matthias314:
> `anchor.nth` is important when two links happen to have the same `.id`

As I said, my understanding is limited. But are two identical anchor ids allowed? How can you distinguish between the various targets  if you refer to an id that occurs more than once?

> I wonder if we could use `\label` for both

`\label` in LaTeX can only refer to special locations in the document, for example `\chapter`, `\section` and so on as well as `\item`. The question is whether all Markdown anchors can be covered this way.

> could you factor out the id generation into a separate function?

For anchors there is already https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/Anchors.jl#L130-L137 One could make the `#` a second argument that would be the empty string for LaTeX. (Or one keeps it as it is invisible the user anyway.)

src/Writers/LaTeXWriter.jl
matthias314: I believe that link ids should use `.nth`, but currently it's not the case. (That's why the links didn't work.) How would one add that to, say, https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/Writers/LaTeXWriter.jl#L686-L690 in the function `latexinline(io::IO, md::Markdown.Link)`?

mortenpi:
> But are two identical anchor ids allowed? How can you distinguish between the various targets if you refer to an id that occurs more than once?

You mean in at-ref links? You can't, and if you try, it [leads to an error](https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/CrossReferences.jl#L100). If that is necessary, the users can use the at-id feature to give a heading a unique label.

However, this is not a problem as most anchors do not explicitly get referred to. But we still want to generate the anchors (e.g. in HTML) so that users could copy-paste the URLs.

> `\label` in LaTeX can only refer to special locations in the document, for example `\chapter`, `\section` and so on as well as `\item`. The question is whether all Markdown anchors can be covered this way.

Yep, that's fair. Let's stick to the macro.

> For anchors there is already

Yeah, but LaTeXWriter does additional processing that is specific to it, so I think there should be a function in `LaTeXWriter` that does the hashing etc. I don't think it's a good idea to go down the rabbit hole of trying to generalize the `Anchors` module in this PR.

src/Writers/LaTeXWriter.jl
mortenpi: Right, so we don't really have access to the underlying `Anchor` or docstring objects when we're generating the links. I'd say we just make sure that the anchors we generate in `LaTeXWriter` just hash a string that matches the URL fragment, which is generated here:

https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/CrossReferences.jl#L98
https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/CrossReferences.jl#L153

matthias314: But the question remains: How do we figure out what `nth` should be? It seems to me that we have to look up the `AnchorMap`s `doc.internal.headers` and `doc.internal.docs` where `doc::Document` is the first argument of `LaTeXWriter.render`. (`doc` is already carried along in quite a few `latex` functions.) Would you agree?

mortenpi: I don't believe that's necessary -- the `.nth` is already included in the URL fragment, no?

https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/CrossReferences.jl#L98
https://github.com/JuliaDocs/Documenter.jl/blob/6c92d59a19348b9ec4f1936bdce91d34511e4c18/src/Anchors.jl#L130-L137

Also, for what it's worth, it looks like for docstrings we ignore `.nth`, since for those we shouldn't get duplicates. And the `latex(io::IO, anchor::Anchors.Anchor, page, doc)` method only generated anchors/targets for headings, not docstrings, so we don't need to worry about distinguishing those two cases here.

matthias314: Maybe I'm missing something. Here is what I understand: We need to generate the `nth` both for the target and for the link. For the target it's easy. For the link we have (at least) the code snipped from `latexinline(io::IO, md::Markdown.Link)` that I copied above. As far as I can tell, that code is used both for referring to headers and to docstrings. In that code the `nth` is not included at present for otherwise links would work in the PDF file. So we have to add it for headers. The code snippets you have given above both use an argument of type `Anchor`, but we currently don't have an `Anchor` when we generate the link.

mortenpi: Unless I am misreading the code myself, `latexinline(io::IO, md::Markdown.Link)` hashes the fragment part of `md.url`, which is determined by the `fragment()` function in `CrossReferences`. This is means that it already includes `.nth` (well, if `.nth` is `>= 2`). We just have to make sure that the string we hash when we generate the targets (and I guess also the at-contents block entires) is consistent with what `CrossReferences` puts into the `.url` field of the `Link` objects.

Basically, I don't think we need to make any changes to the `latexinline(io::IO, md::Markdown.Link)` method.

mortenpi: Isn't the reason the links are not working right now that when we generate the targets, we use `.nth` in the string we hash even when `.nth == ``. But it that case it is not included in the fragment, which is what we hash in the link generation method?

matthias314:
OK, I think that now the treatment of `anchor.nth` is correct. My impression is that the problem arose because the labels were generated are several places in the source code (and have to match!), and at one point (when the `nth` was omitted if it equals `1`) the change was not done everywhere in the code.  I therefore thought that the cleanest approach would be to introduce a new function `Anchors.label` to generate the labels for all anchor and links; also `Anchors.fragments` uses it now. For uniformity, I've also created a function `_hash` in `LaTeXWriters.jl` that is called whenever a hash is computed in that file. This new PR still fixes the broken links in the Julia documentation.

Let me know what you think. If you think it's ready for merging, then I can first make a rebase to get a single clean commit.

assets/latex/documenter.sty
```diff
@@ -85,3 +85,7 @@
 \usepackage{graphicx}
 \usepackage[export]{adjustbox}
 %
+
+% insert \hyperlink or, if a corresponding label exists, \hyperref
```
mortenpi, line 89:
```suggestion
% Some internal link targets are implemented with \label, some with \hypertarget,
% but they require different links. This inserts a \hyperref if a corresponding label exists,
% and \hyperlink if it doesn't.
```

mortenpi (approved):
This is excellent! I like the splitting of the logic into these separate functions.

Could you also add a note into `CHANGELOG.md`?

matthias314:
I've updated the changelog and rebased the branch. (It took me a few attempts to figure out how links work in the changelog file.) Should be ready now for merging.

mortenpi:
Thanks @matthias314!

PR #1919, opened 2022-09-05
Restrict the type of the last expression in at-eval blocks

mortenpi:
Currently, you are generally allowed to have at-eval to return anything, and the resulting object will get spliced into the document somehow, but we do not really define how we treat arbitrary objects (HTML just `repr`s them and dumps the resulting string as text in the document).

In practice, and based on our docs, most of the time you're expected to either return `nothing` (so nothing gets displayed) or `Markdown.MD` (so you get properly rendered Markdown). This makes that requirement general and explicit, making it easier to support this block in the writers (that now know that it will always be either `nothing` or `Markdown.MD`, and don't have to deal with arbitrary types).

This can be breaking because it was possible to get somewhat reasonable output for some return types and it's possible that people are using it (in particular, I am concerned about cases where people might be returning various `Markdown.*` nodes without wrapping them in `Markdown.MD`). For that reason, this does not make that case quite an error, but instead you get a warning, and the resulting object is `show`ed into a text representation and displayed as a code block (so there is behavior change for users).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 
   **For upgrading:** To keep using the Markdown backend, refer to the [DocumenterMarkdown package][documentermarkdown]. That package might not immediately support the latest Documenter version, however.
 
+* ![BREAKING][badge-breaking] `@eval` blocks now require the last expression to be either `nothing` or of type `Markdown.MD`, with other cases now issuing a warning and falling back to a text representation in a code block. ([#1919][github-1919])
+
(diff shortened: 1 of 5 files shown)


fredrikekre:
Perhaps possible to display the same as an example block?

mortenpi:
> Yea, and maybe it makes sense to keep that restriction. If you want an example block you can just use an example block (the only advantage would be hiding the code by default I suppose).

Let's leave that as an open issue for now.

PR #1921, opened 2022-09-06
make binding support `->` function

clouds56:
So that it works for ``[`x -> sin(x)`](@ref)``,
As well as ``[`sin(_)`](@ref)`` if https://github.com/JuliaLang/julia/pull/46633 checked in

### DIFF
--- a/src/DocSystem.jl
+++ b/src/DocSystem.jl
@@ -62,6 +62,7 @@ end
 binding(m::Module, x::Expr) =
     Meta.isexpr(x, :.) ? binding(getmod(m, x.args[1]), x.args[2].value) :
     Meta.isexpr(x, [:call, :macrocall, :curly]) ? binding(m, x.args[1]) :
+    Meta.isexpr(x, :->) ? binding(m, x.args[2].args[end]) :
     Meta.isexpr(x, :where) ? binding(m, x.args[1].args[1]) :
         error("`binding` cannot understand expression `$x`.")
 


mortenpi:
I am a little hesitant about this.. this feels like a niche usage, and in those rare cases wouldn't it be okay to just write it as ``[`x -> sin(x)`](@ref sin)``? Also, isn't `x -> sin(x)` technically different from `sin`?

clouds56:
We both hold the view that `x -> sin(x)` is the same as `sin`, so there's no logical problem.
The only question here is shall we support this corner case / would it be useful since nobody would write `x -> sin(x)`.

It's common to write `sin(_)` and after somehow we implement `_` as anonymous function it might be translate to `x -> sin(x)`, you could see errors in the build https://buildkite.com/julialang/julia-master/builds/15482#01830cdc-697e-4664-a67f-354ab826efea for more details.
And at any level it doesn't hurt since in no case we read `x -> sin(x)` different from `sin`.

If you still not sure just leave it until `_` is about to land.

odow:
> We both hold the view that x -> sin(x) is the same as sin, so there's no logical problem
> since in no case we read x -> sin(x) different from sin.

No, I think @mortenpi and I hold the opposite view. `x -> sin(x)` is a very different object to `sin`, any they may have different docstrings.

I am not in favor of this.

fingolfin:
I agree these are different functions, and I would suggest this be closed

PR #1923, opened 2022-09-07
fix error in relpath_from_repo_root when file not exists

clouds56:
now `relpath_from_repo_root` would fail if file not exists

```
ERROR: LoadError: IOError: cd("/cache/build/default-amdci5-1/julialang/julia-master/usr/share/julia/stdlib/v1.9/CRC32c/src"): no such file or directory (ENOENT)
Stacktrace:
  [1] uv_error
    @ ./libuv.jl:100 [inlined]
  [2] cd(dir::String)
    @ Base.Filesystem ./file.jl:91
  [3] cd(f::Documenter.Utilities.var"#3#4"{String}, dir::String)
    @ Base.Filesystem ./file.jl:111
  [4] relpath_from_repo_root
    @ /cache/build/default-amdci4-6/julialang/julia-master/doc/deps/packages/Documenter/GS7EF/src/Utilities/Utilities.jl:423 [inlined]
  [5] source_url(repo::Documenter.Utilities.Remotes.GitHub, mod::Module, file::String, linerange::UnitRange{Int64})
    @ Documenter.Utilities /cache/build/default-amdci4-6/julialang/julia-master/doc/deps/packages/Documenter/GS7EF/src/Utilities/Utilities.jl:471
```

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -420,6 +420,7 @@ Returns the path of `file`, relative to the root of the Git repository, or `noth
 file is not in a Git repository.
 """
 function relpath_from_repo_root(file)
+    isfile(file) || return nothing
     cd(dirname(file)) do
         root = repo_root(file)
         root !== nothing && startswith(file, root) ? relpath(file, root) : nothing
@@ -427,6 +428,7 @@ function relpath_from_repo_root(file)
 end
 
 function repo_commit(file)
+    isfile(file) || return nothing
     cd(dirname(file)) do
         readchomp(`$(git()) rev-parse HEAD`)
     end
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -420,7 +420,7 @@ Returns the path of `file`, relative to the root of the Git repository, or `noth
 file is not in a Git repository.
 """
 function relpath_from_repo_root(file)
-    isfile(file) || return nothing
+    isfile(file) || error("relpath_from_repo_root called with nonexistent file: $file")
     cd(dirname(file)) do
         root = repo_root(file)
         root !== nothing && startswith(file, root) ? relpath(file, root) : nothing
@@ -428,7 +428,7 @@ function relpath_from_repo_root(file)
 end
 
 function repo_commit(file)
-    isfile(file) || return nothing
+    isfile(file) || error("repo_commit called with nonexistent file: $file")
     cd(dirname(file)) do
         readchomp(`$(git()) rev-parse HEAD`)
     end
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -191,8 +191,8 @@ end
             @test Documenter.Utilities.repo_root(filepath; dbdir=".svn") == nothing
             @test Documenter.Utilities.relpath_from_repo_root(filepath) == joinpath("src", "SourceFile.jl")
             # We assume that a temporary file is not in a repo
-            @test Documenter.Utilities.repo_root(tempname()) == nothing
-            @test Documenter.Utilities.relpath_from_repo_root(tempname()) == nothing
+            @test_throws ErrorException Documenter.Utilities.repo_root(tempname())
+            @test_throws ErrorException Documenter.Utilities.relpath_from_repo_root(tempname())
         end
 
         # Test worktree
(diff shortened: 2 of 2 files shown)


mortenpi:
We should indeed handle this case. How did you run into this by the way? Was that when building the Julia manual? Locally or on CI?

I am not sure returning `nothing` fixes it though.. don't you now get a `MethodError` from `repofile`?

clouds56:
> Was that when building the Julia manual? Locally or on CI?

On CI, just use the master julia and master Document.jl would repro this

Before the change: https://buildkite.com/julialang/julia-master/builds/15556#018313da-c7ff-4313-903a-e52390834827
after the change: https://buildkite.com/julialang/julia-master/builds/15581#01831796-6847-4ab9-9c7e-ce63d813c377
(other changes are unrelated, just focus on doctest.

clouds56:
The key is julia is not installed at prefix it build, so it would trying to find file (/cache/build/**default-amdci5-1**/julialang/julia-master/usr/share/julia/stdlib/v1.9/CRC32c/src)
but there's only /cache/build/**default-amdci4-6**/julialang/julia-master/doc/deps/packages/Documenter/GS7EF/src

src/Utilities/Utilities.jl
```diff
@@ -420,13 +420,15 @@ Returns the path of `file`, relative to the root of the Git repository, or `noth
 file is not in a Git repository.
 """
 function relpath_from_repo_root(file)
+    isfile(file) || return nothing
```
mortenpi, line 423:
```suggestion
    isfile(file) || error("relpath_from_repo_root called with nonexistent file: $file")
```

src/Utilities/Utilities.jl
```diff
@@ -420,13 +420,15 @@ Returns the path of `file`, relative to the root of the Git repository, or `noth
...
     cd(dirname(file)) do
         root = repo_root(file)
         root !== nothing && startswith(file, root) ? relpath(file, root) : nothing
     end
 end
 
 function repo_commit(file)
+    isfile(file) || return nothing
```
mortenpi, line 431:
```suggestion
    isfile(file) || error("repo_commit called with nonexistent file: $file")
```

mortenpi (commented):
In reference to #1964.

mortenpi:
It looks like there are some tests that are outdated now due to this change. Could you change these to `@test_throws ErrorException ...`? I don't have write access to your branch, so I can't do it myself.

* https://github.com/clouds56-contrib/Documenter.jl/blob/2bcd70716dd84e4bcba404114e40727ea9778397/test/utilities.jl#L193-L195
* https://github.com/clouds56-contrib/Documenter.jl/blob/2bcd70716dd84e4bcba404114e40727ea9778397/test/utilities.jl#L215-L217
* https://github.com/clouds56-contrib/Documenter.jl/blob/2bcd70716dd84e4bcba404114e40727ea9778397/test/utilities.jl#L250-L252

test/utilities.jl
```diff
@@ -191,8 +191,8 @@ end
             @test Documenter.Utilities.repo_root(filepath; dbdir=".svn") == nothing
             @test Documenter.Utilities.relpath_from_repo_root(filepath) == joinpath("src", "SourceFile.jl")
             # We assume that a temporary file is not in a repo
-            @test Documenter.Utilities.repo_root(tempname()) == nothing
-            @test Documenter.Utilities.relpath_from_repo_root(tempname()) == nothing
+            @test_throws ErrorException Documenter.Utilities.repo_root(tempname())
```
mortenpi, line 194:
```suggestion
            @test Documenter.Utilities.repo_root(tempname()) == nothing
```

test/utilities.jl
```diff
@@ -213,8 +213,8 @@ end
...
```
mortenpi, line 216:
```suggestion
            @test Documenter.Utilities.repo_root(tempname()) == nothing
```

test/utilities.jl
```diff
@@ -248,8 +248,8 @@ end
...
```
mortenpi, line 251:
```suggestion
            @test Documenter.Utilities.repo_root(tempname()) == nothing
```

mortenpi (commented):
Uh, sorry, my bad, it was `repo_commit` we changed to throw an error, not `repo_root`.

mortenpi:
Thanks @clouds56!

PR #1932, opened 2022-09-14
Style admonitions with tcolorbox in LaTeXWriter

### LINKED ISSUE #1931: Styling of admonitions in LaTeXWriter
The styling of admonitions in the latex writer does't make them stand out: 

image

It's just a `quote` with `textbf` for the header:
https://github.com/JuliaDocs/Documenter.jl/blob/6b8f8f917bdfdbb3a095fe1eb380b4c614b489cc/src/Writers/LaTeXWriter.jl#L568-L576

If we used something like https://ctan.org/pkg/tcolorbox?lang=en, we could have:
```
\begin{tcolorbox}[colback=red!5!white,colframe=red!75!black,title=Tip]
If you aren't sure if you should use JuMP, read "Should I use JuMP?"
\end{tcolorbox}
```

image

Is there a workflow for adding new dependency packages? Do we have to update the docker image?

odow:
TODOs before merging:
 - [x] NEWS
 - [x] Tests? I'm not really sure how to test this.

Closes #1931 

## Before

image

## After

image


odow:
Ah. Test failures are because I was developing on the latest patch, and then rebased onto master.

src/Writers/LaTeXWriter.jl
```diff
@@ -566,13 +566,18 @@ function latex(io::Context, node::Node, ::MarkdownAST.BlockQuote)
...
-        _println(io, "\n")
-        latex(io, node.children)
+    color = "default"
+    if md.category in ("danger", "warning", "note", "info", "tip", "compat")
+        color = "admonition-$(md.category)"
     end
+    _print(io, "\\begin{tcolorbox}[")
+    _print(io, "colback=$(color)!5!white,colframe=$(color)!75!black,")
+    _print(io, "title={")
+    latexinline(io, md.title)
```
odow, line 576:
```suggestion
    latexesc(io, md.title)
```

assets/latex/documenter.sty
```diff
@@ -91,3 +91,16 @@
 % and \hyperlink if it doesn't.
 \def\hyperlinkref#1#2{\@ifundefined{r@#1}{\hyperlink{#1}{#2}}{\hyperref[#1]{#2}}}
 %
+
+% styling of admonitions. tcolorbox is available in the texlive-latex-extra
+% package
+\usepackage{tcolorbox}
+% These colors are taken from documeter-light.css in the HTML assets.
```
odow, line 98:
```suggestion
% These colors are taken from documenter-light.css in the HTML assets.
```

mortenpi:
To fix the reference tests, you can run the tests locally with [`DOCUMENTER_FIXTESTS` set (to anything)](https://github.com/JuliaDocs/Documenter.jl/blob/00b9f75ee7275f72f1ed4d289d8239308db8769d/test/examples/tests.jl#L54-L57).

odow:
Okay, that seems to have done the trick.

mortenpi:
Hmm, it looks like there is some bad interaction with the surrounding elements here, effectively making the admonitions inline elements:

image

You can get the PDF from [the PDFs artifact here](https://github.com/JuliaDocs/Documenter.jl/actions/runs/3056775567), and and [the relevant markdown is here](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/src.latex_showcase/showcase.md).

mortenpi:
Oh, and this is getting into bikeshedding territory, but what about (1) bold text in the title, and (2) any way to get it vertically centered?

odow:
Oh, nice. I didn't notice the artifacts. I'll admit that I didn't look at the PDF of the showcase, I was just looking at the MOI and JuMP docs. 

It looks like this might be some weird behavior with the use of `\subparagraph` for the `Note admonition` headers in bold?

Other parts seem normal:

image

image

Bikeshed away. Bold and vertical centering look doable: https://tex.stackexchange.com/questions/482103/tcolorbox-vertically-center-title-box

odow:
It's because `paragraph` and `subparagraph` don't create new paragraphs after themselves. I guess we need an explicit `\indent`. This would affect things other than admonitions, like tables:
```
\subparagraph{New subparagraph}

\begin{tabular}{| c | c |}
\hline
foo & bar \\
\hline
\end{tabular}

\subparagraph{New subparagraph with indent}\indent

\begin{tabular}{| c | c |}
\hline
foo & bar \\
\hline
\end{tabular}
```

image

odow:
Okay, this is now:

image

I don't understand the centering of the titles. It looks like some interplay with how we're redefining paragraph spacing etc.

I'd be in favor of merging this as-is, and coming back to it as a second-order thing.

Issue #1933, opened 2022-09-15
Styling of code blocks in LaTeXWriter


odow:
The styling of code blocks in the PDF output could be improved: it's hard to distinguish code from text, and code that should be typed from output. Here are some examples from the JuMP documentation: https://jump.dev/JuMP.jl/stable/JuMP.pdf.

 ---

<img width="374" alt="image" src="https://user-images.githubusercontent.com/8177701/190283301-dcef540f-67c5-4d3e-99b4-0a896bccfbd3.png">

---

<img width="487" alt="image" src="https://user-images.githubusercontent.com/8177701/190283207-9ab18680-2452-43ea-aa4a-9d93febd2c6f.png">

---

## Suggestions

This is pretty easy to fix, because it just involves changing 
https://github.com/JuliaDocs/Documenter.jl/blob/00b9f75ee7275f72f1ed4d289d8239308db8769d/assets/latex/documenter.sty#L34-L48

So the hard part is deciding what it should be changed to. Here are some suggestions:

<img width="463" alt="image" src="https://user-images.githubusercontent.com/8177701/190283827-a951ace9-3822-47da-8d54-6554b5e96e5d.png">

I'd probably vote for 3 or 6.

odow:
I realized we can implement this with a `custom.sty` file for Option 6. 

Here's the PR to MOI: https://github.com/jump-dev/MathOptInterface.jl/pull/1999

Preview link: https://jump.dev/MathOptInterface.jl/previews/PR1999/MathOptInterface.pdf

I think it looks a lot better:

<img width="504" alt="image" src="https://user-images.githubusercontent.com/8177701/190290756-d078e511-09e9-47be-8d8b-46965e522ff1.png">

but there are a few weird top-margin issues (I think it happens when the code is the first item in a paragraph?):

<img width="484" alt="image" src="https://user-images.githubusercontent.com/8177701/190290862-8e6c0ac4-d64d-487b-93b4-06b4ad154cce.png">

and docstrings expose the fact that the first indent is treated as `{julia}` but the rest are treated as unstyled code block. (This is our fault.)

<img width="484" alt="image" src="https://user-images.githubusercontent.com/8177701/190290907-193d6964-e32a-44cd-948f-d394ee06ddec.png">

mortenpi:
Yeah, I think having just some background color looks much better. And I think it actually looks better to not have any lines/borders. What are you thinking about doing with the outputs though?

[...]

PR #1935, opened 2022-09-15
Change styling of code blocks in LaTeXWriter

odow:
The exact settings are still somewhat up for discussion.

Closes https://github.com/JuliaDocs/Documenter.jl/issues/1933.

Here are two examples where I've been overriding the settings with a `custom.sty`:

 * MathOptInterface:  https://github.com/jump-dev/MathOptInterface.jl/pull/1999
   * https://jump.dev/MathOptInterface.jl/previews/PR1999/MathOptInterface.pdf
 * JuMP: https://github.com/jump-dev/JuMP.jl/pull/3075
   * https://jump.dev/JuMP.jl/previews/PR3075/JuMP.pdf

I think we could still change the styling of non-minted blocks, but I'm not sure the best settings. Here's what it looks like currently:

image

### DIFF
--- a/assets/latex/documenter.sty
+++ b/assets/latex/documenter.sty
@@ -36,10 +36,10 @@
     basicstyle = \small\ttfamily,
     breaklines = true,
     columns = fullflexible,
-    frame = leftline,
+    backgroundcolor = \color{codeblock-color},
+    frame = none,
     keepspaces = true,
     showstringspaces = false,
-    xleftmargin = 3pt,
 }
 
 \setminted{
--- a/assets/latex/documenter.sty
+++ b/assets/latex/documenter.sty
@@ -22,6 +22,7 @@
 \definecolor{dark-green}{HTML}{3b972e}
 \definecolor{light-purple}{HTML}{aa7dc0}
 \definecolor{dark-purple}{HTML}{945bb0}
+\definecolor{codeblock-background}{gray}{0.96}  
 %
 
 % maths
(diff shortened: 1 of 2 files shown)


odow:
I though it might be helpful to look around for inspiration:

| Project | link | Example |
| ------- | --- | ---------|
| AMPL | https://ampl.com/BOOK/CHAPTERS/07-tut4.pdf | image |
| Pyomo | https://link.springer.com/book/10.1007/978-3-030-68928-5 | image|
| Python | https://bugs.python.org/file47781/Tutorial_EDIT.pdf | image |
| Rust | https://doc.rust-lang.org/book/ | image |
| Clojure | https://riptutorial.com/Download/clojure.pdf | image |

The best option might be adding a `When evaluated in the REPL, this will print:` to all the `@example` blocks.

assets/latex/documenter.sty
```diff
@@ -35,16 +36,17 @@
     basicstyle = \small\ttfamily,
     breaklines = true,
     columns = fullflexible,
-    frame = leftline,
+    backgroundcolor = \color{codeblock-color},
```
odow, line 39:
```suggestion
    backgroundcolor = \color{codeblock-background},
```

odow:
This is now
image

mortenpi:
> The best option might be adding a `When evaluated in the REPL, this will print:` to all the `@example` blocks.

It's not a bad idea to guide the reader, but I am also a little hesitant to generate text that is not written by the author. I think this would be trickier to achieve, but some sort of "output" tag to the output box would maybe be better?

odow:
It currently produces:

image

Take a look through the JuMP docs: https://jump.dev/JuMP.jl/dev/JuMP.pdf. I think it works pretty well.

mortenpi:
Looks great!

PR #1937, opened 2022-09-15
Wrap tabular environments in adjustbox to resize to page width

### LINKED ISSUE #1930: Overfull elements in LaTeXWriter
The JuMP documentation (https://jump.dev/JuMP.jl/stable/JuMP.pdf) has a bunch of examples where the text is overfull. Are there any tools in Documenter to improve the layout of things? 

I guess the hard part is dealing without put from types (like dataframes) which implement `show(io, ::MIME"text/latex", x::MyType)`, but perhaps every output should have a `\resizebox{\textwidth}{!}{...}` around it? It's probably better to have things too small than to go off the page.

## Equations

image

## DataFrame tables

image

## In-line code `snippets`

image

odow:
This most commonly affects tables produced by DatFrames.jl, and resizes them to fit the page if they have many columns that would otherwise run off the right-hand edge of the page.

Closes #1930

This is what it looks like now:

image

I added a `{table}` environment with `\centering`, but I can remove if people object.


odow:
Linkcheck failure seems like a network issue. It isn't related to this PR.

mortenpi:
There is still a table in the showcase that manages to overflow.. but that is due to the cells themselves overflowing I guess, which is harder to fight.

image

Could we add one or two examples with tables to the showcase for this?

Also from the showcase, big equations can be problematic too:

image

Do you think `adjustbox`ing them could also work to avoid this?

odow:
I don't know if LaTeX has a way of automatically wrapping text in a table without manually specifying the column widths.

odow:
Math issue is https://github.com/JuliaDocs/Documenter.jl/issues/1470, which was fixed for HTML but not for LaTeX.

mortenpi:
I am happy to merge this as is too, and leave the equations etc. for the future. If so, this could just use a CHANGELOG note.

odow:
Added. 

p.s. I'm just about to head to the airport, so I'll only be online sporadically until mid-next week.

mortenpi:
Do you think we should also put an adjustbox around the usual tables (as opposed to ones from `show` methods)? I.e.

https://github.com/JuliaDocs/Documenter.jl/blob/674c02ea48bb1674d6ba4dfe3c1f81d72404cf4f/src/Writers/LaTeXWriter.jl#L640-L657

odow:
The normal markdown tables already use the `{tabulary}` environment with the `linewidth` option, so I think we're okay without? 

The problematic tables are really `DataFrames` because it implements `show(::IO, ::MIME"text/latex", ::DataFrame)` that produces a `{tabular}` environment (since it can't use external latex packages).

mortenpi:
They can still overflow if they're big enough though (you can look at the at-eval generated table I added to the showcase). However, I'm also happy to leave it for a future improvement.

odow:
I don't really understand why the `@eval` doesn't get resized, but the `@example` does?

image

It's another second-order thing. We're heading in the right direction, and getting everything perfect every time is a really hard job.

mortenpi:
It's a different `latex` method that gets called (we print the `Markdown[AST].Table` directly, and don't rely on the `show` method), so it doesn't get wrapped in `adjustbox` right now.

https://github.com/JuliaDocs/Documenter.jl/blob/674c02ea48bb1674d6ba4dfe3c1f81d72404cf4f/src/Writers/LaTeXWriter.jl#L640-L657

But as I said, I'm more than happy to merge this as is. As you said, mostly it's DataFrames etc. that cause problems, which the current PR already takes care of.

odow:
> But as I said, I'm more than happy to merge this as is. As you said, mostly it's DataFrames etc. that cause problems, which the current PR already takes care of.

:+1: I'm keen to build the JuMP docs once these PRs are all in to see how much of a difference it makes.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 * ![Enhancement][badge-enhancement] Admonitions are now styled with color in the LaTeX output. ([#1931][github-1931], [#1932][github-1932])
 * ![Enhancement][badge-enhancement] Improved the styling of code blocks in the LaTeXWriter. ([#1933][github-1933], [#1935][github-1935])
+* ![Enhancement][badge-enhancement] Resize oversize `tabular` enviroments in the LaTeXWriter. ([#1930][github-1930], [#1937][github-1937])
```
mortenpi, line 16:
```suggestion
* ![Enhancement][badge-enhancement] Automatically resize oversize `tabular` environments from `@example` blocks in LaTeXWriter. ([#1930][github-1930], [#1937][github-1937])
```

PR #1938, opened 2022-09-15
Improve styling of tables in LaTeXWriter

odow:
The current tables have all borders and don't distinguish headers from subsequent rows.

## Current

image

image

## New

image

### DIFF
--- a/assets/latex/documenter.sty
+++ b/assets/latex/documenter.sty
@@ -107,3 +107,7 @@
 \definecolor{admonition-tip}{HTML}{22c35b}
 \definecolor{admonition-compat}{HTML}{1db5c9}
 %
+
+% Styling of tables.
+\usepackage{booktabs}
+%
--- a/src/Writers/LaTeXWriter.jl
+++ b/src/Writers/LaTeXWriter.jl
@@ -643,7 +643,7 @@ function latex(io::Context, node::Node, table::MarkdownAST.Table)
     _println(io, "\n\\begin{table}[h]\n\\centering")
     _print(io, "\\begin{tabulary}{\\linewidth}")
     _println(io, "{", uppercase(join(spec_to_align.(table.spec), ' ')), "}")
-    _println(io, "\\hline")
+    _println(io, "\\toprule")
     for (i, row) in enumerate(rows)
         for (j, cell) in enumerate(row.children)
             j === 1 || _print(io, " & ")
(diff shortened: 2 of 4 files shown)


KristofferC:
Maybe worth to use `\usepackage{booktabs}` and use ` \toprule`, `\midrule` and `\bottomrule` instead of `\hline`. Looks nicer imo.

mortenpi:
For larger tables I do find that it improves readability if you have some horizontal lines to guide you between the rows. Do you think it would still look good to have either light or dark gray lines between the rows? I do agree that it makes sense to distinguish the header from the body rows though.

odow:
Before

image

After

image

KristofferC:
It's much better than before but the horizontal lines in between the rows is not very nice IMO. I kind of agree with

> For larger tables I do find that it improves readability if you have some horizontal lines to guide you between the rows. 

so how about something like

gnome-shell-screenshot-hnfvcc

with a very faint grey background to guide you? Generated with:

```latex
\documentclass[12pt]{article}
\usepackage{booktabs}

\usepackage{color, colortbl}
\usepackage[table]{xcolor}
\definecolor{Gray}{gray}{0.93}
\usepackage[first=0,last=9]{lcg}
\newcommand{\ra}{\rand0.\arabic{rand}}

\begin{document}

\rowcolors{2}{Gray}{white}
\begin{table}[ht]
\centering
\begin{tabular}{cccccccc}
\toprule
\rowcolor{white}
&col1 &col2 &col3 &col4 & col5 &col6 &col7\\
\midrule
row1& \ra & \ra & \ra & \ra & \ra & \ra & \ra \\
row2& \ra & \ra & \ra & \ra & \ra & \ra & \ra \\
row2& \ra & \ra & \ra & \ra & \ra & \ra & \ra \\
row2& \ra & \ra & \ra & \ra & \ra & \ra & \ra \\
row2& \ra & \ra & \ra & \ra & \ra & \ra & \ra \\
row3& \ra & \ra & \ra & \ra & \ra & \ra & \ra \\
\hline
\end{tabular}
\caption{My Table}
\end{table}

\end{document}
```

odow:
I'll let @mortenpi decide. But I'd vote no lines and no color. 

Is this really less readable?

image

mortenpi:
First, color >> lines, so that's a great suggestion!

As for the readability of color vs no color: I think for small, dense tables no color maybe even more readable. But I am concerned about large tables, with potentially some whitespace. E.g. this example

image

If you would make it even bigger and have more variation in the line lengths, I think it would become tricky to read without any horizontal guides.

odow:
Here's a big table from the JuMP docs:

## No color 

image

## With color 

image

For now, I think the current state of the PR is an improvement in all cases, but it's hard to be perfect in every case. In the interest of increment change, I'd vote we merge, wait to see how things go in the ecosystem, and come back for more fixes/if anyone complains.

mortenpi:
Sounds good. Yea, there is always more bikeshedding that can be done with these things. But all the PRs you made are already a massive improvement to the look of the PDF, so thank you very much for taking them all on @odow!

PR #1940, opened 2022-09-17
Issue a more helpful warning if #output contains trailing whitespace in doctest

### LINKED ISSUE #1939: jldoctest fails if # output has trailing space
Currently named doctests fail, if there is a trailing space after the beginning the jldoctest block. Example

```jldoctest mytest
a = 1
b = 2
a + b

# output 
3
```
where I have one trailing space behind the `# output ` line fails with 
```
┌ Warning: invalid doctest block in ...
│ Requires `julia> ` or `# output`
│
│ ```jldoctest mytest
```

which took me quite a while to find.

jlbosse:
Fixes #1939

### DIFF
--- a/src/DocTests.jl
+++ b/src/DocTests.jl
@@ -192,6 +192,18 @@ function doctest(ctx::DocTestContext, block_immutable::MarkdownAST.CodeBlock)
             eval_repl(block, sandbox, ctx.meta, ctx.doc, ctx.file)
         elseif occursin(r"^# output\s*$"m, block.code)
             eval_script(block, sandbox, ctx.meta, ctx.doc, ctx.file)
+        elseif occursin(r"^# output\s+$"m, block.code)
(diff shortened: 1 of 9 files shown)


src/DocTests.jl
```diff
@@ -190,8 +190,20 @@ function doctest(ctx::DocTestContext, block_immutable::MarkdownAST.CodeBlock)
         end
         if occursin(r"^julia> "m, block.code)
             eval_repl(block, sandbox, ctx.meta, ctx.doc, ctx.file)
-        elseif occursin(r"^# output$"m, block.code)
+        elseif occursin(r"^# output\s*$"m, block.code)
```
mortenpi, line 193: I think we need to take this away, otherwise the next `if` never triggers?

```suggestion
        elseif occursin(r"^# output$"m, block.code)
```

mortenpi:
LGTM, thanks!

PR #1941, opened 2022-09-18
Fix render-this example in LaTeX showcase

odow:
This PR fixes a small bug in the LaTeX tests. I initially suspected I'd borked something while working on the other PRs I have going, but it's just a typo.

Here's what it was previously:

image


mortenpi:
Good catch, thanks!

PR #1954, opened 2022-09-29
Small style tweaks to LaTeXWriter

odow:
I built the JuMP docs with the `master` branch of Documenter (https://github.com/jump-dev/JuMP.jl/pull/3094, https://jump.dev/JuMP.jl/previews/PR3094/JuMP.pdf), and I have a couple of small tweaks to PRs that went in. 

 * I tweaked the colors of #1932 a bit to make them more similar to the HTML. They were previous a bit too dark
 * I removed the `\midrule` from the tables in #1938. I think we intended to do this, but I forgot to update the PR again. I think I got a bit confused because we kept changing back and forward. 

I can split into two PRs if needed.

On the whole though, things are much better. The table resizing is great, the admonitions stand out a lot more, and the code cells are nicely highlighted.


mortenpi:
LGTM, thanks @odow!

PR #1955, opened 2022-09-29
Fix title margin of admonitions in LaTeXWriter

### LINKED ISSUE #1946: Properly center the adminition title vertically in LaTeX output
See https://github.com/JuliaDocs/Documenter.jl/pull/1932#issuecomment-1250185278

odow:
Closes #1946 

The default should be 0mm, so it must be some package that's mucking with this? I couldn't find the exact height is being modified, so let's just check how this looks in the PDFs created by CI. It looks okay locally.


odow:
Better?

image

mortenpi (approved):
It definitely looks better. You're just narrowing the title bar? I think in an ideal world there would be a little bit more padding above and below the text (so keep the old height I guess, but center the text vertically), but this is better than the status quo. Happy to merge as is if you're happy with it.

odow:
> You're just narrowing the title bar?

I think some other package is mucking with the margins above the title, so it's not that the title isn't centered, it's that there is extra space above. So this gets things back to the default. 

We can go the other way and try to increase the margin underneath the title if you like. I don't have a preference.

mortenpi:
Hmm. Could we try removing just 1mm from the top and adding 1mm to the bottom, if we can mix it like that easily?

odow:
Better?

image

mortenpi:
Yep, I think that looks perfect!

mortenpi:
Thank you for keeping on getting nerdsniped with this stuff :laughing:

PR #1956, opened 2022-09-29
Make code block styling in LaTeXWriter consistent with HTML

### LINKED ISSUE #1944: Make code block styling in LaTeX consistent with HTML
Right now, LaTeX shows Julia with `minted` and everything else with `listings`, which leads to a different style depending of if the code block is Julia or not. We should print everything always on the gray background, and only the example outputs in white boxes (like in HTMLWriter).

X-ref: https://github.com/JuliaDocs/Documenter.jl/issues/1933#issuecomment-1250009121

odow:
Closes #1944 

Now
````
```
A code block
```

```@example
1 + 1
```
````

produces

image

The raw code blocks have a gray background `{minted}` with `text` as the language, and `text/plain` outputs are left as `{lstlisting}`s.

I also pulled in the margins of the `{lstlisting}` because they were slightly wider than the `{minted}` environments.


odow:
This isn't quite right, because ` ````markdown ` blocks are white:

image

in HTML they are

image

mortenpi (approved):
Looks great. It's good to go from your end?

odow:
> It's good to go from your end?

Yes, I think so.

PR #1957, opened 2022-09-30
Switch to use {minted} for code output in LaTeXWriter

### LINKED ISSUE #1936: Greek letters mangled in LaTeXWriter
These are `@example` blocks in the JuMP documentation:

````
```@example
π
```
```@example
typeof(π)
```
````

image

Note the new line after the first π and the incorrect location of π in the output of the `typeof` cell.

REPL is
```julia
julia> π
π = 3.1415926535897...

julia> typeof(pi)
Irrational{:π}
```

odow:
Closes #1936

Fixes formatting issues like this from the JuMP documentation: https://jump.dev/JuMP.jl/previews/PR3094/JuMP.pdf

image


odow:
So this renders nicely. But the margins are different?!?

image

odow:
This seems to have done the trick

image

odow:
Okay, I think I'm happy:

image

mortenpi:
LGTM, thanks!

PR #1964, opened 2022-10-08
Don't throw in source_url() due to a non-existing source file

mortenpi:
@clouds56: sorry it took me this long to get back to you on #1923, but I think this fix is more correct. `relpath_from_repo_root` and `repo_commit` are only called in `edit_url` and `source_url`, and they should never get called with a non-existing file. That `source_url` sometimes does was an oversight from #1808.

It might still be a good idea to add an `isfile(file) || error("...")` to those functions though, just to explicitly catch this case in the future, and you're welcome to update the PR with that.

This should fix the CI errors you reported, which I'm testing over at https://github.com/JuliaLang/julia/pull/47105.

### DIFF
--- a/src/Utilities/Utilities.jl
+++ b/src/Utilities/Utilities.jl
@@ -467,7 +467,7 @@ function source_url(repo, mod, file, linerange)
             Base.GIT_VERSION_INFO.commit
         end
         repofile(julia_remote, ref, "base/$file", linerange)
-    else
+    elseif isfile(file)
         path = relpath_from_repo_root(file)
         # If we managed to determine a remote for the current file with getremote,
         # then we use that information instead of the user-provided repo (doc.user.remote)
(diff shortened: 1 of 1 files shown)


fredrikekre:
WIll the fact that these are not found mean links will be broken in the julia docs?

clouds56:
Thanks!

mortenpi:
> WIll the fact that these are not found mean links will be broken in the julia docs?

Yes, there are broken links, already now. In particular, the standard libraries are problematic, e.g. see [anything in LinearAlgebra](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.ZeroPivotException). That's next on my list to look into, but goes a bit out of scope of this one, which just aims to fix the regression where it now throws and kills Documenter.

PR #1970, opened 2022-10-18
Expand code-like blocks in nested contexts

ettersi:
Fixes https://github.com/JuliaDocs/Documenter.jl/issues/491. This PR is heavily inspired by the previous work in https://github.com/JuliaDocs/Documenter.jl/pull/1637 and https://github.com/JuliaDocs/Documenter.jl/pull/1747

_Edit by @mortenpi: close #1747 ._


fredrikekre:
Can you add a note in the changelog?

ettersi:
Done!

ettersi:
Not sure why some tests fail, but I suspect it's caused by something outside the scope of this PR. The tests used to pass before the last two commits, and since then I've only changed the changelog.

src/Utilities/Selectors.jl
```diff
@@ -163,7 +163,7 @@ Selectors.dispatch(MySelector, args...)
 """
 function dispatch(::Type{T}, x...) where T <: AbstractSelector
     types = get!(selector_subtypes, T) do
-        sort(subtypes(T); by = order)
+        sort(leave_subtypes(T); by = order)
```
mortenpi, line 166:
```suggestion
        sort(leaf_subtypes(T); by = order)
```

src/Utilities/Selectors.jl
```diff
@@ -174,6 +174,28 @@ function dispatch(::Type{T}, x...) where T <: AbstractSelector
...
 end
 
+"""
+Return a list of all subtypes of `T` which do not have further subtypes.
+
+The returned list includes subtypes of subtypes, and it does not distinguish
+between concrete types (i.e. types which are guaranteed not to have subtypes)
+and abstract types (which may or may not have subtypes).
+"""
+function leave_subtypes(::Type{T}) where T
```
mortenpi, line 184:
```suggestion
function leaf_subtypes(::Type{T}) where T
```

mortenpi (commented):
This looks great! Just a couple of tiny nitpicks.

Could you add the new example page to `pages` here: https://github.com/JuliaDocs/Documenter.jl/blob/998d26e568a20e3b1764d0c2cef30ebe4d2563ae/test/examples/make.jl#L181-L213 ?

The test failures indeed look unrelated.

mortenpi:
I can replicate the test failures locally and outside of Julia as well, and it looks like it's this Ubuntu bug: https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586

ettersi:
Done

mortenpi:
Thanks a lot @ettersi!

ettersi:
Thank you! Documenter is a great tool, and I'm very grateful to everyone who has helped shape it into what it is today!

ettersi:
Btw, is there a rough guesstimate for when the next version will be released?

mortenpi:
> Btw, is there a rough guesstimate for when the next version will be released?

Not really, no. I did some thinking, and I would like to get [a few more things into 0.28](https://github.com/JuliaDocs/Documenter.jl/milestone/107). People get somewhat annoyed with breaking Documenter releases, so I would like batch up as many (breaking or potentially breaking) things as we can.

However, a good thing about Documenter deployments is that you can just commit a `Manifest.toml` and run with `Documenter#master` for your docs. There is no particular need to depend on released versions.

PR #1989, opened 2022-12-06
Support regex => substitution pairs in doctest filters

fredrikekre:
This patch adds support for using regex => substitution pairs in doctest filters. This is useful when you do not want to replace the full match, but only remove some part. As an example, consider a test where the output may differ in the third decimal:

````
```jlddoctest
julia> 1.124
1.123
```
````

With this patch it is still possible to test the first 2 decimals using the filter `r"(\d\.\d\d)\d*" => s"\1"`.


mortenpi:
You can probably steal some tests from #1271 for this. That PR has some refactoring and more extensive docs, but I think we should merge the feature here as is, and I can revisit the other PR afterwards.

fredrikekre:
Oh wow, wasn't aware of that PR -- we even used the same new variable names :D

Issue #1321, opened 2020-05-13
Generate Meta Tag


logankilpatrick:
Documenter currently does not generate an html Meta tag, which it optionally should: https://web.dev/meta-description/?utm_source=lighthouse&utm_medium=devtools

mortenpi:
Easy to implement, but the question is what tags with what content. Or do you specifically mean to have a way for the user to populate a `<meta name="description">` tag?

Also, possibly related: #1145, #1184

ztangent:
I created a [(hacky) workaround](https://gist.github.com/ztangent/282885ed7d44a88f23c73b5e2e48d4fe) for this issue today, so that the [PDDL.jl documentation](https://juliaplanners.github.io/PDDL.jl/dev/) shows a nice image preview on social media.

It might be a starting point for changes to Documenter.jl itself! Though ideally, I think default meta tag values should be automatically generated, possibly based on the specific contents on of each page, and I'm not sure how best to do that.

mortenpi:
If someone could list the `<meta>` tags and values we want, implementing them here would be easy.

ztangent:
Here's the functionality I would really want to ensure that social media / search engine / Slack link previews are nicely displayed and customizable:

1. Ability to specify a preview image URL (maybe as a relative asset link), that auto-populates the `og:image` and `twitter:image` meta tags.
    - If the preview image is specified, the `twitter:card` tag should also be have the value `"summary_image_large"`.
2. Auto-populate `title`, `og:title` and `twitter:title` with the same `page_title` value that gets used in the `<title>` tag.
    - Maybe this can optionally be overridden by a user-specified string, but then the value would end up being constant across all pages in the documentation, which is not ideal.
3. Ability to specify the `description`, `og:description` and `twitter:description` meta tags with a string.
    - Maybe this can be auto-generated for each page by default, but I'm not sure how best to do that -- maybe the first few sentences of the page contents?
4. Auto-populate the `og:url` and `twitter:url` metatags with the URL of that page.
    - This one is less important, because both Twitter and Facebook automatically infer it without causing issues with the preview.
    
I'm using https://metatags.io/ as a reference for what meta tags to support. I think one design question here is how much Documenter.jl should be hard-coded to support meta tags for specific social media providers, (e.g. Facebook's OpenGraph, or Twitter), versus be agnostic about the social media provider.

[...]

PR #1991, opened 2022-12-11
Add support for generating meta tags.

ztangent:
This PR addresses issue #1321 by adding support for generated meta tags.

As discussed in that issue, OpenGraph and Twitter meta-tags are generated to enable link previews. The following functionality is provided:

1. Ability to specify a preview image, that auto-populates the `og:image` and `twitter:image` meta tags.
    - This is done by creating `preview.{png,webp,gif,jpg,jpeg}` in the `assets` directory. Because meta tags need to link an absolute URL, the canonical URL for the image is constructed from the `canonical` option in the `HTML` constructor. 
2. Auto-populate title, `og:title` and `twitter:title` with the same page_title value that gets used in the `<title>` tag.
3. Ability to specify the description, `og:description` and `twitter:description` meta tags with a string.
    - A site-wide description can be specified as a keyword argument to `makedocs`, and can be over-ridden in a page-specific manner using the `Description` keyword in an `@meta`-block.
5. Auto-populate the `og:url` and `twitter:url` metatags with the canonical URL of that page, as long as `canonical` is set.

Documentation has been updated accordingly. The only thing I haven't done is write test cases for this functionality, because I'm not quite sure where to add them or how to test for the existence of the correct meta-tags in some generated HTML file. Advice would be appreciated!


oxinabox:
Can we fallback to the logo if the preview image isn't found?

ztangent:
@mortenpi suggested not having a fallback because the preview doesn't look the best (see https://github.com/JuliaDocs/Documenter.jl/issues/1321#issuecomment-1105919028), but it'd be an easy change either way!

src/makedocs.jl
```diff
@@ -106,6 +106,9 @@ is enabled by default.
 
 **`sitename`** is displayed in the title bar and/or the navigation menu when applicable.
 
+**`description`** is the site-wide description that displays in page previews and search
+engines. Defaults to "Documentation for \$sitename".
```
fredrikekre, line 110: Since this is specific to the HTML output (?) perhaps it should be passed in `format = HTML(...)` like some of the other HTML-specific options.

mortenpi: Agreed. We can always promote it to a `makedocs` argument later, and have `HTML` default to the `makedocs` value.

src/html/HTMLWriter.jl
```diff
@@ -921,18 +977,11 @@ function warning_script(src, ctx)
...
-   if canonical_link === nothing
-      return Tag(Symbol("#RAW#"))("")
-   else
-      canonical_link_stripped = rstrip(canonical_link, '/')
-      href = "$canonical_link_stripped/$src"
-      return link[:rel => "canonical", :href => href]
-   end
+ 
+function empty_node()
+    return Tag(Symbol("#RAW#"))("")
```
mortenpi, line 984: This is a nice to have, but to make things a little neater here: do you think we could move this into the `DOM` module? We could have it be just a global `const VOID = Tag(Symbol("#RAW#"))("")` I think (as for the name, there is some precedent in HTML with "void elements").

We could also slightly modify the `@tags` macro to allow bringing `VOID` into the local context.

The `DOM` module has [some unit tests](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/dom.jl) which could be used to make sure it all works as intended as well.

src/html/HTMLWriter.jl
```diff
@@ -850,13 +862,30 @@ function render_head(ctx, navnode)
...
+
+        # Description meta tags
+        meta[:name => "description", :content => description],
+        meta[:property => "og:description", :content => description],
+        meta[:property => "twitter:description", :content => description],
+
+        # Canonical URL tags
+        isnothing(canonical) ? empty_node() : meta[:property => "og:url", :content => canonical],
+        isnothing(canonical) ? empty_node() : meta[:property => "twitter:url", :content => canonical],
+        isnothing(canonical) ? empty_node() : link[:rel => "canonical", :href => canonical],
```
mortenpi, line 880: One option here would be to assign these tags into a 

```
meta_urls = isnothing(canonical) ? [VOID] : [
    meta[...], ...
]
```

and them just splat that here. Either way is fine though.

mortenpi (commented):
Other than changing it to be an argument to `HTML`, this looks great as is. Everything else is just bikeshedding.

One last thing that would be nice is to have a [couple of small `occursin` tests](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests.jl), mostly to make sure that the conditional behaviors are handled correctly.

ztangent:
Made all of the suggested changes! 

I can also add this test preview image to the real documentation for Documenter.jl if you'd like:

preview

(*Note that it will only start showing up as a preview image with the next stable release of Documenter.jl, due to how the preview image URL is constructed from the canonical URL.)

ztangent:
Have tested that this actually works with the documentation site for PDDL.jl: https://juliaplanners.github.io/PDDL.jl/dev/

You can enter the URL above into [metatags.io](https://metatags.io/) to check, or just inspect the meta tags of the generated docs for `dev`.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 * ![Enhancement][badge-enhancement] The at-ref links are now more flexible, allowing arbitrary links to point to both docstrings and section headings. ([#781][github-781], [#1900][github-1900])
 * ![Enhancement][badge-enhancement] Code blocks like `@example` or `@repl` are now also expanded in nested contexts (e.g. admonitions, lists or block quotes). ([#491][github-491], [#1970][github-1970])
 * ![Enhancement][badge-enhancement] The new `pagesonly` keyword to `makedocs` can be used to restrict builds to just the Markdown files listed in `pages` (as opposed to all `.md` files under `src/`). ([#1980][github-1980])
+* ![Enhancement][badge-enhancement] Search engine and social media link previews are now supported by generating the relevant HTML `meta` tags. ([#1321][github-1321], [#1991][github-1991])
```
mortenpi, line 34:
```suggestion
* ![Enhancement][badge-enhancement] Search engine and social media link previews are now supported, with Documenter generating the relevant HTML `meta` tags. ([#1321][github-1321], [#1991][github-1991])
```

mortenpi (approved):
LGTM. 

> I can also add this test preview image to the real documentation for Documenter.jl if you'd like:

I you don't mind, that would be awesome. We can merge as soon as that is done.

ztangent:
Done!

mortenpi:
Thanks!

PR #1992, opened 2022-12-12
Add a few examples from SlimGroup

mloubout:
Packages from [slimgroup](https://github.com/orgs/slimgroup) that uses documenter.

Thanks for the great work


mortenpi:
Thanks!

Issue #1291, opened 2020-04-13
Release builds for packages in subdirectories


fredrikekre:
Documenter needs support for packages in a subdirectory of a repo, ref https://github.com/JuliaRegistries/Registrator.jl/issues/230

mortenpi:
Is there anything else other than making sure that we interpret the tags correctly?

PR #1993, opened 2022-12-13
Support custom tag prefix for release docs

### LINKED ISSUE #1792: Support custom tags for release docs
If I have a subpackage in a repo (possibly more than one), I want to be able to build releases for a package based on tags like `MyPackage-v0.1.0` instead of just `v0.1.0` (since different packages might need different tags).

I thought this is what `versions` did in `deploydocs`, e.g.
```julia
versions = ["stable" => "MyPackage-v^", "MyPackage-v#.#", "dev" => "dev"])
```
but it's not. That's just for mapping from the words in the version-selector to the tags, but not about which tags it will actually deploy on. (Thanks @fredrikekre for clarifying this).

Right now, Documenter [uses](https://cs.github.com/JuliaDocs/Documenter.jl/blob/d2ed9b8b171e487835f456524e52799aa9da1d15/src/deployconfig.jl#L427) `Base.VERSION_REGEX` for determining if a tag is a version:
```julia
julia> Base.VERSION_REGEX
r"^
    v?                                      # prefix        (optional)
    (\d+)                                   # major         (required)
    (?:\.(\d+))?                            # minor         (optional)
    (?:\.(\d+))?                            # patch         (optional)
    (?:(-)|                                 # pre-release   (optional)
    ([a-z][0-9a-z-]*(?:\.[0-9a-z-]+)*|-(?:[0-9a-z-]+\.)*[0-9a-z-]+)?
    (?:(\+)|
    (?:\+((?:[0-9a-z-]+\.)*[0-9a-z-]+))?    # build         (optional)
    ))
$"ix

julia> occursin(Base.VERSION_REGEX, "MyPackage-v0.1.0")
false
```
But if we remove the `^` at the start,
```julia
julia> r = r"v?                                      # prefix        (optional)
           (\d+)                                   # major         (required)
           (?:\.(\d+))?                            # minor         (optional)
           (?:\.(\d+))?                            # patch         (optional)
           (?:(-)|                                 # pre-release   (optional)
           ([a-z][0-9a-z-]*(?:\.[0-9a-z-]+)*|-(?:[0-9a-z-]+\.)*[0-9a-z-]+)?
           (?:(\+)|
           (?:\+((?:[0-9a-z-]+\.)*[0-9a-z-]+))?    # build         (optional)
           ))
       $"ix;

julia> occursin(r, "MyPackage-v0.1.0")
true
````
it will match.

I don't think `version_tag_strip_build` will require any other changes to keep working, but I don't know the broader implications of such a change.

(Well, one is multiple subpackages could deploy on the same tags, so one might need to ensure only the right package builds on the right tag, e.g. by using the `tags` deploy trigger on GHA to specify a pattern)

### PR DESCRIPTION
**Pull request by hannahilea**
Implement #1792. Is non-breaking.

- [x] support tag_prefix:
  - [x] in `version_tag_strip_build`
    - [x] tests
  - [x] thread through `deploy_folder`
    - [x] tests
  - [x] thread through `deploydocs`
    - [x] tests
- [x] docstrings updated
- [x] docs (examples) updated
  - [x] add test for monorepo behavior as described in docs
- [x] do lint pass
- [x] update changelog

_Edit by @mortenpi: close #1792, close #1291_

### CODE REVIEW

**Comment by hannahilea** [comment]
wip - next step: update/redo current pr's approach to `deploydocs` to save in _subdir_ and not prefix links

**Comment by hannahilea** [comment]
This is my first PR for this repo, so let me know if there are additional tests/docs/etc that I should include here!

This implementation is non-breaking, so could potentially be included in (or back-ported to) a v0.27 patch. Since the changelog shows the next unreleased version as 0.28.0 (and I see that this issue has been added to the [0.28.0 milestone](https://github.com/JuliaDocs/Documenter.jl/milestone/107)) I've only added it to the changelog's 0.28.0 section.

Also, from #1792 I considered

> Another possibility would be to allow the user to pass a full regex, requiring it to have one capture group which would have to capture the pure version number (i.e. something like tag_regex = r"^MyPackage-([0-9-+\.]+)$"; optionally with a named capture group). More general, but at the same time more complicated for the user.

I started with the basic prefix option, with the idea that once I got a hang for the structure of the codebase I'd know how complicated the regex option would be to plumb through. After adding the basic prefix option, though, I think that it is sufficient for the original motivating feature request that @ericphanson described (in regards to our particular monorepo). I also think the regex option could introduce complicated implementation around versioned tags that maybe isn't worth taking on until there's a motivating use case without a simpler solution. (I suspect that it would likely be relevant for a non-versioned tag workflow.)

File: docs/src/man/hosting.md [hunk 1]
```diff
@@ -535,6 +537,90 @@ Preview builds are still deployed to the `previews` subfolder.
...
+```
+# In ./docs/make.jl
+deploydocs(; repo = "github.com/USER_NAME/PACKAGE_NAME.jl.git",
+            # ...any additional kwargs
+            )
+
+# In ./PackageA.jl/docs/make.jl
+deploydocs(; repo = "github.com/USER_NAME/PACKAGE_NAME.jl.git",
+             dirname="PackageA",
+             tag_prefix="PackageA-",
```
fredrikekre, line 576: I haven't followed TagBot lately. Does TagBot generate tags like this by default, or is it configurable on that end too?

fredrikekre:
Closes #1291.

docs/src/man/hosting.md
ericphanson: I don't think TagBot supports subdir packages at all (xref https://github.com/JuliaRegistries/TagBot/issues/157), but this PR would allow manual tags at least. Ideally tagbot could be updated too, although @christopher-dG labelled that issue with "rewrite" so maybe it would need a lot of work.

christopher-dG: Correct, tbh if that feature is added (whether in a full rewrite or not) it is unlikely that I will be the one who does it.

mortenpi:
> I'm not sure why the changelog CI is failing

You just need to add the Markdown link definition to the PR at the bottom of the file.

hannahilea:
thanks! added.

docs/src/man/hosting.md
hannahilea: Opened a PR to add this feature: https://github.com/JuliaRegistries/TagBot/pull/252/files

(I thought it would be a fun way to learn how GHA are written, and instead fought to remember python. you win some, you lose some!)

mortenpi (approved):
This looks excellent! I presume it's good to go?

Do you happen to have an example run somewhere with the branch?

hannahilea:
> Do you happen to have an example run somewhere with the branch?

Nothing public facing, so I started an example public monorepo (https://github.com/hannahilea/ExampleMonorepo.jl); I ran out of time to try it on this branch but should have the example running tomorrow. Will update when that's up!

hannahilea:
> Do you happen to have an example run somewhere with the branch?

Here's an example monorepo using this Documenter branch---note that each package (ExampleMonorepo.jl, SubpackageA.jl, and SubpackageB.jl) was each tagged with a separate version so that it's clear that each docs version is correct:
- Top level (manual git tag: `v0.4.0`): https://hannahilea.github.io/ExampleMonorepo.jl/stable/
- SubpackageA.jl (manual git tag: `SubpackageA-v0.2.0`): https://hannahilea.github.io/ExampleMonorepo.jl/SubpackageA/stable/
- SubpackageB.jl (manual git tag: `SubpackageA-v0.3.0`): https://hannahilea.github.io/ExampleMonorepo.jl/SubpackageB/stable/

mortenpi:
That looks awesome. Thanks a lot!

PR #1995, opened 2022-12-13
Allow automated test missingdocs

KeithWM:
We wish to use the missingdocs functionality in a CI/CD environment. For this, it is useful to have the missingdocs function not only print out missing documentation, but also return the number of missing documents. This can then be verified using a `@test missingdocs(...) == 0`.


mortenpi:
LGTM, thanks! Something to note here though is that `missingdocs` is not actually part of Documenter's API. But it's also unlikely to go anywhere or change it's semantics any time soon.

Issue #2010, opened 2023-01-08
Section titles with [`foo`](@id foo) fail to render


odow:
Reported by @ParadaCarleton in JuMP: https://github.com/jump-dev/JuMP.jl/issues/3179

On the latest `master` commit:

## index.md

```md
# [A `@foo` title](@id foo)

Hello. World!
```

## make.jl 

```julia
import Documenter
Documenter.makedocs(sitename = "test", pages = ["index.md"])
```

## Render

<img width="1262" alt="image" src="https://user-images.githubusercontent.com/8177701/211222040-93a1028d-37d9-440e-b816-565a1a073496.png">

I'll take a look at this, unless anyone has an immediately obvious idea what the problem is.

odow:
Also happens in the LaTeXWriter:

<img width="657" alt="image" src="https://user-images.githubusercontent.com/8177701/211222200-1f8cb27c-3def-4710-b029-f7a0bdc94861.png">

odow:
Fix incoming

PR #2011, opened 2023-01-08
Fix expanding of x-ref targets containing a code block

odow:
Took me a while to find the right place, but I think this is the correct fix. 

It renders the example from https://github.com/JuliaDocs/Documenter.jl/issues/2010 okay.

image

Not sure how/where to test this?

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -277,9 +277,7 @@ function Selectors.runner(::Type{TrackHeaders}, node, page, doc)
             # the tree.
             link_node = first(node.children)
             MarkdownAST.unlink!(link_node)
-            for child in collect(link_node.children)
-                push!(node.children, child)
-            end
+            append!(node.children, link_node.children)
             match(NAMEDHEADER_REGEX, link_node.element.destination)[1]
         else
             # TODO: remove this hack (replace with mdflatten?)
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -277,7 +277,9 @@ function Selectors.runner(::Type{TrackHeaders}, node, page, doc)
             # the tree.
             link_node = first(node.children)
             MarkdownAST.unlink!(link_node)
-            append!(node.children, link_node.children)
+            for child in collect(link_node.children)
+                push!(node.children, child)
+            end
             match(NAMEDHEADER_REGEX, link_node.element.destination)[1]
         else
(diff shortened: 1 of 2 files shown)


mortenpi:
I think this should actually be fixed in https://github.com/JuliaDocs/MarkdownAST.jl/pull/16. I'll tag that shortly, and then we can check if that fixes #2010.

odow:
The release of MarkdownAST didn't fix the problem, because there's also this method I just found:

https://github.com/JuliaDocs/Documenter.jl/blob/354b0f936fe7cdc99a579bc4d83c16eca233aa46/src/Expanders.jl#L766-L771

mortenpi:
I'd also merge the test changes of this PR.

odow:
> I'd also merge the test changes of this PR.

How do I test it though? You kind need some reference HTML to compare against. This won't throw an error, it'll just render the wrong thing.

mortenpi:
Ah, no, I just meant that let's leave this case in the example file. No need to test if it actually gets rendered properly. Ideally we'd test the rendering, but we don't really have the infrastructure for it at the moment.

test/examples/src/xrefs.md
```diff
@@ -14,3 +14,5 @@ Named x-refs:
 ## X-ref target
 
 ## [X-ref target with id](@id xreftarget)
+
+## [x-ref with `@code` block](@id xrefcodeblock)
```
mortenpi, line 18:
```suggestion
## [x-ref with `@code` block](@id xrefcodeblock)

This should render as ```x-ref with `@code` block```: #2010, #2011.
```

src/Expanders.jl
```diff
@@ -277,7 +277,9 @@ function Selectors.runner(::Type{TrackHeaders}, node, page, doc)
             # the tree.
             link_node = first(node.children)
             MarkdownAST.unlink!(link_node)
-            append!(node.children, link_node.children)
+            for child in collect(link_node.children)
```
odow, line 280:
```suggestion
            append!(node.children, link_node.children)
```

src/Expanders.jl
```diff
@@ -277,7 +277,9 @@ function Selectors.runner(::Type{TrackHeaders}, node, page, doc)
...
+                push!(node.children, child)
```
odow, line 281:
```suggestion
```

mortenpi:
Thanks!

PR #2012, opened 2023-01-09
catch existing symlinks when adding new ones

jkrumbiegel:
`ispath(existing_symlink)` returns `false` if the target of the symlink doesn't exist anymore, but `symlink(target, existing_symlink)` errors with `file already exists`. I had this problem after I deleted something that the `stable` symlink pointed to.


fredrikekre (approved):

mortenpi:
Thanks!

PR #2014, opened 2023-01-10
Remove pirated method from Expanders.jl

odow:
Closes https://github.com/JuliaDocs/Documenter.jl/pull/2011

We should potentially update the lower bound on MarkdownAST to 0.1.1 as well?


mortenpi (approved):
Good catch!

PR #2018, opened 2023-01-17
Explicitly avoid committing SSH key to gh-pages

mortenpi:
Rather than writing the key to the repository, let's write it to `homedir()`. I think ideally we would avoid writing a temporary file altogether, but I think the only way to do it then is with `ssh-agent`?

cc @simeonschaub


src/deploydocs.jl
```diff
@@ -379,14 +379,16 @@ function git_push(
...
+        keyfile = abspath(joinpath(homedir(), ".documenter-identity-file.tmp"))
+        ispath(keyfile) && error("Keyfile not cleaned up from last run: $(keyfile)")
         try
             if is_preview
                 keycontent = documenter_key_previews(deploy_config)
             else
                 keycontent = documenter_key(deploy_config)
             end
             write(keyfile, base64decode(keycontent))
+            chmod(keyfile, 0o600) # user-only rw permissions
```
fredrikekre, line 391: This is also happening outside the try, remove from there?

src/deploydocs.jl
```diff
@@ -379,14 +379,16 @@ function git_push(
         # Get the parts of the repo path and create upstream repo path
         user, host, upstream = user_host_upstream(repo)
 
-        keyfile = abspath(joinpath(root, ".documenter"))
+        keyfile = abspath(joinpath(homedir(), ".documenter-identity-file.tmp"))
```
fredrikekre, line 382: Better with temp?
```suggestion
        keyfile = abspath(joinpath(mktempdir(), ".documenter-identity-file.tmp"))
```

simeonschaub: Yeah, I think it's not inconceivable someone might try to upload the entire home directory, e.g. on a CI service where there's nothing much there anyways. If we wanted to be on the safe side, maybe even check if `tempdir` is a subdirectory of the target and error loudly if it is?

fredrikekre:
Good idea to add the file to `$GIT_DIR/info/exclude` (or perhaps use `git add -A -- :!.documenter-identity-file.tmp`)

src/deploydocs.jl
mortenpi: `tempdir` would indeed be the most convenient, but what I was concerned about is that, if somehow it doesn't get cleaned up, it's better to leave the file into `/home` (where hopefully only the current user can read it), rather than `/tmp`. Especially if you're maybe running Documenter on 

Checking for the subdirectory seems reasonable though.

src/deploydocs.jl
mortenpi: Good catch, I missed that.

src/deploydocs.jl
fredrikekre: Could the file not be created as read-write only for current user, then after the key is written, change to read-only for current user?

mortenpi: I guess the permissions are fine already (user-only rw protects as much). I guess my main concern is some shared system where you might have the keyfile sitting around in the `/tmp` of some random node. In your home directory you might be more likely to notice it.

simeonschaub (approved):
Thanks for the quick fix! Not very familiar with Documenter internals, but this seems like the right approach to me

src/deploydocs.jl
```diff
@@ -357,7 +365,7 @@ function git_push(
         end
 
         # Add, commit, and push the docs to the remote.
-        run(`$(git()) add -A .`)
+        run(`$(git()) add -A -- ':!.documenter-identity-file.tmp' ':!**/.documenter-identity-file.tmp'`)
```
mortenpi, line 368: @fredrikekre Does this look right to you? `:!.documenter-identity-file.tmp'` seemed to only exclude the top-level file.

mortenpi:
Okay, I kinda changed my mind now. I think checking that tmpdir/homedir is a subdirectory of `target` is not too useful, since it doesn't catch most bad cases. What we might want to do is to check that `target` is a subdirectory of `root` though.

On the other hand, the `git` solution seems to nicely avoid pushing the key file, so I changed back to writing it into `root` (i.e. `docs/` in most cases). I think between `homedir()` and `root`, the latter is preferred.

src/deploydocs.jl
fredrikekre: Looks like that works, but I wonder why the first isn't sufficient. In a standard `.gitignore` file a line with `.documenter-identity-file.tmp` would exclude all files with that name.

fredrikekre (approved):

PR #2033, opened 2023-02-06
Refactor CHANGELOG.md to the Keep a Changelog style

odow:
There has been some discussion on Slack about package changelogs.

The general consensus was that following https://keepachangelog.com/en/1.0.0/ is good practice.

Existing examples:

 * @fredrikekre's https://github.com/fredrikekre/Literate.jl/blob/master/CHANGELOG.md
 * JuMP's https://jump.dev/JuMP.jl/dev/release_notes/

The JuMP release notes are slightly complicated because they are built on-demand from a simplified file:
 * https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/changelog.md
 * https://github.com/jump-dev/JuMP.jl/blob/451740ada26f634edb09cc5ec19b2c4d48031cfe/docs/make.jl#L214-L249

I've followed the JuMP approach here. Open to suggestions/discussions for what we could do differently, and how this could be incorporated more directly into Documenter.

cc @giordano

I guess the preview won't push, so here's what my local copy looks like:
image

I haven't gone through and looked up the date each release was tagged, but that would be good to have.

Ideally, we'll aim for something standard here so that other packages can copy.

Perhaps @mortenpi can take this branch over and push it to the proper remote so that we have preview.


fredrikekre:
I think having a CHANGELOG.md file at the repo root is pretty useful.

odow:
I've moved the `CHANGELOG.md` back to the repo root: https://github.com/odow/Documenter.jl/blob/od/changelog/CHANGELOG.md

odow:
Closing in favor of #2035 which is now on the `JuliaDocs` remote.

PR #2035, opened 2023-02-07
Refactor CHANGELOG.md to the Keep a Changelog style

odow:
I'll add the preview link once it runs: https://documenter.juliadocs.org/previews/PR2035/release_notes/

There has been some discussion on Slack about package changelogs.

The general consensus was that following https://keepachangelog.com/en/1.0.0/ is good practice.

Existing examples:

 * @fredrikekre's https://github.com/fredrikekre/Literate.jl/blob/master/CHANGELOG.md
 * JuMP's https://jump.dev/JuMP.jl/dev/release_notes/

The JuMP release notes are slightly complicated because they are built on-demand from a simplified file:
 * https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/changelog.md
 * https://github.com/jump-dev/JuMP.jl/blob/451740ada26f634edb09cc5ec19b2c4d48031cfe/docs/make.jl#L214-L249

I've followed the JuMP approach here. Open to suggestions/discussions for what we could do differently, and how this could be incorporated more directly into Documenter.

cc @giordano

I haven't gone through and looked up the date each release was tagged, but that would be good to have.

Ideally, we'll aim for something standard here so that other packages can copy.


CHANGELOG.md
```diff
@@ -1,182 +1,313 @@
...
+```@meta
+CurrentModule = Documenter
+```
+
+# Release notes
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Version `v0.28.0` (unreleased)
```
odow, line 10:
```suggestion
## Unreleased
```

CHANGELOG.md
```diff
@@ -346,241 +517,317 @@
...
+### Breaking
 
-* ![BREAKING][badge-breaking] `DocTestSetup`s that are defined in `@meta` blocks no longer apply to doctests that are in docstrings. ([#774][github-774])
+* `DocTestSetup`s that are defined in `@meta` blocks no longer apply to doctests that are in docstrings. (#774)
 
   - Specifically, the pattern where `@docs` or `@autodocs` blocks were surrounded by `@meta` blocks, setting up a shared `DocTestSetup` for many docstrings, no longer works.
 
   - Documenter now exports the `DocMeta` module, which provides an alternative way to add `DocTestSetup` to docstrings.
 
   **For upgrading:** Use `DocMeta.setdocmeta!` in `make.jl` to set up a `DocTestSetup` that applies to all the docstrings in a particular module instead and, if applicable, remove the now redundant `@meta` blocks. See the ["Setup code" section under "Doctesting"](https://juliadocs.github.io/Documenter.jl/v0.23.0/man/doctests/#Setup-Code-1) in the manual for more information.
```
odow, line 748:
```suggestion
  **For upgrading:** Use `DocMeta.setdocmeta!` in `make.jl` to set up a `DocTestSetup` that applies to all the docstrings in a particular module instead and, if applicable, remove the now redundant `@meta` blocks. See the ["Setup code" section under "Doctesting"](https://documenter.juliadocs.org/v0.23/man/doctests/#Setup-Code-1) in the manual for more information.
```

docs/make.jl
```diff
@@ -10,6 +10,43 @@ if haskey(ENV, "DOCSARGS")
...
+    return line
+end
+
+open(joinpath(dirname(@__DIR__), "CHANGELOG.md"), "r") do in_io
+    open(joinpath(@__DIR__, "src", "release_notes.md"), "w") do out_io
+        for line in readlines(in_io; keep = true)
+            write(out_io, fix_release_line(line))
+        end
+    end
+end
```
mortenpi, line 48: Everything above this is also duplicated in JuMP docs, right? Should we create a `Changelog.jl`?

odow: The exact regexes are slightly different. But yeah, we could standardize and then add this as a function to Documenter. I don't think it needs to be a package.

odow:
The other issue is that if you click "Edit" in the documentation, it tries to redirect you to https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/release_notes.md. Is it possible to change the edit link for a particular page?

mortenpi:
> Is it possible to change the edit link for a particular page?

We should generate an at-meta block with an `EditURL` in `release_notes.md`.

docs/make.jl
mortenpi: You're right, this is quite Documenter-specific right now. We could drop it into Documenter as an undocumented feature right now (we might want to iterate on the exact form of the public API, but let's not do it as part of this PR).

CHANGELOG.md
```diff
@@ -1,182 +1,313 @@
-# Documenter.jl changelog
+```@meta
+CurrentModule = Documenter
+```
```
mortenpi, line 3: Could we add this in `make.jl`?

odow: Done

docs/make.jl
odow: I think we can have some standard rewrites:

 * `(abc#XXXX)` -> `([abc#XXXX](github.com/abc/issue/XXXX))`
 * `(#XXXX)` -> `([#XXXX](github.com/default_url/issue/XXXX))`
 * `## vX.Y.Z` -> `## [vX.Y.Z](url/releases/tag/vX.Y.Z)`

JuMP also has

 * `(@abc)` -> `([@abc](github.com/abc))`

The main issue is the formatting, e.g., do we expect `## Version vX.Y.Z`, `## vX.Y.Z` or `## Version vX.Y.Z`?

mortenpi: I do kind of like having the issue references working in `CHANGELOG.md` as well though (like we do right now here). Unfortunately, it doesn't look like GitHub has any syntax in Markdown files that makes these links work, so one would need to maintain reference links.

I suppose one approach would be to generate both `CHANGELOG.md` and `release_notes.md` from the same source file (and have CI make sure `CHANGELOG.md` is up to date). The rewriting rules look good to me.

Mainly just thinking about loud here though -- this all adds complexity.

> e.g., do we expect ## Version vX.Y.Z, ## vX.Y.Z or ## Version vX.Y.Z

I think it might be good if the version heading line is a bit longer -- so something like `# Version [v1.1.0](...) - 2019-02-15` seems reasonable to me?

odow: Losing the links in CHANGELOG is a downside. But then again... they are in the online documentation which most users would look at.

I considered options like keeping CHANGELOG in the proper format, and having a script that rebuilt it in-place, but that just added more work for the maintainer because you'd have to remember to periodically rebuild it. You might as well just force people writing the changelog to add all the links.

A benefit of this approach is that you can also reference docstrings and stuff in the CHANGELOG because it's now part of Documenter. JuMP does this, so when we add a new function, clicking on it takes the user to the docstring;

image

odow:
Added the dates with

```
$ git log --tags --simplify-by-decoration --pretty="format:%(describe:tags=yes) - %cs
```

```Julia
tags ="""
# Copy-pasted output from git script above
"""
file = read("CHANGELOG.md", String);
for line in split(tags, "\n")
    tag, date = String.(split(line, " - "))
    file = replace(file, "## Version $tag\n" => "## Version $tag - $date\n")
end
write("CHANGELOG.md", file);
```

docs/make.jl
mortenpi: Yep, I think that's fair. Let's stick to this approach.

docs/make.jl
```diff
@@ -10,6 +10,84 @@ if haskey(ENV, "DOCSARGS")
...
+                write(out_io, fix_release_line(line; repo = repo))
+            end
+        end
+    end
+    return
+end
+
+rewrite_changelog(
+    changelog_filename = joinpath(dirname(@__DIR__), "CHANGELOG.md"),
+    release_notes_filename = joinpath(@__DIR__, "src", "release_notes.md"),
```
mortenpi, line 85:
```suggestion
    release_notes_filename = joinpath(@__DIR__, "src", "release-notes.md"),
```

docs/make.jl
```diff
@@ -63,6 +141,7 @@ makedocs(
             ),
         ],
         "contributing.md",
+        "release_notes.md",
```
mortenpi, line 144:
```suggestion
        "release-notes.md",
```

.gitignore
```diff
@@ -17,3 +17,4 @@ docs/dev/
 docs/build/
 docs/build-pdf/
 docs/site/
+docs/src/release_notes.md
```
mortenpi, line 20:
```suggestion
docs/src/release-notes.md
```

docs/make.jl
```diff
@@ -10,6 +10,84 @@ if haskey(ENV, "DOCSARGS")
...
+        line = replace(line, m.match => "([$new_repo#$id](https://github.com/$new_repo/issues/$id))")
+    end
+    # Rule: (#XXXX) -> ([#XXXX](https://github.com/url/issue/XXXX))
+    # Description: Replace issue/PR nnumbers with a link to the default repo
+    # Example: (#123) -> ([#123](https://github.com/JuliaDocs/Documenter.jl/issues/123))
+    # There is no need to distinguish between PRs and Issues because GitHub
+    # redirects.
+    while (m = match(r"\(\#([0-9]+)\)", line)) !== nothing
+        id = m.captures[1]
+        line = replace(line, m.match => "([#$id](https://github.com/$repo/issues/$id))")
```
mortenpi, line 37: Documenter actually handles references issue references to the current repo, so this should be sufficient:

```suggestion
        line = replace(line, m.match => "([#$id](#$id))")
```

mortenpi (commented):
Alright, I think this is basically good to go. I'll just make a couple of tiny modifications.

mortenpi:
Alright, let's get this merged. We can do any code organizational changes for re-use in a separate PR. Thanks a lot @odow, especially for taking the time to convert the existing changelog!

odow:
This looks good: https://documenter.juliadocs.org/dev/release-notes/. We should definitely encourage packages to adopt the style.

PR #2038, opened 2023-02-13
Bump checkout to v3 and mention needed permissions

JeffFessler:
My cleanup job kept failing and it took me a while to realize it was a repo permissions issue.
Hopefully having a brief note here will help others (and my future self).

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -417,16 +417,9 @@ jobs:
 
 _This workflow was taken from [CliMA/ClimaTimeSteppers.jl](https://github.com/CliMA/ClimaTimeSteppers.jl/blob/0660ace688b4f4b8a86d3c459ab62ccf01d7ef31/.github/workflows/DocCleanup.yml) (Apache License 2.0)._
 
-For this action to execute,
-you can give GitHub workflows write permissions
-under the rep settings, e.g.,
-https://github.com/<USER>/<REPO>.jl/settings/actions.
-Alternatively,
-you can modify the yaml file above
-to include the following lines before the `steps:` line
-to grant permissions
-per
-[GitHub Docs][https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#assigning-permissions-to-a-specific-job]:
+For this action to execute, you can give GitHub workflows write permissions under the rep settings, e.g. `https://github.com/<USER>/<REPO>.jl/settings/actions`.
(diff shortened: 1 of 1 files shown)


fredrikekre:
I believe it should be possible to set the permissions only for this job, see e.g. https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#assigning-permissions-to-a-specific-job

mortenpi:
Yes, I think this should be sufficient?

```yaml
permissions:
  contents: write
```

JeffFessler:
Ok, I updated the PR to list both options.
I have several of github actions that use GITHUB_TOKEN, like TagBot and CompatHelper and of course Documenter, so to me it seems easier just to give them all the write permissions using the repo settings.  But I agree it is also worth explaining how to do it in the way that applies only to this job.  If you prefer to emphasize the job-specific way, then we probably should update the example yaml file to include the permissions in it explicitly so that users can easily copy/paste that example and have it work.

docs/src/man/hosting.md
```diff
@@ -417,6 +417,22 @@ jobs:
...
+For this action to execute,
+you can give GitHub workflows write permissions
+under the rep settings, e.g.,
+https://github.com/<USER>/<REPO>.jl/settings/actions.
+Alternatively,
+you can modify the yaml file above
+to include the following lines before the `steps:` line
+to grant permissions
+per
+[GitHub Docs][https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#assigning-permissions-to-a-specific-job]:
```
mortenpi, line 429: Just formatting:

```suggestion
For this action to execute, you can give GitHub workflows write permissions under the rep settings, e.g. `https://github.com/<USER>/<REPO>.jl/settings/actions`.
Alternatively, you can modify the yaml file above to include the following lines before the `steps:` line to grant permissions per [GitHub Docs][https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#assigning-permissions-to-a-specific-job]:

```

mortenpi:
> If you prefer to emphasize the job-specific way, then we probably should update the example yaml file to include the permissions in it explicitly so that users can easily copy/paste that example and have it work.

Actually, if you don't mind, that would be good. In my opinion, in general, you want the additional permissions to be as close to where you are using them. So I think we should suggest `permissions:` by default, but can also mention the repo-wide settings as an alternative.

JeffFessler:
Done.  Feel free to reformat further if you really like those long lines.  The problem with long lines (IMHO) is that they lead to big diffs in the repo when anyone changes one small part of a long line.  See the current file diff where I changed "taken from" to "based on" - that tiny change causes a big diff.  But it's up to you :)

mortenpi:
LGTM, thanks!

PR #2056, opened 2023-02-23
Fix empty lines being generated in LaTeX output for hidden `@repl` entries

### LINKED ISSUE #1521: Empty line with #hide in at-repl block outputs
Originally reported by @knuesel in #1509: trying to hide an expression with `... # hide` in an at-repl block introduces a spurious newline. Especially annoying if you might want to do something like this:

````
```@repl
using LinearAlgebra # hide
A = rand(10, 10)
eigen(A)
```
````

### LINKED ISSUE #2054: `#hide` results in empty lines in code blocks in PDF output
For example, this:

```
` ` `@repl abstract_field
myComplexFunc(MyStructFieldParamRestricted(data)) # hide
myComplexFunc(MyStructFieldParamRestricted(lazy_data)) # hide
@time myComplexFunc(MyStructFieldParamRestricted(data))
` ` `
```

results in this:

image

I suspect either an issue in https://github.com/JuliaDocs/Documenter.jl/blob/b50d74bc02d6883ef36aa19f19129e7a2d675ea9/src/Expanders.jl#L774 or in LaTeXWriter (though I think the latter is less likely, since the same doesn't happen with `@example` blocks, only `@repl`).

### LINKED ISSUE #2399: Hidden leading lines in @repl blocks add an extra line
Using Docmenter `v1.2.1`, observe the difference in output between these three blocks:

    ```@example
    using .Iterators # hide
    reduce(*, repeated("foo ", 4))
    ```

    ```@repl
    using .Iterators # hide
    reduce(*, repeated("foo ", 4))
    ```

    ```@repl
    reduce(*, Iterators.repeated("foo ", 4))
    ```

The `@repl` block with `# hide` has an extra line before the prompt, it would be preferable if it weren't there.

### PR DESCRIPTION
**Pull request by Seelengrab**
Fixes #2054, Fixes #1521, Fixes #2399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -349,6 +349,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `deploydocs` now correctly handles version symlinks where the destination directory has been deleted. ([#2012])
 
+* Entries in `@repl` blocks that were hidden with `# hide` no longer produce erronous empy lines (#2054)
+
 ### Other
 
 * Documenter now uses [MarkdownAST](https://github.com/JuliaDocs/MarkdownAST.jl) to internally represent Markdown documents. While this change should not lead to any visible changes to the user, it is a major refactoring of the code. Please report any novel errors or unexpected behavior you encounter when upgrading to 0.28 on the [Documenter issue tracker](https://github.com/JuliaDocs/Documenter.jl/issues). ([#1892], [#1912], [#1924], [#1948])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Fixed
 
 * Don't require custom themes to set a color for the 'todo' admonition. ([#2576])
-* Entries in `@repl` blocks that were hidden with `# hide` no longer produce erroneous empty lines ([#2054], [#1521], [#2399])
+* Entries in `@repl` blocks that were hidden with `# hide` no longer produce erroneous empty lines ([#1521], [#2054], [#2399])
 
 ## Version [v1.9.0] - 2025-03-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Now the cursor remain focused on search box even after selecting the filter. ([#2410])
 
+* Entries in `@repl` blocks that were hidden with `# hide` no longer produce erroneous empty lines ([#2054], [#1521], [#2399])
+
 ## Version [v1.9.0] - 2025-03-17
 
 ### Added
@@ -348,8 +350,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * The default decision for whether to deploy preview builds for pull requests have been changed from `true` to `false` when not possible to verify the origin of the pull request. ([#1969])
 
 * `deploydocs` now correctly handles version symlinks where the destination directory has been deleted. ([#2012])
-
-* Entries in `@repl` blocks that were hidden with `# hide` no longer produce erronous empy lines (#2054)
 
 ### Other
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1678,6 +1678,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#1518]: https://github.com/JuliaDocs/Documenter.jl/issues/1518
 [#1519]: https://github.com/JuliaDocs/Documenter.jl/issues/1519
 [#1520]: https://github.com/JuliaDocs/Documenter.jl/issues/1520
+[#1521]: https://github.com/JuliaDocs/Documenter.jl/issues/1521
 [#1526]: https://github.com/JuliaDocs/Documenter.jl/issues/1526
 [#1527]: https://github.com/JuliaDocs/Documenter.jl/issues/1527
 [#1529]: https://github.com/JuliaDocs/Documenter.jl/issues/1529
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Expanders.jl [hunk 1]
```diff
@@ -839,8 +839,8 @@ function Selectors.runner(::Type{REPLBlocks}, node, page, doc)
 
         outstr = String(take!(out))
         # Replace references to gensym'd module with Main
-        outstr = remove_sandbox_from_output(outstr, mod)
-        push!(multicodeblock, MarkdownAST.CodeBlock("documenter-ansi", rstrip(outstr)))
+        outstr = remove_sandbox_from_output(outstr, mod) |> rstrip
```
mortenpi, line 842:
```suggestion
        outstr = rstrip(remove_sandbox_from_output(outstr, mod))
```

mortenpi (commented):
This seems reasonable to me. The bug also affects HTML as far as I can tell. Do you mind adding a quick CHANGELOG note?

mortenpi:
A few additional notes that go beyond this PR: I actually have quite a few question marks about the use of `#hide` in at-repl.

First, you can omit lines in multi-line prompts, which make the input and output inconsistent, e.g.:

````
```@repl
begin
    println("A")
    println("B") # hide
    println("C")
end
```
````

I think it's fine to allow that though, since maybe someone has a use case for it.

If you hide a whole single-line prompt, then we still print the output (if there is any). So

````
```@repl
println("A") # hide
println("B")
```
````

will show up as something like

```
A

julia> println("B")
B
```

This seems a bit counter-intuitive, and I think we should maybe hide the output as well when `isempty(input)`.

**Comment by mortenpi** [comment]
> I think the explicit use of `println` is a sort of edge case,

I wouldn't say so. At-repl is meant to emulate the REPL, including the stdout outputs. But regardless, you can also get the same inconsistency with return values:

````
```@repl
begin
    x = 1
    x += 1 # hide
    x
end
```
````

**Comment by Seelengrab** [comment]
> I think it's fine to allow that though, since maybe someone has a use case for it.

Turns out, yes I very much have a usecase for it :joy: What I'm currently doing is showing the nice pretty `@code_warntype` invocation, but in reality capturing that with a `redirect_stdout(devnull)` that's in a `# hide`. Then, I `code_warntype(IOBuffer()..)` the invocation again, so I can filter/shorten it a bit to make it more readable.

So removing that would be a big issue indeed :)

> I wouldn't say so. At-repl is meant to emulate the REPL,

I don't have a strong opinion on this :shrug: If `@repl` is meant to emulate the REPL and `# hide` is sort-of-kind-of like a trailing semicolon there, having `A` be printed is consistent:

```
julia> println("A");
A
```

It's a sideeffect after all, not the return value.

**Comment by mortenpi** [comment]
> If `@repl` is meant to emulate the REPL and `# hide` is sort-of-kind-of like a trailing semicolon there

`# hide` is not the semicolon though.. semicolon is the semicolon. `# hide`, I think, should be just a way to run extra code without showing the input nor the output to the user. Caveat being that in multi-line inputs things get weird, but it's up to the user to make sure they do the right thing then.

**Comment by Seelengrab** [comment]
@fingolfin Should be updated & hopefully work - I noticed that I hadn't added a test, is there some place I could add one? Not too familiar with the architecture/tests here :)

**Comment by fingolfin** [comment]
I am certainly not an expert myself but I noticed `examples/src/index.md` contains many sections of the form `# Issue 1392` or `# Issue653` or similar (albeit all jumbled up in nor order discernible to me). Anyway, if it was me I'd try to fit in there, perhaps?

**Comment by fingolfin** [comment]
Ah, however, you really need to be able to check that output is right. I am not sure if there is already a system for that in place? There are the files
```
test/examples/references/latex_showcase.tex
test/examples/references/latex_simple.tex
```
which look to me as if they are perhaps compared to the generated files?


What I'd expect to see (but doesn't seem to be there) is some tests were the `.html` files generated from certain inputs are stored as reference ("expected" data) and simply compared against what is generated ("actual" data). E.g. the git test suite (as in, git.git) uses such an approach (and of course doctests do it in a sense, too). Coupled with a script that makes it easy to regenerate the "expected" files if necessary.

**Comment by fingolfin** [comment]
If @mortenpi or @fredrikekre can suggest a good way to add such a test, that's great, but otherwise I'd say this should not be delayed over this; this PR has been open for over 2 years, only changes two lines and the underlying issue has been reported in at least four issues (one of those is already closed).

**Comment by mortenpi** [comment]
I took the liberty of pushing a test for this (I was double checking that `latex_simple` is the right place for this, so I had the changes anyway).

File: CHANGELOG.md [hunk 2]
```diff
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Fixed
 
 * Don't require custom themes to set a color for the 'todo' admonition. ([#2576])
-
+* Entries in `@repl` blocks that were hidden with `# hide` no longer produce erroneous empty lines ([#2054], [#1521], [#2399])
```
fingolfin, line 15: Order of the issue seems odd? But of course that's not important
```suggestion
* Entries in `@repl` blocks that were hidden with `# hide` no longer produce erroneous empty lines ([#1521], [#2054], [#2399])
```

test/examples/src.latex_simple/index.md
```diff
@@ -131,3 +131,14 @@ lines**.
 ## Issue 2300
 
 You Shall Not Break! You Shall Not Break! You Shall Not Break!
+
+## PR 2056
+
+Issues: `#2054`, `#1521`, `#2399`
```
fingolfin, line 137:
```suggestion
Issues: `#1521`, `#2054`, `#2399`
```

test/examples/references/latex_simple.tex
```diff
@@ -254,4 +254,26 @@ \section{Issue 2300}
...
 
 
+\section{PR 2056}
+
+
+
+\label{1659457410429464311}{}
+
+
+Issues: \texttt{\#2054}, \texttt{\#1521}, \texttt{\#2399}
```
fingolfin, line 264:
```suggestion
Issues: \texttt{\#1521}, \texttt{\#2054}, \texttt{\#2399}
```

fingolfin (approved):
Still looks good to me. Order of the three issues seems arbitrary but that's not so important

CHANGELOG.md
mortenpi: In the changelog at least I do like ordering them, so good catch :)

mortenpi:
Thank you @Seelengrab and @fingolfin!

Issue #1899, opened 2022-08-08
Menu's on-hover background height feels small


pankgeorg:
![image](https://user-images.githubusercontent.com/8681967/183393031-59cbd912-068e-46a5-840a-7b843070221b.png)

pfitzseb:
Could also do with a bit of horizontal padding.

Hetarth02:
After increasing `padding` from `0` to `0.3rem` it looks much better.

| Before | ![image](https://user-images.githubusercontent.com/80420115/222243134-9e413fae-e13a-4460-80bb-8149fa2559ba.png) |
|---|---|
|  After  | ![image](https://user-images.githubusercontent.com/80420115/222241777-e60366bc-ddcc-4a47-bbde-e09382ce637a.png)  |

Hetarth02:
@mortenpi Can you take a look at this too?

mortenpi:
I feel that, rather than adding padding, we should expand the gray highlight to cover the gap between the list items:

![image](https://user-images.githubusercontent.com/147757/224578427-df5f32ed-5f81-46df-8a89-92b6a43fe74c.png)

Otherwise it will always "feel small" relative to the list item itself. But I think for that we need to add background color to the `li`, rather than the `a`.

> Could also do with a bit of horizontal padding.

@pfitzseb Where do you actually mean? In front of the list bullet?

pfitzseb:
Yeah, just add a bit of horizontal and vertical padding to the `li` (and add the hover effect for that instead of the anchor):
![image](https://user-images.githubusercontent.com/6735977/224670327-fab1e2b0-afc0-450c-829d-1e5b2dd3383a.png)

Hetarth02:
@pfitzseb I cannot understand what you are saying. Are you implying to add the hover effect to `li` tag after adding padding to the it?

If so, doing that results in this on hover,
![image](https://user-images.githubusercontent.com/80420115/226102422-db004ff0-fd9a-4b2f-8e82-d50701dff899.png)

mortenpi:
> Are you implying to add the hover effect to `li` tag after adding padding to the it?

The idea is to fill the gap between the items, without (necessarily) changing the spacing between the items.

![image](https://user-images.githubusercontent.com/147757/227097529-69c8b0d2-6275-4957-92c7-3fd76e76d0d5.png)

[...]

PR #2065, opened 2023-03-12
Adding padding to menu's element on-hover

Hetarth02:
This PR aims to close Issue #1899.


Hetarth02:
@mortenpi I don't know why it is showing 17 commits?

I have my forked synced with the `master` branch. So, to make things clear I have only made changes to one `scss` file which could been seen in the `Files changed` menu.

mortenpi:
Let's discuss how we want to change this in #1899.

---

You also need to update the CSS files. I don't think we have the process documented anywhere, but you can [use DocumenterTools](https://github.com/JuliaDocs/DocumenterTools.jl/blob/c1af03618b9d1e9b50e9b6a472d2c4115dab6be5/src/Themes.jl#L68-L72) for that. I usually would do something like:

```
pkg> activate --temp
# Assuming you're in Documenter/, this will add your Documenter dev checkout as a dev. dependency:
pkg> dev .
# You need a special DocumenterTools tag (because the standard versions are not compatible with 0.28):
pkg> add DocumenterTools#documenter-v0.1.14+0.28.0-DEV
julia> using DocumenterTools
julia> DocumenterTools.Themes.compile_native_themes()
```

> I have my fork synced with the master branch.

You have extra commits on your master branch though, which is why they show here. The general practice is to create a new branch from the current master for each PR you want to open. That way you don't run into this. It's fine for this PR though -- we'll squash merge the PR anyway, so the intermediate commits won't affect anything.

Hetarth02:
> You also need to update the CSS files. I don't think we have the process documented anywhere, but you can [use DocumenterTools](https://github.com/JuliaDocs/DocumenterTools.jl/blob/c1af03618b9d1e9b50e9b6a472d2c4115dab6be5/src/Themes.jl#L68-L72) for that. I usually would do something like:

It would also be helpful to add some documentation regarding this. Do we have something lilke a `CONTRIBUTING.md`?

mortenpi:
> It would also be helpful to add some documentation regarding this. Do we have something lilke a `CONTRIBUTING.md`?

We have this: https://documenter.juliadocs.org/stable/contributing/

Hetarth02:
I am closing this PR for the time being. I will make new PR/reopen this PR later.

PR #2076, opened 2023-03-18
Move Hamburger to left

### LINKED ISSUE #1312: Collapsed view menu button should be on the same side as the menu
I don't know if this is standard, but I saw it used in docs.juliaplots.org so reporting it here, presuming this package generated it.

Annotation 2020-05-04 153736

Menu appears on the left but the menu button is on the top right. This just doesn't work well. Please keep them on the same side.

inkydragon:
fix #1312

<details>
<summary>old style</summary>

image
image

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - jquery has been updated from `v3.6.0` to `v3.6.4`.
   - MathJax 2 has been updated  from `v2.7.7` to `v2.7.9`.
 
-* Move the sidebar icon (Hamburger) from the right side to the left side.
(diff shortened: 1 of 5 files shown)


mortenpi:
The HTML implementation LGTM, but could we also retain the original size and positioning? Right now it looks a bit small, and is not quite properly aligned. Also needs a CHANGELOG note.

inkydragon:
- [x] it looks a bit small
- [x] not quite properly aligned
- [x] needs a CHANGELOG note

image
image

The text actually does center on it, if you type a lowercase x and then look at the centerline of the x and the centerline of the hamburger.
However, since the letters of our title are mainly above the baseline, the hamburger icon needs to be moved up a bit

mortenpi (approved):
Awesome, love it! I was actually still a little bit in two minds on this, but when I pulled a demo built up on my phone, the left feels far more natural (partially probably because I'm so used to Material).

PR #2077, opened 2023-03-18
Added Auto Theme Switch

### LINKED ISSUE #1745: Reenable auto-switching between light and dark theme
**Problem description:**
I find that once I’ve manually selected either the dark or light theme for a documentation there is no easy way to get back to automatic switching (based on my MacOS light/dark mode).

**Potential workaround:**
I guess one could get back auto-switching if one deletes the cache or so. However, that’s just annoyingly inconvenient.

**Suggested solution:**
Would it be possible to simply add an "auto" option (default) besides the light and dark theme options? That would be the simplest and most natural solution IMO.

(For reference: has been mentioned in [this discourse post](https://discourse.julialang.org/t/automatic-light-dark-theme-change-in-documentation/73864/3?u=carstenbauer) already.)

shravanngoswamii:
Close #1745.

### DIFF
--- a/assets/html/themeswap.js
+++ b/assets/html/themeswap.js
@@ -44,20 +44,20 @@ function set_theme_from_local_storage() {
     // Store the style sheets of inactive themes so that we could disable them
     if(themename !== theme) disabled.push(ss);
   }
-  if(active !== null) {
-    // If we did find an active theme, we'll (1) add the theme--$(theme) class to <html>
-    document.getElementsByTagName('html')[0].className = "theme--" + theme;
-    // and (2) disable all the other theme stylesheets
-    disabled.forEach(function(ss){
-      ss.disabled = true;
-    });
-  }
-  else if(darkTheme !== null && darkPreference === true) {
-    // If we did find an active theme, we'll (1) add the theme--$(theme) class to <html>
-    document.getElementsByTagName('html')[0].className = "theme--" + darkTheme;
-    // and (2) disable all the other theme stylesheets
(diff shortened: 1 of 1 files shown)


assets/html/themeswap.js
```diff
@@ -44,23 +44,23 @@ function set_theme_from_local_storage() {
...
-    disabled.forEach(function(ss){
-      ss.disabled = true;
-    });
-  }
-  else if(darkTheme !== null && darkPreference === true) {
-    // If we did find an active theme, we'll (1) add the theme--$(theme) class to <html>
-    document.getElementsByTagName('html')[0].className = "theme--" + darkTheme;
-    // and (2) disable all the other theme stylesheets
-    disabled.forEach(function(ss){
+  if (theme === "auto" || (active === null && darkPreference === true && darkTheme !== null)) {
```
mortenpi, line 47: Umm, `theme` comes from local storage, and will never be `"auto"`, because we never set it to that.

Pretty sure this script does not need to be changed. I think it should be enough to add an "auto" dropdown item to the theme switcher, that wipes the value from `localStorage` when you switch to it. I.e. in here: https://github.com/JuliaDocs/Documenter.jl/blob/master/assets/html/js/themes.js

shravanngoswamii: Ohhk i added the "auto" dropdown item there, please verify if its ok!!|
https://github.com/JuliaDocs/Documenter.jl/pull/2085#issue-1637542656

mortenpi:
Closing in favor of #2085.

PR #2078, opened 2023-03-18
Add "collapse/expand" functionality

### LINKED ISSUE #1393: Add "collapse/expand all" functionality
For people that already know other programming languages, it would help a lot to get an overview of what's available in a package/module by looking at the list of available functions/types (without all the detailed explanations). However, in the current implementation of documenter template where each article is implemented like a card (with header + section) and a margin between each of them, it is (at least for me) not possible to do that easily. I have already seen that there is the index functionality but it doesn't seem to be in use that much and it adds a bit of clutter that can be avoided by collapsing the sections of each article. This would allow a somewhat implicit implementation of the index with additional information for different function groups with markdown in-between. I have created a Tampermonkey userscript to evaluate this idea as a proof-of-concept: 

collapse_expand

 (Tampermonkey userscript gist: https://gist.github.com/cloudgiant/d02dca129d79558cab1a8cbbd820f0a7  . You can try it out on official julia docs after installing: https://docs.julialang.org/en/v1/base/collections/)

This script just adds an `expand/collapse articles` button on top right (It will of course need some css love for the proper implementation) and adds an onclick listener for the headers to expand/collapse single articles (Single article expand/collapse functionality can also be triggered by a button if an onclick listener for the header is a bit too much).

In this crude implementation, the methods are shown duplicated. This problem can be solved by either showing signatures or grouping the methods together for the function name (with number of methods maybe?). Ideally the header would also show a single sentence summary in the collapsed state.

I think a variation of this functionality would be beneficial for reference part of every package documentation, especially for newcomer julia developers.

Hetarth02:
- Add "collapse/expand all" functionality with a icon in navbar
- Add "collapse/expand" functionality to sensible individual element
- This PR aims to close #1393. 

https://user-images.githubusercontent.com/80420115/226101724-5479e0be-159a-483e-8bc1-4eafc9a94acf.mp4


Hetarth02:
@mortenpi I tried creating another branch as per the discussion in #2065 but it still displays all of the commit history. So, once again to clarify for anyone looking at this PR please check the `Files changed` menu.

mortenpi:
I actually really like this! It actually makes these long docstring lists readable. And the open/close animation feels pretty smooth.

A couple of thoughts though. What do you think about the following:

* Having the collapse/expand all button be  the rightmost item in the top bar?
* Could we try a variant where the docstring collapse/expand button is also on the top-right of the docstring?
* I am wondering if we should have them be collapsed by default?
* We should do a demo build of the Julia manual with this.

inkydragon:
https://www.mathworks.com/help/matlab/ref/sin.html
image

> 1. Having the collapse/expand all button be the rightmost item in the top bar?

Agree.
Would it be clearer to use text + icons? Just like Matlab doc ①.

> 2. Could we try a variant where the docstring collapse/expand button is also on the top-right of the docstring?

For document strings, we can modify the collapsed style, the cursor style when the mouse is selected (③), and also add hint text to highlight that this is a collapsed area that can be clicked to interact with.

> 3. I am wondering if we should have them be collapsed by default?

Maybe collapse by default for automatically generated doc str.
We can also add a keyword argument to control the default collapse behavior.

> 4. We should do a demo build of the Julia manual with this.

I would say that even though I've done it many times, building julia's documentation is still a bit complicated. Maybe try building the documentation for Documenter.jl first.

Hetarth02:
> I really like this! It makes these long docstring lists readable. And the open/close animation feels pretty smooth.

Thanks!

> * Having the collapse/expand all button be  the rightmost item in the top bar?

The thing is when in responsive mode the side-menu button will be the last.

> * Could we try a variant where the docstring collapse/expand button is also on the top-right of the docstring?

I think we can do that, but the problem then would be when a large name comes it will mess up the design.

> * I am wondering if we should have them be collapsed by default?

Personally I would want them to be expanded, but I am fine with either.

> * We should do a demo build of the Julia manual with this.

I tried but had some issues in generating one. I removed some parts from the julia library to mitigate the errors and was successfully able to build a partial working site. The video I have attached above is the same site I built locally.

Hetarth02:
> Would it be clearer to use text + icons? Just like Matlab doc ①.

The thing is we don't have much space in the navbar or the docstring title on smaller devices. It is much likely to break the UI on smaller devices when we have text + icons combination.

> We can also add a keyword argument to control the default collapse behavior.

Yeah, this is a good idea.

mortenpi:
> The thing is when in responsive mode the side-menu button will be the last.

That just hopped on the other side though, so it's not a problem anymore :slightly_smiling_face:  https://github.com/JuliaDocs/Documenter.jl/pull/2076

> I tried but had some issues in generating one. I removed some parts from the julia library to mitigate the errors and was successfully able to build a partial working site. The video I have attached above is the same site I built locally.

I didn't realize that it was already showing the Julia manual :facepalm: I'm happy to look into doing the full build though. The errors might be due to some changes we've had on master here.

> Personally I would want them to be expanded, but I am fine with either.

Let's leave them expanded for now then. Adding a per-block option is a good idea. We could also have a per-page option via an at-meta block. But let's do those in a follow-up PR -- let's not make this PR too complicated.

> The thing is we don't have much space in the navbar or the docstring title on smaller devices. It is much likely to break the UI on smaller devices when we have text + icons combination.

I tend agree with this. I think we're kinda forced to go with icons right now.

> I think we can do that, but the problem then would be when a large name comes it will mess up the design.

That is fair. I feel that maybe we should follow Matlab's lead here though and have the collapsed icon be `>` and expanded `v`.

> For document strings, we can modify the collapsed style, the cursor style when the mouse is selected (③), and also add hint text to highlight that this is a collapsed area that can be clicked to interact with.

Yeah, some mover and cursor stuff might be worth trying out.

Hetarth02:
It auto-closed the PR when I synced my fork. I have reopened the PR. Sorry for this!

Hetarth02:
- [x] Cursor will be a pointer when hovering over article `header`.
- [x] Shifted the Nav-bar toggle to last position
- [x] Changed the `>` as a indication instead of `^` for articles(kept the Nav-bar icons as it is)

Hetarth02:
> I'm happy to look into doing the full build though. The errors might be due to some changes we've had on master here.

@mortenpi If possible can you do this and check if everything works fine or there changes needed?

assets/html/scss/documenter/components/_docstring.scss
```diff
@@ -21,6 +22,10 @@
       background-color: transparent;
     }
 
+    .docstring-article-toggle-button {
+      padding: 0.2rem 0.2rem 0.2rem 0;
```
mortenpi, line 26: The rest of the docstring header moves right now when you expand/collapse it (because the icon character is of different width).

```suggestion
      padding: 0.2rem 0.2rem 0.2rem 0;
      min-width: 1.1rem; // to make sure it does not resize when icon changes
```

assets/html/js/collapse.js
```diff
@@ -0,0 +1,55 @@
...
+      .find(".docstring-article-toggle-button")
+      .removeClass("fa-chevron-down")
+      .addClass("fa-chevron-right");
+  } else {
+    $(this)
+      .find(".docstring-article-toggle-button")
+      .removeClass("fa-chevron-right")
+      .addClass("fa-chevron-down");
+
+    articleToggleTitle = "Collapse article";
```
mortenpi, line 20:
```suggestion
    articleToggleTitle = "Collapse docstring";
```

assets/html/js/collapse.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  $(this).find(".docstring-article-toggle-button").prop("title", articleToggleTitle);
+  $(this).siblings("section").slideToggle();
+});
+
+$(document).on("click", ".docs-article-toggle-button", function () {
+  let articleToggleTitle = "Expand article";
+  let navArticleToggleTitle = "Expand all Articles";
```
mortenpi, line 29:
```suggestion
  let articleToggleTitle = "Expand docstring";
  let navArticleToggleTitle = "Expand all docstrings";
```

assets/html/js/collapse.js
```diff
@@ -0,0 +1,55 @@
...
+    $(".docstring section").slideUp();
+  } else {
+    $(this).removeClass("fa-chevron-down").addClass("fa-chevron-up");
+    $(".docstring-article-toggle-button")
+      .removeClass("fa-chevron-right")
+      .addClass("fa-chevron-down");
+
+    isExpanded = true;
+    articleToggleTitle = "Collapse article";
+    navArticleToggleTitle = "Collapse all Articles";
```
mortenpi, line 48:
```suggestion
    articleToggleTitle = "Collapse docstring";
    navArticleToggleTitle = "Collapse all docstrings";
```

assets/html/js/collapse.js
```diff
@@ -0,0 +1,55 @@
+// libraries: jquery
+// arguments: $
+
+var isExpanded = true;
+
+$(document).on("click", ".docstring header", function () {
+  let articleToggleTitle = "Expand article";
```
mortenpi, line 7:
```suggestion
  let articleToggleTitle = "Expand docstring";
```

mortenpi (commented):
Awesome, this is great! Sorry for the delay, but I think this is nearly good to go if you're done with it.

* Just some minor suggestions in other comments.
* Could you add a CHANGELOG entry for this?

In principle, the JS code seems to have some duplication (e.g. the expand/collapse strings), so it could probably be refactored a bit to make it a bit cleaner. But that's pretty minor and not very important here, so more than happy to merge as is.

mortenpi:
Here's a full Julia manual build: http://mortenpi.eu/juliadocs-demos/pr-2078-julia-manual-demo/

image

CHANGELOG.md
```diff
@@ -77,6 +77,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Move the mobile layout sidebar toggle (hamburger) from the right side to the left side. (#1312, #2076)
 
+* Added the ability to expand/collapse individual as well as all docstrings. (#1393)
```
mortenpi, line 80:
```suggestion
* Added the ability to expand/collapse individual as well as all docstrings. (#1393, #2078)
```

mortenpi:
Thanks a lot @Hetarth02!

PR #2081, opened 2023-03-20
Overhaul remote repository link handling

### LINKED ISSUE #1975: Source links to standard library functions missing in Julia manual
E.g. anything on the [LinearAlgebra](https://docs.julialang.org/en/v1.9-dev/stdlib/LinearAlgebra/) or [Tar](https://docs.julialang.org/en/v1.9-dev/stdlib/Tar/) pages (and most others). Related to https://github.com/JuliaLang/julia/issues/43307 (although we now just omit the source links, rather than pointing to the wrong place).

At least for some we need to create a way to point to (the right commit) of an external repository (e.g. [Tar.jl](https://github.com/JuliaIO/Tar.jl)). I am pretty sure the paths in the docstring metadata point to the Julia repository, as the code gets copied there during build. LinearAlgebra I am less sure about right now, as it still lives in the Julia repo, but clearly something is also going wrong there.

Slightly related to #1808. However, properly fixing this will likely involve some more refactoring of the way we handle these URLs in Documenter, and lead to some behavior changes. So I would like to hold off on 0.28 until we have this.

mortenpi:
I think I have settled on the designed for this now. Basically, this aims to add a new `makedocs` argument called `remotes` (in additional to `repo`), that can be used to specify `directory => repo` pairs, that will then be used to determine the remote source links.

The code fixes the original issue, but is still WIP, since it now needs to be updated to match the new docs & design. Feedback or questions would be most welcome, in particular on the explainer docs in [`remote-links.md`](https://github.com/JuliaDocs/Documenter.jl/compare/mp/julia-stdlibs?expand=1#diff-6d62983b8c6f4f3647d9c9854e20cb295ea55824d60e3aa703041141bbefefc4).

This will fix #1975, and should pretty much unblock 1.0.


mortenpi:
Alright, I think I need to do another pass on the code, and look into adding a few new tests for edge cases. But generally, I think this is almost ready to go. Any eyes reviewing this would always be welcome.

mortenpi:
Tests are passing. Time to merge this and deal with the consequences later.

PR #2085, opened 2023-03-23
Add "auto" theme option to theme picker dropdown

shravanngoswamii:
"auto" theme is added in theme picker, is it ok?
https://github.com/JuliaDocs/Documenter.jl/pull/2077#discussion_r1145651759


mortenpi:
Yeah, this looks like what I had in mind. But it also needs to have the corresponding entry in the dropdown for this to work.

Also, please don't open separate PRs, but iterate on the code in the existing PR.

shravanngoswamii:
Entry in dropdown is added, will it work now!!

assets/html/js/themes.js
```diff
@@ -34,10 +34,20 @@ function set_theme(theme) {
 
 // Theme picker setup
 $(document).ready(function() {
+  // Add "auto" option to the dropdown
+  $('#documenter-themepicker').append($('<option>', { 
+    value: 'auto',
+    text : 'Auto'
```
mortenpi, line 40:
```suggestion
    text : 'Automatic (OS preference)' 
```

mortenpi (commented):
A few more things:

* It should also update the theme if you change to `auto`. I.e. basically needs a call to `set_theme`, but might need to be updated a bit to handle this case.
* It would be great to have a small note in `CHANGELOG.md` about this as well.

mortenpi:
The switching back to OS theme doesn't seem to work. Does it work for you when you run it locally?

shravanngoswamii:
Actually, I am having an error while running Documenter.jl locally, i update the code soon if OS theme doesn't work!!

PR #2086, opened 2023-03-25
Clean docstrings

pitmonticone:


mortenpi:
An amazing number of typos have slipped through over the years :laughing: Thank you for cleaning it up!

PR #2088, opened 2023-03-27
Fix various typos.

goggle:
Corrects various typos that I've found in the repository.


KristofferC:
Thanks, what tool are you running?

mortenpi:
> It can also be integrated in Github Actions, but then it needs some configuration to prevent it from reporting false positives.

If anyone is interested in setting it up, we could have it run as an advisory check on PRs.

mortenpi:
Thanks @goggle! My comment from https://github.com/JuliaDocs/Documenter.jl/pull/2086#issuecomment-1484280442 was apparently still relevant :smile:

PR #2092, opened 2023-03-29
Add Dockerfile for LaTeX build

### LINKED ISSUE #2073: Migrate LaTeX Docker image to GCHR
We publish the [LaTeX Docker image on Docker Hub](https://hub.docker.com/r/juliadocs/documenter-latex), but [Docker is nuking free team repos there](https://web.docker.com/rs/790-SSB-375/images/privatereposfaq.pdf), so users won't be able to pull the image from April onwards.

We should probably move the image hosting to GitHub's GHCR:
 * https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages
 * https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

While we're at it, we should also move the [`Dockerfile`](https://github.com/JuliaDocs/DocumenterLaTeX.jl/tree/master/docker) to this repo I think.

inkydragon:
Fix #2073 

- Do we want to update the version of Ubuntu? Maybe use `ubuntu:latest`
- Do we need a `tectonic` docker image? `juliadocs/documenter-tectonic`
	And rename texlive-based image to: `juliadocs/documenter-texlive`  
	Estimated size is 100MB~. And current image is 430MB+

todo
- [ ] run PDF build for `JuliaDocs/Documenter.jl`
- [ ] run PDF build for `JuliaLang/julia`
- [ ] release docker image, and update `DOCKER_IMAGE_TAG` in `src/latex/LaTeXWriter.jl`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,25 @@
+# SPDX-License-Identifier: MIT OR CC0-1.0
+# build image:
+#   docker build -t juliadocs/documenter-latex:1.0  .
+FROM ubuntu:18.04
+
+RUN useradd --create-home zeptodoctor
+RUN set -eux; \
+    apt-get update; \
+    apt-get install --no-install-recommends -y \
+    # minimal texlive
+        texlive-luatex \
+        texlive-latex-base \
+        texlive-latex-recommended \
+        texlive-latex-extra \
+    # latexmk: LaTeX build tool
+        latexmk \
+    # code highlighting
+        python-pygments \
+    # refresh the font cache
+        fontconfig \
+        ; \
+    rm -rf /var/lib/apt/lists/*; \
+    apt-get purge -y --auto-remove;
+
+WORKDIR /home/zeptodoctor/
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,10 @@
+# documenter-latex
+
+Docker image for compiling pdf output with
+[`Documenter.jl`](https://github.com/JuliaDocs/Documenter.jl)
+
+
+## Whats in the image?
+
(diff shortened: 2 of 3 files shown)


mortenpi (commented):
Don't we need the fonts, [like in the old image](https://github.com/JuliaDocs/DocumenterLaTeX.jl/blob/081470dbdee0896436a9872fe4aa53d52cb52a11/docker/Dockerfile#L11-L20)?

We also had [`ENV DEBIAN_FRONTEND noninteractive`](https://github.com/JuliaDocs/DocumenterLaTeX.jl/blob/081470dbdee0896436a9872fe4aa53d52cb52a11/docker/Dockerfile#LL3-L3C35). I am not 100% sure why though. Maybe @fredrikekre remembers?

mortenpi:
> Do we need a tectonic docker image? juliadocs/documenter-tectonic. And rename texlive-based image to: juliadocs/documenter-texlive. Estimated size is 100MB~. And current image is 430MB+

Tectonic should be used via a JLL I think, so shouldn't be necessary to have an image?

> And I'd like to rename the docker usename to `juliadocs`

Sounds reasonable.

> * I choose `MIT OR CC0-1.0` for Dockerfile, Because I think it makes it easy for people to copy and paste

Fine by be. I feel that it's not particularly copyrightable anyway, for what it's worth. So CC0 seems suitable.

mortenpi:
We also need a `LABEL` for GHCR? https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package#connecting-a-repository-to-a-container-image-using-the-command-line

PR #2109, opened 2023-04-19
ci: run master branch of crate-ci/typos

mortenpi:
X-ref: https://github.com/JuliaDocs/Documenter.jl/pull/2105#issuecomment-1515435429

The seems to be their official recommendation anyway: https://github.com/crate-ci/typos/blob/master/docs/github-action.md

cc @goggle

### DIFF
--- a/.github/workflows/SpellCheck.yml
+++ b/.github/workflows/SpellCheck.yml
@@ -10,4 +10,4 @@ jobs:
       - name: Checkout Actions Repository
         uses: actions/checkout@v3
       - name: Check spelling
-        uses: crate-ci/typos@v1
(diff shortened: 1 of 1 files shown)


goggle:
I think this looks reasonable.
The project [`polar-rs/polars`](https://github.com/pola-rs/polars) has it like this in their workflow.
The only thing is that unrelated failures can happen, because [`crate-ci/typos`](https://github.com/crate-ci/typos) updates their dictionary once a month, so it is possible that previously unnoticed typos in the source tree get suddenly discovered. But this would happen anyways when `dependabot` updates the version.

PR #2111, opened 2023-04-22
Add `@diagram` support

pedromxavier:
This pull request aims to add support for native diagrams in Julia docs by leveraging [Kroki.jl](https://github.com/bauglir/Kroki.jl/).

It adds a `@diagram <format>` code block allowing the user to select a format from those supported by [kroki](https://kroki.io).

PS: documentation skectch for `@diagram` already included.

Task list:
- [x] Implement `@diagram` block
- [ ] Add LaTeX support
- [ ] Write unit tests?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,6 +11,7 @@ DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 IOCapture = "b5f81e59-6552-4d32-b1f0-c071b021bf89"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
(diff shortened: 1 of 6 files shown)


mortenpi:
This should be done as a plugin package, for the time being at least. Literally yesterday I put together one for Mermaid (https://github.com/JuliaDocs/DocumenterMermaid.jl), which should work as a guide if need be. But it's literally just a matter of moving the about putting the new definitions into a separate module.

That said, this is cool! I know that @bauglir might find this interesting. I presume this works by connecting to the `https://kroki.io/` endpoint?

pedromxavier:
Thanks for the guidance, @mortenpi! I will get this running asap :)

Yes, it does rely on the kroki.io API and yes, you do need an active internet connection during the build (which should be fairly OK for most applications i.e. gh-pages)

pedromxavier:
Here it is @mortenpi @bauglir: https://github.com/pedromxavier/DocumenterDiagrams.jl

I think this PR can be closed.

PR #2112, opened 2023-04-23
Add DocumenterDiagrams, DocumenterMermaid links to README

mortenpi:
At some point I would actually like to have a documentation page with plugins.

cc @pedromxavier


pedromxavier:
@mortenpi do you foresee any automatic interface for that? I mean, some `@docs`-like mechanism to just list the plugins i.e.

```markdown
# Documenter Plugins

```@docs
DocumenterCitations
DocumenterMermaid
DocumenterDiagrams
...
```

Maybe this could work from module docs. Food for thought.

PR #2116, opened 2023-04-29
Changed display to none on docs-left

pranavvatsa:
Hey, I went ahead and fixed the bug by replacing the display: inline-block to display: none as it was mentioned. It would be wonderful if someone could validate this.


mortenpi:
Sorry for a delay in the review. Unfortunately, this is not the correct solution, since it will _always_ hide the hamburger, which is definitely not what we want.

PR #2134, opened 2023-05-26
[LaTeXWriter] fix escaping in header titles

odow:
I think this is the culprit of https://github.com/jump-dev/JuMP.jl/pull/3386. I'll test on this branch before adding tests etc.

### DIFF
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -83,6 +83,7 @@ Context(io, doc) = Context{typeof(io)}(io, false, Dict(), 1, "", doc)
 
 _print(c::Context, args...) = Base.print(c.io, args...)
 _println(c::Context, args...) = Base.println(c.io, args...)
+_print(io, args...) = Base.print(io, args...)
 
 # Labels in the TeX file are hashes of plain text labels.
 # To keep the plain text label (for debugging), say _hash(x) = x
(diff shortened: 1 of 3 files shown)


mortenpi:
Added a test here that should hopefully fail on current master: https://github.com/JuliaDocs/Documenter.jl/commit/e1396e557829ff96e70a29fdc4b1a32865793d77.

src/latex/LaTeXWriter.jl
```diff
@@ -83,6 +83,7 @@ Context(io, doc) = Context{typeof(io)}(io, false, Dict(), 1, "", doc)
 
 _print(c::Context, args...) = Base.print(c.io, args...)
 _println(c::Context, args...) = Base.println(c.io, args...)
+_print(io, args...) = Base.print(io, args...)
```
odow, line 86: I guess the single-argument `latexesc(s)` method wasn't actually tested, because it hit this method error.

mortenpi:
I got the test to finally fail on master, so I took the liberty of pushing it here.

odow (commented):
LGTM

mortenpi:
It looks like the JuMP build also passed with this fix?

odow:
Yip. Although now I need to tweak again, because it made every docstring a Chapter 😆

mortenpi:
Let's hope it doesn't reveal more bugs -- I'll go ahead and merge this :laughing:

PR #2168, opened 2023-07-04
Documenter semver guarantee docs

mortenpi:
This adds a section to the manual that aims clarify and codify what Documenter behaviors, strictly speaking, are part of the public API and what aren't.

This is relevant for the 1.X era, since I expect us to be living on version `v1` for a while before we put out another breaking release. So we need to balance being able to make changes to the way Documenter behaves, vs. not breaking things too badly.

Would be good to hear any thoughts on this (cc @fredrikekre).


docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+[^1]: Eventually, of course, Documenter `2.0` may break everything. But we don't expect a breaking release in the near future.
+
+However, Documenter is relatively complex, and it can sometimes be unclear what constitutes a _breaking change_.
+For example, is changing the CSS classes in the HTML themes allowed?
+What if the user relied on a CSS class in an `@raw html`-block?
+If their HTML is no longer rendered correctly, is that a breaking change?
+Is completely changing the LaTeX compiler a breaking change?
+What if it breaks a PDF build somewhere due to some math block using a feature that only works with `pdflatex`?
+
+This page aims to clarify what **is** and **is not** covered by the Documenter semver guarantees, both as information for users, and as guidance for developers.
```
fredrikekre, line 15:
```suggestion
This page aims to clarify what **is** and **is not** covered by the Documenter SemVer guarantees, both as information for users, and as guidance for developers.
```

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+
+!!! note "This page is not complete!"
+
+    If you need to rely on something that is currently an internal API or undocumented behavior, open an issue or a pull request to get it documented!
+    There is a good chance that the behavior is already a _de factor_ semver guarantee, or can easily be cleaned up and made public API.
+    The goal is to, over time, add and document additional semver guarantees.
```
fredrikekre, line 21:
```suggestion
    There is a good chance that the behavior is already a _de factor_ SemVer guarantee, or can easily be cleaned up and made public API.
    The goal is to, over time, add and document additional SemVer guarantees.
```

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+  The behavior of undocumented edge cases may change, but only in accordance with what is documented.
+
+!!! note "Experimental APIs"
+
+    Note that some APIs may explicitly marked experimental.
+    In that case, you can only rely on them _withing a minor version_.
+    The next minor version release may completely change or remove experimental features and APIs.
+
+
+## What is not covered by semver
```
fredrikekre, line 40:
```suggestion
## What is not covered by SemVer
```

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+* The HTML, TeX, or file structure of the generated documents (unless explicitly documented).
+  This includes the CSS of the HTML themes, and as such any custom CSS overrides may stop working at any point.
+  However, there are many _de facto_ guarantees here that should get documented over time (e.g. for custom themes).
+* The look and feel of the HTML UI and the generated PDF may change considerably from one minor version to another.
+* Anything explicitly marked experimental is not guaranteed to remain stable (see the note above).
+
+
+!!! note "Patch versions are probably okay"
+
+    If you are relying on some non-semver behaviors, features, or internals, it probably fine to expect things not to break _within a patch release_.
```
fredrikekre, line 58:
```suggestion
    If you are relying on some non-SemVer behaviors, features, or internals, it probably fine to expect things not to break _within a patch release_.
```

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+    In this case, you should add a `[compat]` entry to your `Project.toml` files with a [tilde specifier](https://pkgdocs.julialang.org/v1/compatibility/#Tilde-specifiers) fixing Documenter's to a specific minor version, e.g.
+
+    ```toml
+    Documenter = "~1.X"
+    ```
+
+    where `X` is the minor version you are developing against.
+
+    Alternatively, if this is for package documentation, and your `docs/make.jl` script is relying on some non-semver behavior, you can also check in a `docs/Manifest.toml` file to fully fix the Documenter version.
```
fredrikekre, line 67:
```suggestion
    Alternatively, if this is for package documentation, and your `docs/make.jl` script is relying on some non-SemVer behavior, you can also check in a `docs/Manifest.toml` file to fully fix the Documenter version.
```

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+    There is a good chance that the behavior is already a _de factor_ semver guarantee, or can easily be cleaned up and made public API.
+    The goal is to, over time, add and document additional semver guarantees.
+
+
+## Documenter's API guarantees
+
+The following APIs and behaviors are guaranteed not to change:
+
+* Standard promises about the Julia APIs (public Documenter functions and their documented arguments).
+  In a nutshell, any `make.jl` making use of just public, documented parts of the Documenter API should always continue working (i.e. builds should complete).
```
fredrikekre, line 29: Could this be clarified? Does it mean all of Documenters `export`s essentially?

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+    The next minor version release may completely change or remove experimental features and APIs.
+
+
+## What is not covered by semver
+
+In principle, anything that is not covered by the previous section is, by definition, _not_ part of the public API and is _not_ guaranteed not to break.
+
+However, it is worth mentioning a few things explicitly, in particular things that are currently not part of the API, but should be added (in some way) in the future:
+
+* Any time you hook into Documenter's Julia internals some way.
```
fredrikekre, line 46:
```suggestion
* Any time you hook into Documenter's internals in some way.
```

fredrikekre (approved):

docs/src/lib/semver.md
mortenpi: I'll think about how to expand on this. It's not just exports though -- we have some unexported public stuff too, like `Documenter.HTML` etc. So it's really anything that's under the `Library > Public` section in the manual.

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+
+* Standard promises about the Julia APIs (public Documenter functions and their documented arguments).
+  In a nutshell, any `make.jl` making use of just public, documented parts of the Documenter API should always continue working (i.e. builds should complete).
+* Any explicitly documented behaviors (e.g. the way Documenter [determines remote repository links is documented](@ref "Remote links")).
+  The behavior of undocumented edge cases may change, but only in accordance with what is documented.
+
+!!! note "Experimental APIs"
+
+    Note that some APIs may explicitly marked experimental.
+    In that case, you can only rely on them _withing a minor version_.
```
mortenpi, line 36:
```suggestion
    In that case, you can only rely on them _within a minor version_.
```

mortenpi:
Thanks @fredrikekre! Let's leave this open for a little while -- I might want to do another pass on it too. We can merge this close to the release.

docs/src/lib/semver.md
```diff
@@ -0,0 +1,68 @@
...
+    The goal is to, over time, add and document additional SemVer guarantees.
+
+
+## Documenter's API guarantees
+
+The following APIs and behaviors are guaranteed not to change:
+
+* Standard promises about the Julia APIs (public Documenter functions and their documented arguments).
+  In a nutshell, any `make.jl` making use of just public, documented parts of the Documenter API should always continue working (i.e. builds should complete).
+* Any explicitly documented behaviors (e.g. the way Documenter [determines remote repository links is documented](@ref "Remote links")).
```
mortenpi, line 30: Note to self: this at-ref needs to be updated after #2081 is merged.

docs/src/lib/semver.md
fredrikekre: But also things like the markdown syntax (`@example` etc) I guess? Perhaps it would make sense to restructure the docs a bit. Would be good to add "docstrings" to those things too and include in the API section.

mortenpi: Ah, yes, should definitely reference those. And yeah, we should think about how to present them better. Not sure we strictly speaking need docstrings -- the `Manual > Syntax` page is basically the reference manual for those. But it should probably be moved under the `API` section.

fredrikekre: Yea, it always feels a bit strange to go to `Manual > Syntax` when looking up these.

mortenpi: I slightly expanded that bullet point, and also opened #2246 to remember to reorganize the at-block page.

Issue #2141, opened 2023-06-12
Changing the search engine


Hetarth02:
Issue #1457, Issue #1749 This addresses the problem of getting insufficient or non-ideal search results after the search.

Issue #1832: This issue also proposes the idea of using a different search engine instead of LunrJs to reduce initial lag which happens during the building of the indexes.

Moreover, lunrjs repo has been stale for some time.

I mostly suggest minisearchjs.

- It is designed for memory-constrained cases.
- Is inspired by Lunrjs, so it will be easier to shift to it by not having to re-implement the entire search logic.
- Supports suggestions if you want to improve the search UI in future.

**16 June, 2023**
For anyone wanting to test it out, here is the link to compare both results from both engines.

[Compare Searches](https://documenter-search.netlify.app/compare-search.html)

Hetarth02:
For the cold start times of lunrjs, currently we can have something of an opt-in pre-building of indexes and just load up the serialised index on client-side.

A simple benchmark I did yielded these results,

![img-20230524-wa0001-1.jpg](https://github.com/JuliaDocs/Documenter.jl/assets/80420115/3e103815-3a59-4bf6-86d6-3272269b050a)

![img-20230524-wa0002-1.jpg](https://github.com/JuliaDocs/Documenter.jl/assets/80420115/64efe85a-0d4c-4caf-ba89-25128e304818)

The load times, went to around ~2900ms to ~800ms.

The serialised index size is around 7.6mb

For minisearchjs,

Serialised Index size is around 6.4mb (Around 16% less than lunrjs)

And Instantiation time with prebuild index is around ~500ms.

The only downside for prebuilding index is that we require NodeJs as an environment.

mortenpi:
Looks good, I think we definitely want this on our roadmap. It would be good to have side-by-side demos with the same index for e.g. the Julia manual, to see how the search results differ. And also maybe have it print out the timings and index sizes in the consoles.

Hetarth02:
For anyone wanting to test it out, here is the link to compare both results.

[Compare Searches](https://documenter-search.netlify.app/compare-search.html)

mortenpi:
Alright, I did a few small tweaks to the compare page (https://github.com/mortenpi/documenter-updated-search/commit/04c766c953e09949f618d1ecc3240e1ad0f1a50c; [deployed here](https://mortenpi.eu/documenter-updated-search/compare-search.html)) and tried out a bunch of cases.

[...]

PR #2172, opened 2023-07-05
Change from LunrJs to MinisearchJs for client-side search

Hetarth02:
This PR aims to close Issue #2141.


mortenpi:
Something I noticed while testing is that it likes to repeat page headings a lot:

image

The lunr based approach doesn't do that:

image

---

One other note: the PR has also a lot of non-substantial formatting changes. While I don't disagree with the changes, it's generally a good idea to make the formatting changes independently. In this case, though, it's a little less important -- we're kinda replacing the whole `search.js` anyway. But it is a little tricky to see right now how the implementations differ, and where they're the same.

Hetarth02:
I have fixed the duplication issue.

Additionally, I have also updated the UI for the search results. The new UI looks like this,

image

image

The new UI provides context(if possible) as to where the search matches, while also displaying the category badge on the right and title on the left with link at the bottom.

Hetarth02:
I have fixed the duplication issue.

Additionally, I have updated the search result UI to provide better context for the results like displaying not only title but the matched string(if found) and the category badge too.

The new UI looks like this,

image

image

mortenpi (changes requested):
Awesome! Mostly just a bit of bikeshedding. The main thing would be moving the search layout stuff into a shared SCSS file, to avoid duplication, but then I'd pretty much be happy to merge I think.

Hetarth02:
Alright, will clean up the code as suggested.

assets/html/scss/documenter-light.scss
```diff
@@ -40,4 +40,57 @@ code.language-julia-repl > span.hljs-meta {
...
+  color: whitesmoke;
+  background-color: #33415580;
+  border-radius: 0.6rem;
+}
+
+.search-result-highlight {
+  background-color: hsl(48, 100%, 67%);
+}
+
+.search-divider {
```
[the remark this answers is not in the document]
Hetarth02: Hmm, this won't on `a` tag because we are rounding off the borders so it will look like this.

image

Two ways to solve this,

1. Remove the `border-radius` from `a` tag
2. Keep this as it is.

assets/html/scss/documenter-light.scss
```diff
@@ -40,4 +40,57 @@ code.language-julia-repl > span.hljs-meta {
...
+  font-style: italic;
+  text-transform: none !important;
+  line-height: 1.5;
+  color: whitesmoke;
+  background-color: #33415580;
+  border-radius: 0.6rem;
+}
+
+.search-result-highlight {
+  background-color: hsl(48, 100%, 67%);
```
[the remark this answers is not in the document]
Hetarth02: This color suits well in both light and dark mode, so I will chose this instead of `#edff00`.

assets/html/scss/documenter-dark.scss
```diff
@@ -122,4 +122,58 @@ html.theme--#{$themename} {
...
+    background-color: #33415580;
+    border-radius: 0.6rem;
+  }
+  
+  .search-result-title {
+    color: whitesmoke;
+  }
+  
+  .search-result-highlight {
+    background-color: greenyellow;
```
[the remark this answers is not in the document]
fredrikekre: Wouldn't the orange one from the light theme work in the dark theme too? Or perhaps a slightly darker orange

assets/html/scss/documenter-light.scss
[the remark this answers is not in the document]
mortenpi: Fair point. Let's keep it as is.

assets/html/scss/documenter-light.scss
```diff
@@ -40,4 +40,4 @@ code.language-julia-repl > span.hljs-meta {
 
 // Workaround to compile in highlightjs theme, so that we could have different
 // themes for both
-@import "highlightjs/default"
+@import "highlightjs/default";
```
mortenpi, line 43:
```suggestion
@import "highlightjs/default";

```

mortenpi, line 43: Needs a newline at the end as well.

assets/html/scss/documenter/_variables.scss
```diff
@@ -89,3 +89,15 @@ $documenter-docstring-header-padding: 0.5rem $documenter-container-left-padding;
 $documenter-docstring-body-padding-h: $documenter-container-left-padding;
 $documenter-docstring-body-padding-v: 0.75rem;
 $documenter-docstring-body-padding: $documenter-docstring-body-padding-v $documenter-docstring-body-padding-h;
+
+// Search Results variables
+$search-result-link-hover: rgba(0, 128, 128, 0.1) !default;
+
+$search-333: #333 !default;
+$search-f1f5f9: #f1f5f9 !default;
+$search-whitesmoke: whitesmoke !default;
```
mortenpi, line 98: I think the variables should have a bit more "semantic" names. What role do they fill?

assets/html/scss/documenter-dark.scss
```diff
@@ -122,4 +122,13 @@ html.theme--#{$themename} {
...
   }
+  
+  .search-result-link:hover .property-search-result-badge, .search-result-link:focus .property-search-result-badge {
+    color: $search-333 !important;
+    background-color: $search-f1f5f9 !important;
+  }
+  
+  .search-result-title {
+    color: $search-whitesmoke !important;
+  }
```
mortenpi, line 133: I think we should just be able to do this by overriding the variables in `documenter-dark.scss`, not the whole selectors?

assets/html/scss/documenter/layout/_search.scss
```diff
@@ -14,5 +14,59 @@
...
+    }
+    
+    .search-result-link:hover .property-search-result-badge, .search-result-link:focus .property-search-result-badge {
+      color: $search-f1f5f9;
+      background-color: $search-333;
+    }
+    
+    .search-result-highlight {
+      background-color: $search-result-highlight;
+      color: black !important;
```
mortenpi, line 49: Are these `!important`s important? I think ideally we should avoid them, unless we really need them. But this can always be "fixed" later as well.

assets/html/scss/documenter/layout/_search.scss
```diff
@@ -14,5 +14,59 @@
     .docs-highlight {
       background-color: yellow;
     }
```
mortenpi, line 16: Any of the old ones we can remove? Is e.g. `.docs-highlight` used anywhere?

mortenpi (approved):
This LGTM, thank you @Hetarth02! I need to get the Prettier workflow working properly, so I'll hold off on merging until that passes.

PR #2176, opened 2023-07-06
add --project to julia command using --color

mdmaas:
Just a tiny fix for the guide. I copy-pasted the command that included colored output, and of course it didn't work for a large project with several dependencies, so I was wandering what had happened to my environment only to then realize that --project was not part of the command.


mortenpi:
Thank you!

PR #2182, opened 2023-07-12
Add `@evalraw` synthax for programmatically generate raw contents

disberd:
This PR is motivated by the discussion started in [this discourse thread](https://discourse.julialang.org/t/create-raw-blocks-programmatically-from-documenter-jl/101522/6)

It simply adds a new synthax block called `@evalraw` that allows to generate raw html (or LaTeX) contents from Julia that are then rendered as if they were written *verbatim* inside a
````
```@raw
contents...
```
````
block

I think this functionality can be very useful for generating arbitrary HTML from Julia, especially for people that do not have html/js expertise but can rely on packages that already provide HTML outputs from Julia.

The synthax is quite simple, I mostly reused the code for `@eval` blocks with the simple checks from `@raw` and just put a hard constraints that the result has to be a `String`, then the resulting string is used to generate a `Documenter.RawNode` element.

At the moment I did put a normal `error` when the result is not a string as I am not well versed in the Documenter internals and how the `@docerror` macro should be used.

If the maintainers are open to the idea of adding this functionality I will write tests and docs where needed for this.

### DIFF
--- a/src/Expanders.jl
+++ b/src/Expanders.jl
@@ -180,6 +180,7 @@ Markdown.parse("![Plot](plot.svg)")
 """
 abstract type EvalBlocks <: NestedExpanderPipeline end
 
+abstract type EvalRawBlocks <: NestedExpanderPipeline end
 abstract type RawBlocks <: NestedExpanderPipeline end
 
 """
@@ -246,7 +247,8 @@ Selectors.order(::Type{ContentsBlocks}) = 7.0
 Selectors.order(::Type{ExampleBlocks})  = 8.0
 Selectors.order(::Type{REPLBlocks})     = 9.0
 Selectors.order(::Type{SetupBlocks})    = 10.0
-Selectors.order(::Type{RawBlocks})      = 11.0
+Selectors.order(::Type{EvalRawBlocks})  = 11.0
+Selectors.order(::Type{RawBlocks})      = 12.0
 
 Selectors.matcher(::Type{TrackHeaders},   node, page, doc) = isa(node.element, MarkdownAST.Heading)
 Selectors.matcher(::Type{MetaBlocks},     node, page, doc) = iscode(node, "@meta")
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mortenpi** [comment]
I am not necessarily against this, and I think it's a use case we should support. However, I wonder if could re-use the at-eval block somehow, without having to create a new one?

On #master / 1.0, Documenter will be very picky about what objects you can create in an at-eval:

https://github.com/JuliaDocs/Documenter.jl/blob/7c97a86a31e360d7d22082a9a783b0cab24163b5/src/Expanders.jl#L606-L629

So I think we could have some sort of an interface where a special returned object gets rendered as HTML or something along those lines. It could even support the case where a single block gets included in different formats depending on the writer (e.g. HTML or LaTeX). Not 100% sure what the interface should be right now though.

**Comment by disberd** [comment]
Hi @mortenpi, thanks for the fast reply
What about something along these lines?
```julia
      result = if isnothing(result)
          nothing
      elseif isa(result, Base.HTML)  # Added branch
          node.element = Documenter.RawNode(:html, result)
          return
      elseif isa(result, Markdown.MD)
          convert(Node, result)
      else
```

I do not know exactly how the Writers work and how the LaTeX part is handled.
The nice thing about `Base.HTML` is that in in Base so I believe it could be a good and safe type to use for checking whether this should became a rawnode.

I don't know if there is something similar for LaTeX

fredrikekre:
Could this not be handled similarly to the result from `@example` blocks?

mortenpi:
I am hesitant to just rely on the the `show` methods like at-example does. In particular, for `Markdown.MD`, that I think would change the behaviour. Right now, we incorporate the actual AST into the Documenter AST, and so it gets rendered "natively" (i.e. exactly as Documenter would render that Markdown). If you change to `show` methods, then it would use the `Markdown` stdlibs HTML rendering, which is different from Documenter's.

I also don't think we should special case `Markdown.MD`, because that creates complexity, and we almost certainly want to "special case" other things in the future (e.g. `MarkdownAST.Node`s), but using `show` methods implies that we should treat all objects the same.

We _could_ still have a way to explicitly tell Documenter to fall back to `show` methods though, e.g. with something like this:

`````markdown
```@eval
result = f()
Documenter.UseShowMethods(result)
```
`````

So I would advocate being explicit here somehow. `Base.HTML` is kind-of deprecated and an unfortunate artifact that should not really be used.. but if stick to at-eval only accepting certain types, we _could_ still add support for that in that `if` -- it wouldn't really hurt, I don't think.

disberd:
Hi @mortenpi,

would you have a chance to review the latest changes and give some feedback on whether the approach here is on the right track?

fingolfin:
Sadly this PR has been a bit neglected and now has a bunch of conflicts...

On the upside, we maybe have new opportunities: I am in the process of adding uniform parsing of the "language" part of codeblocks; we already have `@eval NAME` now; we could easily add `@eval NAME ; key1 = value1, key2 = value2` syntax. This then could be used to do things like this:
````md
```@eval ; format=:html, raw=true
````
to signal: this eval block is only mean to be used if the output format is `html`; and the output is "raw" (meaning it should be rendered using `show` / `sprint`). Of course one could also do something like `render=show` vs. `render=markdown` or whatever.

PR #2184, opened 2023-07-15
Change the changelog infrastructure

fredrikekre:
This patch changes how the changelog is written in order to support links in both the CHANGELOG.md as viewed on GitHub, and in the generated HTML. The changelog related code is moved into docs/changelog.jl which now has two purposes: i) generate the Documenter friendly markdown for the manual, and ii) generate links to add to the footer of the CHANGELOG.md file directly.


fredrikekre:
I was getting annoyed by the missing links when viewing the changelog directly on github. After this patch, entries in the changelog should be of the form `[#XXX]` instead of `(#XXX)`. The list of links at the bottom of CHANGELOG.md can be updated by running eg
```
julia -L docs/changelog.jl -e 'write_link_footer(; changelog_inputfile = "CHANGELOG.md", repo = "JuliaDocs/Documenter.jl")'
```

docs/changelog.jl
```diff
@@ -0,0 +1,134 @@
+# =================================================================================
+#  Utilities for managing CHANGELOG.md and generating a Documneter friendly version
```
mortenpi, line 2:
```suggestion
#  Utilities for managing CHANGELOG.md and generating a Documenter friendly version
```

docs/changelog.jl
```diff
@@ -0,0 +1,134 @@
...
+    open(changelog_outputfile, "w") do io
+        write(io, header)
+        write(io, content)
+    end
+
+    return
+end
+
+# Collect and add all links to the CHANGELOG.md footer.
+# julia -L docs/changelog.jl -e 'write_link_footer(; changelog_inputfile = "CHANGELOG.md", repo = "JuliaDocs/Documenter.jl")'
```
mortenpi, line 89: Any thoughts on having like an `isinteractive() || main(ARGS)` type pattern, so that you wouldn't have to copy this whole thing into the terminal? Although that would clash with the `include` in `make.jl` though (in case that file is loaded in an interactive session)..

docs/changelog.jl
```diff
@@ -0,0 +1,134 @@
...
+    # Description: Replace version headers with a link to the GitHub release
+    # Example: ## Version [v0.27.0] -> https://github.com/JuliaDocs/Documenter.jl/releases/tag/v0.27.0
+    for m in eachmatch(r"^\#\# Version (?<token>\[(?<tag>v[0-9]+.[0-9]+.[0-9]+)\])"m, content)
+        linkmap[m["token"]] = "https://github.com/$(repo)/releases/tag/$(m["tag"])"
+    end
+
+    return linkmap
+end
+
+# Rewrite CHANGELOG.md to a Documenter friendly markdown file
```
mortenpi, line 52:
```suggestion
# Rewrite CHANGELOG.md to a Documenter-friendly markdown file
```

mortenpi (commented):
LGTM, although it would be good to get @odow's thoughts here as well. Having a `Makefile` to easily trigger this seems like a good idea though.

docs/changelog.jl
fredrikekre: `make changelog` seems easier.

docs/changelog.jl
```diff
@@ -0,0 +1,138 @@
+# =================================================================================
+#  Utilities for managing CHANGELOG.md and generating a Documenter-friendly version
+# =================================================================================
```
odow, line 3: it seems like this file could even be in `src` so it can be re-used by any project using Documenter?

mortenpi: I think this should be re-used, yes, but I lean towards this being in a separate package, so that it could be versioned independently and hopefully more rapidly than Documenter itself.

That said, we _could_ still drop it into `src/` if you want to re-use this for e.g. JuMP right away, and just not promise this as official API (but secretly promise to you that we won't drop it on the `1.0.x` branch, for example)?

mortenpi: Or, to be fair, if we generally agree that this should be a separate package, we could just copy-paste it into that new package now, and tag that as 0.1.0, and then iterate on that as needed.

mortenpi:
I went ahead and merged this as is, so that we wouldn't have to deal with conflicts dues to the formatting changes in the CHANGELOG. We can always shuffle the code around in a follow-up PR.

mortenpi:
Just used `make changelog` in #2187 and it's amazing, particularly when compared against the old way where you had to manage those references by hand (pre Oscar's changes) :rocket::rocket::rocket:

fredrikekre:
FYI, I started working on https://github.com/JuliaDocs/Changelog.jl so we can maybe move to that once that is finished.

PR #2194, opened 2023-07-26
Replace `strict` with `warnonly` and make `makedocs` builds fail by default

### LINKED ISSUE #2051: strict=true & enable linkcheck by default
And any other checks we might have. This is to encourage people to fix the errors they have in their docs, and in particular to catch any new ones that pop up on CI. It's quite fair to say that no-one checks the warnings on CI.

This well be very breaking for existing workflows though.

X-ref: https://julialang.slack.com/archives/C67TK21LJ/p1676982324363569

mortenpi:
Close #2051.


mortenpi:
This should be good to go now.

As for `linkcheck`: I've been going back and forth in my head about this, but my feeling right now is that we should keep that as `false` by default, because it's a bit flaky and also environment dependent. The checks that are on by default should be consistent. So the user should consciously opt-in to the online linkchecks.

One thing to note, though, is that if you set `linkcheck = true`, then linkcheck failures now do fail the build (unless you set `warnonly = :linkcheck`).

Issue #2103, opened 2023-04-16
Docstring collapse/expand icon not changing when clicking rapidly


mortenpi:
I was playing with the collapse/expand feature a bit, and I noticed that the icon doesn't change correctly if you click on it rapidly a bunch of times:

https://user-images.githubusercontent.com/147757/232263663-b283ed48-0777-4ea0-b1a3-049cb2544b75.mp4

Extremely minor, unlikely to be run into in practice. But might indicate some sort of a race condition or something in the JS.

cc @Hetarth02

Hetarth02:
I think we can put sort of a safety lock so that it doesn't get triggered when its already in triggered(animating) state.

PR #2027, opened 2023-01-29
Added keyboard shortcut for search box

Hetarth02:
Also, added .vscode to .gitignore
Link to keyboard shortcut is [here](https://github.com/JuliaDocs/Documenter.jl/issues/1536).

To focus into search bar press `ctrl` + `/` or `cmd` + `/` keys.
To focus out of search bar press `Esc` key.

_Edit by @mortenpi: close #1536._

PR #2202, opened 2023-08-06
New Search Modal UI

### LINKED ISSUE #1437: Show more than just a single word on search results page
I think the searchability with documenter could be improved by doing including more info (like a search engine) in the results page. Currently, this is what shows when I search for `sin` in the Julia Docs: 
Screen Shot 2020-10-16 at 5 37 03 AM

It's not terrible but I have to click around a lot and not really know if what I am clicking on is what I actually need. I propose something like this: 
Screen Shot 2020-10-16 at 5 37 03 AM
where the actual info from the page the link goes to is previewed on the search page either to the right or below the title of the item. 

I think this will significantly improve the ability for search to be done effectively with documenter.

### LINKED ISSUE #2147: Updating the search UI
Currently the search UI looks like this,

This doesn't display much information about the search results.

I think that the UI should display more information regarding search results like, what category is this from, where did the term match.

Moreover, rather than having a separate page for searching we should have something like a search modal which can be accessible from anywhere. This would be a lot helpful. We can take inspiration from other documentations and how have they implemented their search UI.

[Issue #1437](https://github.com/JuliaDocs/Documenter.jl/issues/1437), [Issue #1986](https://github.com/JuliaDocs/Documenter.jl/issues/1986): These issues also address that the search results should display some more information rather than just the title of the search result.

Hetarth02:
_Edit by @mortenpi: added the TODO list:_
_Last Edit by @Hetarth02 : 2023-08-20:_

- [x] Have the modal be full size all the time on desktop too.
- [x] Distinguish the `search-result-title` somehow (`font-weigth: 600`).
- [x] Have the `search-result-title` for code items (like function names and such; anything except `Page`, `Section`) in monospace (but make font a bit smaller, to match what we do in main body).
- [x] Add section titles to search results (add to the bottom near the URL somehow).
- [x] Fix the bug with some URLs incorrectly being cut off (search "parametric")
- [x] Remove / replace the "No recent searches!" string, as it's confusing. Replace with `Type something to get started`.
- [x] Get rid of the search input box (or make it inactive, with click triggering the search modal)

_Edit by @mortenpi:_ close #1437, close #2147

---

Well, the search UI has had a complete overhaul, with a fresh new modal UI s mentioned in Issue #2147 with better context for search results and a filter mechanism to remove those pesky unwanted results.

Here is the full demo,

https://github.com/JuliaDocs/Documenter.jl/assets/80420115/ad7ebc4e-4148-49b6-ba53-8a3baa39cea8

- Additionally, the modal is mobile responsive!

Search will now look like this,

### Light Mode:
Without Filter:
image

With Filter: 
image

### Dark Mode:
Without Filter:
image

With Filter: 
image

~Since these we very small fixes, I did them while we are updating the search UI altogether.~
~- #2215 Removed some extra space between navbar header icons.~
image

~- #2103: Fixed, docstring collapse/expand icon not changing when clicking rapidly. Now click all you want!~


Hetarth02:
I think the only thing to do now would be,

- To add a search icon/input box in the navbar
- Remove the search input from the sidebar

mortenpi:
Awesome! A couple of thoughts:

* Let's reduce the top and bottom margins. Something on the order of `1-3rem` should be good.
* Let's also not resize the height of the modal with content. Let's have it be full height, even if there are no results (and then have the overflow scrollbar as needed, just like now).
* On mobile the modal could actually be full-screen with an `X` somewhere? This is just a thought though.
* Let's distinguish the `search-result-title` somehow (but not with a badge :smile:). Maybe just bold, or `<code>`/monospace (I guess code might not be appropriate for page/section titles though).

Also cc @fredrikekre -- what are your thoughts on this UX in general (i.e. replacing a separate search page with a modal)?

The other two fixes -- could you open separate PRs for those please?

sloede:
Great work! From the invitation on Slack to give feedback, I took the demo site for a test drive, and these are the things I noticed (using Firefox on an M1 Macbook):

1. GitHub recently changed their search shortcut to press just `/` (or was it always that way and I just didn't know 🤔?). Why not use the same here (`/` instead of `Ctrl+/`) to make it easier on our muscle memory?
2. The modal is IMHO a little too generous with vertical white space. I'd prefer a more compact result such that it is easier to scan for results further down, especially when I already know more or less what I am looking for.
3. At the moment there is very little to go for when trying to decide which result is the best. Maybe showing a little snippet with each result would be helpful? This might be out of scope for the PR though.
4. Compared to the current search info, some information seems to be missing. For example, the previous search for "parametric" also showed the section name for each result.
5. Right now, all filters are shown no matter what. Would it be possible to show only those filters that would return a non-empty set of search results?
6. Some links seem to be broken. For example, when I search for "parametric", the first result "Troubleshooting"... leads to `manual/faq.html#Troubleshooting-`, while the correct link should read `manual/faq.html#Troubleshooting-"method-not-matched":-parametric-type-invariance-and-MethodErrors`. This used to work before.
7. The order of the results seems to have changed, at least compared to https://docs.julialang.org/en/v1/. Not a problem, I just wasn't sure if this is intentional or not.
8. While an interactive modal dialogue is nice from a UI perspective, it also has some disadvantages. For example, it is not possible to bookmark the result of a search, or share it with someone else, since no URL exists. Also, when closing/reopening tabs, the search is lost. This is a structural disadvantage of modals, so nothing you can do about it per se. But maybe it would be possible to combine the best of both worlds, by allowing a user to hit "Enter" and then open a static search page (similar to how GH does it)? Then again, this may be out of scope for this PR.

image

Please don't take these points as criticism but rather as suggestions for further improvements. As stated above, overall great work!

KristofferC:
Grabbag of unfiltered thoughts trying it out.

------

When I click down on the text box the browser shows some auto complete entries until I do mouse down and then they disappear and the text box shows up. It also somehow feels quite surprising to have a input text box spawn a new floating window.

image

-------

These files like they should be code quoted like they are in the manual itself.

image

-----

The scroll bar is very narrow for me and hard to click on with the mouse. If I am on the right half of the scroll bar it seems to not reach and I have to position the mouse on the left-most part of it

image

------------

There is no need for the ... here? ctime is the full name of the function?

image


----

Is there any value in the short cut command for search being shown here? To me it makes it feels like it makes sense to use that short cut at this current moment

image

----

The recent searches list doesn't seem to ever show anything

image

Hetarth02:
Edit: 2023-08-18

Made an Issue regarding extra spacing and the PR for the same.
~#2215 Removed some extra space between navbar header icons.~
**Moved to #2216.**

~#2103: Fixed, docstring collapse/expand icon not changing when clicking rapidly. Now click all you want!~
**Moved to #2217**

Hetarth02:
@sloede 
> 1. GitHub recently changed their search shortcut to press just `/` (or was it always that way and I just didn't know 🤔?). Why not use the same here (`/` instead of `Ctrl+/`) to make it easier on our muscle memory?

We had discussion over this before. #2027

> 2. The modal is IMHO a little too generous with vertical white space. I'd prefer a more compact result such that it is easier to scan for results further down, especially when I already know more or less what I am looking for.

Do you mean the gap between the search results?

> 3. At the moment there is very little to go for when trying to decide which result is the best. Maybe showing a little snippet with each result would be helpful? This might be out of scope for the PR though.

While, I do not understand what you exactly mean but by hunch do you want to display a code snippet or something like that? If so, that would require some thinking(yeah, out of scope for this PR 😅)

> 4. Compared to the current search info, some information seems to be missing. For example, the previous search for "parametric" also showed the section name for each result.

We have opted to go with, title, page link and the category as for now. @mortenpi Let me know what are your thoughts on this?

> 5. Right now, all filters are shown no matter what. Would it be possible to show only those filters that would return a non-empty set of search results?

Actually, this was my initial thought but I scraped it because as per the current UI the filters changed too much(every 2 or 3 keystrokes) which can sometimes be jarring for someone, Hence I opted out from it.

> 6. Some links seem to be broken. For example, when I search for "parametric", the first result "Troubleshooting"... leads to `manual/faq.html#Troubleshooting-`, while the correct link should read `manual/faq.html#Troubleshooting-"method-not-matched":-parametric-type-invariance-and-MethodErrors`. This used to work before.

This is on purpose, because if I try to display the full link as it is in `<pre>` tag sometimes big links it would break the modal UI.

> 7. The order of the results seems to have changed, at least compared to https://docs.julialang.org/en/v1/. Not a problem, I just wasn't sure if this is intentional or not.

This might be due to change in search engine. #2172

> 8. While an interactive modal dialogue is nice from a UI perspective, it also has some disadvantages. For example, it is not possible to bookmark the result of a search, or share it with someone else, since no URL exists. Also, when closing/reopening tabs, the search is lost. This is a structural disadvantage of modals, so nothing you can do about it per se. But maybe it would be possible to combine the best of both worlds, by allowing a user to hit "Enter" and then open a static search page (similar to how GH does it)? Then again, this may be out of scope for this PR.

We can add features like bookmarks and search history, essentially storing it in the local storage of the browser but local storage has character limitations which vary based on devices.

Hetarth02:
> When I click down on the text box the browser shows some auto complete entries until I do mouse down and then they disappear and the text box shows up. It also somehow feels quite surprising to have a input text box spawn a new floating window.
> 
> image
> 

Look forward to future PRs, this might be removed entirely and replaced with a icon or input tag in the top nav.

> These files like they should be code quoted like they are in the manual itself.
> 
> image
> 

@mortenpi If we were to do this, is there some way to know if some title is code quoted or not?

> The scroll bar is very narrow for me and hard to click on with the mouse. If I am on the right half of the scroll bar it seems to not reach and I have to position the mouse on the left-most part of it
> 
> image
> 

Will be fixed.

> There is no need for the ... here? ctime is the full name of the function?
> 
> image
> 

Will look into that.

> Is there any value in the short cut command for search being shown here? To me it makes it feels like it makes sense to use that short cut at this current moment
> 
> image
> 

No, those are for informative purposes only.

> The recent searches list doesn't seem to ever show anything
> 
> image

This is just a placeholder. There is no feature for storing search history as of now. 😓

Hetarth02:
> * Let's reduce the top and bottom margins. Something on the order of `1-3rem` should be good.
Done

> * On mobile the modal could actually be full-screen with an `X` somewhere? This is just a thought though.
Added the "X", but it is not full-screen.

> * Let's distinguish the `search-result-title` somehow (but not with a badge 😄). Maybe just bold, or `<code>`/monospace (I guess code might not be appropriate for page/section titles though).
Suggestions are appreciated... 😄 

> The other two fixes -- could you open separate PRs for those please?
Done

mortenpi:
> > 7\. Compared to the current search info, some information seems to be missing. For example, the previous search for "parametric" also showed the section name for each result.
> 
> We have opted to go with, title, page link and the category as for now. @mortenpi Let me know what are your thoughts on this?

I think this is a fair point. It would be good to keep that information somewhere in the new UI.

> > 15\. While an interactive modal dialogue is nice from a UI perspective, it also has some disadvantages. For example, it is not possible to bookmark the result of a search, or share it with someone else, since no URL exists. Also, when closing/reopening tabs, the search is lost. This is a structural disadvantage of modals, so nothing you can do about it per se. But maybe it would be possible to combine the best of both worlds, by allowing a user to hit "Enter" and then open a static search page (similar to how GH does it)? Then again, this may be out of scope for this PR.
> 
> We can add features like bookmarks and search history, essentially storing it in the local storage of the browser but local storage has character limitations which vary based on devices.

I think this is also a good point. In the current implementation, you can use the `?q=` query parameter, e.g.: https://docs.julialang.org/en/v1/search/?q=ccall

That said, I just noticed it's not fully working -- if you update the search query while on the search site, it doesn't update the URL.

All in all, I think this would be nice to have, but is probably not critical. And since the current implementation is only half working, then removing this is maybe not even a regression. It would also be good to know if there are any other modal-ish searches that have a way of linking to searches.

I don't think we want to implement any UI here specifically for bookmarking though -- that should be left to the browsers etc.

> Look forward to future PRs, this might be removed entirely and replaced with a icon or input tag in the top nav.

I think replacing the search input box should be done as part of this PR, since the input box doesn't really make sense if we have the modal.

>> Right now, all filters are shown no matter what. Would it be possible to show only those filters that would return a non-empty set of search results?
>
> Actually, this was my initial thought but I scraped it because as per the current UI the filters changed too much(every 2 or 3 keystrokes) which can sometimes be jarring for someone, Hence I opted out from it.

Let's gray out the irrelevant filters maybe? That would be less jarring than DOM elements hopping in and out of existence.

> @mortenpi If we were to do this, is there some way to know if some title is code quoted or not?

I think we have to look at the class? Functions and types would be code, sections and such would be text. It's also possible to have code in section titles, but let's ignore that for now.

>> Some links seem to be broken. For example, when I search for "parametric", the first result "Troubleshooting"... leads to manual/faq.html#Troubleshooting-, while the correct link should read manual/faq.html#Troubleshooting-"method-not-matched":-parametric-type-invariance-and-MethodErrors. This used to work before.
>
> This is on purpose, because if I try to display the full link as it is in <pre> tag sometimes big links it would break the modal UI.

There's a bug here too -- note the `href` here which is wrong (it's cut off).

```
<a href="./manual/faq.html#Troubleshooting-" method-not-matched":-parametric-type-invariance-and-methoderrors"="" class="search-result-link w-100 is-flex is-flex-direction-column gap-2 p-4">
        <div class="w-100 is-flex is-flex-wrap-wrap is-justify-content-space-between is-align-items-center">
          <div class="search-result-title has-text-weight-semi-bold">Troubleshooting "method not matched": parametric type invariance and MethodErrors</div>
          <div class="property-search-result-badge">section</div>
        </div>
        <p>
          
        </p>
        <div class="has-text-left" style="font-size: smaller;" title="manual/faq.html#Troubleshooting-" method-not-matched":-parametric-type-invariance-and-methoderrors"="">
          <i class="fas fa-link"></i> manual/faq.html#Troubleshooting-"method-not-matche...
        </div>
      </a>
```

Of course, Documenter also generates pretty crazy ids/fragments sometimes. But I think it _should_ work as is, and point to `https://mortenpi.eu/juliadocs-demos/2023-08-10-search-demo/manual/faq.html#Troubleshooting-%22method-not-matched%22:-parametric-type-invariance-and-MethodErrors`.

sloede:
@Hetarth02 

> > 2. The modal is IMHO a little too generous with vertical white space. I'd prefer a more compact result such that it is easier to scan for results further down, especially when I already know more or less what I am looking for.
> 
> Do you mean the gap between the search results?

Yes. There is a lot of vertical whitespace that IMHO would be put to better use by being able to view more results at once (i.e., by reducing the vertical whitespace).

> > 3. At the moment there is very little to go for when trying to decide which result is the best. Maybe showing a little snippet with each result would be helpful? This might be out of scope for the PR though.
> 
> While, I do not understand what you exactly mean but by hunch do you want to display a code snippet or something like that? If so, that would require some thinking(yeah, out of scope for this PR 😅)

For example, [Read the Docs](https://docs.readthedocs.io/en/stable/tutorial/index.html?rtd_search=search) provides a small glimpse at the _content_ of the search result, such that one can more easily decide which of the results is likely to be most useful to the user:
image

> > 6. Some links seem to be broken. For example, when I search for "parametric", the first result "Troubleshooting"... leads to `manual/faq.html#Troubleshooting-`, while the correct link should read `manual/faq.html#Troubleshooting-"method-not-matched":-parametric-type-invariance-and-MethodErrors`. This used to work before.
> 
> This is on purpose, because if I try to display the full link as it is in `\<pre\>` tag sometimes big links it would break the modal UI.

The _displayed_ part of the link, i.e., the `123` in `<a href="abc">123</a>`, should be shortened, yes, but not the actual URL in the link (`abc` in this example), right? Otherwise, clicking on those links might become useless very quickly.
 

> > 8. While an interactive modal dialogue is nice from a UI perspective, it also has some disadvantages. For example, it is not possible to bookmark the result of a search, or share it with someone else, since no URL exists. Also, when closing/reopening tabs, the search is lost. This is a structural disadvantage of modals, so nothing you can do about it per se. But maybe it would be possible to combine the best of both worlds, by allowing a user to hit "Enter" and then open a static search page (similar to how GH does it)? Then again, this may be out of scope for this PR.
> 
> We can add features like bookmarks and search history, essentially storing it in the local storage of the browser but local storage has character limitations which vary based on devices.

No, don't store anything in the browser. But maybe you can do something like [Read the Docs](https://docs.readthedocs.io/en/stable/tutorial/index.html?rtd_search=search) does: When you type in a search term, e.g., "search" (see picture above), they also automatically update the URL:
image

This allows one to close the browser tab and reopen it, and it will be exactly at the same place again. Also, it allows me to copy the URL and share it with someone else (like [so](https://docs.readthedocs.io/en/stable/tutorial/index.html?rtd_search=search)) and they will be able to access the same search results.

sloede:
@mortenpi 
> All in all, I think this would be nice to have, but is probably not critical. And since the current implementation is only half working, then removing this is maybe not even a regression. It would also be good to know if there are any other modal-ish searches that have a way of linking to searches.

Yes, e.g., [Read the Docs](https://docs.readthedocs.io/en/stable/tutorial/index.html?rtd_search=search).

> I don't think we want to implement any UI here specifically for bookmarking though -- that should be left to the browsers etc.

I agree and this is not what I had in mind. I hope in my answer above it becomes clearer.

mortenpi:
> > > 5\. At the moment there is very little to go for when trying to decide which result is the best. Maybe showing a little snippet with each result would be helpful? This might be out of scope for the PR though.
> > 
> > 
> > While, I do not understand what you exactly mean but by hunch do you want to display a code snippet or something like that? If so, that would require some thinking(yeah, out of scope for this PR 😅)
> 
> For example, [Read the Docs](https://docs.readthedocs.io/en/stable/tutorial/index.html?rtd_search=search) provides a small glimpse at the _content_ of the search result, such that one can more easily decide which of the results is likely to be most useful to the user:

Hethart's implementation does show context in some cases now:

image

Although it does looks like for the Julia manual a lot of them are missing somehow. @Hetarth02 any ideas why? 

In some cases we're probably missing the context in the index. E.g. if we're matching just the title or function name, we should probably show like the first bit of the section or docstring. But I don't think we have that information currently in the index. That should be looked at in a follow-up PR though -- let's not change the index here.

mortenpi:
> > It would also be good to know if there are any other modal-ish searches that have a way of linking to searches.
> 
> Yes, e.g., [Read the Docs](https://docs.readthedocs.io/en/stable/tutorial/index.html?rtd_search=search).

Ah, cool. Yeah, I think we can take inspiration from that, and just keep a query parameter (either in this PR, or a follow-up).

Side note: RTD also uses `/` for the search hotkey.

Hetarth02:
Edit: 2023-08-20

@mortenpi 

- [x] Have the modal be full size all the time on desktop too.

Desktop:
image

Mobile:
image

- [x] Distinguish the `search-result-title`. (Made it bold)

image

- [x] Have the `search-result-title` for code items (like function names and such; anything except `Page`, `Section`) in monospace.

image

- [x] Add section titles to search results (add to the bottom near the URL somehow).

image

- [x] Remove / replace the "No recent searches!" string, as it's confusing. Replace with `Type something to get started`.

image

- [x] Get rid of the search input box (Replaced the input with `div`).
- [x] Fix the bug with some URLs incorrectly being cut off (search "parametric")

This should be working, I have tested it locally but it would be helpful if someone can verify this with different urls.

thazhemadam:
After you have already searched for something and navigated to the respective documentation, if you search for something else that falls in the same section, the search modal doesn't go away. The documentation itself apparently navigates properly within the same section, but the search modal still stays on the screen.
This is not consistent with the behavior observed while searching for and navigating to different sections of the documentation (which imo is a nicer user experience - not having to manually close the search modal every time they search something).

mortenpi:
One more option for what to put in the search modal when it's not showing any results :stuck_out_tongue: 

image

mortenpi:
> After you have already searched for something and navigated to the respective documentation, if you search for something else that falls in the same section, the search modal doesn't go away.

Good catch! I can also replicate this. cc @Hetarth02 -- this would be good to get fixed.

assets/html/scss/documenter/layout/_search.scss
```diff
@@ -1,69 +1,133 @@
...
+.search-min-width-50 {
+  min-width: 50%;
+}
+
+.search-min-height-100 {
+  min-height: 100%;
+}
+
+.search-modal-card-body {
+  max-height: 500px;
```
mortenpi, line 26: I don't see why we wouldn't want to fill up pretty much the whole vertical space? I was thinking maybe we could do about `margin-top/bottom` instead of `max-height`, but this would work too:

```suggestion
  max-height: calc(100vh - 15rem);
```

image
image

assets/html/scss/documenter/layout/_search.scss
```diff
@@ -1,69 +1,133 @@
...
+  border-bottom: 1px solid $border;
+}
+
+.search-result-title {
+  width: 85%;
+  color: $search-result-title-text-color;
+}
+
+.search-result-code-title {
+  font-size: 15px;
```
mortenpi, line 94: This applies to multiple places, but let's not use `px` units, but instead use `rem` (except for like borders, shadows, rounded corners and such). Among other things, that (should be) consistent with how the rest of SCSS is written.

Also, this should be the equivalent of `14px` (to be consistent with the main body). We can probably just replicate the `rem`s of the main body (there should be a variable for that).

assets/html/scss/documenter/layout/_search.scss
```diff
@@ -1,69 +1,133 @@
...
+  gap: 0.5rem;
+}
+
+.gap-4 {
+  gap: 1rem;
+}
+
+.gap-8 {
+  gap: 2rem;
 }
```
mortenpi, line 133: We can leave this as is, but in general I am not a fan of these kinds of helpers -- essentially you're using HTML to do the styling again.

src/html/HTMLWriter.jl
```diff
@@ -1090,14 +1092,7 @@ function render_sidebar(ctx, navnode)
...
     push!(navmenu.nodes,
-        form[".docs-search", :action => navhref(ctx, ctx.search_navnode, navnode)](
-            input[
-                "#documenter-search-query.docs-search-query",
-                :name => "q",
-                :type => "text",
-                :placeholder => "Search docs (Ctrl + /)",
-            ],
-        )
+        div["#documenter-search-query.docs-search-query.input.is-rounded.is-small.is-clickable.my-2.mx-auto.py-1.px-2"]("Search docs (Ctrl + /)")
```
mortenpi, line 1095: Semantically, wouldn't a `<button>` be more suitable here?

Also, let's merge all the CSS into `#documenter-search-query`. I don't think we reuse `.docs-search-query` anywhere?

CHANGELOG.md
```diff
@@ -123,6 +123,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * The client-side search engine has been changed from LunrJs to MinisearchJs. Additionally, the search results will now contain additional context and have an improved UI. ([#2141])
 
+* The search UI has had a complete overhaul, with a fresh new modal UI with better context for search results and a filter mechanism to remove those pesky unwanted results. ([#2147])
```
mortenpi, line 126:
```suggestion
* The search UI has had a complete overhaul, with a fresh new modal UI with better context for search results and a filter mechanism to remove unwanted results. The client-side search engine has been changed from LunrJs to MinisearchJs. ([#2141], [#2147])
```

mortenpi (commented):
Nice! A few more comments, now that I'm looking more closely at the code:

- Is there any reason to have `shortcut.js` separate from `search.js`? Although, maybe the files are getting big enough that it does make sense to split them up.
- I think we also still need to remove the generation of `search/index.html`? I didn't see that change in the Julia code?

mortenpi:
Here's an updated full-site demo: https://mortenpi.eu/juliadocs-demos/2023-08-21-search-demo/

assets/html/scss/documenter/layout/_search.scss
Hetarth02: I find these kinds of helper more useful, maybe it has to do something with me using bootstrap normally...

Hetarth02:
@mortenpi 

> * Is there any reason to have `shortcut.js` separate from `search.js`? Although, maybe the files are getting big enough that it does make sense to split them up.

If possible, let us keep this as it is for the moment.

> * I think we also still need to remove the generation of `search/index.html`? I didn't see that change in the Julia code?

Done

Hetarth02:
@mortenpi @thazhemadam 
> > After you have already searched for something and navigated to the respective documentation, if you search for something else that falls in the same section, the search modal doesn't go away.
> 
> Good catch! I can also replicate this. cc @Hetarth02 -- this would be good to get fixed.

Done!

mortenpi (approved):
LGTM! In my opinion we can merge this. Just needs the CSS files to be updated.

mortenpi (approved):
Awesome! Thanks a lot @Hetarth02 for iterating on this! And thanks everyone for the feedback!

PR #2217, opened 2023-08-17
Fixed docstring collapse/expand icon not changing when clicking rapidly

Hetarth02:
Now click all you want!

_Edit by @mortenpi: close #2103_


CHANGELOG.md
```diff
@@ -131,6 +131,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Docstring doctests now properly get checked on each `makedocs` run, when run multiple times in the same Julia session. ([#974], [#1948])
 
+* Fixed, docstring collapse/expand icon not changing when clicking rapidly. Now click all you want! ([#2103])
+
```
mortenpi, line 135:
```suggestion
```

CHANGELOG.md
```diff
@@ -1619,6 +1621,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2081]: https://github.com/JuliaDocs/Documenter.jl/issues/2081
 [#2085]: https://github.com/JuliaDocs/Documenter.jl/issues/2085
 [#2100]: https://github.com/JuliaDocs/Documenter.jl/issues/2100
+[#2103]: https://github.com/JuliaDocs/Documenter.jl/issues/2103
```
mortenpi, line 1624:
```suggestion
```

mortenpi (commented):
This shouldn't have a CHANGELOG because it's not a change relative to a previous versions (though we could add the issue and PR reference to the original changelog entry).

mortenpi:
So two things I notice here now:

1. There is now this weird lag after you click (once) which doesn't seem ideal. I think it should react immediately to the first click and then ignore the following ones.
2. If you're now clicking rapidly, nothing happens at all until you stop. This is not a big deal though, since there's not much reason to normally rapidly click anyway.

Hetarth02:
> 1. There is now this weird lag after you click (once) which doesn't seem ideal. I think it should react immediately to the first click and then ignore the following ones.

This is because it is wrapped in the debounce function with a 300ms delay. And debounce runs everytime, so best we can do is decrease the delay to 100ms.

> 2. If you're now clicking rapidly, nothing happens at all until you stop. This is not a big deal though, since there's not much reason to normally rapidly click anyway.

Again, due to debounce.

mortenpi:
> This is because it is wrapped in the debounce function with a 300ms delay. And debounce runs everytime, so best we can do is decrease the delay to 100ms.

I am not sure the lag is an improvement over the current status quo. The bug with rapid clicking is an unlikely edge case, but the lag is something the users will run into every time they collapse/open something. And reducing the debounce time doesn't seem like the right approach either. As I said, I think the UI should react right away to the first click, and then debounce the following clicks.

Hetarth02:
Ok, maybe a mix of throttling and debounce could be the way to go. Let me see what I can do.

(As far as I know, lodash debounce has some options relating to this.)

Hetarth02:
@mortenpi Would you be able take a look?

mortenpi (approved):
LGTM, thanks @Hetarth02!

PR #2223, opened 2023-08-18
fixed text not in warning box

rroelz:
Fixed a warning box in hosting.md that had the text displayed under instead of in it.


mortenpi:
Thank you!

Issue #2143, opened 2023-06-15
An option to write at-example outputs to file


mortenpi:
Related to #2142, but one way to manage huge figures is to have a way for HTMLWriter to put them in files. I imagine this probably being either a global HTMLWriter option.. or possibly per-block. Needs some thought on how we want to handle filenames etc.

Also, I think this will be limited to only some MIME types.. and now that I think about it, I am not sure we could use it for `text/html`, since there is no way to actually include it..? It could work for huge SVG images though.

cc @bauglir

bauglir:
The scenarios that I have seen that are causing issues are primarily the huge SVG images already mentioned, but also GIF animations that are base64 encoded into `img` tags.

> Needs some thought on how we want to handle filenames etc.

A very quick thought, but perhaps derive them from the source filename and keeping an incremental counter for that file is sufficient?

mortenpi:
Not to make this too complicated, but one more option here would be to have a threshold value in bytes, above which we start writing files. And we could make that user-configurable, so `0` means you write everything to file.. and `Inf` or `nothing` means everything goes into the HTML. And Documenter by default uses some "reasonable" default, so that we wouldn't write a gazillion files when you have a lot of tiny examples, but we would use files automatically if we go over that threshold.

> A very quick thought, but perhaps derive them from the source filename and keeping an incremental counter for that file is sufficient?

Yep, this seems pretty reasonable. Although, with pretty URLs, we have a whole directory to play with. So an image on `foo/bar.md` could become `foo/bar/example-0001.png` next to `foo/bar/index.html`. With non-pretty URLs, we could do `foo/bar-example-0001.png` next to `foo/bar.html`. And if it clashes with a user-provided file, we just bump the counter.

mortenpi:
One complication I discovered when implementing #2247 with Plots.jl: we prioritize the `text/html` show method over all the other ones (which is reasonable -- `text/html` is probably the best representation usually, and we _can_ display it, so we should prefer it):

https://github.com/JuliaDocs/Documenter.jl/blob/ae3a4a85e4002fa493a7326a7138c4fd2f9b4f53/src/html/HTMLWriter.jl#L2137-L2138

However, we can't write `text/html` to a file. This means that, in practice, most cases where we need #2247 (large figures), we still end up including the figure directly in the HTML. It seems like we'd somehow need to either automatically/automagically or with a toggle disable the `text/html` `show` method.

PR #2228, opened 2023-08-21
Set version to 1.0.0

mortenpi:
_Not meant to be just merged yet, opening this as a PSA._

The 1.0.0 release will contain quite a few breaking changes, so it would be good to get some feedback from the community. It would be amazing if people could play with `Documenter #master` for your own docs a bit (just `pkg> add Documenter#master` to `docs/Project.toml`), before we foist it on everyone with a release.

In particular, it would be good to know how painful upgrading to 1.0 is, and if there is anything that should be added to the release notes (including on how to upgrade). But also, any feedback or thoughts are very welcome.

* You can find the release notes here: https://documenter.juliadocs.org/dev/release-notes/#Unreleased
* Note that there are also a few outstanding issues [on the 1.0.0 milestone](https://github.com/JuliaDocs/Documenter.jl/milestone/107).


odow:
The JuMP docs are building on a very recent commit (https://github.com/JuliaDocs/Documenter.jl/tree/e2bd1940ad9a8ad1320252a2fc3b4468b770e3a4), so I'm happy.

I understand the motivation for the page size limit, but I still found it quite annoying.

mortenpi:
@odow What is the main source of the large pages in the JuMP manual? Is figured that we could fix with https://github.com/JuliaDocs/Documenter.jl/issues/2143?

odow:
It's both. We have some very long pages, and some pages with lots of figures and data tables.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## Version [v1.0.0] - 2023-08-??
```
```suggestion
## Version [v1.0.0] - 2023-09-15
```

PR #2231, opened 2023-08-23
docs: update remote links CHANGELOG entry

mortenpi:
@Datseris I slightly rephrased the "For upgrading" part of the repo link overhaul CHANGELOG entry. Do you think it improves it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Documenter is now also more strict about the cases where it is unable to determine the URLs, and therefore previously successful builds may break.
 
-  **For upgrading:** As the fallbacks to CI variables have been removed, make sure that you have your source checked out as a proper Git repository when building the documentation, so that Documenter could determine the repository link automatically from the Git `origin` URL. On GitHub Actions, this should normally already be the case. In other cases, you can configure the `repo` and/or `remotes` options to `makedocs` appropriately.
+  **For upgrading:** As the fallbacks to CI variables have been removed, make sure that you have your source checked out as a proper Git repository when building the documentation, so that Documenter could determine the repository link automatically from the Git `origin` URL. On GitHub Actions, this should normally already be the case, and user action is unlikely to be necessary. In other cases, if necessary, you can configure the `repo` and/or `remotes` keywords of `makedocs` appropriately.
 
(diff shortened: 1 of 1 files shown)


mortenpi:
> This won't work anymore, right?

Well, the `modules` part is unchanged, but you do need to make sure that the packages are `Pkg.develop`ed, not `Pkg.add`ed (or you set `remotes` by hand). That said, there is #2232.. so maybe we'll get a better approach for this.

Datseris:
I am confused. WHy do you need the remote for the packages? Is it so that one can access the source code? I thought this didn't work for "external" packages, clicking source. Does this mean that now clicking "source" button actually works on external packages as long as they are developed?

Datseris:
In any case, `develop`ing all included modules is totally fine for me. I anyways build all documentations from a centralized repository which is cloned by all packages, so for me it is a simple change of altering the build function to first develop all given modules.

(https://github.com/JuliaDynamics/doctheme  and add a  simple `Pkg.dev` loop over `modules` in this line: https://github.com/JuliaDynamics/doctheme/blob/master/build_docs_with_style.jl#L42 )

mortenpi:
> I thought this didn't work for "external" packages, clicking source. Does this mean that now clicking "source" button actually works on external packages as long as they are developed?

Yes!! E.g. you can look at the DocumenterTools docstrings here: https://documenter.juliadocs.org/dev/lib/public/#DocumenterTools

I'll just note here that the major motivator was actually the Julia standard libraries that are are hosted externally from the main Julia repo.

Datseris:
Isn't there the (potential) downside that the links to the functiosn are resolved based on the _development_ version of a package, while in the majority of times you would want the source code of the _latest stable_ to be linked?

Datseris:
Actually, this is an important point. Linking to stable versions is hugely important when building the documentation via a tagged release.

mortenpi:
> Isn't there the (potential) downside that the links to the functiosn are resolved based on the development version of a package, while in the majority of times you would want the source code of the latest stable to be linked?

Sure. So you need to make sure you have the tag checked out, no main/master :slightly_smiling_face: It's true that this does require a bit more care when adding the packages though, and this should maybe be emphasized in the docs. But also https://github.com/JuliaDocs/Documenter.jl/pull/2232 would pretty much solve that.

That said, this question about what version of the other package to link to is exactly why I am actually not at all a fan of including these "external" docstrings like this.

PR #2232, opened 2023-08-24
Automatically get remote for Pkg.add(...) packages

frankier:
This will only work for nicely tagged packages hosted on GitHub for now. It works by getting the repository URL and tree hash from data in the registry and getting the version number from the manifest. The GitHub API is then used to get a commit id from a tag corresponding to the version v$VERSION. These tags are the type created by the widely used `JuliaRegistries/TagBot`. This is robust against the tag being incorrect (e.g. if it was created manually but incorrectly) since the package tree hash is checked against the tree hash of the commit.

I think it would be nice to have this before 0.28 / 1.0.0 since everyone's missing remotes will start erroring then.

Happy to give this another pass if need be to try and address any concerns.

### DIFF
--- a/docs/src/lib/remote-links.md
+++ b/docs/src/lib/remote-links.md
@@ -34,8 +34,7 @@ To handle those cases, the `remotes` keyword to [`makedocs`](@ref) can be used t
 The local directory is assumed to correspond to the root of the Git repository, and any subpath within that directory is then resolved to the corresponding path in the remote repository.
 If there are nested `remotes` configured, Documenter will use the one that matches first as it walks up the directory tree from the original path.
 
-As the common case is a locally checked out Git repository cloned from GitHub, Documenter will also try to determine such remotes automatically.
-This is done regardless whether any additional remotes have been specified with `remotes` or not.
+As the common cases are a locally checked out Git repository (added with `Pkg.develop` to the docs environment), or a released package which is hosted on GitHub (`Pkg.add`ed to the environment), Documenter will also try to determine such remotes automatically.
 
 * When Documenter walks up the directory tree, it checks whether the directory is a root of a Git repository (by looking for the presence of a `.git` directory or file).
   Once it finds a valid local repository root, it tries to read its [`origin` remote URL](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes).
@@ -44,6 +43,13 @@ This is done regardless whether any additional remotes have been specified with 
     In this case, the remote should be configured explicitly with `remotes`.
 
 You can think of it as Documenter automatically populating `remotes` with any cloned GitHub repositories it finds.[^3]
+
+For released packages (those added using `Pkg.add(...)` rather than `Pkg.develop(...)`), the version and repository can be determined from the package metadata, but a commit hash is not readily available.
+In this case, Documenter will guess that a tag `v$VERSION` exists in the repository on GitHub.
+Note that these tags are created automatically by the widely used JuliaRegistries/TagBot action.
+Since this is sometimes not the case, and could cause dead or incorrect links, setting the `linkcheck` keyword to `true` to [`makedocs`](@ref) will check these guessed links have an existing target and that the existing target matches the published package.
+(Note this will also all other external links from your documentation.)
+Note that enabling this option can cause documentation builds to fail due to network errors or intermittent downtime of external services.
 
 !!! note
 
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -44,7 +44,7 @@ const NO_KEY_ENV = Dict(
 
 # Names of possible internal errors
 const ERROR_NAMES = [:autodocs_block, :cross_references, :docs_block, :doctest,
-                     :eval_block, :example_block, :footnote, :githubcheck, :linkcheck,
+                     :eval_block, :example_block, :footnote, :linkcheck_remotes, :linkcheck,
                      :meta_block, :missing_docs, :parse_error, :setup_block]
 
 """
--- a/src/Documenter.jl
+++ b/src/Documenter.jl
@@ -14,11 +14,14 @@ $(EXPORTS)
 module Documenter
 
 import AbstractTrees
+import Downloads
 import IOCapture
 import Markdown
 using MarkdownAST: MarkdownAST, Node
 import REPL
 import Unicode
+import Pkg
+import RegistryInstances
 # Additional imported names
 using Test: @testset, @test
 using DocStringExtensions
@@ -41,8 +44,8 @@ const NO_KEY_ENV = Dict(
 
 # Names of possible internal errors
 const ERROR_NAMES = [:autodocs_block, :cross_references, :docs_block, :doctest,
-                     :eval_block, :example_block, :footnote, :linkcheck, :meta_block,
-                     :missing_docs, :parse_error, :setup_block]
+                     :eval_block, :example_block, :footnote, :githubcheck, :linkcheck,
+                     :meta_block, :missing_docs, :parse_error, :setup_block]
 
 """
     abstract type Plugin end
--- a/src/documents.jl
+++ b/src/documents.jl
@@ -343,6 +343,8 @@ struct Internal
     indexnodes    :: Vector{IndexNode}
     locallinks :: IdDict{MarkdownAST.Link, String}
     errors::Set{Symbol}
+    src_to_uuid::Dict{String, Base.UUID} # These two are used to cache information from Pkg
+    uuid_to_version_info::Dict{Base.UUID, Tuple{VersionNumber, String}}
 end
 
 # Document.
@@ -451,7 +453,9 @@ function Document(plugins = nothing;
         [],
         [],
         Dict{Markdown.Link, String}(),
-        Set{Symbol}()
+        Set{Symbol}(),
+        Dict{String, String}(),
+        Dict{String, Tuple{String, String}}()
     )
 
     plugin_dict = Dict{DataType, Plugin}()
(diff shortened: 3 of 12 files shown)


frankier:
Ref: https://github.com/JuliaDocs/Documenter.jl/pull/2081

odow:
:+1: I just ran into this yesterday https://github.com/jump-dev/JuMP.jl/pull/3455

src/documents.jl
```diff
@@ -688,6 +688,93 @@ function lt_remotepair(r1::RemoteRepository, r2::RemoteRepository)
...
+                return registry_info(pkg)
+            end
+        end
+    end
+end
+
+function uuid_to_repo(uuid)
+    env = EnvCache()
+    package_entry = manifest_info(env.manifest, uuid)
+    pkginfo = get_pkginfo(uuid)
```
mortenpi, line 720: These `Pkg` functions.. are they public API? I suspect not? Or, minimally, they're probably not documented? I am okay with relying on them but, if they're not clearly documented, we should add some tests that explicitly test their existence and the interface we expect them to have. Something like: https://github.com/JuliaDocs/DocStringExtensions.jl/blob/a970c7d7f8c8aa89a3cdae9b4da3bdb49fb12add/test/tests.jl#L22-L68

mortenpi (commented):
Thanks for the thorough PR! It seems like there's enough demand for automating this, so let's try to get a solution into 1.0.

I am happy to use the registry to get the remote for these. However, I am not a fan of using GitHub for the commit, since that means `makedocs` will try to make a bunch of flaky network requests (by default) and and you also have to be aware of GitHub rate limits.

How about we just use the `v$(package_version)` string, and pray that the package has properly run TagBot? It's not going to be 100% true, but should be good enough?

If we're _really_ concerned about that this will introduce invalid URLs, we could pass on these URLs to online linkcheck and let it double check it (including that the tree hashes match). Key being that those checks and network traffic are opt-in.

frankier:
Okay -- that makes a lot of sense and I didn't realise I could directly use the tag in the URL like that. I agree this way is a better set of tradeoffs.

I have switched to https://github.com/GunnarFarneback/RegistryInstances.jl which works on Julia 1.6 and avoids the internal API, added some sanity checks for the manifest internal API, and moved to the "just make a link the the guessed tag" approach. I have then moved the GitHub commit hash checking stuff to an extension module, which people can either introduce into their build, or run manually as they like.

frankier:
Hm still got some Julia 1.6 failures -- taking a look

frankier:
Okay. Now I've updated so it doesn't use any internal APIs at all. I've also updated the tests to test the basic functionality without checking when offline + on older Julia versions.

The one thing I'm not sure of is the current organisation of the tests. Happy to receive feedback on this. Hopefully the rest is okay now.

docs/src/lib/remote-links.md
```diff
@@ -34,8 +34,7 @@ To handle those cases, the `remotes` keyword to [`makedocs`](@ref) can be used t
 The local directory is assumed to correspond to the root of the Git repository, and any subpath within that directory is then resolved to the corresponding path in the remote repository.
 If there are nested `remotes` configured, Documenter will use the one that matches first as it walks up the directory tree from the original path.
 
-As the common case is a locally checked out Git repository cloned from GitHub, Documenter will also try to determine such remotes automatically.
-This is done regardless whether any additional remotes have been specified with `remotes` or not.
+As the common case is a locally checked out Git repository cloned from GitHub, or a released package which is hosted on GitHub, Documenter will also try to determine such remotes automatically. This is done regardless whether any additional remotes have been specified with `remotes` or not.
```
mortenpi, line 37:
```suggestion
As the common cases are a locally checked out Git repository (added with `Pkg.develop` to the docs environment), or a released package which is hosted on GitHub (`Pkg.add`ed to the environment), Documenter will also try to determine such remotes automatically.
This is done regardless whether any additional remotes have been specified with `remotes` or not.
In both cases, the package must be hosted on GitHub for this to work automatically.
```

docs/src/lib/remote-links.md
```diff
@@ -45,6 +44,18 @@ This is done regardless whether any additional remotes have been specified with
 
 You can think of it as Documenter automatically populating `remotes` with any cloned GitHub repositories it finds.[^3]
 
+For released packages (those added using `Pkg.add(...)` rather than `Pkg.develop(...)`), the version and repository can be determined from the package metadata, but a commit hash is not readily available. In this case, Documenter will guess that a tag `v$VERSION` exists in the repository on GitHub. Note that these tags are created automatically by the widely used JuliaRegistries/TagBot action. Since this is sometimes not the case, and could cause dead or incorrect links, the `GitHubCheckExt` extension module is available for Julia 1.9 once the `GitHub` module is installed. You can use it to check the links as part of your documentation build. Note that this will make GitHub API requests and so could cause builds to fail due to network errors or GitHub downtime. Simply run `import Pkg; Pkg.add("GitHub")` and then modify `build.jl` to look like the following:
```
mortenpi, line 47: For cleaner diffs, I have started following a one line = one sentence structure in Markdown docs:

```suggestion
For released packages (those added using `Pkg.add(...)` rather than `Pkg.develop(...)`), the version and repository can be determined from the package metadata, but a commit hash is not readily available.
In this case, Documenter will guess that a tag `v$VERSION` exists in the repository on GitHub.
Note that these tags are created automatically by the widely used JuliaRegistries/TagBot action.
Since this is sometimes not the case, and could cause dead or incorrect links, the `GitHubCheckExt` extension module is available for Julia 1.9 once the `GitHub` module is installed.
You can use it to check the links as part of your documentation build.
Note that this will make GitHub API requests and so could cause builds to fail due to network errors or GitHub downtime.
Simply run `import Pkg; Pkg.add("GitHub")` and then modify `build.jl` to look like the following:
```

src/Documenter.jl
```diff
@@ -19,6 +19,9 @@ import Markdown
 using MarkdownAST: MarkdownAST, Node
 import REPL
 import Unicode
+import Pkg
+import Pkg.Types
```
mortenpi, line 23: Not directly used I think?

```suggestion
```

src/documents.jl
```diff
@@ -688,6 +688,63 @@ function lt_remotepair(r1::RemoteRepository, r2::RemoteRepository)
...
+    global src_to_uuid
+    global uuid_to_version
+    src_to_uuid = Dict()
+    uuid_to_version = Dict()
+
+    for (uuid, dep) in pairs(Pkg.dependencies())
+        src_to_uuid[dep.source] = uuid
+        uuid_to_version[uuid] = dep.version
+    end
+end
```
mortenpi, line 704: Globals, let's try to avoid those :grimacing: Could we move these to being fields of `doc.internal`? We'd be re-populating them if you call `makedocs` multiple times in a REPL session, but that's fine I think?

mortenpi (commented):
Awesome! I think I'm quite happy with the main behavior now :slightly_smiling_face: 

* One thing I would like your thoughts on is integrating the new online check with the existing linkcheck. I am not sure it's worth adding extra API for this. If you're running `linkcheck`, you probably want to check these links too?

  We could still use the weak deps here, just that if you have online checks enabled (with `linkcheck=true`) _and_ you have one of these `Pkg.add`ed URLs, then we try to load the extension module in Documenter. And if `GitHub` is not available in the environment, we  fail the online test with an error saying "please add GitHub" to the environment.

* On the testing front, could you add the docstring for `IOCapture.iocapture` into `test/examples/src`? That should be a `Pkg.add`ed dependency of the test environment (may need to import the package in `test/examples/make.jl` though), and we control it, so it should be a pretty stable test we can add.

frankier:
> We could still use the weak deps here, just that if you have online checks enabled (with linkcheck=true) and you have one of these Pkg.added URLs, then we try to load the extension module in Documenter. And if GitHub is not available in the environment, we fail the online test with an error saying "please add GitHub" to the environment.

This was my original idea, but I couldn't work out how to do it. It seems like a common enough thing to want to do but I couldn't find any docs about it anywhere. [I've posted a question to StackOverflow.](https://stackoverflow.com/questions/77014386/how-can-a-julia-package-load-its-own-package-extension-at-runtime-based-on-a-con)

mortenpi:
Sooo... one more request, after thinking and discussing this a bit: could we get rid of GitHub.jl altogether in favor of raw Downloads.jl `request` calls? :sweat_smile: The main issue here is that GitHub.jl is actually pretty heavy (it would quadruple the indirect dependency count of Documenter). The bonus would be that we don't have to figure out the weak dependency problem :slightly_smiling_face: I understand if that maybe puts a bit too much work on your plate for this PR though, so I am also happy to take a look at implementing that part here some time next week.

Although.. something I just noticed -- did you by any chance forget to push the `GitHubCheckExt` module? I don't see it in the diff right now.

frankier:
> Sooo... one more request, after thinking and discussing this a bit: could we get rid of GitHub.jl altogether in favor of raw Downloads.jl request calls? 😅 The main issue here is that GitHub.jl is actually pretty heavy (it would quadruple the indirect dependency count of Documenter). The bonus would be that we don't have to figure out the weak dependency problem 🙂 I understand if that maybe puts a bit too much work on your plate for this PR though, so I am also happy to take a look at implementing that part here some time next week.

No worries, I think I will still try and finish this one off according to your advice now-ish. I guess again we got to a nice solution in the end (and more discussion could have been front loaded from my side anyway). 

> Although.. something I just noticed -- did you by any chance forget to push the GitHubCheckExt module? I don't see it in the diff right now.

Whoops, but no need now.

frankier:
Done

mortenpi (approved):
I'll make a couple of very tiny cosmetic changes, but this looks great now! Thanks a lot @frankier for iterating on this!

PR #2236, opened 2023-08-28
Remove search.js as separate overridable asset

mortenpi:
Since we now load `search.js` anyway on every HTML page, we can just combine it into `documenter.js`, so simplify our assets handling. cc @Hetarth02 -- can you think of any reason we shouldn't do this?

This is technically breaking, since previously you could actually provide a custom `search.js` file that overrides Documenter's implementation. But I _assume_ it is not used very much.. and anyway those files would be broken by the new UI.


Hetarth02:
@mortenpi 
I think yes, we can definitely move the `search.js` to `documenter.js`.

PR #2247, opened 2023-09-08
feat: write at-example outputs over threshold to a file

### LINKED ISSUE #2143: An option to write at-example outputs to file

mortenpi:
Building on top of @jvaverka's contribution. Close #2143.

- [x] Currently only implemented for PNG. Expand for other binary images.
- [x] Expand for SVG -- it's a bit more complicated, so needs testing.
- [x] Tests-tests-tests.


mortenpi:
This should essentially be good to go, except for the complication mentioned in https://github.com/JuliaDocs/Documenter.jl/issues/2143#issuecomment-1713611873.

src/html/HTMLWriter.jl
```diff
@@ -2137,13 +2239,18 @@ function domify(dctx::DCtx, node::Node, d::Dict{MIME,Any})
...
     elseif haskey(d, MIME"image/svg+xml"())
+        @tags img
         svg = d[MIME"image/svg+xml"()]
         svg_tag_match = match(r"<svg[^>]*>", svg)
         if svg_tag_match === nothing
             # There is no svg tag so we don't do any more advanced
             # processing and just return the svg as HTML.
             # The svg string should be invalid but that's not our concern here.
             rawhtml(svg)
+        elseif length(svg) > dctx.ctx.settings.example_size_threshold
```
mortenpi, line 2250:
```suggestion
        elseif length(svg) >= dctx.ctx.settings.example_size_threshold
```

mortenpi:
@odow Would you by any chance have time to test this branch with JuMP, see if it helps with the page sizes? I actually suspect it might not be as useful as I had hoped due to https://github.com/JuliaDocs/Documenter.jl/issues/2143#issuecomment-1713611873.

odow:
Let me have a go

mortenpi:
One possibility to work around the `text/html` issue:

> Check the size of `text/html`. If it is over the size limit, check if `show`s for any of the file-based image ones are available. If yes, use one of those (with a warning saying that `text/html` MIME was ignored due to size). If not, write `text/html`, but issue a warning that it's above the size threshold.

src/html/HTMLWriter.jl
```diff
@@ -512,11 +522,16 @@ struct HTML <: Documenter.Writer
         elseif size_threshold_warn > size_threshold
             throw(ArgumentError("size_threshold_warn ($size_threshold_warn) must be smaller than size_threshold ($size_threshold)"))
         end
+        if isnothing(example_size_threshold)
+            example_size_threshold = typemax(Int)
+        elseif example_size_threshold <= 0
```
odow, line 527:
```suggestion
        elseif example_size_threshold < 0
```
Docstring suggests it can be set to `0`.

src/html/HTMLWriter.jl
```diff
@@ -360,6 +362,12 @@ value is `"en"`.
 **`warn_outdated`** inserts a warning if the current page is not the newest version of the
 documentation.
 
+**`example_size_threshold`** specifies the size threshold above which the `@example` and other block
+outputs get written to files, rather than being included in the HTML page. This mechanism is present
+to reduce the size of the generated HTML files that contain a lot of figures etc.
+Setting it to `nothing` will disable writing to files, and setting to `0` means that all files
```
odow, line 368:
```suggestion
Setting it to `nothing` will disable writing to files, and setting to `0` means that all examples
```

odow:
Aren't the images just `svg` blocks? Couldn't we write them to files?
https://github.com/jump-dev/JuMP.jl/blob/104d2a4e8fbfec887f1702170a39b90b25fb9f11/v1.14.1/tutorials/getting_started/getting_started_with_data_and_plotting/index.html#L14C12-L14C12

mortenpi:
> Aren't the images just `svg` blocks?

Yes, but they are generated using the `text/html` `show` method, so we can't really reliably determine whether they are pure SVG or not.

mortenpi:
One user-level hack here is to define

```julia
Base.showable(::MIME"text/html", ::Plots.Plot) = false
```

in your `make.jl`, which will make Documenter ignore the `text/html` representation.

mortenpi:
Documenter should now print just a single warning that looks something like this:

```
┌ Warning: For 3 @example blocks, the 'text/html' representation of the resulting
│ object is above the the threshold (example_size_threshold: 8192 bytes).
│ - 1 blocks had no image MIME show() method representation as an alternative.
│   Sticking to the 'text/html' representation (largest block is 121034 bytes).
│   On pages: index.md
│ - 2 blocks had 'image/svg+xml' fallback image representation available, using that.
│   On pages: index.md, plots.md
└ @ Documenter.HTMLWriter ~/juliadocs/Documenter/src/html/HTMLWriter.jl:779
```

PR #2249, opened 2023-09-10
Revise Plugins API, add `plugins` keyword arg for `makedocs` and `doctest`

### LINKED ISSUE #2245: Add a `plugins` keyword argument for `doctests` (and `makedocs`)
From https://github.com/JuliaDocs/DocumenterCitations.jl/issues/34, downstream from https://github.com/CliMA/ClimaAtmos.jl/issues/2068

At least in principle, it would be conceivable that a Documenter plugin could do something that would affect doctests. It would be good if there was some way to pass plugins to the `doctest` function.

As a matter of fact (since there's a breaking 1.0 release coming), it might be worth reconsidering how Documenter handles plugins in general. The way plugins are passed as positional arguments is undocumented, see https://github.com/JuliaDocs/Documenter.jl/issues/2113. Maybe it would be better to have a `plugins` keyword argument for both `makedocs` and `doctests`?

goerz:
* Add tests for the `Plugin`/`getplugin` API (testing all documented behavior)
* Fix the documentation for `Plugin`/`getplugin`. Specifically, the documentation claimed that it was possible to pass `Plugin`-*types* instead of *objects* to `makedocs`. This didn't match either the old or the new implementation, so the docstrings have been corrected to match what is actually implemented
* Add a `plugins` keyword argument to `makedocs` that replaces passing instances of `Plugin` as positional arguments (which was completely undocumented)
* Add a `plugins` keyword argument to `doctest` to address https://github.com/JuliaDocs/DocumenterCitations.jl/issues/34. ~~This is implemented by forwarding *all* unknown keyword arguments from `doctest` to the internal call of `maketest`. The feature is documented as "use only as directed". Alternatively, I could implement this by fowarding *only* `plugins`, but generally forwarding `kwargs` seems more future-proof, avoid problems where in some specific context `maketest` requires some argument that `doctest` isn't giving it.~~

The introduction of the `plugins` keyword argument is (at least arguably) a breaking change, and thus should be included in the 1.0 release. The old behavior with regards to plugins wasn't documented, but anyone who might have been passing `Plugin` objects as positional arguments to `makedocs` has to change that to a `plugins` keyword argument.

Most notably, users of [DocumenterCitations](https://github.com/JuliaDocs/DocumenterCitations.jl) will be affected by this.

In any case, this PR makes the API for plugins significantly more explicit.

Closes #2245

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   **For upgrading:** The JS file can still be include by passing it to the `assets` keyword of `format = HTML(...)`. However, it will likely conflict with Documenter's default search implementation. If you require an API to override Documenter's search engine, please open an issue.
 
+* Plugin objects which were formally passed as (undocumented) positional keyword arguments to `makedocs` are now given as elements of a list `plugins` passed as a keyword argument ([#2245], [#2249])
+
+  **For upgrading:** If you are passing any plugin objects to `makedocs` (positionally), pass them via the `plugins` keyword instead.
+
 ### Added
 
 * Doctest filters can now be specified as regex/substitution pairs, i.e. `r"..." => s"..."`, in order to control the replacement (which defaults to the empty string, `""`). ([#1989], [#1271])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * The search UI has had a complete overhaul, with a fresh new modal UI with better context for search results and a filter mechanism to remove unwanted results. The client-side search engine has been changed from LunrJs to MinisearchJs. ([#1437], [#2141], [#2147], [#2202])
 
+* The `doctest` routine can now receive the same `plugins` keyword argument as `makedocs`. This enables `doctest` to run if any plugin with a mandatory `Plugin` object is loaded, e.g., [DocumenterCitations](https://github.com/JuliaDocs/DocumenterCitations.jl). ([#2245])
+
 ### Fixed
 
 * Line endings in Markdown source files are now normalized to `LF` before parsing, to work around [a bug in the Julia Markdown parser][julia-29344] where parsing is sensitive to line endings, and can therefore cause platform-dependent behavior. ([#1906])
--- a/src/doctest.jl
+++ b/src/doctest.jl
@@ -46,12 +46,15 @@ manual pages can be disabled if `source` is set to `nothing`.
 
 # Keywords
 
-**`testset`** specifies the name of test testset (default `Doctests`).
+**`testset`** specifies the name of test testset (default `"Doctests"`).
 
 **`doctestfilters`** vector of regex to filter tests (see the manual on [Filtering Doctests](@ref))
 
 **`fix`**, if set to `true`, updates all the doctests that fail with the correct output
 (default `false`).
+
+**`plugins`** is a list of [`Documenter.Plugin`](@ref) objects to be forwarded to
+[`makedocs`](@ref). Use as directed by the documentation of a third-party plugin.
 
 !!! warning
     When running `doctest(...; fix=true)`, Documenter will modify the Markdown and Julia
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   **For upgrading:** The JS file can still be include by passing it to the `assets` keyword of `format = HTML(...)`. However, it will likely conflict with Documenter's default search implementation. If you require an API to override Documenter's search engine, please open an issue.
 
+* Plugin objects which were formally passed as (undocumented) positional keyword arguments to `makedocs` are now given as elements of a list `plugins` passed as a keyword argument.
```
mortenpi, line 61:
```suggestion
* Plugin objects which were formally passed as (undocumented) positional keyword arguments to `makedocs` are now given as elements of a list `plugins` passed as a keyword argument. ([#2245], [#2249])
```

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
```
mortenpi, line 62:
```suggestion

  **For upgrading:** if you are passing any plugin objects to `makedocs` (positionally), pass them via the `plugins` keyword instead.

```

src/doctest.jl
```diff
@@ -59,13 +59,18 @@ manual pages can be disabled if `source` is set to `nothing`.
...
+[`makedocs`](@ref). These should only be used as directed, e.g., in the documentation of a
+third-party plugin.
 """
 function doctest(
         source::Union{AbstractString,Nothing},
         modules::AbstractVector{Module};
         fix = false,
         testset = "Doctests",
         doctestfilters = Regex[],
+        kwargs...
```
mortenpi, line 73: Hmm. I feel like we should be conservative here and only pass `plugins`, not all of them. If at some point we discover that we should be passing others, then we add them as needed.

src/documents.jl
```diff
@@ -376,6 +376,7 @@ function Document(plugins = nothing;
         pages    :: Vector           = Any[],
         pagesonly:: Bool             = false,
         expandfirst :: Vector        = String[],
+        plugins  :: Vector           = Any[],
```
mortenpi, line 379:
```suggestion
        plugins  :: Vector           = Plugin[],
```

src/documents.jl
```diff
@@ -455,14 +456,12 @@ function Document(plugins = nothing;
...
-        for plugin in plugins
-            plugin isa Plugin ||
-                throw(ArgumentError("$(typeof(plugin)) is not a subtype of `Plugin`."))
-            haskey(plugin_dict, typeof(plugin)) &&
-                throw(ArgumentError("only one copy of $(typeof(plugin)) may be passed."))
-            plugin_dict[typeof(plugin)] = plugin
-        end
+    for plugin in plugins
+        plugin isa Plugin ||
+            throw(ArgumentError("$(typeof(plugin)) is not a subtype of `Plugin`."))
```
mortenpi, line 461:
```suggestion
            throw(ArgumentError("$(typeof(plugin)) in plugins= is not a subtype of `Documenter.Plugin`."))
```

src/documents.jl
```diff
@@ -798,9 +797,10 @@ end
 """
     getplugin(doc::Document, T)
```
mortenpi, line 798:
```suggestion
    Documenter.getplugin(doc::Document, T) -> Plugin
```

mortenpi (commented):
This looks great, thank you @goerz! Other than being conservative about what kwargs to pass from `doctest` -> `makedocs`, the other comments are purely cosmetic.

src/doctest.jl
goerz: Sure! I was on the fence about this one, but just passing `plugins` should be totally fine.

goerz: Ok, force-pushed a new version

goerz:
Ok, I made all the changes from @mortenpi's comments.

Should be ready for merge, unless there's any more changes related to the revised more conservative approach of only passing `plugins` in `doctest`.

mortenpi (approved):
LGTM, thanks! I'll leave it open for a day or two, in case there is any other feedback, but otherwise I'm be happy to merge this as is.

PR #2252, opened 2023-09-12
Add `size_threshold_ignore` to `HTML`

### LINKED ISSUE #2211: Add `size_threshold_ignore` to HTML, to ignore specific pages
Suggested by @odow in https://github.com/JuliaDocs/Documenter.jl/pull/2205#discussion_r1296645306, to have a bit more fine grained way to ignore some pages going over the size threshold.

Behavior-wise, I would suggest that the ignored pages still warn if they go over the `size_threshold_warn` limit, as this is supposed to be a temporary solution. So effectively it sets `size_threshold = Inf` for those pages.

mortenpi:
Fix #2211.


mortenpi:
@odow Does this look good for the JuMP use case?

odow (approved):
Yeah I think so.

PR #2268, opened 2023-09-18
Add backticks to size_threshold_msg

jaakkor2:
before
image

after
image

### DIFF
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -1769,11 +1769,11 @@ function write_html(ctx::HTMLContext, navnode::Documenter.NavNode, page_html::DO
     isdir(dirname(path)) || mkpath(dirname(path))
     file_size = open(io -> write(io, take!(buf)), path; write=true)
     size_threshold_msg(var::Symbol) = """
-    Generated HTML over $(var) limit: $(navnode.page)
+    Generated HTML over `$(var)` limit: `$(navnode.page)`
         Generated file size: $(file_size) (bytes)
-        size_threshold_warn: $(ctx.settings.size_threshold_warn) (bytes)
-        size_threshold:      $(ctx.settings.size_threshold) (bytes)
-        HTML file:           $(page_path)"""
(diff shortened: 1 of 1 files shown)


mortenpi:
This coloring does not seem to be standard... where is it coming from? 
image


While there is some precedent in Documenter to apply the backticks in log messages, I am a little hesitant to accept this, since for the standard user (and in CI logs), it will hurt readability.

jaakkor2:
Sorry, should have mentioned I use `TerminalLoggers.jl` in my `startup.jl` to get colored output.
```
try
    using Logging: global_logger
    using TerminalLoggers: TerminalLogger
    global_logger(TerminalLogger())
catch e
    @warn "Error initializing TerminalLogger" exception=(e, catch_backtrace())
end
```

mortenpi:
That's neat and useful. But I am not sure we should expect users to be using it, and so I am still hesitant to reduce readability for the more common case.

jaakkor2:
Default user experience should have the priority, hence I'd close the PR.

My use of TerminalLoggers.jl comes from https://docs.sciml.ai/DiffEqDocs/stable/features/progress_bar/#Using-Progress-Bars-Outside-Juno

Based on a quick search, I did not find how it would be possible to set a logger based on a module, say if Documenter.jl could force `SimpleLogger()` or `ConsoleLogger()`.

mortenpi:
> Based on a quick search, I did not find how it would be possible to set a logger based on a module, say if Documenter.jl could force `SimpleLogger()` or `ConsoleLogger()`.

I am not sure, but wouldn't that go against the philosophy of the Logging module? My impression has been that the point is that you use logging macros in your code normally, and then can use whichever logger you want externally? So Documenter probably shouldn't be injecting its own logger, but it should be up to the user?

Instead, I'm thinking that maybe we can apply some default styling on strings ourselves, e.g. with the help of the work in https://github.com/JuliaLang/julia/pull/49586

> hence I'd close the PR

I'll close it then. Thank you regardless!

PR #2275, opened 2023-09-20
hosting.md: recommend current codecov/codecov-action version

fingolfin:
Version 1 is deprecated and may not even work anymore


mortenpi:
Do as I do, don't do as I say :smile:

https://github.com/JuliaDocs/Documenter.jl/blob/fea343ce987fac480567539a3b6e683cc9176fc5/.github/workflows/CI.yml#L42-L44

Thank you!

PR #2276, opened 2023-09-20
Fix debug logging for doctest filtering

fingolfin:
While debugging a doctest filter with `ENV["JULIA_DEBUG"] = "Documenter"` I noticed that the "Evaluated output (filtered)" was clearly not derived from the "Evaluated output". I think this here is the cause.


mortenpi:
Good catch, thank you!

Issue #2280, opened 2023-09-21
Support in-line LaTeX math output in HTMLWriter


odow:
`index.md`
````
```@example
struct Foo end
function Base.show(io::IO, ::MIME"text/latex", ::Foo)
    return print(io, "\$ x_{1} + x_{2} \$")
end
Foo()
```
````

`make.jl`
````
import Documenter
Documenter.makedocs(
    sitename = "Test",
    format = Documenter.HTML(;
        prettyurls = get(ENV, "CI", nothing) == "true",
    ),
    pages = ["index.md"],
)
````

Produces

![image](https://github.com/JuliaDocs/Documenter.jl/assets/8177701/f5ac704d-f61e-4579-bfb8-266b42e6792d)

Here's a case from the wild: https://jump.dev/JuMP.jl/stable/tutorials/linear/knapsack/#JuMP-formulation

odow:
Closed by #2281

PR #2281, opened 2023-09-21
Fix in-line math output in HTMLWriter

odow:
Before I add tests etc for #2280, is this something that would be considered? Or should all `MIME"text/latex"` outputs be display math?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+* Fixed display of inline LaTeX math `$ ... $` in HTMLWriter ([#2281])
+
 ## Version [v1.0.1] - 2023-09-18
 
 ### Fixed
@@ -1683,6 +1689,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2259]: https://github.com/JuliaDocs/Documenter.jl/issues/2259
 [#2260]: https://github.com/JuliaDocs/Documenter.jl/issues/2260
 [#2269]: https://github.com/JuliaDocs/Documenter.jl/issues/2269
+[#2281]: https://github.com/JuliaDocs/Documenter.jl/issues/2281
 [JuliaLang/julia#29344]: https://github.com/JuliaLang/julia/issues/29344
 [JuliaLang/julia#36953]: https://github.com/JuliaLang/julia/issues/36953
 [JuliaLang/julia#38054]: https://github.com/JuliaLang/julia/issues/38054
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Fixed display of inline LaTeX math `$ ... $` in HTMLWriter ([#2281])
+* Fixed display of inline LaTeX math `$ ... $` from `show` methods in HTMLWriter. ([#2280], [#2281])
 * Fixed a crash in GitHub remote link checking when `remotes = nothing`. ([#2274], [#2285])
 
 ## Version [v1.0.1] - 2023-09-18
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -2395,7 +2395,7 @@ function domify(dctx::DCtx, node::Node, d::Dict{MIME,Any})
         latex = d[MIME"text/latex"()]
         # Make sure to match multiline strings!
         m_bracket = match(r"\s*\\\[(.*)\\\]\s*"s, latex)
-        m_dollars = match(r"\s*\$\$(.*)\$\$\s*"s, latex)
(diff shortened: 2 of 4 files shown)


src/html/HTMLWriter.jl
```diff
@@ -2395,7 +2395,7 @@ function domify(dctx::DCtx, node::Node, d::Dict{MIME,Any})
         latex = d[MIME"text/latex"()]
         # Make sure to match multiline strings!
         m_bracket = match(r"\s*\\\[(.*)\\\]\s*"s, latex)
-        m_dollars = match(r"\s*\$\$(.*)\$\$\s*"s, latex)
+        m_dollars = match(r"\s*\$\$?(.*?)\$?\$\s*"s, latex)
```
mortenpi, line 2398:
```suggestion
        m_dollars = match(r"\s*\$\$?(.*?)\$?\$\s*"s, latex)
```

Wait, why the middle `?`?

mortenpi (commented):
This feels fine to me. We'll show it as display math.. I think..? Or at least should if we don't, for consistency (at-blocks should output blocks). But no harm in being permissive with the input here I think.

src/html/HTMLWriter.jl
odow: Because otherwise it will eagerly expand the middle `()` and capture the optional `\$` at the end:
```julia
julia> match(r"\s*\$\$?(.*?)\$?\$\s*"s, "abc \$ x_1 \$ def")
RegexMatch(" \$ x_1 \$ ", 1=" x_1 ")

julia> match(r"\s*\$\$?(.*?)\$?\$\s*"s, "abc \$\$ x_1 \$\$ def")
RegexMatch(" \$\$ x_1 \$\$ ", 1=" x_1 ")

julia> match(r"\s*\$\$?(.*)\$?\$\s*"s, "abc \$ x_1 \$ def")
RegexMatch(" \$ x_1 \$ ", 1=" x_1 ")

julia> match(r"\s*\$\$?(.*)\$?\$\s*"s, "abc \$\$ x_1 \$\$ def")
RegexMatch(" \$\$ x_1 \$\$ ", 1=" x_1 \$")   # <-- wrong
```

odow:
> We'll show it as display math.. I think..?

Yeah, I think so. But that's okay.

src/html/HTMLWriter.jl
mortenpi: Ah, makes sense. Let's add this this as a comment though!

test/htmlwriter.jl
```diff
@@ -295,5 +295,50 @@ end
...
+                    " $left$content$right",
+                    "$left$content$right ",
+                    "\t$left$content$right  ",
+                    " \t$left$content$right\t\t",
+                ]
+                    @test _strip_latex_math_delimiters(input) == (false, input)
+                end
+            end
+        end
+    end
```
odow, line 343:
```suggestion
    end
end
```

test/htmlwriter.jl
```diff
@@ -295,5 +295,50 @@ end
...
+                    "$left$content$right" => (true, content),
+                    " $left$content$right" => (true, content),
+                    "$left$content$right " => (true, content),
+                    "\t$left$content$right  " => (true, content),
+                    " \t$left$content$right\t\t" => (true, content),
+                ]
+                    @test _strip_latex_math_delimiters(input) == output
+                end
+            end
+            # Test that mis-matched delimiters are not treated as math
```
odow, line 318:
```suggestion
            # Test that miss-matched delimiters are not treated as math
```

src/html/HTMLWriter.jl
```diff
@@ -2415,6 +2413,21 @@ function domify(dctx::DCtx, node::Node, d::Dict{MIME,Any})
...
 end
 
+function _strip_latex_math_delimiters(latex::AbstractString)
+    # Make sure to match multiline strings!
+    m_bracket = match(r"\s*\\\[(.*)\\\]\s*"s, latex)
+    if m_bracket !== nothing
+        return true, m_bracket[1]
+    end
+    # This regex matches the strings "$ ... $" and "$$ ... $$", but not
+    # strings with mis-matched dollar signs such as "$ ... $$" or "$$ ... $"
```
odow, line 2423:
```suggestion
    # strings with miss-matched dollar signs such as "$ ... $$" or "$$ ... $"
```

odow:
I refactored a bit and used a different regex.

odow:
Thoughts @mortenpi?

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+* Fixed display of inline LaTeX math `$ ... $` in HTMLWriter ([#2281])
```
mortenpi, line 10:
```suggestion
* Fixed display of inline LaTeX math `$ ... $` from `show` methods in HTMLWriter. ([#2280], [#2281])
```

mortenpi (commented):
Sorry, Documenter went on the backburner for a few days. But this looks great. And the refactoring makes the code much cleaner!

Let me just add a quick thing into the example builds.

mortenpi (approved):
This LGTM. Did you want to do anything else or is this good to go?

odow:
Good by me

PR #2283, opened 2023-09-25
Fix malformed link in CHANGELOG.md

mgkurtz:
The malformed link syntax currently shows up at https://documenter.juliadocs.org/stable/release-notes/#Fixed-2. Curiously GitHub instead renders the link as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,7 +150,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


mgkurtz:
Since `changelog.jl` does not work with `[text](target)` link syntax, I moved the reference to the end.

mortenpi:
I'd say we just need to fix this in Changelog.jl, so that it would pick up these inline labels too. It's more reader-friendly if the link is inline. @fredrikekre, any thoughts?

Issue #2274, opened 2023-09-19
Error with `remotes=nothing` and `linkcheck=true`


JackDunnNZ:
Running with the following settings on Documenter v1 generates the following error:

```
using Documenter
makedocs(remotes=nothing, sitename="abc", linkcheck=true)
```

```
[ Info: CheckDocument: running document checks.
ERROR: LoadError: MethodError: no method matching iterate(::Nothing)

Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen})
   @ Base range.jl:880
  iterate(::Union{LinRange, StepRangeLen}, ::Integer)
   @ Base range.jl:880
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}}
   @ Base dict.jl:698
  ...

Stacktrace:
  [1] githubcheck(doc::Documenter.Document)
    @ Documenter ~/.julia/packages/Documenter/Meee1/src/docchecks.jl:324
```

It looks like iteration [here](https://github.com/JuliaDocs/Documenter.jl/blob/fea343ce987fac480567539a3b6e683cc9176fc5/src/docchecks.jl#L324) needs to guard against a `nothing` (or prevent `nothing` from getting here)

PR #2285, opened 2023-09-26
Stop githubcheck() error when remotes is nothing

frankier:
Fixes https://github.com/JuliaDocs/Documenter.jl/issues/2274


mortenpi (approved):
LGTM, thanks!

PR #2299, opened 2023-10-09
Update actions/checkout version in docs

fingolfin:


mortenpi:
Side note: we should maybe load these dynamically from a YAML file, so that dependabot could pick them up.

Issue #2293, opened 2023-10-05
Long text in docstring does not wrap on smaller screens


Hetarth02:
![image](https://github.com/JuliaDocs/Documenter.jl/assets/80420115/d3ee4789-9571-4873-8d3c-a32fcb2655d1)

Hetarth02:
One solution would be to `overflow: auto` the `header` tag just like the code content. (I prefer this.)
![image](https://github.com/JuliaDocs/Documenter.jl/assets/80420115/a1f776ba-2305-4b7a-b7a3-ecfa67005a5f)


Another one would be to `word-break: break-all`, but that doesn't look good.
![image](https://github.com/JuliaDocs/Documenter.jl/assets/80420115/2aefb8c6-5eb4-4e80-8810-00e1ececb059)

mortenpi:
Yeah, `overflow: auto` seems better.

PR #2307, opened 2023-10-24
Fix: Enable text wrapping in docstring header on smaller screens (#2293)

QuantuM410:
### Changes:
overflow: auto CSS property added to the header tag.

[screen-recorder-tue-oct-24-2023-13-13-39.webm](https://github.com/JuliaDocs/Documenter.jl/assets/85943021/0a3d734a-9b8c-4d84-9e58-611f43c6b85c)


mortenpi (commented):
Thanks! This basically LGTM, but could you:

1. Undo the unrelated formatting changes? I think the original formatting is preferred and more consistent with the rest of the code, and also I would prefer not to mix unrelated changes in PRs.
2. Add a quick note about this into `CHANGELOG.md`.

QuantuM410:
> Thanks! This basically LGTM, but could you:
> 
>     1. Undo the unrelated formatting changes? I think the original formatting is preferred and more consistent with the rest of the code, and also I would prefer not to mix unrelated changes in PRs.
> 
>     2. Add a quick note about this into `CHANGELOG.md`.

While I am adding the note `CHANGELOG.md`, do I also have to run `make changelog`. Once I am running make changelog two lines are being added (one unrelated to the issue but is related to the previous fix in `v1.1.2`)
image

image

mortenpi:
Yeah, if you don't mind, run `make changelog`. And let's add the 2300 too here -- that was an oversight in the earlier PR.

QuantuM410:
@mortenpi I have updated the files with required changes

assets/html/scss/documenter/components/_docstring.scss
```diff
@@ -71,4 +72,4 @@
   & > section:hover a.docs-sourcelink {
     opacity: 1;
   }
-}
+}
```
mortenpi, line 75:
```suggestion
}

```

mortenpi (approved):
Thank you!

PR #2310, opened 2023-10-24
give better file:lineno in the log

simonbyrne:
Log now prints:
```
└ @ Documenter ~/src/ClimaCore.jl/docs/src/api.md:210
```
instead of
```
└ @ Documenter ~/src/Documenter/src/utilities/utilities.jl:44
```

which makes it easier to jump to in my editor (can ctrl-click to go there)

### DIFF
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -299,7 +299,10 @@ function Selectors.runner(::Type{Expanders.MetaBlocks}, node, page, doc)
                     ```$(x.info)
                     $(x.code)
                     ```
-                    """, exception = err)
+                    """, exception = err,
+                    _file = joinpath(pwd(), page.source),
+                    _line = lines.first,
+                )
             end
         end
     end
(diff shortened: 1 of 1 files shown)


fredrikekre:
The location should also be printed in the message itself already. Is the problem that we print a line range there and not just a single number?

mortenpi:
I guess the question here is whether we actually want the location info from the logging macro to _not_ match the actual macro invocation? Arguably, this might be confusing, especially if we happen to mess up the line numbers somehow (due to a bug or something). On the other hand, the links to Documenter source are pretty useless 90% of the time, and this feels like a UX improvement.

mortenpi:
> The location should also be printed in the message itself already. Is the problem that we print a line range there and not just a single number?

Since we do already print the location information for all (?) of these blocks, then maybe we should just disable the location information from the macro, if that's possible? Does `_file = nothing, _line = nothing` work?

PR #2313, opened 2023-10-27
added recursive list search for headings

### LINKED ISSUE #2308: Markdown heading level handling in docstring
Consider docstring:

```
"""
    foo(x, f, k)

Do stuff

# Arguments
- `x`: some argument
- `f`: some function
    ## Pattern
    - `f(t)`
    - `f(t, u)`
    - `f(t, u, c)`
- `k`: some other argument
"""
function foo(x, f, k)
end
```

In REPL help, the format is as expected: \"Pattern\" is indented and a lower level heading as \"Arguments\"

When making the documentation with Documenter on the other hand, \"Pattern\" gets translated to \<h2\> while \"Arguements\" only becomes \<strong\>, not \<h1\>. Therefore it looks quite off.

Using \<h1\> for \"Arguments\" is obviously not a good solution either
=> Idea: translate all heading levels in docstrings to \<strong\> instead of \<h_\>

0815Creeper:
fixes #2308
added search through markdown-lists in docstrings to find headings for rewrite from heading to strong


src/expander_pipeline.jl
```diff
@@ -961,12 +961,13 @@ function create_docsnode(docstrings, results, object, page, doc)
     slug = Documenter.slugify(object)
     anchor = Documenter.anchor_add!(doc.internal.docs, object, slug, page.build)
     docsnode = DocsNode(anchor, object, page)
-    # Convert docstring to MarkdownAST, convert Heading elements, and push to DocsNode
-    for (markdown, result) in zip(docstrings, results)
-        ast = convert(Node, markdown)
-        doc.user.highlightsig && highlightsig!(ast)
-        # The following 'for' corresponds to the old dropheaders() function
+    function recursive_heading_to_bold!(ast)
```
mortenpi, line 964:
```suggestion
    function recursive_heading_to_bold!(ast::MarkdownAST.Node)
```

mortenpi: Maybe let's pull this into a separate function, not a closure. It doesn't depend on any of the variables from `create_docsnode`, right?

mortenpi (commented):
In addition to the comment, could you please:

1. Add a CHANGELOG.md note.
2. Add a test case. I think the best place for the test in this case would be one that shows up in https://documenter.juliadocs.org/stable/showcase/#Docstrings, so it should be somewhere in https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/DocumenterShowcase.jl

src/expander_pipeline.jl
0815Creeper: done

0815Creeper:
> In addition to the comment, could you please:
> 
>     1. Add a CHANGELOG.md note.
Done


> 
>     2. Add a test case. I think the best place for the test in this case would be one that shows up in https://documenter.juliadocs.org/stable/showcase/#Docstrings, so it should be somewhere in https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/DocumenterShowcase.jl
Not quite shure if I get what you said, but I tried

src/expander_pipeline.jl
```diff
@@ -990,3 +981,21 @@ function highlightsig!(node::Node)
...
+        end
+        node.element isa MarkdownAST.Heading || continue
+        # node is a headingnode
+        boldnode = Node(MarkdownAST.Strong())
+        for textnode in collect(node.children)
+            push!(boldnode.children, textnode)
+        end
+        node.element = MarkdownAST.Paragraph()
+        push!(node.children, boldnode)
+    end
```
odow, line 1000:
```suggestion
    end
    return
```

src/expander_pipeline.jl
```diff
@@ -990,3 +981,21 @@ function highlightsig!(node::Node)
...
+end
```
odow, line 1001:
```suggestion
end

```

src/expander_pipeline.jl
```diff
@@ -990,3 +981,21 @@ function highlightsig!(node::Node)
...
 end
+
+function recursive_heading_to_bold!(ast::MarkdownAST.Node)
+    for node in ast.children
+        if node.element isa MarkdownAST.List
+            for child in node.children
+                recursive_heading_to_bold!(child)
+            end
+        end
+        node.element isa MarkdownAST.Heading || continue
```
odow, line 992:
```suggestion
        if !(node.element isa MarkdownAST.Heading)
            continue
        end
```

docs/DocumenterShowcase.jl
```diff
@@ -8,6 +8,8 @@ This is the [`DocumenterShowcase`](@ref), which contains
 
     This showcase page is included in Documenter version 0.24.0 and above.
 
+Headings that are part of docstrings are not rendered as headings but rather as bold text. Just like:
```
mortenpi, line 11:
```suggestion
Headings that are part of docstrings are not rendered as headings but rather as bold text:
```

docs/src/showcase.md
```diff
@@ -324,6 +324,12 @@ We then see the same docstring as above
 DocumenterShowcase.bar
 ```
 
+We can also use nested lists and headings in docstrings
+
+```@docs
+DocumenterShowcase.baz
```
mortenpi, line 330:
```suggestion
DocumenterShowcase.baz
```

docs/DocumenterShowcase.jl
```diff
@@ -63,6 +65,26 @@ Docstring for `bar(::AbstractString)`.
 """
 bar(::AbstractString) = nothing
 
+"""
+    baz(x, f, k)
+
+Docstring for a more complex function.
```
mortenpi, line 71:
```suggestion
Function with a more complex docstring.
```

mortenpi (approved):
Did some cosmetic changes to the docs, but PR LGTM. Thank you @0815Creeper!

PR #2323, opened 2023-11-01
Add regression-tests.yml for manually testing dependent packages

### LINKED ISSUE #2314: GitHub Actions workflows for downstream testing
The goal would be to set up standard CI workflow that we could use to downstream-test some packages and documentation builds. These would act as canaries for any breaking changes to internal we might be doing:

- [ ] A couple of candidates for packages: DocumenterCitations, DocumenterMermaid (pick a commit, run the package's test suite, but with Documenter#master as dependency)
- [ ] We could also try to build some documentations, but not sure who are good candidates. The ones that would be good canaries (like JuMP) are very expensive to build.
- [ ] If we can, we could also try to build the Julia manual. Maybe we can just pull the tarball of the latest release? This is also a bit expensive, but possibly worth it.

For the more expensive downstream tests, we could add some [concurrency options](https://docs.github.com/en/actions/using-jobs/using-concurrency) to avoid too many unnecessary builds.

It's a nice self-contained issue for anyone who wants to dabble with GitHub Actions.

odow:
Closes #2314

I don't know if I got this right, so we'll need to test.

Running the dependent package tests is easy. Checking that we can build other package docs is...less easy. For example, JuMP currently pins the version of Documenter, https://github.com/jump-dev/JuMP.jl/blob/908275f105d7902d0065aad8a56a2caf80e391cf/docs/Project.toml#L43, so trying to build here will fail.

If we knew there were regression tests in Documenter we might relax that though.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -1,0 +1,79 @@
+name: regression-tests
+on:
+  workflow_dispatch:
+jobs:
+  test-dependent-packages:
+    name: run-tests-${{ matrix.package }}
+    runs-on: ubuntu-latest
+    env:
+      PACKAGE: ${{ matrix.package }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - package: 'DocumenterCitations'
+          - package: 'DocumenterMermaid'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1'
+      - uses: actions/cache@v1
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - uses: julia-actions/julia-buildpkg@v1
+      - name: Test
+        shell: julia --color=yes {0}
+        run: |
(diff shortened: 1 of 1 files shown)


odow:
Now I remember: this is a problem with GitHub. You can only kick off a workflow-dispatch once the workflow has been added. and then once it has, you can create a branch, edit, open a PR, and then run it against the branch.

So the only way to tell if this works is to merge it...which isn't a big deal, because it can only be run manually.

mortenpi:
One quick thing to double check: the code running in this workflow doesn't have access to user secrets, right? We would have to explicitly pass them to the jobs via `env`? We're running arbitrary user code here, so just want to be extra sure.

odow:
It only has access to the normal environment variables, which shouldn't be sensitive. But it's something to be mindful of. You probably don't want to try and run arbitrary packages.

When I've needed to, I've isolated packages into their own job and given them access to a secret:
https://github.com/jump-dev/MathOptInterface.jl/blob/b4398223041e4427d9122cbb971648bffe6b80d3/.github/workflows/solver-tests.yml#L77-L112

mortenpi:
Yeah, but I don't think the regression tests need access to secrets anyway.

Should we fix the packages to specific versions/commits though?

mortenpi (approved):
By the way, feel free to merge this, if you want to trigger this for testing.

odow:
> Should we fix the packages to specific versions/commits though?

Nah. We typically just pull in the latest `master`. That assumes `master` is always good, and supports the latest Documenter. But this is a sanity check anyway, so the odd failure doesn't matter. It just prods you to go and take a look at the failure.

odow:
This seems to have worked https://github.com/JuliaDocs/Documenter.jl/actions/runs/6728161571

PR #2324, opened 2023-11-01
Warn on unsupported arguments in contents, index, autodocs, and meta blocks

### LINKED ISSUE #2306: Throw an error if unexpected keys are provided
I thought I stumbled over this bug https://github.com/JuliaDocs/Documenter.jl/issues/936 but what I actually missed is the initial upper case of Pages in the @contents block. 

It would be good to throw an error when unexpected or misspelled keys are provided.

`@autodocs` only accepts: `Modules`, `Filter`, `Pages`, `Private`, `Public`, `Order`. Example keys that would error: `Module`, `pages`, `DocTestSetup`
`@contents` should only accept `Pages` and `Depth`
...

odow:
Closes #2306

I don't know the best way to test this. I also decided to warn, rather than error, because doing so might break working builds if they have added some additional arguments in there, intentionally or otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * `id` anchors may now start with a numeric digit. ([#744]), ([#2325])
+* Documenter now warns when it encounters invalid keys in the various key-value at-blocks. ([#2306], [#2324])
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


mortenpi (approved):
This LGTM. Anything else you wanted to do here @odow or can we merge this?

odow:
I don't have any other plans. Except I wasn't really sure how to test properly

mortenpi:
> Except I wasn't really sure how to test properly

In principle, some `@test_logs` calls in `test/errors` should probably do it? But it would probably require reorganizing the test, so I'm happy to merge this as is.

odow:
:+1:

Issue #744, opened 2018-06-16
Header names starting with numbers get the number stripped when the HTML `id` is generated


anthony-wang:
I may have noticed a bug when using headers that start with numbers.

Here is the MWE:
```markdown
# Header test
Hi

## Header 1
Hello

## 2nd header
Welp
```

which generates the HTML snippet
```html
<h2><a class="nav-anchor" id="nd-header-1" href="#nd-header-1">2nd header</a></h2>
```

The number "2" is stripped from the header and as a result, the anchor `id` becomes `nd-header-1`.

It should be allowable that an anchor `id` starts with a number, according to [W3C](https://www.w3.org/TR/html5/dom.html#element-attrdef-global-id)

> § 3.2.5. Global attributes
> There are no other restrictions on what form an ID can take; in particular, IDs can consist of just digits, start with a digit, start with an underscore, consist of just punctuation, etc.

Thanks!

odow:
I confirm that this is still a problem with the latest Documenter v1.1.2.

PR #2325, opened 2023-11-01
Fix slugify removing leading numeric digits

odow:
Closes https://github.com/JuliaDocs/Documenter.jl/issues/744.

I don't know if there was a good reason for this? It looks very intentional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
-### Added
+### Changed
 
 * `id` anchors may now start with a numeric digit. ([#744]), ([#2325])
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
+### Added
+
+* `id` anchors may now start with a numeric digit. ([#744]), ([#2325])
+
 ### Fixed
 
 * Enabled text wrapping in docstring header on smaller screens. ([#2293], [#2307])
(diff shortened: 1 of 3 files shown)


mortenpi:
I don't think HTML `id`s can't start with a number.

odow:
Seemed to work for me

image

odow:
okay, but https://www.w3schools.com/html/html_id.asp says `The id name must contain at least one character, cannot start with a number, and must not contain whitespaces (spaces, tabs, etc.).`

odow:
So perhaps we should just prefix `_` rather than stripping the numbers?

mortenpi:
It looks

odow:
It also looks like any duplicate headers are handled appropriately, so maybe we just need to add the tests and close the issue.

mortenpi:
We do 

```html
<!DOCTYPE html>
```

i.e. we say we're HTML5. I'd say let's escape the stone age and allow `id`s starting with numbers. We can consider it a `Change` instead of a `Fix` and put it into the next minor.

test/utilities.jl
```diff
@@ -706,6 +706,36 @@ end
...
+            # Nonstrings get converted to strings
+            1 => "1",
+            # Good strings stay good
+            "a" => "a",
+            "my-heading" => "my-heading",
+            "documenter.jl/abc" => "documenter.jl/abc",
+            "https://documenter.jl/abc" => "https://documenter.jl/abc",
+            "2nd" => "2nd",
+            "2nd-2" => "2nd-2",
+            "123a" => "123",
```
odow, line 721:
```suggestion
            "123" => "123",
            "123a" => "123a",
```

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
+### Added
```
mortenpi, line 8: Keep-a-changelog suggests this as another category. Might we worth adopting?

```suggestion
### Changed
```

mortenpi (approved):

CHANGELOG.md
odow: I haven't been using it, so up to you.

mortenpi: We can always change our minds.

Issue #2288, opened 2023-09-30
Improve warning when there are $ in doc strings / .md files in Documenter 1.0


kellertuer:
I was a bit uncareful with my LaTeX writing in the doc strings and the markdown files, cf. e.g.

https://github.com/JuliaManifolds/Manopt.jl/blob/0865eb8b3a9a6f937c0599354bd3bbbc4c5e6c6d/docs/src/solvers/ChambollePock.md?plain=1#L6

While the “old” Documenter (pre 1.0) was quite fine with this (or tolerant to `$...$` LaTeX, now one gets the warning

```
┌ Warning: Unexpected Julia interpolation of type Symbol in the Markdown.
│   value = :p
└ @ Documenter.HTMLWriter ~/.julia/packages/Documenter/9kOxY/src/html/HTMLWriter.jl:2215
```

(which luckily someone hunted down for me to the above line). I think that warning could maybe improved?
It also seems that this does not always happen … so maybe also only for `$` in Doc strings but not in md files?
The reason I assume that is, that some of my Markdown files are generated by Quarto before the Documenter run – that Markdown contains (for laziness and Quarto/Pandoc preferring `$...$`) normal LaTeX, and that renders without warning.

Sorry if this is a bit vague, but I am still not 100% sure how to reproduce this.

odow:
x-ref https://github.com/JuliaDocs/Documenter.jl/issues/872

kellertuer:
Yes, kind of, while for a missing reference I can _both_ search for the reference mentioned _and_ know the file (which often at least helps me, I often see what is wrong with the ref already) – here I have neither.

But sure, even having a line number would be a dream.

odow:
I think we could improve all error reporting, not just this specific one, or the reference one.

mortenpi:
https://github.com/JuliaDocs/Documenter.jl/pull/2327

odow:
Yeah I wasn't sure if the warning closes this. It's a more helpful but still not very helpful warning.

mortenpi:
That's fair actually. Let's make having some source file / line information in the warning the actionable item here.

PR #2327, opened 2023-11-02
Improve the warning for an unexpected dollar sign in a Markdown file

odow:
I was looking into https://github.com/JuliaDocs/Documenter.jl/issues/2288

The issue is that by the time we notice the value, we've lost the original source file and line information.

@mortenpi might have some ideas for how to get it back?

cc @kellertuer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-* `id` anchors may now start with a numeric digit. ([#744]), ([#2325])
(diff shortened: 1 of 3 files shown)


fredrikekre (approved):

mortenpi:
Another way would be to do the check somewhere earlier in the pipeline though. Would also avoid code duplication.

odow:
Good by me. (Except for the status failures?)

mortenpi:
> Except for the status failures?

I think they're unrelated..

mortenpi:
It looks like the Coveralls failures resolved themselves.

PR #2330, opened 2023-11-03
Fix `linkcheck` for docstrings

### LINKED ISSUE #2329: Linkcheck does not traverse into docstrings
When `makedocs` is called with `linkcheck=true`, the [`Documenter.linkcheck`](https://github.com/JuliaDocs/Documenter.jl/blob/c6abf9d7f00acd0293a6bad90267ea0669c01ab7/src/docchecks.jl#L168) routine does not traverse into `DocsNode`. That is, broken links inside of docstrings will not be reported.

The reason for this is

https://github.com/JuliaDocs/Documenter.jl/blob/c6abf9d7f00acd0293a6bad90267ea0669c01ab7/src/docchecks.jl#L183-L184

Since the `node.element` for a docstring is a `DocsNode`, nothing happens.

To fix this, `linkcheck(node::MarkdownAST.Node, doc::Document; method::Symbol=:HEAD)` should receive `element=node.element` as an explicit second parameter, so that we can dispatch on that. Other places in `Documenter` follow the same approach, e.g., the [`domify`](https://github.com/JuliaDocs/Documenter.jl/blob/c6abf9d7f00acd0293a6bad90267ea0669c01ab7/src/html/HTMLWriter.jl#L1598) function in the `HTMLWriter`.

An alternative solution would be for [`AbstractTrees.PreOrderDFS`](https://github.com/JuliaDocs/Documenter.jl/blob/c6abf9d7f00acd0293a6bad90267ea0669c01ab7/src/docchecks.jl#L172) to traverse into `DocsNode`, but that might be a lot trickier. Besides, [`DocumenterCitations` has the exact same problem](https://github.com/JuliaDocs/DocumenterCitations.jl/issues/58), and there either `linkcheck` would have to dispatch for elements of type [`BibliographyNode`](https://github.com/JuliaDocs/DocumenterCitations.jl/blob/426a649d9f7b465c8fe3f5864ea4f86a641a7737/src/bibliography_node.jl#L32) or `AbstractTrees.PreOrderDFS` would have to traverse into `BibliographyNode`.

I would maybe also refactor [`Documenter.linkcheck`](https://github.com/JuliaDocs/Documenter.jl/blob/c6abf9d7f00acd0293a6bad90267ea0669c01ab7/src/docchecks.jl#L183) a little bit. Right now it is difficult to unit-test. The call to `curl` should be probably be isolated more into a `linkcheck(url::String)` method that returns [`(status, scheme, location)`](https://github.com/JuliaDocs/Documenter.jl/blob/c6abf9d7f00acd0293a6bad90267ea0669c01ab7/src/docchecks.jl#L225C13-L225C37): I don't think that network access on Github workflow runners (or anywhere else the test suite runs) is necessarily robust. For unit-testing, you'd want to mock the call to `curl` and use a pre-defined `(status, scheme, location)` for a given `url`.

We could easily add a `cache` keyword argument to `linkcheck(doc)` and `linkcheck(node, doc; method=:HEAD)` (which should be `linkcheck(node, element, doc; method=:HEAD)`, as discussed above). This `cache` would a `Dict` mapping `url::String` => `(status, scheme, location)`. It would be internal use only (specifically for unit testing, that is). If a `url` is in the `cache` (set up in a test), we'd just use the cached values instead of calling `linkcheck(url)` that goes out to `curl` and the network.

goerz:
Dispatch `Documenter.linkcheck` on the type of `node.element`, allowing it to process a `Documenter.DocsNode`

Closes #2329

### DIFF
--- a/src/docchecks.jl
+++ b/src/docchecks.jl
@@ -169,9 +169,7 @@ function linkcheck(doc::Document)
     if doc.user.linkcheck
         if hascurl()
             for (src, page) in doc.blueprint.pages
-                for node in AbstractTrees.PreOrderDFS(page.mdast)
-                    linkcheck(node, doc)
-                end
+                linkcheck(page.mdast, doc)
             end
         else
             @docerror(doc, :linkcheck, "linkcheck requires `curl`.")
@@ -180,9 +178,20 @@ function linkcheck(doc::Document)
     return nothing
 end
 
-function linkcheck(node::MarkdownAST.Node, doc::Document; method::Symbol=:HEAD)
-    node.element isa MarkdownAST.Link || return
-    link = node.element
+function linkcheck(mdast::MarkdownAST.Node, doc::Document)
+    for node in AbstractTrees.PreOrderDFS(mdast)
+        linkcheck(node, node.element, doc)
+    end
+end
+
+function linkcheck(node::MarkdownAST.Node, element::MarkdownAST.AbstractElement, doc::Document)
+    # The linkcheck is only active for specific `element` types
+    # (`MarkdownAST.Link`, most importantly), which are defined below as more
+    # specific methods
+    return nothing
+end
+
+function linkcheck(node::MarkdownAST.Node, link::MarkdownAST.Link, doc::Document; method::Symbol=:HEAD)
 
     # first, make sure we're not supposed to ignore this link
     for r in doc.user.linkcheck_ignore
@@ -244,7 +253,7 @@ function linkcheck(node::MarkdownAST.Node, doc::Document; method::Symbol=:HEAD)
             elseif protocol === :HTTP && status == 405 && method === :HEAD
                 # when a server doesn't support HEAD requests, fallback to GET
                 @debug "linkcheck '$(link.destination)' status: $(status), retrying without `-I`"
-                return linkcheck(link, doc; method=:GET)
+                return linkcheck(node, link, doc; method=:GET)
             else
                 @docerror(doc, :linkcheck, "linkcheck '$(link.destination)' status: $(status).")
             end
@@ -254,6 +263,14 @@ function linkcheck(node::MarkdownAST.Node, doc::Document; method::Symbol=:HEAD)
     end
     return false
 end
+
+
+function linkcheck(node::MarkdownAST.Node, docs_node::Documenter.DocsNode, doc::Document)
+    for mdast in docs_node.mdasts
+        linkcheck(mdast, doc)
+    end
+end
+
 
 linkcheck_ismatch(r::String, url) = (url == r)
 linkcheck_ismatch(r::Regex, url) = occursin(r, url)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -21,10 +21,6 @@ end
     @info "Building missingdocs/make.jl"
     include("missingdocs/make.jl")
 
-    # Test invalid links (linkcheck option)
-    @info "Building invalidlinks/make.jl"
-    include("invalidlinks/make.jl")
-
     # Error reporting.
     @info "Building errors/make.jl"
     @quietly include("errors/make.jl")
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -21,6 +21,10 @@ end
     @info "Building missingdocs/make.jl"
     include("missingdocs/make.jl")
 
+    # Test invalid links (linkcheck option)
+    @info "Building invalidlinks/make.jl"
+    include("invalidlinks/make.jl")
+
     # Error reporting.
     @info "Building errors/make.jl"
     @quietly include("errors/make.jl")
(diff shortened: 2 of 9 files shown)


goerz:
This solves the problem that links inside docstrings were not being checked.

As soon as this is merged and released, I'll be able to address https://github.com/JuliaDocs/DocumenterCitations.jl/issues/58

I haven't done any of the further refactoring that I proposed https://github.com/JuliaDocs/Documenter.jl/issues/2329#issue-1976691911. It's probably best to keep this PR minimal and get it out the door.

Maybe I can do the refactoring in a separate PR at some point in the future. The tests I've included in this PR are okay because I'm testing with non-existent domains, which are always going to fail (even if the CI doesn't have network access). If we wanted to do more elaborate testing of the `linkcheck` function, e.g., for links with different return codes, we'd definitely need to be able to mock out the calls to `curl`, or any actual network access.

goerz:
Oh, and I should still update the changelog...

src/docchecks.jl
```diff
@@ -180,9 +178,21 @@ function linkcheck(doc::Document)
...
 
-function linkcheck(node::MarkdownAST.Node, doc::Document; method::Symbol=:HEAD)
-    node.element isa MarkdownAST.Link || return
-    link = node.element
+function linkcheck(mdast::MarkdownAST.Node, doc::Document)
+    for node in AbstractTrees.PreOrderDFS(mdast)
+        linkcheck(node, node.element, doc)
+    end
+end
+
```
mortenpi, line 186:
```suggestion
```

src/docchecks.jl
```diff
@@ -255,6 +265,17 @@ function linkcheck(node::MarkdownAST.Node, doc::Document; method::Symbol=:HEAD)
     return false
 end
 
+
+function linkcheck(node::MarkdownAST.Node, docs_node::Documenter.DocsNode, doc::Document)
+    success = true
+    for mdast in docs_node.mdasts
+        success &= !(linkcheck(mdast, doc) === false)
+        # `linkcheck` may return `true` / `false` / `nothing`
```
mortenpi, line 273: Just personally, I'm used to having the explanation before the behaviour:

```suggestion
        # `linkcheck` may return `true` / `false` / `nothing`
        success &= !(linkcheck(mdast, doc) === false)
```

mortenpi (approved):
LGTM. Thanks for catching this! This is a pretty bad regression.

As for the refactoring: yea, I think it would be better to do that in a separate PR, as not to tangle up the fix and refactor.

src/docchecks.jl
```diff
@@ -255,6 +265,17 @@ function linkcheck(node::MarkdownAST.Node, doc::Document; method::Symbol=:HEAD)
...
+    end
+    return success
+end
```
mortenpi, line 276: Is it actually necessary to keep track of `success` here, since we're kinda discarding the information anyway in the main loop?

test/runtests.jl
```diff
@@ -21,6 +21,10 @@ end
     @info "Building missingdocs/make.jl"
     include("missingdocs/make.jl")
 
+    # Test invalid links (linkcheck option)
+    @info "Building invalidlinks/make.jl"
+    include("invalidlinks/make.jl")
```
mortenpi, line 26: After writing up https://github.com/JuliaDocs/Documenter.jl/issues/2329#issuecomment-1794116134, I now  think that we should actually move these tests into `online_linkcheck.jl`, since while they fail, they do try to talk to the internet, which can be flaky in other ways too probably.

src/docchecks.jl
goerz: I suppose the options are:

1. Always return `nothing` (i.e., no `return` statements). That works fine (since we discard the information in the main loop, as you say), but is a little "cosmetically" inconsistent with other `linkcheck` methods, in that some of the methods will return `false` on failure and others won't. Also, returning something informative might help in the context of (future) unit tests.
2. Return `false` early as soon as on of the elements in `mdasts` has a failing link. That's probably most consistent with the other `linkcheck` methods, but it seemed better to get a list of *all* broken links in one go, as opposed to having to iterate the edit-push cycle multiple times.
3. The above, keeping track of `success`

So I'd got with option 1 or 3 over 2, but let me know which you'd prefer

test/runtests.jl
goerz: Makes sense!

I'll clean up the other stuff – let me know about https://github.com/JuliaDocs/Documenter.jl/pull/2330#discussion_r1384076737; and then see if I can move it over.

src/docchecks.jl
goerz: Well, option 1 actually seems fine for the moment. If someone wants "informative returns" for better unit testing in the future, it can be re-added. Also, in that case we'd probably want to consistently return `true`/`false`, everywhere, not `nothing`/`false`. So, I'll just got with that…

test/runtests.jl
goerz: OK, I moved it over to `test/online_linkcheck.jl` (and renamed `test/invalidlinks/make.jl` -> `test/docstring_links/make.jl`)

goerz:
Okay, should be good for merging or further review.

The coverage went down from moving the test into `online_linkcheck.jl`, so I added coverage collection to that CI job, in the last of the commits.

mortenpi:
LGTM, thanks @goerz

PR #2335, opened 2023-11-08
breaking changes: Woodpecker CI updates and deprecations

uncomfyhalomacro:
- Woodpecker CI is no longer backwards compatible with Drone due to lots of deprecated syntax and env vars

https://woodpecker-ci.org/docs/next/migrations#next-200

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -439,7 +439,7 @@ uppercase environment variables to your pipeline. Next, create a new pipeline
 configuration file called `.woodpecker.yml` with the following contents:
 
 ```yaml
-pipeline:
+steps:
     docs:
     when:
         branch: main  # update to match your development branch
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -1056,12 +1056,12 @@ function authenticated_repo_url(cfg::Woodpecker)
     # If so, we just split the `http(s)://` from the string we want 
     # e.g. `https://github.com/JuliaDocs/Documenter.jl` to `github.com/JuliaDocs/Documenter.jl`.
     if haskey(ENV, "FORGE_URL")
-        return "https://$(ENV["CI_REPO_OWNER"]):$(ENV["PROJECT_ACCESS_TOKEN"])@$(ENV["FORGE_URL"])/$(cfg.woodpecker_repo).git"
+        return "https://$(ENV["PROJECT_ACCESS_TOKEN"])@$(ENV["FORGE_URL"])/$(cfg.woodpecker_repo).git"
     else
         if occursin(cfg.woodpecker_repo, cfg.woodpecker_forge_url)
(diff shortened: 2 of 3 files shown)


uncomfyhalomacro:
I will fix this over the weekend. :)

mortenpi:
So if I understand correctly, this will now only work in pipelines running on Woodpecker 2? And it would start failing on any Woodpecker 1 installations?

uncomfyhalomacro:
> So if I understand correctly, this will now only work in pipelines running on Woodpecker 2? And it would start failing on any Woodpecker 1 installations?

Yes and No. The previous PR i made before was pre-1.x, i think specifically 0.15.x (?). The latest stable release version 1.0.0 starts to deprecate but not remove some old environmental variables e.g. `CI_BUILD*` -> `CI_PIPELINE*`. In the 2.0.0 `next` branch which is not released yet, those deprecated becomes "removed" so those env vars won't exist anymore.

mortenpi:
Ah, right-right. But this would fail for anyone using Woodpecker 0.15? And this is theoretically possible, since it can be self-hosted.. actually.. I am a little bit confused -- does Woodpecker even offer a central service? Or is it always self-hosted?

I feel that to do this properly, we would need to version guard things, if that's something that can be done. I.e. use different sets of environment variables as needed. Can we detect 0.15 vs 1? Otherwise, I feel that this would constitute dropping 0.15 support, which would be a breaking change for Documenter.

uncomfyhalomacro:
> Ah, right-right. But this would fail for anyone using Woodpecker 0.15? And this is theoretically possible, since it can be self-hosted.. actually.. I am a little bit confused -- does Woodpecker even offer a central service? Or is it always self-hosted?
> 
> I feel that to do this properly, we would need to version guard things, if that's something that can be done. I.e. use different sets of environment variables as needed. Can we detect 0.15 vs 1? Otherwise, I feel that this would constitute dropping 0.15 support, which would be a breaking change for Documenter.

Sure. I can help on that. I will check out the docs later. :D

uncomfyhalomacro:
> Ah, right-right. But this would fail for anyone using Woodpecker 0.15? And this is theoretically possible, since it can be self-hosted.. actually.. I am a little bit confused -- does Woodpecker even offer a central service? Or is it always self-hosted?
> 
> I feel that to do this properly, we would need to version guard things, if that's something that can be done. I.e. use different sets of environment variables as needed. Can we detect 0.15 vs 1? Otherwise, I feel that this would constitute dropping 0.15 support, which would be a breaking change for Documenter.

Seems we can detect the version with `CI_SYSTEM_VERSION` I will see how to work with it once I am done with things. Possibly I will ping again for a code review next week

mortenpi:
Re #2339, this one we can close, right?

uncomfyhalomacro:
Yes. Closed now. As for the superseding PR, 1.0.x and 2.0.x are compatible except some removed support for other environmental variables pre-1.0.x

Issue #2272, opened 2023-09-19
Express sizes in more human-readable form?


cormullion:
Congratulations on Documenter 1.0 - I've been using it without issues for various packages. It's awesome!

I think that this output:

<img width="1110" alt="Screenshot 2023-09-19 at 09 18 01" src="https://github.com/JuliaDocs/Documenter.jl/assets/52289/e10b8f94-6931-4436-a570-37b6b96654d3">

could be made better if the "bytes" were converted to something more familiar. These days I'd be happy with MB, but KB could be OK. And perhaps it could be rounded to the nearest sensible amount, the precision here seems unnecessary (and we'd avoid the Kibibytes versus Kilobytes issue too).

mortenpi:
Yeah, absolutely. I was just too lazy to implement it :slightly_smiling_face:

mortenpi:
The only thing is that there is a bit of range to the values, so I think it should pick the appropriate unit based on the order of magnitude (don't need to worry about anything other than `B`, `KiB` and `MiB` though).

Hetarth02:
I would be happy to contribute.

@mortenpi Can you guide me, what we want to achieve(display in kb, mb, etc.) and where should I start looking?

mortenpi:
@Hetarth02 The relevant line of code should be here: https://github.com/JuliaDocs/Documenter.jl/blob/5dafb6488f90d173ca4fcfeead0396332bdc6de6/src/html/HTMLWriter.jl#L1771-L1776

And as far as what we want to achieve: make sure that it prints the file sizes in `B`, `KiB` and `MiB` as appropriate when printing those values. No need to support `GiB` and bigger, since if the user has pages that big, they have bigger problems.

cormullion:
`9223372036854775807` was the worst one 😂

mortenpi:
> `9223372036854775807` was the worst one 😂

That probably needs to be special cased though. I'm using `typemax(Int)` to indicate "no limit".

https://github.com/JuliaDocs/Documenter.jl/blob/5dafb6488f90d173ca4fcfeead0396332bdc6de6/src/html/HTMLWriter.jl#L524

mortenpi:
#2344

cormullion:
Thanks!

PR #2344, opened 2023-11-13
Updated format to display File sizes

Hetarth02:
This PR aims to close Issue #2272.

Before:

268888423-e10b8f94-6931-4436-a570-37b6b96654d3

After:

image


src/html/HTMLWriter.jl
```diff
@@ -1790,6 +1795,27 @@ function write_html(ctx::HTMLContext, navnode::Documenter.NavNode, page_html::DO
     return true
 end
 
+"""
+Calculates and converts bytes to appropriate format.
+"""
+function format_units(size, unit = "bytes")
```
odow, line 1801: Just make this function return a `String`. It doesn't get used elsewhere

src/html/HTMLWriter.jl
```diff
@@ -1768,11 +1768,16 @@ function write_html(ctx::HTMLContext, navnode::Documenter.NavNode, page_html::DO
...
+    file_size_format_results = format_units(file_size)
+    size_threshold_format_results = format_units(ctx.settings.size_threshold)
+    size_threshold_warn_format_results = format_units(ctx.settings.size_threshold_warn)
+
     size_threshold_msg(var::Symbol) = """
     Generated HTML over $(var) limit: $(navnode.page)
-        Generated file size: $(file_size) (bytes)
-        size_threshold_warn: $(ctx.settings.size_threshold_warn) (bytes)
-        size_threshold:      $(ctx.settings.size_threshold) (bytes)
+        Generated file size: $(file_size_format_results.size) $(file_size_format_results.unit == "" ? "" : "($(file_size_format_results.unit))")
```
odow, line 1778:
```suggestion
        Generated file size: $file_size_format_results
```

odow (commented):
Needs tests. Really just a few simple ones for `format_units`

src/html/HTMLWriter.jl
```diff
@@ -1790,6 +1795,27 @@ function write_html(ctx::HTMLContext, navnode::Documenter.NavNode, page_html::DO
...
+        size = size / 1024
+        unit = "KiB"
+
+        if size >= 1024
+            size = size / 1024
+            unit = "MiB"
+        end
+    end
+
+    return (size = round(size, digits = 2), unit = unit)
```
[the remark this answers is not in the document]
Hetarth02: Since this is a utility function, I would argue against doing this, as it can be useful somewhere else in the future.

Hetarth02:
> Needs tests. Really just a few simple ones for `format_units`

@odow I really don't know how to write tests. Can anyone guide me here?

CC: @mortenpi

src/html/HTMLWriter.jl
Hetarth02: Since this is a utility function, I would argue against doing this, as it can be useful somewhere else in the future.

mortenpi: I'm with @odow here: it's better to have the string formatting complexity just once in this utility function, rather than reproduced three times in `size_threshold_msg`. The latter is not very readable right now. The function is quite simple and internal, so we can always change things up if we need to reuse this somehow.

Hetarth02: > The function is quite simple and internal, so we can always change things up if we need to reuse this somehow.

Alright, no problem will refactor and update the PR.

mortenpi:
@Hetarth02 The tests for `HTMLWriter` utility functions live in https://github.com/JuliaDocs/Documenter.jl/blob/master/test/htmlwriter.jl

I thought that the Julia manual would have a tutorial on testing, but it doesn't look like it does. But these resources might help get started if you haven't touched tests before:

* https://julia.quantecon.org/software_engineering/testing.html#more-on-writing-tests
* https://docs.julialang.org/en/v1/stdlib/Test/
* https://pkgdocs.julialang.org/dev/creating-packages/#Adding-tests-to-the-package

Hetarth02:
I have added some test. Please review them and let me know if they are what you expected.

test/htmlwriter.jl
```diff
@@ -24,6 +24,25 @@ function verify_redirect_file(redirectfile, version)
...
+        unit = "KiB"
+        if size >= 1024
+            size = size / 1024
+            unit = "MiB"
+        end
+    end
+
+    return string(round(size, digits = 2), " (", unit, ")")
+end
+
```
mortenpi, line 45: The tests should test the actual function you've implemented.

```suggestion
```

test/htmlwriter.jl
```diff
@@ -296,6 +315,11 @@ end
         @test html.size_threshold_warn == 1234
     end
 
+    @testset "HTML: size_threshold_result_formats" begin
+        @test typeof(format_units(1024)) == String
+        @test typeof(format_units(typemax(Int))) == String
```
mortenpi, line 320: Slightly expanding, to cover more edge cases, and also making sure (1) we test the actual function, and (2) actually testing the output (which is the interesting thing to test for):

```suggestion
        @test HTMLWriter.format_units(0) == "0 (bytes)"
        @test HTMLWriter.format_units(1) == "1 (bytes)"
        @test HTMLWriter.format_units(1023) == "1023 (bytes)"
        @test HTMLWriter.format_units(1024) == "1 (KiB)"
        @test HTMLWriter.format_units(4000) == "4 (KiB)"
        @test HTMLWriter.format_units(2^20 + 123) == "@ (MiB)"
        @test HTMLWriter.format_units(typemax(Int) == "(no limit)"
```

Not 100% sure if these tests pass actually.

test/htmlwriter.jl
```diff
@@ -296,6 +315,11 @@ end
         @test html.size_threshold_warn == 1234
     end
 
+    @testset "HTML: size_threshold_result_formats" begin
```
mortenpi, line 318:
```suggestion
    @testset "HTML: format_units" begin
```

src/html/HTMLWriter.jl
```diff
@@ -1790,6 +1795,25 @@ function write_html(ctx::HTMLContext, navnode::Documenter.NavNode, page_html::DO
     return true
 end
 
+"""
+Calculates and converts bytes to appropriate format.
+"""
+function format_units(size, unit = "bytes")
+    if size == typemax(Int)
+        return "No Limit"
```
mortenpi, line 1803:
```suggestion
        return "(no limit)"
```

src/html/HTMLWriter.jl
```diff
@@ -1790,6 +1795,25 @@ function write_html(ctx::HTMLContext, navnode::Documenter.NavNode, page_html::DO
     return true
 end
 
+"""
+Calculates and converts bytes to appropriate format.
+"""
+function format_units(size, unit = "bytes")
```
mortenpi, line 1801: It doesn't really make sense to pass this as an argument here.

```suggestion
function format_units(size)
    unit = "bytes"
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `id` anchors may now start with a numeric digit. ([#744], [#2325])
 * Documenter prints a more informative warning now if there is unexpected Julia interpolation in the Markdown (e.g. from errant `$` signs). ([#2288], [#2327])
 * Documenter now warns when it encounters invalid keys in the various key-value at-blocks. ([#2306], [#2324])
+* File sizes are now expressed in more Human-readable format. ([#2272])
```
mortenpi, line 13:
```suggestion
* File sizes are now expressed in more human-readable format. ([#2272], [#2344])
```

mortenpi (commented):
Partly just a bit of bikeshedding, but also fixing the tests.

Hetarth02:
Updated as per the feedbacks.

test/htmlwriter.jl
```diff
@@ -296,6 +296,18 @@ end
         @test html.size_threshold_warn == 1234
     end
 
+    @testset "HTML: format_units" begin
+        @test typeof(HTMLWriter.format_units(1024)) == String
+        @test typeof(HTMLWriter.format_units(typemax(Int))) == String
```
mortenpi, line 301:
```suggestion
```

Hetarth02:
@mortenpi 

Do I need to reiterate on this, or is this done?

Can I help in anyway in merging this?

mortenpi (approved):
No, this is good to go. Just trying to get the CI green (#2351).

thazhemadam:
Just saw this and wanted to note that [`Base.format_bytes`](https://github.com/JuliaLang/julia/blob/0c46852901c63b33f0603b0afd58ff1da687d760/base/timing.jl#L131-L140) could/should probably have been used here.
It's significantly more performant and covers more cases as well (as unnecessary as that may be), all the while lifting the (seemingly minor) burden of maintenance off your shoulders. 🙂

Hetarth02:
> Just saw this and wanted to note that Base.format_bytes could/should probably have been used here.
It's significantly more performant and covers more cases as well (as unnecessary as that may be), all the while lifting the (seemingly minor) burden of maintenance off your shoulders. 🙂

@mortenpi Should we reopen this?

mortenpi:
> [Base.format_bytes](https://github.com/JuliaLang/julia/blob/0c46852901c63b33f0603b0afd58ff1da687d760/base/timing.jl#L131-L140) could/should probably have been used here.

It's good to know that it exists, and could have been used, but don't think it's worth changing now that this is merged. Something to note, though, is that `Base.format_bytes` is not public API.

Issue #2343, opened 2023-11-12
Coveralls fails to submit coverage


mortenpi:
E.g. https://github.com/JuliaDocs/Documenter.jl/actions/runs/6843332387/job/18605813768

Thought that this is maybe some weird backend on their part, but doesn't look like it. _Probably_ a race condition on our part w.r.t. parallel builds. It started after #2330, but it's not clear how that would break things.

mortenpi:
#2347 will remove the Coveralls submission from online linkcheck. Unsure if it will fix things. But one way or another, the actionable item here is to figure out why it's failing and re-enable it.

PR #2347, opened 2023-11-13
Disable coveralls submission in online linkcheck

mortenpi:
Just a test to debug #2343.


mortenpi:
Let's merge this and see what happens. I still see the Covecov failures quite often if you have like multiple PRs running in parallel.

PR #2350, opened 2023-11-18
1341 latex engines

### LINKED ISSUE #1341: Is it possible to add an options to use other latex engines?
My documents include Mandarin Chinese characters.
So I'm wondering whether it is possible to add an argument to allowing switching latex engines?

Thanks.

BambOoxX:
Adds `texify` as a latex engine to allow compilation when `latexmk` is not available, and makes `native` decide automatically which is available between `latexmk` and `texify`.

Partially fixes #1341

### DIFF
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -48,10 +48,10 @@ considered to be deprecated), or to an empty string if `TRAVIS_TAG` is unset.
 
 See [Other Output Formats](@ref) for more information.
 """
-struct LaTeX <: Documenter.Writer
+mutable struct LaTeX <: Documenter.Writer
     platform::String
-    version::String
-    tectonic::Union{Cmd,String,Nothing}
+    const version::String
+    const tectonic::Union{Cmd,String,Nothing}
     function LaTeX(;
             platform = "native",
             version  = get(ENV, "TRAVIS_TAG", ""),
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -56,7 +56,7 @@ struct LaTeX <: Documenter.Writer
             platform = "native",
             version  = get(ENV, "TRAVIS_TAG", ""),
             tectonic = nothing)
-        platform ∈ ("native", "tectonic", "docker", "none") || throw(ArgumentError("unknown platform: $platform"))
+        platform ∈ ("native", "latexmk","texify", "tectonic", "docker", "none") || throw(ArgumentError("unknown platform: $platform"))
         return new(platform, string(version), tectonic)
     end
 end
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -177,6 +177,17 @@ const DOCKER_IMAGE_TAG = "0.1"
 
 function compile_tex(doc::Documenter.Document, settings::LaTeX, fileprefix::String)
     if settings.platform == "native"
+        @info "LaTeXWriter: attempting to find native platform."
+        native_platform = hastex()
+        if !isempty(native_platform)
+            @info "LaTeXWriter: found native platform $native_platform."
+            settings.platform = native_platform
+            return compile_tex(doc,settings,fileprefix)
+        else
+            @error "LaTeXWriter: no native platform found."
+            return false
+        end
+    elseif settings.platform == "latexmk"
         Sys.which("latexmk") === nothing && (@error "LaTeXWriter: latexmk command not found."; return false)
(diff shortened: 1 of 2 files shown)


test/examples/make.jl
```diff
@@ -20,7 +20,7 @@ else
     ["html", "html-meta-custom", "html-mathjax2-custom", "html-mathjax3", "html-mathjax3-custom",
     "html-local", "html-draft", "html-repo-git", "html-repo-nothing", "html-repo-error",
     "html-sizethreshold-defaults-fail", "html-sizethreshold-success", "html-sizethreshold-ignore-success", "html-sizethreshold-override-fail", "html-sizethreshold-ignore-success", "html-sizethreshold-ignore-fail",
-    "latex_texonly", "latex_simple_texonly", "latex_showcase_texonly", "html-pagesonly"]
```
mortenpi, line 23: We don't want to run the full LaTeX builds as part of the main test suite (because of the dependencies). But we have a special test file for running those:

https://github.com/JuliaDocs/Documenter.jl/blob/7251be3e45579591f755b668e8e364b367f5b7a6/test/examples/tests_latex.jl#L3-L7

Which we run as part of this workflow:

https://github.com/JuliaDocs/Documenter.jl/blob/7251be3e45579591f755b668e8e364b367f5b7a6/.github/workflows/CI.yml#L52-L83

src/latex/LaTeXWriter.jl
```diff
@@ -48,15 +48,15 @@ considered to be deprecated), or to an empty string if `TRAVIS_TAG` is unset.
 
 See [Other Output Formats](@ref) for more information.
 """
-struct LaTeX <: Documenter.Writer
+mutable struct LaTeX <: Documenter.Writer
     platform::String
-    version::String
-    tectonic::Union{Cmd,String,Nothing}
+    const version::String
+    const tectonic::Union{Cmd,String,Nothing}
```
mortenpi, line 54: This doesn't work on 1.6. But regardless, I don't see much value in making `.platform` mutable anyway -- just call `hastex()` here in the constructor.

src/latex/LaTeXWriter.jl
```diff
@@ -87,7 +87,14 @@ _hash(x) = string(hash(x))
...
 
-hastex() = (try; success(`latexmk -version`); catch; false; end)
+hastex() = (
+    try
+        success(`latexmk -version`) && return "latexmk"
+        success(`texify --version`) && return "texify"
+    catch
+        return ""
+    end
+)
```
mortenpi, line 97: Cleaner if it's using just the standard function syntax:

```suggestion
function hastex()
    try
        success(`latexmk -version`) && return "latexmk"
        success(`texify --version`) && return "texify"
    catch
        return ""
    end
end
```

test/examples/make.jl
```diff
@@ -703,6 +703,59 @@ else
...
+        source="src.latex_simple",
+        pages=["Main section" => ["index.md"]],
+        doctest=false,
+        debug=true,
+    )
+else
+    @info "Skipping build: LaTeXWriter/latex_simple_latekmk"
+    @debug "Controlling variables:" EXAMPLE_BUILDS get(ENV, "DOCUMENTER_TEST_EXAMPLES", nothing)
+    nothing
+end
```
mortenpi, line 740: This pretty much duplicates `latex_simple_nondocker` (which, to be fair, I'm not 100% sure if we're running on CI right now..). So let's just make that explicitly `platform = "latexmk"`, and test the `hastex` behavior with a unit test.

mortenpi (changes requested):
Sorry for the delay in reviewing this.

I think I'm okay with this -- adding `texify` wouldn't add much maintenance burden. However, I would like to see the texify build pass on CI and upload a PDF as an artifact, to see what it looks like.

Also would need docs and a CHANGELOG entry.

src/latex/LaTeXWriter.jl
BambOoxX: Do you mean you'd prefer keeping this an immutable `struct` ?

test/examples/make.jl
BambOoxX: But if there is no dedicated CI for `texify` wouldn't it mean that if some options change in `texify`, which is unlikely though, that would not be caught in tests ?

src/latex/LaTeXWriter.jl
BambOoxX: This was reverted in favor of a `hastex()` call in the constructor as suggested

test/examples/make.jl
BambOoxX: This was reverted

PR #2351, opened 2023-11-19
fix doctests on nightly

mortenpi:


test/doctests/stdouts/31.v1_11.stdout
```diff
@@ -0,0 +1,33 @@
...
+│ ```
+│ 
+│ Subexpression:
+│ 
+│ baz(20)
+│ 
+│ Evaluated output:
+│ 
+│ ERROR: UndefVarError: `baz` not defined in `Main`
+│ Suggestion: check for spelling errors or missing imports. No global of this name exists in this module.
```
mortenpi, line 17: The difference in `31` is straighforward -- Julia now outputs this `Suggestion` with some errors. So I think we can just support version-specific reference files, rather than trying to hack this with regexes.

mortenpi:
Well. This was fun.

PR #2356, opened 2023-11-27
Add AzureDevOps remote

mmiller-max:
Adding `AzureDevOps` as a `Remote`. The logic was already there for URL parsing so just brought it to the non-deprecated interface.

So far I've reused `format_commit` to avoid duplicating the commit suffix, but if this function will be removed let me know and I can insert this logic directly into `fileurl`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
+### Added
+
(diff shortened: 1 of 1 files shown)


mortenpi (commented):
LGTM, but do you mind adding some tests into https://github.com/JuliaDocs/Documenter.jl/blob/master/test/remotes.jl ?

fingolfin:
@mmiller-max are you still interested in this, and adding those tests?

fingolfin:
No response by @mmiller-max in the past 2 years, closing

PR #2371, opened 2023-12-08
fix typo in HTMLWriter.HTML

pierre-haessig:
I just spotted a tiny typo in the docstring of [`HTMLWriter.HTML`](https://documenter.juliadocs.org/stable/lib/internals/writers/#Documenter.HTMLWriter.HTML)

it mention the `remotes` argument of `makedoc` which should be plural.


mortenpi:
Thank you!

PR #2372, opened 2023-12-10
Fix handling of remote URLs for standard libraries, tag 1.2.2

mortenpi:
These changes are necessary to fix the generation of source URLs for standard libraries in the Julia manual.

* In `relpath_from_remote_root` we no longer expand symlinks in the docstring metadata filepaths. This is because for normal builds, the standard libraries in `usr/share/julia/stdlib/...` are symlinked to `stdlib/`, and this way we can handle both the CI build and normal build the same way. At the same time, it also doesn't seem important for the packages, since I wouldn't expect there to be any symlinks there normally.
* In the CI builds, the filepaths in the docstring metadata for standard libraries are actually the `/cache/*` ones. So if the path doesn't exists, we'll call `Base.fixup_stdlib_path` on it, to see if that allows us to find the actual source path.

X-ref: https://github.com/JuliaLang/julia/pull/47105#issuecomment-1842130476

### DIFF
--- a/src/documents.jl
+++ b/src/documents.jl
@@ -788,9 +788,15 @@ function relpath_from_remote_root(doc::Document, path::AbstractString)
     end
     ispath(path) || error("relpath_from_repo_root called with nonexistent path: $path")
     isabspath(path) || error("relpath_from_repo_root called with non-absolute path: $path")
-    # We want to expand the path properly, including symlinks, so we call realpath()
-    # Note: it throws for non-existing files, but we just checked for it.
-    path = realpath(path)
+    # We'll normalize the path, removing `..`-s etc.
+    #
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Version [v1.2.2] - 2023-12-10
```
```suggestion
## Version [v1.2.2] - 2024-01-15
```

fingolfin:
@mortenpi what's the status of this PR?

mortenpi:
Excellent question. I assume it's a fix we'd want. But I don't really have any memory of what's going on here.

Issue #2373, opened 2023-12-10
Search should not be case-sensitive


mortenpi:
In Documenter's own docs, searching for `editurl` doesn't yield any results:

![image](https://github.com/JuliaDocs/Documenter.jl/assets/147757/9b0dd38a-d9ae-4655-89d6-860f3b79ff73)

But `EditURL` does:

![image](https://github.com/JuliaDocs/Documenter.jl/assets/147757/f9982fd5-edd2-4481-ae53-24bc988d5aea)

We probably want to prioritize case-matches if we can, but I think we should not exclude results just because the case does not match.

FWIW, this seems to be a regression relative to the old search:

![image](https://github.com/JuliaDocs/Documenter.jl/assets/147757/15f03065-5ab5-427b-a559-99f5327ba087)

cc @Hetarth02

Hetarth02:
Yup, I agree this should not be the case. I will try if I can fix this quickly.

PR #2374, opened 2023-12-10
Made search case insensitive

Hetarth02:
This PR aims to fix Issue #2373.

image


Hetarth02:
@mortenpi  Before merging, where do I add the entry in `CHANGELOG.md`? In version 1.2.1 or any other place?

fredrikekre:
Vim has `smartcase` which I have started to like. Basically, if the search query contains at least one upper case letter, then the search is case sensitive, otherwise not. Perhaps that is something to mimic here?

fredrikekre:
> where do I add the entry in CHANGELOG.md? In version 1.2.1 or any other place?

Above 1.2.1 in a new header with `### Unreleased`

LilithHafner:
Honestly, this seems like an upstream bug in mini-search. I would hope that even a very light fuzzy search would fix this, but apparently they use Levenshtein edit distance to compute fuzzy matches which is not a case-aware system (i.e. Editlink and editlink are equally far apart as editWink and editlink).

source: https://lucaong.github.io/minisearch/types/MiniSearch.SearchOptions.html#__type.weights.__type-10.fuzzy-1

I think smartcase seems like a reasonable idea and it's also worth trying out `fuzzy: true` to see if minisearch's default fuzzy parameters are enough to fix this.

LilithHafner (approved):
This seems like an unambiguous improvement. LGTM.

Smartcase is certainly an interesting idea to try out, but does not block this PR.

CHANGELOG.md
```diff
@@ -3,10 +3,9 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
-
 ### Changed
 
+* Made search case insensitive ([#2373], [#2374])
```
mortenpi, line 8:
```suggestion
* The search in the HTML output is now case-insensitive. ([#2373], [#2374])
```

mortenpi (approved):
LGTM as well, sorry for the delay. Thank you!

CHANGELOG.md
```diff
@@ -3,10 +3,9 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
-
 ### Changed
```
```suggestion
## Unreleased

### Changed
```

Issue #1675, opened 2021-08-15
`@gif` block


ericphanson:
I think it would be awesome if there was an `@gif` block that acted like an `@repl` but instead of printing the commands linearly, it generated an asciinema gif and inserted it. I made a prototype of sorts in https://github.com/ericphanson/TravelingSalesmanExact.jl/pull/16. See it in action here: https://ericphanson.github.io/TravelingSalesmanExact.jl/v0.3.7/#Example, or in a screenshot:

<img width="1155" alt="Screenshot 2021-08-15 at 21 25 33" src="https://user-images.githubusercontent.com/5846501/129490054-db999ca6-2c50-42ee-b718-60d349bd79de.png">

There are a lot of things to be unhappy with in my hacky version:

1. asciinema is a python dependency, so this might be kind of annoying if it was installed by default
    * perhaps it can be opt-in somehow?
2. I spin up a new Julia process to run the code for the gif
    * so with this method it would be very hard to support named blocks (`@repl 1` etc).
    * and every time making a gif hits first-call compile time latency (which usually I amortize by calling `include("docs/make.jl")` repeatedly when developing docs). I therefore setup caching system to prevent re-making the same gif over and over, but I'm not sure it's robust enough for Documenter.jl's large userbase
3. I didn't mess with any Documenter internals to actually make a new block; I just use `@example` and run a string macro.
4. I ran into #1433 and worked around it in a very fragile way
5. right now, I create the gif at macro-expansion time; probably it should be at runtime? They are not very different times with example blocks though and doing stuff at top-level scope
6. asciinema is licensed GPL v3, so maybe there would be license issues?

However, I still think the result is super cool! Any ideas on how to get some (better?) version of this into Documenter.jl?

ericphanson:
I realized asciinema is based on a simple JSON-based specification: https://github.com/asciinema/asciinema/blob/asciicast-v2/doc/asciicast-v2.md. So I quickly whipped up a Julia implementation, which lets us avoid issues 1 and 6. The inline HTML player is apache licensed so no issues there. Then I went ahead and looked at the Documenter internals and made an `@gif` block based on the code for `@repl`. This resolved issues 2, 3, and 5. The result is https://github.com/ericphanson/Asciicast.jl.

So the only remaining issue AFAIK is getting the JS loaded in the correct order, and not relying too much on Documenter.jl internals. And testing it.

fredrikekre:
[...]

PR #2376, opened 2023-12-12
add Asciicast to regression tests

ericphanson:
Based on https://github.com/SciML/SciMLBase.jl/blob/master/.github/workflows/Downstream.yml

Briefly discussed a few years ago here: https://github.com/JuliaDocs/Documenter.jl/issues/1675#issuecomment-899858954

I've only added my new [Asciicast.jl](https://github.com/ericphanson/Asciicast.jl) so far, but I imagine we could add all the documenter plugins, if there is interest


odow:
See https://github.com/JuliaDocs/Documenter.jl/blob/master/.github/workflows/regression-tests.yml

.github/workflows/Downstream.yml
```diff
@@ -0,0 +1,51 @@
...
+          try
+            # force it to use this PR's version of the package
+            Pkg.develop(PackageSpec(path="."))  # resolver may fail with main deps
+            Pkg.update()
+            Pkg.test(coverage=true)  # resolver may fail with test time deps
+          catch err
+            err isa Pkg.Resolve.ResolverError || rethrow()
+            # If we can't resolve that means this is incompatible by SemVer and this is fine
+            # It means we marked this as a breaking change, so we don't need to worry about
+            # Mistakenly introducing a breaking change, as we have intentionally made one
```
odow, line 44: I feel like this just encourages people to make breaking changes :cry:

ericphanson:
Oops, missed that! Ok, I updated this PR to just add Asciicast there

.github/workflows/Downstream.yml
ericphanson: eh, I think you need something like this if you're going to put the workflow on 20+ repos, like SciML does (where I copied it from). Because it also could be some unrelated dependency has made a breaking change, and the package being tested has updated for it, but one of the regressionees has not, and then you can't resolve, so you can get failing CI unrelated to either package making a breaking release.

Agreed it's probably not necessary here though

ericphanson:
do we only want workflow-dispatch for regression-tests.yml? That seems OK if we remember to run it, but otherwise we could add it to just run every time

odow:
I think we should at least attempt to be mindful of CI resources. I'm not overly fond of running the tests of a bunch of unrelated packages every time we fix a typo in Documenter.

And if you're touching internals, Documenter is under no obligation to ensure you don't break. In general, I use the `workflow-dispatch` in JuMP and MOI quite liberally for anything that seems like it might cause issues. Here's the most recent: https://github.com/jump-dev/MathOptInterface.jl/pull/2359#issuecomment-1855098514.

We also talked about adding a TODO item to the release checklist:
https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/checklists.md#making-a-release

mortenpi:
I'd say let's stick to running this manually. Those tests are informational and not strictly part of the Documenter test suite.

> We also talked about adding a TODO item to the release checklist

I thought we already did, apparently not. We definitely should.

ericphanson:
bump

odow (approved):

mortenpi:
Thank you!

Issue #2303, opened 2023-10-16
`doctest(Module; fix=true)` produces weird results


lkastner:
I made a minimal working example. My doctest looks similar to:
````
``` jldoctest label
julia> a = ("a", "b", "c");                       

julia> a

```
````
(I left the output empty on purpose to trigger the error.)
Running Documenter.jl to fix the doctest gives:
````
```jldoctest label
julia
("a", "b", "c")> a = ("a", "b", "c");

julia> a

```
````
It seems like some regex is replacing the first appearance of `a` with the command + correct output, instead of the actual command + (wrong) output.

PR #2378, opened 2023-12-14
Add a prefix when replacing for doctest=:fix

lkastner:
Attempt fixing #2303 
The logic is as follows: The repl session is split into two parts, the first part goes from the start until the current command, and the second part is the rest. The old output is removed from the second part and replaced by the correct output. However if the command e.g. was just `a` for displaying the variable `a`, then the split could incorrectly happen at the `a` in `julia>`. The new logic introduces a prefix containing the entire repl until the current command, so the split happens after (prefix + command) is found.


mortenpi:
Sorry for the delay here. I can confirm that the bug is there, but it doesn't look like the PR fixes it right now. I'm getting this after fixing:

````
```jldoctest
julia> a
("a", "b", "c") = ("a", "b", "c");

julia> a

```
````

Which is slightly different than the original bug, but still wrong.

cc @fredrikekre: if you have any bandwith to help with this?

**Comment by lkastner** [comment]
> Sorry for the delay here. I can confirm that the bug is there, but it doesn't look like the PR fixes it right now. I'm getting this after fixing:
> 
> ```
> ```jldoctest
> julia> a
> ("a", "b", "c") = ("a", "b", "c");
> 
> julia> a
> 
> ```
> ```
This should work now. The testcases I added never had the output suppressed with `;` in the first command which kept the prefix from initializing. And unfortunately I made up the example for the PR test without running it. Murphy. I am sorry about that.

**Comment by lkastner** [comment]
There are some other cleaner approaches but these would require a larger refactoring, so I just went for a minimal approach.

**Comment by lkastner** [comment]
@benlorenz pointed out to me, that the tests are failing now because Julia has changed the output format for errors. Here is the difference:
First 1.6:
```
julia> a
ERROR: UndefVarError: a not defined
```
Now current Julia:
```
julia> a
ERROR: UndefVarError: `a` not defined
```
i.e. the backticks were added. I added a commit to account for this in the tests by having a separate file for 1.6.

**Review by fingolfin — APPROVED** [review]
This looks fine to me (I made two trivial and optional change suggestions to match the style of the surrounding code).

It would be a pity if this fix for a real-world problem would be wasted :-(

**Review by mortenpi — APPROVED** [review]
Thank you @lkastner! Apologies that it took a while to get this reviewd -- this is a great PR! We'll have it in 1.4.0.

I took the liberty of adding a CHANGELOG and also editing the code a bit, to simplify it a bit and make it a bit more Julia (in my opinion).

**Comment by mortenpi** [comment]
It looks like there's still a regression, or some issue, here. This doctest

````
```jldoctest
julia> a
ERROR: UndefVarError: `a` not defined

julia> b
```
````

does not fix itself on nightly (and causes the test failures). But it (more or less) works with the latest Documenter.  I say more or less, since it will get replaced with

````
```jldoctest
julia> a
ERROR: UndefVarError: `a` not defined

julia> b
ERROR: UndefVarError: `b` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
```
````

so the first doctest is not being fixed for some reason.

**Comment by fingolfin** [comment]
Could someone with the relevant permissions (e.g. @mortenpi) please allow the CI tests to run?

**Comment by lkastner** [comment]
I managed to fix the tests on nightly. The main reason for failure was that the error messages in nightly have an 
```
in `Main`
```
added. I modified the error messages in the tests and have the `test.jl` remove these parts for older versions of Julia.

mortenpi (approved):
I'm happy to merge this. Thank you very much for sticking with this, and sorry it took so long to get this done and merged!

PR #2381, opened 2023-12-18
Add regression-tests to release checklist

odow:
x-ref https://github.com/JuliaDocs/Documenter.jl/pull/2376#issuecomment-1855177533


mortenpi:
LGTM, thanks!

Issue #2384, opened 2024-01-01
docs build failing on obscure `@assert`


JeffFessler:
I have a Documenter build that is failing on this `@assert` in `parseblock`:
https://github.com/JuliaDocs/Documenter.jl/blob/d9865b4b73e4b0e37d242b65f77a87c1b7ffff4d/src/utilities/utilities.jl#L167

It seems quite hard to diagnose the cause of this error because `@assert` gives no context (other than the Stacktrace that simply says it originates in the `makedocs` call in `make.jl`).

Would you be open to a PR where we change the `@assert` to use a `throw` where info like the `file` could be shown as part of the error message?  That would help narrow down the search a lot.  Or other suggestions?

In case it helps, here is the build log:
https://github.com/JeffFessler/book-la-demo/actions/runs/7378584668/job/20073887407?pr=48

Thanks

goerz:
The `@assert` isn't much different from a `throw` in that `@assert` can also emit a custom message with extra information. Try dev-installing `Documenter` and changing the relevant line 167 in `src/utilities/utilities.jl` to something like

```
@assert (length(exs) == 2 * length(results)) "Invalid code at $linenumbernode:\n$code"
```

These `@assert`'s are *internal* errors, though, so we should figure out what is happening here and then determine what the appropriate place for catching this with a helpful user-facing error message would be.

Documenter is running into some kind of parsing error while evaluating an `@example` block generated by Literate. If you can take a look at the `.md` files that `Literate` generates based on your demos, maybe you'll see something that sticks out.

schillic:
This is probably a duplicate of #2206. In my case there was a space missing in a comment line, as in:
```julia
code...
#incorrect comment with missing leading space
# correct comment
code...
```

JeffFessler:
Sorry for the duplicate, and thanks for the very helpful suggestions!  I will close this one then.

JeffFessler:
> Try dev-installing `Documenter`

I took this suggestion and found that the culprit, which is a good reason to modify the assert so that it reports more information!

The issue is a paragraph consisting of two Julia comments like this:
```julia
## rank(Xr)
## svdvals(Xr)
```
with blank lines before and after.

These get translated into markdown like this:

```markdown
````@example lrmc-m
# rank(Xr)
# svdvals(Xr)
```

(There are four more backticks right after but I can't get github to show those.)
[...]

Issue #2206, opened 2023-08-11
AssertionError: length(exs) == 2 * length(results)


kbarros:
I get the above assertion error with the following input, `fei2_classical.md`:

`````
```@meta
EditURL = "../../../examples/fei2_classical.jl"
```

````@example fei2_classical
# Calculating S
````

---

*This page was generated using [Literate.jl](https://github.com/fredrikekre/Literate.jl).*
`````

This input was generated by running Literate on the following `fei2_classical.jl` file:

```
## Calculating S
```

I would expect to get an error message rather than an assertion failure.

Full stack trace:

```
ERROR: AssertionError: length(exs) == 2 * length(results)
Stacktrace:
  [1] parseblock(code::String, doc::Documenter.Document, file::String; skip::Int64, keywords::Bool, raise::Bool, linenumbernode::LineNumberNode)
    @ Documenter ~/.julia/dev/Documenter/src/utilities/utilities.jl:172
  [2] parseblock
    @ Documenter ~/.julia/dev/Documenter/src/utilities/utilities.jl:135 [inlined]
  [3] #parseblock#18
    @ Documenter ~/.julia/dev/Documenter/src/documents.jl:62 [inlined]
  [4] runner(::Type{Documenter.Expanders.ExampleBlocks}, node::MarkdownAST.Node{Nothing}, page::Documenter.Page, doc::Documenter.Document)
    @ Documenter.Expanders ~/.julia/dev/Documenter/src/Expanders.jl:713
  [5] dispatch(::Type{Documenter.Expanders.ExpanderPipeline}, ::MarkdownAST.Node{Nothing}, ::Vararg{Any})
    @ Documenter.Selectors ~/.julia/dev/Documenter/src/utilities/Selectors.jl:170
  [6] expand(doc::Documenter.Document)
    @ Documenter.Expanders ~/.julia/dev/Documenter/src/Expanders.jl:44
  [7] runner(::Type{Documenter.Builder.ExpandTemplates}, doc::Documenter.Document)
    @ Documenter.Builder ~/.julia/dev/Documenter/src/Builder.jl:232
  [8] dispatch(::Type{Documenter.Builder.DocumentPipeline}, x::Documenter.Document)
    @ Documenter.Selectors ~/.julia/dev/Documenter/src/utilities/Selectors.jl:170
  [9] #41
    @ Documenter ~/.julia/dev/Documenter/src/makedocs.jl:237 [inlined]
[...]

PR #2388, opened 2024-01-06
Add tiny help message to @assert error

JeffFessler:
To help anyone hitting issues #2206 / #2384.

I know it is tiny change, but to quote Emily Dickinson:
```poem
If I can stop one heart from breaking,
I shall not live in vain;
```
https://dailypoetry.me/emily-dickinson/if-stop-heart-breaking

### DIFF
--- a/src/utilities/utilities.jl
+++ b/src/utilities/utilities.jl
@@ -164,7 +164,7 @@ function parseblock(code::AbstractString, doc, file; skip = 0, keywords = true, 
     end
     if linenumbernode isa LineNumberNode
         exs = Meta.parseall(code; filename=linenumbernode.file).args
-        @assert length(exs) == 2 * length(results)
+        @assert length(exs) == 2 * length(results) "Issue at $linenumbernode:\n$code"
         for (i, ex) in enumerate(Iterators.partition(exs, 2))
             @assert ex[1] isa LineNumberNode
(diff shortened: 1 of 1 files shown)


goerz:
Looks good to me! I'm assuming you've "tested" this by actually triggering this assert while debugging #2384. Can you paste in the text or a screenshot of how this ends up looking in practice? Does the `$linenumbernode` result in something readable?

JeffFessler:
Yes, I tested it and that is how I found the "issue" in my docs examples.
BTW those examples worked fine in Julia 1.9 but something happened with Julia 1.10.
I had already fixed up my code
https://github.com/JeffFessler/book-la-demo/pull/48/commits/0fb45d4830fc6ea02f764825a5161e35f099f56d
so I had to "recreate" the problem situation and found that the following specific scenario reliably causes the `@assert` to file.  Something like this:
```julia
# Text for markdown

## A Julia comment

# More markdown text
```

The middle "paragraph" gets translated into an isolated `@example` in the generate .md file:

```
````@example myfilename
# A Julia comment
````

Obviously that "example" generates no output, so then the `@assert` fails with a message like this:

```
ERROR: LoadError: AssertionError: Issue at #= myfilename.md:116 =#:
# A Julia comment
```
which was plenty enough bread crumbs for me to find a work around.

PR #2389, opened 2024-01-06
Fix `version` hiding the version selection menu

### LINKED ISSUE #2385: The `version` argument for `makedocs`
The `makedocs` function currently takes an undocumented `version` argument (which the HTML writer can access as `ctx.doc.user.version`).

I was hoping to start encouraging the use of that argument to pass the version of the package being documented. The [inventory files](https://juliadocs.org/DocumenterInterLinks.jl/dev/write_inventory/) I'm currently prototyping preferably should [write this version information into the inventory](https://github.com/JuliaDocs/DocumenterInterLinks.jl/blob/7183ef3b0cdf83a9e2c1cc4166d03757642353a6/src/write_inventory.jl#L33).

Alas, it turns out that setting the `version` parameter [actually disables the version selection menu entirely](https://github.com/JuliaDocs/Documenter.jl/blob/d9865b4b73e4b0e37d242b65f77a87c1b7ffff4d/src/html/HTMLWriter.jl#L1133) (it gets set to "visible", which [the javascript then toggles to "invisible"](https://github.com/JuliaDocs/Documenter.jl/blob/d9865b4b73e4b0e37d242b65f77a87c1b7ffff4d/assets/html/js/versions.js#L63)).

Is this intentional? The way this happens is pretty round-about, so it may not be intentional. What was the idea behind the `version` argument when it was originally implemented? Is this being used anywhere in the wild?

Since `version` is completely undocumented, is there any chance that we can remove

https://github.com/JuliaDocs/Documenter.jl/blob/d9865b4b73e4b0e37d242b65f77a87c1b7ffff4d/src/html/HTMLWriter.jl#L1133-L1137

without this being a "breaking change"?

Would it be possible to officially make the `version` argument to `makedocs` be for "you can (should) pass the version of the package being documented" and document it as such?

goerz:
This fixes the issue with the version menu disappearing when someone specifies `version` in `makedocs`.

This PR is pretty minimal and hopefully uncontroversial. It would be great if we could merge/release this, as the underlying issue breaks https://github.com/JuliaDocs/DocumenterInterLinks.jl/blob/master/src/write_inventory.jl

I've adapted the documentation to say that `version` is not currently used in the `HTMLWriter`. If I add the inventory writing to Documenter later, that would become the first instance of where it *is* used (it gets written into the inventory).

Maybe we can also use `version` in other ways. The `LaTeXWriter` `version` parameter should probably fall back to the `version` in `makedocs` (without removing `LaTeXWriter`'s `version`, as that would be a breaking change).

I'd also still be in favor of showing a non-empty `version` in the sidebar and/or the footer, but those things would go into a separate PR, as they might be more controversial.

Closes #2385

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+* Fixed the `version` parameter to `makedocs` inadvertently hiding the version selection menu ([#2385])
+
+
 ## Version [v1.2.1] - 2023-12-02
 
 ### Fixed
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Documenter"
 uuid = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-version = "1.2.1"
+version = "1.2.2-dev"
 
 [deps]
 ANSIColoredPrinters = "a4c015fc-c6ff-483c-b24f-f7ea428134e9"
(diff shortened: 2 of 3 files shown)


mortenpi:
I think this might be outdated, based on the discussions in #2389?

goerz:
Yes, I’ll make another PR to address this

PR #2394, opened 2024-01-08
Add support for `CollapsedDocStrings` in `@meta` block

### LINKED ISSUE #2282: Add keyword argument for folding all docstrings by default
I think this would be a very cool feature, it was mentioned in #2078 but then left aside

goerz:
Setting `DocStringsCollapsed = true` in the `@meta` block of a particular page essentially clicks the "Collapse all docstrings" in the navigation bar after the page loads, collapsing all docstrings on that page.

I think the approach of just having JavaScript click the collapse button is the right one here: First, I'm not sure that it would be possible to write out the page in a collapsed state, or at least not without doing some pretty gnarly rewrites of the `HTMLWriter`. More importantly, rendering the page expanded and then collapsing it with JavaScript seems like the right thing to do from an accessibility standpoint: We'd want the page to still be useful without JavaScript. I've been known to access API documentation in the terminal with `w3m`, so not relying on JavaScript to render the core content is something I actually care about. Screenreaders probably also wouldn't like if all docstrings initially rendered as `hidden`.

Closes #2282

### DIFF
--- a/assets/html/js/metadata.js
+++ b/assets/html/js/metadata.js
@@ -1,11 +1,13 @@
 // libraries: jquery
 // arguments: $
 
-let meta = $("div[data-docstringscollapsed]").data();
+$(document).ready(function () {
+  let meta = $("div[data-docstringscollapsed]").data();
 
-if (meta.docstringscollapsed) {
-  $("#documenter-article-toggle-button").trigger({
-    type: "click",
-    noToggleAnimation: true,
-  });
-}
+  if (meta.docstringscollapsed) {
+    $("#documenter-article-toggle-button").trigger({
+      type: "click",
+      noToggleAnimation: true,
+    });
+  }
+});
--- a/assets/html/js/metadata.js
+++ b/assets/html/js/metadata.js
@@ -4,5 +4,8 @@
 let meta = $("div[data-docstringscollapsed]").data();
 
 if (meta.docstringscollapsed) {
-  $("#documenter-article-toggle-button").click();
+  $("#documenter-article-toggle-button").trigger({
+    type: "click",
+    noToggleAnimation: true,
+  });
 }
--- a/assets/html/js/metadata.js
+++ b/assets/html/js/metadata.js
@@ -1,0 +1,8 @@
+// libraries: jquery
+// arguments: $
+
+let meta = $("div[data-docstringscollapsed]").data();
+
+if (meta.docstringscollapsed) {
+  $("#documenter-article-toggle-button").click();
+}
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -284,7 +284,7 @@ page. Currently recognised keys:
 - `Description`: a page-specific description that gets displayed in search engines and
   link previews. Overrides the site-wide description in [`makedocs`](@ref). 
 - `Draft`: boolean for overriding the global draft mode for the page.
-- `DocStringsCollapsed`: if `true`, collapse all docstrings after loading the page (experimental).
+- `DocStringsCollapsed`: for output formats that support this (i.e. only [`HTML`](@ref Documenter.HTML) currently), if set to `true`, render all docstrings as collapsed by default.
 
 Example:
 
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -284,6 +284,7 @@ page. Currently recognised keys:
 - `Description`: a page-specific description that gets displayed in search engines and
   link previews. Overrides the site-wide description in [`makedocs`](@ref). 
 - `Draft`: boolean for overriding the global draft mode for the page.
+- `DocStringsCollapsed`: if `true`, collapse all docstrings after loading the page (experimental).
 
 Example:
 
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -282,9 +282,9 @@ page. Currently recognised keys:
   but if the source is something else (for example if the `.md` page is generated as part of
   the doc build) this can be set, either as a local link, or an absolute url.
 - `Description`: a page-specific description that gets displayed in search engines and
-  link previews. Overrides the site-wide description in [`makedocs`](@ref). 
+  link previews. Overrides the site-wide description in [`makedocs`](@ref).
 - `Draft`: boolean for overriding the global draft mode for the page.
-- `DocStringsCollapsed`: for output formats that support this (i.e. only [`HTML`](@ref Documenter.HTML) currently), if set to `true`, render all docstrings as collapsed by default.
+- `CollapsedDocStrings`: for output formats that support this (i.e. only [`HTML`](@ref Documenter.HTML) currently), if set to `true`, render all docstrings as collapsed by default.
 
 Example:
 
@@ -365,7 +365,7 @@ As with `@index` if `Pages` is not provided then all pages are included. The def
             "Subsection" => SUBSECTION_PAGES,
         ...
     ```
-    
+
     That variable will exist in the `Main` module and can be reused in the `@contents` and other blocks, e.g.
 
     ````markdown
@@ -373,7 +373,7 @@ As with `@index` if `Pages` is not provided then all pages are included. The def
     Pages = Main.SUBSECTION_PAGES
     ```
     ````
-    
+
     Documenter will then list the contents of the "Subsection" pages, and they will always appear in the same order as they are in the sidebar.
 
 ## `@example` block
--- a/docs/src/showcase.md
+++ b/docs/src/showcase.md
@@ -1,3 +1,7 @@
+```@meta
+DocStringsCollapsed = true
+```
+
 # Showcase
 
 This page showcases the various page elements that are supported by Documenter.
@@ -211,23 +215,23 @@ Lists can also be included in other blocks that can contain block level items
 
 ## Tables
 
-| object | implemented |      value |
-|--------|-------------|------------|
-| `A`    |      ✓      |      10.00 |
-| `BB`   |      ✓      | 1000000.00 |
+| object | implemented | value      |
+| ------ | ----------- | ---------- |
+| `A`    | ✓           | 10.00      |
+| `BB`   | ✓           | 1000000.00 |
 
 With explicit alignment.
 
 | object | implemented |      value |
-| :---   |    :---:    |       ---: |
+| :----- | :---------: | ---------: |
 | `A`    |      ✓      |      10.00 |
 | `BB`   |      ✓      | 1000000.00 |
 
 Tables that are too wide should become scrollable.
 
-| object | implemented |      value |
-| :---   |    :---:    |       ---: |
-| `A`    |      ✓      |      10.00 |
+| object                 |                 implemented                  |                                                      value |
+| :--------------------- | :------------------------------------------: | ---------------------------------------------------------: |
+| `A`                    |                      ✓                       |                                                      10.00 |
 | `BBBBBBBBBBBBBBBBBBBB` | ✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓ | 1000000000000000000000000000000000000000000000000000000.00 |
 
 
--- a/docs/src/showcase.md
+++ b/docs/src/showcase.md
@@ -1,7 +1,3 @@
-```@meta
-DocStringsCollapsed = true
-```
-
 # Showcase
 
 This page showcases the various page elements that are supported by Documenter.
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -837,7 +837,7 @@ function render_page(ctx, navnode)
     article = render_article(ctx, navnode)
     footer = render_footer(ctx, navnode)
     meta_divs = DOM.Node[]
-    if get(getpage(ctx, navnode).globals.meta, :DocStringsCollapsed, false)
+    if get(getpage(ctx, navnode).globals.meta, :CollapsedDocStrings, false)
         # if DocStringsCollapse = true in `@meta`, we let JavaScript click the
         # collapse button after that page has loaded.
         @tags script
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -836,7 +836,17 @@ function render_page(ctx, navnode)
     navbar = render_navbar(ctx, navnode, true)
     article = render_article(ctx, navnode)
     footer = render_footer(ctx, navnode)
-    htmldoc = render_html(ctx, navnode, head, sidebar, navbar, article, footer)
+    meta_divs = DOM.Node[]
+    if get(getpage(ctx, navnode).globals.meta, :DocStringsCollapsed, false)
+        # if DocStringsCollapse = true in `@meta`, we let JavaScript click the
+        # collapse button after that page has loaded.
+        @tags script
+        push!(
+            meta_divs,
+            div[Symbol("data-docstringscollapsed") => "true"]()
+        )
+    end
+    htmldoc = render_html(ctx, navnode, head, sidebar, navbar, article, footer, meta_divs)
     write_html(ctx, navnode, htmldoc)
 end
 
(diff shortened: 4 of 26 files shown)


goerz:
With respect to #2282, this PR does not implement a global flag for collapsing all docstrings. It's not something I'm interested in, although it wouldn't be terribly hard to add it in (but definitely more work to document and test)

src/html/HTMLWriter.jl
```diff
@@ -930,6 +930,17 @@ function render_head(ctx, navnode)
...
+    collapse_script = ""
+    if get(getpage(ctx, navnode).globals.meta, :DocStringsCollapsed, false)
+        # if DocStringsCollapse = true in `@meta`, we let JavaScript click the
+        # collapse button after that page has loaded.
+        collapse_script = script("""
+        window.onload = function() {
+         var toggleButton = document.getElementById('documenter-article-toggle-button');
+         if (toggleButton) {toggleButton.click();}
+        };
+        """)
```
mortenpi, line 942: I feel that ideally we'd do this by statically generating the necessary HTML. It looks like we use `display: none` on the `section` tags.. and then update some other stuff. Ideally, we'd have a class on the `article` tag that toggles everything.

How much effort would you be willing to put into this? I didn't test this yet, but I assume that this works, and we can merge this as is, and refactor later as well.

mortenpi: > More importantly, rendering the page expanded and then collapsing it with JavaScript seems like the right thing to do from an accessibility standpoint: We'd want the page to still be useful without JavaScript.

Ah.. that is a fair point actually. We should indeed make sure that it loads uncollapsed if you don't have JS enabled.

goerz: Right. I'd still advocate for going with the approach in this PR. I started out with trying to write out the collapsed HTML if the option is set until I realized that:

(a) It's *much* more difficult to implement that way. In fact, I'm not sure that I'd be able to do it in any reasonable amount of time.

(b) "Collapsed" means `display: none`, which *removes* the content entirely without JS, which seems like a bad idea.

The test in the PR only checks for the presence of the injected JS. I don't think there's much I can do beyond that. I did test locally that the JS does what it's supposed to (very nicely, in fact). What I could do as an additional test is to *use* the feature in the Documenter documentation. Maybe collapsing all the [Internal API](https://documenter.juliadocs.org/stable/lib/internals/anchors/) pages by default?

goerz:
I added collapse-by-default to the Internal API pages, e.g., https://documenter.juliadocs.org/previews/PR2394/lib/internals/anchors/

Interestingly (and unlike in my previous local testing), I'm finding that it doesn't seem completely reliable. Maybe I'll have to revise the JS a bit. It might be running too early or something.

goerz:
Sorry, this is going to need more work. I'll let you know if I can figure out how to make this work reliably.

src/html/HTMLWriter.jl
mortenpi: Yeah, let's go with the JS approach.

It currently does have the problem that you can see all the docstrings collapsing when you navigate to the page though. Maybe @Hetarth02 has some ideas how to work around that.

mortenpi:
One other thing: I'm a little bit skeptical of putting too many writer-specific options into the top-level scope of an at-meta block. Should we have some way to namespace writer-specific options like this? Not sure what the best syntax would be, but maybe something like:

```julia
Writer.HTML = begin
    CollapsedDocStrings = true
end
```

src/html/HTMLWriter.jl
Hetarth02: Interesting, I will definitely take a look into this.

goerz: @Hetarth02 I'd appreciate that! You might be better at JS than me. I'm having real trouble getting it to load correctly. In my first iteration, it worked fine locally but only 90% of the time in Chrome and only 1% of the time in Safari. I'm guessing the script was running too early. In the current iteration, I've tried to move the `<script>` tag to the end of the file, but that's still behaving the same way (i.e., not working in Safari). I also tried putting it in an external script (I think local scripts load before external scripts, which may be part of the problem), but that was much more complicated to implement, and then just didn't work at all. I can push that branch again somewhere if you'd like to take a look at it.

goerz: > It currently does have the problem that you can see all the docstrings collapsing when you navigate to the page though

Yeah, I can't decide whether it's a problem or whether I actually kinda like it. But in any case, I think jQuery is handling the collapse animation. Maybe there's a way to temporarily disable the animation or at least speed it up?

Hetarth02: Yes, Jquery handles the animation and there is a way to speed it up.

I also think that, initially appearing as open then getting closed is okay.

goerz:
> I'm a little bit skeptical of putting too many writer-specific options into the top-level scope of an at-meta block.

I have no strong feelings on this either way. Although I'd probably keep it flat, mostly because `EditURL`, `Description`, and maybe `Draft` would also seem to belong with `Writer.HTML`, but moving them is probably breaking compatibility.

Assuming I can make get the `CollapsedDocStrings` to work reliably at all, maybe we can mark it as "experimental" and then move it to a namespaced scheme afterward, in a separate PR.

src/html/HTMLWriter.jl
goerz: > I can push that branch again

Pushed to https://github.com/JuliaDocs/Documenter.jl/tree/mg/meta-docstrings-collapsed2. This one doesn't work at all, though, and I'm not sure why.

Hetarth02: @goerz Which branch should I check from?

`mg/meta-docstrings-collapsed2` or `mg/meta-docstrings-collapsed`

goerz: Preferably, `mg/meta-docstrings-collapsed`. The other one was an attempt to put the JS in a separate script file. It was both more complicated and didn't actually work, but I just wanted to keep the branch around in case `mg/meta-docstrings-collapsed` doesn't work out and I need to get back to it.

goerz: `mg/meta-docstrings-collapsed` works perfectly when looking at documentation locally with `LiveServer`, but doesn't work reliably when looking at the online documentation, e.g. at https://documenter.juliadocs.org/previews/PR2394/lib/internals/anchors/. In Chrome, it works most of the time, in Safari, it does not work most of the time. Presumably, the issue is the asynchronous execution of JS and the code the clicks the collapse button firing too early.

Hetarth02: @goerz So, I fiddled around a bit. I have found a way which I think will work.

Basically, what we will do is save all the `@meta` flags in `data` attribute of a `div` element. Something like below,

```html
<div data-docstringscollapsed="true"></div>
```

Then add a `js` script, which will perform the necessary actions for each flags. For example, like this

```js
// assets/html/js/metadata.js

// libraries: jquery
// arguments: $

let docstringscollapsed = $("div[data-docstringscollapsed]").data();

if (docstringscollapsed) {
  $("#documenter-article-toggle-button").click();
}
```

Doing this will always ensure that, all the scripts and DOM content will be loaded.

goerz: That sounds good! Do you want to just push to the `mg/meta-docstrings-collapsed` branch? I'm assuming you have permissions.

Hetarth02: I don't think I have the necessary permissions to do so.

goerz: I can merge from https://github.com/Hetarth02/Documenter.jl/ into my PR then. Let me know which branch.

Hetarth02: Alright, will do so. Let me iron some things out and will ping you when I am done.

goerz: No worries! Thanks for taking a look! I'll check it out.

Hetarth02: @goerz Just FYI, this is just a rudimentary implementation from my side. There is of course room for improvement in logic.😅

goerz: It works perfectly! Thanks for coming up with the approach that makes this reliable!

assets/html/js/metadata.js
```diff
@@ -0,0 +1,8 @@
+// libraries: jquery
+// arguments: $
+
+let meta = $("div[data-docstringscollapsed]").data();
+
+if (meta.docstringscollapsed) {
+  $("#documenter-article-toggle-button").click();
```
goerz, line 7: This seems fine to me. It still shows the close-animation after the page is loaded. I don't mind it that much, but @mortenpi might still prefer if we speed it up.

@Hetarth02 Do you know what to add to control the animation speed?

docs/src/showcase.md
```diff
@@ -211,23 +215,23 @@ Lists can also be included in other blocks that can contain block level items
 
 ## Tables
 
-| object | implemented |      value |
-|--------|-------------|------------|
-| `A`    |      ✓      |      10.00 |
-| `BB`   |      ✓      | 1000000.00 |
+| object | implemented | value      |
+| ------ | ----------- | ---------- |
```
goerz, line 219: Seems a little unrelated to the PR, but I just kept it in

goerz:
Okay, this is back on track now, and could be merged in principle.

We can still tweak this, of course, maybe changing the animation speed, but that's optional as far as I'm concerned.

As before, https://documenter.juliadocs.org/previews/PR2394/lib/internals/anchors/ (and the other "internals" pages) demo the feature.

docs/src/man/syntax.md
```diff
@@ -284,6 +284,7 @@ page. Currently recognised keys:
 - `Description`: a page-specific description that gets displayed in search engines and
   link previews. Overrides the site-wide description in [`makedocs`](@ref). 
 - `Draft`: boolean for overriding the global draft mode for the page.
+- `DocStringsCollapsed`: if `true`, collapse all docstrings after loading the page (experimental).
```
goerz, line 287: Marked as "experimental" so we can still change this even after the PR is merged/released. For example, https://github.com/JuliaDocs/Documenter.jl/pull/2394#issuecomment-1892874802

mortenpi:
I think I'm okay with the collapse being visible. It's a bit jarring, but we can try to improve on it in the future. It's opt-in for the author anyway.

However, there seems to be some bug when you navigate to the page without a hard-refresh? This was from the preview deployment:


https://github.com/JuliaDocs/Documenter.jl/assets/147757/382d5884-ebbd-496d-994c-de98bd7c44f8

goerz:
It's a bit hard to tell what's going on in the video (even if I step through it frame by frame). The way it should work is that if you reload a page, or navigate to a page, the docstrings always get collapsed. What I think I'm seeing in your video looks a little like how things were at the previous (broken) attempt. Did you clear your browser cache? I was definitely seeing wonky behavior until I did that. But now, everything is working to my expectations without any fail, both on Safari and Chrome.

I'll try to record a little video of how things are looking on my end tomorrow or so. Or else, it would have to be a platform/browser issue. Is that Chrome on Linux?

As for the animation, that looks about correct in the video. You still see the full content flash very briefly before it collapses, even with an animation time of zero. It don't know if it's less jarring if it's a slower animation. That might be something we can decide with a screenshare at the next community call.

Hetarth02:
> I think I'm seeing in your video looks a little like how things were at the previous (broken) attempt.

I also think it might be running old code too.

Although, I think the animation should be slow instead of just plain 0. That's just my 2 cents though!😅

mortenpi:
> Did you clear your browser cache?

Good point, I didn't. However, even after clearing the cache, I see the issue. I also tried in incognito. The only way I can get it to work is to disable the cache altogether.. so it does seem like it might cache-related.. but maybe some sort of a race condition?

Just something to double check: is the JS in https://documenter.juliadocs.org/previews/PR2394/ up to date?

And yes, it's on Chrome & Linux. But I also see the issue in Firefox, where I hadn't opened the page before.

> It don't know if it's less jarring if it's a slower animation.

Slower would be more jarring (I think). But as I said, I think it's good enough. What we'd ideally have is no animation.

mortenpi:
> What we'd ideally have is no animation.

For what it's worth, the [themeswap](https://github.com/JuliaDocs/Documenter.jl/blob/master/assets/html/themeswap.js) actually has the same problem, and it solves it by making sure it runs early. But it's not immediately clear how complex it would be to do something similar here.

goerz:
This is how it looks on my end, very reliably, both in Safari in Chrome, as well as on an iPad

https://github.com/JuliaDocs/Documenter.jl/assets/112306/84db6b7c-a474-4760-b746-6f220211709e

This is using the latest JS in https://documenter.juliadocs.org/previews/PR2394/ which hasn't changed since https://github.com/JuliaDocs/Documenter.jl/pull/2394#issuecomment-1909296725

> so it does seem like it might cache-related.. but maybe some sort of a race condition?

Very strange. I wonder if there's a ISP/router cache or CDN or something getting in the way. But since the problem disappears when you disable the browser cache, that doesn't seem to be it.

I guess we might still have some kind of JS concurrency bug, although @Hetarth02's solution really seems to me like it should work reliably, and click the collapse button only after everything else has loaded. I'm not a JS expert, though.

mortenpi:
I noticed I'm getting these in the console when it's not working:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'docstringscollapsed')
    at documenter.js:238:10
    at Object.execCb (require.min.js:1:16727)
    at e.check (require.min.js:1:10499)
    at e.<anonymous> (require.min.js:1:12915)
    at require.min.js:1:1542
    at require.min.js:1:13376
    at each (require.min.js:1:1020)
    at e.emit (require.min.js:1:13344)
    at e.check (require.min.js:1:11058)
    at e.enable (require.min.js:1:13242)
documenter.js:704 Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at documenter.js:704:45
    at Object.execCb (require.min.js:1:16727)
    at e.check (require.min.js:1:10499)
    at e.enable (require.min.js:1:13242)
    at e.init (require.min.js:1:9605)
    at require.min.js:1:15109
```

goerz:
Ah, that at least gives us something to work with!

assets/html/js/metadata.js
```diff
@@ -0,0 +1,11 @@
...
+// arguments: $
+
+let meta = $("div[data-docstringscollapsed]").data();
+
+if (meta.docstringscollapsed) {
+  $("#documenter-article-toggle-button").trigger({
+    type: "click",
+    noToggleAnimation: true,
+  });
+}
```
mortenpi, line 11: This whole thing probably needs to be in a `$(document).ready(function () { ... }`?

goerz: Let's try it! https://documenter.juliadocs.org/previews/PR2394/ should now have that change. Of course, all I can say that it continues to work for me, so you'll have to let us know :-)

mortenpi:
With the fixes, it looks like it works now! @Hetarth02, are you still seeing the mobile search/sidebar issue you mentioned on Slack?

Hetarth02:
@mortenpi I just checked, everything seems to be working fine.

docs/src/man/syntax.md
mortenpi: Eh, yea, let's just keep it on the top-level. Doing namespacing adds other kinds of complexity. Some of these options will just be ignored by some backends.

```suggestion
- `DocStringsCollapsed`: on output formats that support this (i.e. HTML), if set to `true`, render all docstrings as collapsed by default.
```

mortenpi (approved):
@goerz I took the liberty of pushing a `DocStringsCollapsed` -> `CollapsedDocString` bikeshed. Happy to revert it if you prefer `DocStringsCollapsed`. From my side, I'm happy if we merge this!

goerz:
I have no preferences on the name of the option. Looks good to me!

mortenpi:
Then let's merge it! Thank you @goerz and @Hetarth02 for iterating on this!

lmiq:
I'm sorry. How is this keyword supposed to be used? Searching the `Documenter` docs leads me to the release notes page, than to these PRs, but there is not (as far as I could find) documentation on how to use these options.

goerz:
Are you using the `master` branch version of Documenter? Because this has not been released yet.

lmiq:
Uhm, this in the release notes of 1.0.0, so that got me confused:

Added the ability to expand/collapse individual as well as all docstrings. ([#1393](https://github.com/JuliaDocs/Documenter.jl/issues/1393), [#2078](https://github.com/JuliaDocs/Documenter.jl/issues/2078))

goerz:
That’s referring to the feature of being able to collapse and expand docstrings *manually*

PR #2396, opened 2024-01-09
Suggest using julia-actions/cache in the hosting documentation

JamesWrigley:
This should speed up builds, in my case the time to build the docs went from ~2 minutes to ~50s.


JamesWrigley:
RIP CI :smiling_face_with_tear: A lot are failing at julia-actions/julia-buildpkg, but isn't that unnecessary since Documenter.jl doesn't have a build script?

goerz:
Not sure. Might just be a temporary outage, too.

Issue #2159, opened 2023-06-29
Option for CDN-free builds


mortenpi:
The generated HTML/JS/CSS from the documentation builds depend on cdnjs and Google Fonts right now. It might be useful in some cases to build the docs without CDN-ed dependencies.

Something also worth considering is that we are kinda vulnerable to a CDN going offline or becoming inaccessible right now. So we could also consider bundling by default.

TheStarAlight:
Couldn't agree more. I usually work under an unstable network, and the equations and fonts usually load very slowly :(
I think the LocalHTML is a better plan because it not only gets rid of online resources but also ensures that the page looks the same as the version it is built.

exaexa:
Hi, thanks for opening this. I think I'll need that feature too. :]

Some extra reasons why bundling the javascripts together with the doc would be marvelous:
- Serving documentation that links to CDNs may require a GDPR banner in EU. This problem _might_ get softened a bit by at least using the proper referrer policy in the generated HTML, there's currently none. (IANAL tho.)
- Deployment of non-public (e.g., company-local) documentation doesn't leak user access stats and internal referer URLs
- people who like to block javascripts don't need to unblock the whole cloudflare in order to get the doc search running

PR #2401, opened 2024-01-15
HTMLWriter: allow extra attributes for external asset links, default to no-referrer policy

### LINKED ISSUE #2391: Prevent leaking trackable pageview stats to CDNs by default
This is a kinda continuation of my previous comment here: https://github.com/JuliaDocs/Documenter.jl/issues/2159#issuecomment-1879633205 from #2159 .

It would be great if the generated HTML didn't send referrer URLs in headers, thus preventing the CDNs the default access to (very easily trackable) user page load stats by default. In turn, this would allow much better compliance with various privacy protection guidelines, incl. GDPR.

For the beginning, I think that just pasting a `referrerpolicy="no-referrer"` into all `<link>` tags that point to external CDNs would be more than sufficient. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#referrerpolicy for details there.

Thanks!

exaexa:
Should close #2391. cc @mortenpi 

( I don't expect this code to stay for very long, but it's a simple fix that may hold for at least as long as the #2159 isn't fixed. )

### DIFF
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -619,11 +619,13 @@ mutable struct HTMLContext
     search_index_js :: String
     search_navnode :: Documenter.NavNode
     atexample_warnings::Vector{AtExampleFallbackWarning}
+    external_link_attribs :: Vector{Pair{Symbol,String}}
 
     HTMLContext(doc, settings=nothing) = new(
         doc, settings, [], "", "", "", [], "",
         Documenter.NavNode("search", "Search", nothing),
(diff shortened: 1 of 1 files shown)


mortenpi (commented):
Something I noticed: we load most of our JS dependencies via requirejs, and so those are not covered by these changes. So the CDN will still know what site is making the requests. I wonder if using the [`<meta name="referer">` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name) would be the more appropriate solution?

PR #2402, opened 2024-01-15
Caution against dynamic `prettyurls`

### LINKED ISSUE #2395: Remove CI-dependent `prettyurls` and recommend LiveServer.jl?
See the discussion in https://github.com/JuliaCI/PkgTemplates.jl/pull/450#discussion_r1446173607

goerz:
We should steer people away from using

    prettyurls = get(ENV, "CI", nothing) == "true"

(or even `prettyurls = false`) and towards using `LiveServer`.

It is best to have a consistent setting and to always use a webserver when viewing documentation locally. This avoids situations where there might be subtle difference with `prettyurls=true` or `prettyurls=false` and the documentation might work correctly when viewed locally, but be subtly broken (without an error message!) when deploying on GitHub Action. One such situation is a `@raw` block that references a local image. The correct relative path of that image would be different depending on the `prettyurls` setting.

Closes #2395


goerz:
Just a proposal to get things started. I have no particular attachment to any of the text. Feel free to tell me to change something, or directly push to the branch with whatever text you like.

mortenpi:
I'm actually kinda leaning towards just removing the

```
get(ENV, "CI", nothing) == "true"
```

suggestion. Or maybe putting it in a `!!! warning` admonition, saying that "you may run into this, but it's not really recommended".

goerz:
Sounds good!

Updated to https://documenter.juliadocs.org/previews/PR2402/man/guide/#Building-an-Empty-Document

docs/src/man/guide.md
```diff
@@ -139,35 +139,38 @@ build/
...
+    )
+    ```
+
+    The intent behind this is to use `prettyurls=false` when building the documentation
+    locally, for easy browsing, and `prettyurls=true` when deploying the documentation
+    online from GitHub Actions.
+
+    However, this is not recommended. For example, if a
+    [`@raw` block](@ref @raw-format-block) references a local image, the correct relative
+    path of that image would depend on the `prettyurls` setting. Consequently, the
```
mortenpi, line 170:
```suggestion
    path of that image would depend on the `prettyurls` setting ([#921](@ref)). Consequently, the
```

mortenpi (approved):
LGTM, thank you @goerz!

PR #2404, opened 2024-01-16
Apply JuliaFormatter on the codebase

### LINKED ISSUE #2338: question: formatter config in codebase
I have a question if the code of Documenter.jl has a JuliaFormatter configuration. I feel like it's needed since I don't want to annoy the maintainers of this project when contributing and also from future contributors. I just noticed this since I set up my Julia LSP to autoformat and caused a huge diff. Would anyone be interested on setting up at least their own JuliaFormatter config here so people will just use the config when formatting if they contribute?

mortenpi:
Just RFC draft for now. I ran JuliaFormatter [using the DocumenterCitations style](https://github.com/JuliaDocs/DocumenterCitations.jl/blob/master/.JuliaFormatter.toml) (just because we have precedent in JuliaDocs for that style).

Would need a workflow as well before merging.

Would close #2338 if merged.


fingolfin:
Can this be closed given that PR #2591 was merged which applied runic formatting to everything?

Issue #2406, opened 2024-01-19
Search filter toggle buttons have broken UI [regression]


LilithHafner:
The "Filters" toggle buttons' state is no longer visible in the UI.

![Screenshot from 2024-01-19 10-22-41](https://github.com/JuliaDocs/Documenter.jl/assets/60898866/956799ca-ca02-4932-b86b-38619864b3f3)

After clicking some of the buttons a bunch I got this search result, with no indication of how I can get back to a normal result.

New UI introduced in #2202 cc @Hetarth02. The UI looks fine in the PR demo, but now it seems broken on docs.julialang.org on both IceCat and Chromium.

Hetarth02:
Does julialang's docs introduce some custom css?

I think it may be due to that. Also, can you confirm if this also happens in the light mode as well or not?

LilithHafner:
> can you confirm if this also happens in the light mode as well or not?

Can confirm it works for me in light mode.

Hetarth02:
Did a PR which fixes this in dark mode as well.

PR #2408, opened 2024-01-19
Fix Search filter toggle buttons UI

Hetarth02:
This PR aims to fix Issue #2406.

Filter Not selected:
image

Filter selected:
image


LilithHafner (commented):
I find it difficult to review the change to `assets/html/themes/documenter-dark.css`. Would you clarify what that change is? (It seems to me that maybe that file should not be under source control at all, but that may be off topic for this PR)

Hetarth02:
`assets/html/scss/documenter-dark.scss` this is the file with changes.

`.css` files are compiled versions of the `.scss` files.

mortenpi (approved):
LGTM, thank you!

PR #2422, opened 2024-01-23
Do not test against Julia-nightly

goerz:
I don't think we should test against nightly. It fails all the time, and marks all PRs as broken without providing any useful information.

It seems better to manually test against an upcoming version of Julia when there is a release candidate.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,6 @@ jobs:
         version:
           - '1.6'
           - '1'
-          - 'nightly'
         os:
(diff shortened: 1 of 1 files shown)


mortenpi:
No, we just need to fix the doctests that have become outdated (it's somewhere on the TODO list). We definitely want to keep testing against nightly -- nightly changes breaking Documenter is not a common thing, but reasonably regular to justify testing against it.

FWIW, nightly failures do not block a merge (IIRC).

goerz:
Alright, sounds good!

PR #2424, opened 2024-01-24
Write inventory files

### LINKED ISSUE #2366: At-ref proposal with InterSphinx compatibility
I would propose that we add support for external `@ref` references in a way that is fully compatible with [InterSphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) (part of the [Sphinx](https://www.sphinx-doc.org/en/master/index.html) framework that is the Python-equivalent of `Documenter`). Sphinx is used by virtually all Python projects (including Python itself), as well as some C/C++/Fortran projects.

This provides a solution for #688 and #425.

It may also address #319 and #1343.

The proposal has two parts:

1. Every call of `makedocs` creates an `objects.inv` file in the `build` folder that contains a mapping of all symbols in the project to a relative URL. This "inventory" file gets deployed alongside the rest of the `build` folder. Thus, it will be downloadable as, e.g., `https://documenter.juliadocs.org/stable/objects.inv`, `https://documenter.juliadocs.org/v1.1/objects.inv`, etc. See below for the format of the `objects.inv` file.

2. A plugin, `DocumenterInterlinks` (or some other name) provides the ability for a project to link to any other documentation. For example, `DocumenterCitations` might want to link to the main Julia docs and the `Documenter` docs. It would set that up as follows in `docs/make.jl`:

```julia
# make.jl file of DocumenterCitations
# ...
using DocumenterInterlinks

links = DocumenterInterlinks(
    "julia" => "https://docs.julialang.org/en/v1.9/",
    "Documenter" => "https://documenter.juliadocs.org/stable/",
    "matplotlib" => "https://matplotlib.org/stable",  # just for extra fun (see below)
)

makedocs(plugins=[links, ], …)
```

Assuming that both the Julia documentation and the Documenter documentation were built such that `https://docs.julialang.org/en/v1.9/objects.inv` and `https://documenter.juliadocs.org/stable/objects.inv` exists, this would enable to write Markdown like the following the `DocumenterCitations` documentation:

```
... sorting is done via the [`sort`](@ref Base.sort) routine ...

Pass the `plugins` to [`Documenter.makedocs`](@ref).
```

When building `DocumenterCitations`' documentation, these two references would ordinarily not be resolvable. With the `DocumenterInterlinks` plugin, for any reference that cannot be resolved locally, the plugin would go through all the external references (`"julia"`, `"Documenter"`) that were set up above, download the `objects.inv` file, and resolve to the link defined here. See below for details.

In principle, all of the above would work independent of the format of the `objects.inv` file (or `inventory.toml`, or any other name). I would propose though to adopt the exact format defined by InterSphinx, as detailed below. First, the format is fundamentally suitable (not surprisingly, since Sphinx and Documenter are quite similar). Second, it would allow interoperability with Sphinx: Julia projects could link to Python projects (including the Python manual). For example, the [PythonPlot.jl](https://github.com/JuliaPy/PythonPlot.jl) documentation could link directly to the [matplotlib](https://matplotlib.org) documentation. I've also used Sphinx for a [large Fortran project](https://docs.qdyn-library.net) of which `QuantumControl.jl` is a continuation, and I would certainly be interested in being able to link from the `QuantumControl.jl` documentation to the Fortran QDYN documentation. The other way around would work too, of course: Python projects (or anything using Sphinx) could link directly to Julia documentation.

All of this is a bit gratuitous for the original problem #688 of just linking between Julia projects, but since we have to design some kind of inventory file anyway, simply adopting the InterSphinx format would give us a lot of interoperability basically "for free".

The `links` set up in the configuration may choose to link to something like a `stable` target, to a specific version. The former runs the risk of links breaking when a new version of the target package is released, but it's up to the user what they want. In any case, this is strictly better than the current solution of explicit URLs:

```
Pass the `plugins` to [`Documenter.makedocs`](https://documenter.juliadocs.org/stable/lib/public/#Documenter.makedocs).
```

Now, if Documenter 2.0 came out without a `makedocs` function, I'd no longer have to manually fix all the links. Instead, I could just change the `stable` to `v1` in a single location in my `make.jl` file.

Lastly, I would stress that creating and deploying the `objects.inv` should be part of Documenter itself, not of the `DocumenterInterlinks` plugin: We'd want to be able to link to *any* Documenter-based documentation, without the authors having to opt-in via a plugin. An `objects.inv` file (in any format) would also solve other problems. For example, there has been some discussion (@pfitzseb) about VS Code indexing the docstrings of all installed packages. If these docstrings link to other parts of the documentation, an `objects.inv` could be used to resolve these links without VS Code needing to ingest every package's full documentation.

## Creating an Inventory File

Documenter already keeps track of all reference targets (docstrings, headers) in a project. It would simply have to dump these into an `objects.inv` file in a format compatible with InterSphinx. This format has been reverse-engineered here: https://sphobjinv.readthedocs.io/en/stable/syntax.html

Basically, the `objects.inv` is a compressed list of plain text lines with the structure (parsed by a [regex](https://github.com/sphinx-doc/sphinx/blob/2f60b44999d7e610d932529784f082fc1c6af989/sphinx/util/inventory.py#L115-L116))

```
{name} {domain}:{role} {priority} {uri} {dispname}
```

Documenter would set the individual fields as follows:

*   `name`: For a docstring, this would be the fully qualified name, e.g. `Documenter.makedocs`. For headers, it should be the header ID, as described in ["Duplicate  Headers"](https://documenter.juliadocs.org/stable/man/syntax/#Duplicate-Headers), or an auto-generated ID (the "slug") if there is no specified ID.

*   `domain` would always be `jl`. This matches the choice in the [Sphinx-Julia domain](http://bastikr.github.io/sphinx-julia/juliadomain.html). While that project seems to be inactive, their choices are sensible, and we should adopt them.

*   `role` should be one of `type`, `abstract`, `mod`, `func` for docstrings, depending on what kind of object the docstring is for. Again, this is taken from the [Sphinx-Julia domain](http://bastikr.github.io/sphinx-julia/juliadomain.html#roles). This isn't really relevant for linking between Julia projects, since Documenter doesn't differentiate between

    ~~~
    [`makedocs`](@ref Documenter.makedocs) (link to a function)
    [`Documenter.Document`](@ref) (link to a type)
    ~~~

    but for Sphinx projects linking to a Documenter project, this would come into play, and the relevant [reStructuredText](https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html) syntax would be

    ~~~
    :jl:func:`makedocs <Documenter.makedocs>` (link to a function)
    :jl:type:`Documenter.Document` (link to a type)
    ~~~

    For headers, `role` would always be [`ref`](https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref).

*   `priority` would be `1`  (not relevant to Documenter/InterSphinx)

*   `uri` would be the URL to access the resource, relative to the `objects.inv` file. For example, for `name=makedocs` in Documenter's `objects.inv` it would be `lib/public/#Documenter.makedocs`

*   `dispname` would be the same as `name` for docstrings, and the full string of the section title for sections

## The `DocumenterInterlinks` extension

The extension would be opt-in, enabling a project to link out to any other Documenter or Sphinx project. At a minimum, this would handle all references that cannot be resolved locally, and go through all the external resources defined when instantiating the plugin to find one that can resolve the reference.

If we want to go further, we could also extend Documenter's `@ref` syntax to enable some features comparable to that of InterSphinx, see the [`external` role in Sphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#role-external)

All of the following would be valid links in Documenter:

```
* [:external:`makedocs`](@ref)  (don't try to find `makedocs` locally)
* [:external+julia:`sort`](@ref)  (link specifically to the `sort` in the Julia
  manual, cf. for the label "julia" when instantiating `DocumenterInterlinks`)
* [:py:func:`matplotlib.pyplot.subplots`](@ref) (link to a function in the
  Python `matplotlib` project's Sphinx documentation)
* [`subplots`])(@ref :external+matplotlib:py:func:`matplotlib.pyplot.subplots`)
```

I'd consider these features optional, although without them, there may be a risk for "collisions" of different packages having references for the same `name`. The first one found wins, but that may make it impossible to generate certain links. Nonetheless, I would start any implementation of a `DocumenterInterlinks` plugin to just handle otherwise unresolved references and add fancy new features only later.

I would also be open to the possibility of all external links having to use `@extref` instead of `@ref`. That is, the `DocumenterInterlinks` could provide an entirely new command instead of overloading the existing `@ref`.

goerz:
This automatically writes inventory files `objects.inv` ~~and `inventory.toml.gz`~~ to the `build` folder, which should get deployed together with the rest of the documentation. The presence of these files allows the [`DocumenterInterLinks` plugin](https://github.com/JuliaDocs/DocumenterInterLinks.jl/) to link into the documentation from any other project with an `@extref` link.

The PR adds a single new dependency: [`CodecZlib`](https://github.com/JuliaIO/CodecZlib.jl). This is necessary because inventory files should definitely be compressed (they potentially get pulled a lot, every time someone builds a documentation that has an `@extref` into the project). The `CodecZlib` is maintained by the [JuliaIO org](https://github.com/JuliaIO) and should be quite stable.

Also see the discussion in https://github.com/JuliaDocs/Documenter.jl/issues/2366#issuecomment-1879936021

Closes #2366


goerz:
Pending tests, additional documentation, and a decision on whether it's okay to write `objects.inv` *and* `inventory.toml.gz`, or one or the other.

Also, #2385 should be resolved before merging this (preferably via #2389)

goerz:
I've marked this as "Ready for review" since it now has tests and documentation (no CHANGELOG) yet.

I agree with https://github.com/JuliaDocs/Documenter.jl/issues/2366#issuecomment-1913432761, though: we should probably choose one format. I'd also lean towards the Sphinx `objects.inv`, if that's palatable to everyone. The major pros and cons are outlined in https://github.com/JuliaDocs/Documenter.jl/issues/2366#issuecomment-1879936021. On top of that, not writing the `toml.gz` would cut a significant number of lines from this PR, so that might be another pro.

The documentation is already written to only mention `objects.inv`, although the code still generates both `objects.inv` and `inventory.toml.gz`. I'll clean up and adapt after we make a decision.

goerz:
Community call decision was to only write the Sphinx-format `objects.inv`. So I'll remove the code for the toml-writing and update everything, and let you know when that's done.

goerz:
Okay, should be ready for merging or further comments.

There was actually still a bug: `.md` filenames in the documentation with spaces weren't handled correctly (not that I would ever recommend them, but they do work, and now they also have a properly escaped URI in the inventory)

#2433 is also ready. If you like #2433, you could just change the base branch for that PR to `master` and merge it *instead* of this PR. Or let me know, and I can merge it into this PR before we merge this PR into `master`.

mortenpi:
What's the fallback for `Version:` when you can't determine it from `Project.toml`? Empty string? It's still something that's relevant when you're generating the HTML for something that does not have a well-defined version number (like https://github.com/JuliaDocs/juliadocs.github.io).

I know you don't like this idea, but I still feel that the "right way" to do it right now is to patch the inventory file in the `deploydocs` phase. It should be a reasonably simple patch of the file: just `readlines` through the inventory, find the `^Version:` line, and if the version number is empty, add the correct version then?

Using `Project.toml` would be automatic for the user, but it introduces two sources of truth for the version number, and I'm really allergic to that.

mortenpi:
Also, for `main`/`master`, what do you think the `Version:` value should be? In practice, packages are not particularly consistent with whether they switch to `vX.Y.Z-DEV` after a version was tagged. In `deploydocs`, we could enforce something. E.g. I would be okay with `deploydocs` going into `Project.toml`, figuring out the version number, and then making sure there's a `-DEV` at the end, if it's a "dev" version.

goerz:
> What's the fallback for Version: when you can't determine it from Project.toml? Empty string?

Yes, empty string and a warning to set a manual version as a parameter to `HTML`.

> I know you don't like this idea

Indeed :-) 

> but I still feel that the "right way" to do it right now is to patch the inventory file in the deploydocs phase

Alright, fair enough. Let me have a stab at that. I'll make an alternative to #2433.

You're okay with the new `version` parameter, in `HTML`, though, right? So I'll just remove the fallback to `Project.toml`, and add something in `deploydocs` that patches in a version.

goerz:
> Also, for main/master, what do you think the Version: value should be?

I actually have [very strong opinions on that](https://discourse.julialang.org/t/incrementing-package-version-number-between-releases/106966/4), but you're right, most people aren't very strict or consistent about that.

> In deploydocs, we could enforce something. E.g. I would be okay with deploydocs going into Project.toml, figuring out the version number, and then making sure there's a -DEV at the end, if it's a "dev" version.

It's probably not that critical what exactly the `version` inside the inventory is for a dev-version. But I guess `deploydocs` right now only really knows that it's "dev", without any related version number? I'll have to have a look at the code in `deploydocs`...

mortenpi:
> You're okay with the new `version` parameter, in `HTML`, though, right?

Actually... I was kinda hoping we wouldn't need it :sweat_smile: _But_ I would be okay with `version` or `inventory_version`, if you think we need the ability to set the value in cases where `deploydocs` can't help us (it seems like a reasonable thing you want to do sometimes).

One alternative possibility: we can have a public API function that you can use to patch the inventory file, which `deploydocs` will use, but the user can also use if they do need to set the version, and they don't use `deploydocs`. But.. now that I typed this out.. probably better to have just an argument to `HTML`.

On `version` vs `inventory_version`: I would prefer not to make it quite an official to set a version right now, due to the multiple-source-of-truth issue. For better or for worse, I think we need `deploydocs` to be in charge of the version numbers right now. The only user now would be the inventory writing, so let's make the option specific to that?

goerz:
I’m fine with `inventory_version`, and I’ll have a look at `deploydocs`

src/html/write_inventory.jl
```diff
@@ -0,0 +1,139 @@
...
+
+The `objects.inv` file is compatible with [Sphinx](https://www.sphinx-doc.org/en/master/index.html)
+See [DocInventories](https://juliadocs.org/DocInventories.jl/stable/formats/) for a
+description. The file can be used by [Intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html)
+and the [DocumenterInterLinks](https://github.com/JuliaDocs/DocumenterInterLinks.jl/)
+plugin to link into the documentation from other projects.
+"""
+function write_inventory(doc, ctx)
+
+    @info "Writing inventory file."
```
mortenpi, line 15: This happens unconditionally anyway and is part of the writer, so I don't think we need to print this:

```suggestion
```

goerz: Sure, sounds good! I'll update that when we reach consensus on the `version` issue and I wrap everything up.

goerz:
From https://github.com/JuliaDocs/Documenter.jl/pull/2442#issuecomment-1926289407, on the question of how and where to set the `version` metadata for the inventory file, I have no strong preferences except that it meets the following two requirements:

1. There has to be some way to set the version string for the `inventory` in the `makedocs` phase, either manually with `inventory_version` in `HTML` or by reading it from the `Project.toml`
2. Any project with a standard setup should end up with some appropriate (non-empty) `version` in its inventory for both tagged releases and `dev` deployments. This should happen without them having to make any changes in their configuration after updating to the version of `Documenter` that will include the inventory writing.

goerz:
Just to keep the discussion here up-to-date here, we now have three competing PRs for completing this (regarding what is the best way to set the inventory `version`), see https://github.com/JuliaDocs/Documenter.jl/pull/2443#issuecomment-1937911025 for details:

* #2442 – `makedocs` gets the version for the inventory from the closest `Project.toml`/`JuliaProject.toml`/`VERSION` that defines a version. Then `deploydocs` checks whether that version matches the tag for the release that's being deployed (for non-`dev` deployments). There is no need for a manual `inventory_version` option. I'd be extremely happy with this option!
* #2448 – The inventory version is set by `deploydocs`. For tagged releases, it uses the version from the tag, and for `dev`-releases, it falls back to the same logic as #2442 to get the version from the closest `Project.toml`. **The ultimate result is identical to #2442** (at least that's the intent), but the code is a bit more convoluted than necessary, and I'm not enthusiastic that the `objects.inv` file isn't generally complete after the `makedocs` step.
* #2443 – This PR, which sets the version in `deploydocs` without a fallback. As a result, `dev`-deployments would typically have an empty `version` in the inventory, and we'd have to push people to modify their setup and specify a manual `inventory_version`. I'd find this one hard to swallow.

src/html/write_inventory.jl
goerz: Fixed in https://github.com/JuliaDocs/Documenter.jl/pull/2449

goerz:
This has been updated by squash-merging https://github.com/JuliaDocs/Documenter.jl/pull/2449 as the solution for setting the `version` metadata that we reached consensus on.

I've also added a CHANGELOG entry.

goerz:
@mortenpi This should be ready now for a final review (if there's any more bikeschedding) and merging.

There is one issue we might still have to think about: Since we've settled on auto-detecting the version only for "standard setups", that no longer includes the `VERSION` file that Julia itself uses. I'd put a high priority on the inventory file for Julia (including the nightly version) being in good shape, ideally without them having to tweak their setup. I *think* it will work out the box, since they seem to have a custom setup that deploys nightly to https://docs.julialang.org/en/v1.12-dev/, i.e., a folder that includes a proper version, not just the standard `dev`.

On the other hand, they might also pin a specific Documenter version, and in that case we'll have to make a PR to Julia to update that to the new release anyway, which then might as well include setting `inventory_version` manually. I'll try to look into this or find someone on Slack with a better understanding of Julia's documentation setup.

I've rebased everything onto the current `master` and rewritten the history to be somewhat sane. If possible, I would prefer to merge this with a non-fast-forward merge commit (as opposed to a squash-merge) in order to preserve some of the history, especially the commit messages referencing the various alternative PRs about how to set the version. I'm actually not quite sure if there's a button in the Github UI to do that kind of merge (I *think* it would be "Create a merge commit"), but on the command line, it would be `git merge --no-ff mg/write-inventory` from `master`.

I've assumed that this will make it into a `v1.3.0` release eventually (and set `version = "1.3.0-dev"` in the `Project.toml` accordingly). I still have to do some tweaking of the `DocInventories`/`DocumenterInterLinks` documentations to adapt to the new reality of inventory-writing now being built in to Documenter ("As of Documenter 1.3.0…"). For the same reason, I also wouldn't mind terribly if we waited at least until next Monday to actually tag a new Documenter release.

After we make the release, it might make sense to make some kind of announcement on Discourse about the new inventory feature. I'll also want to make a separate announcement of the `DocumenterInterLinks` plugin, but I want to polish that just a little more before I remove the official "experimental, use at your own risk" warning.

goerz:
> I'd put a high priority on the inventory file for Julia (including the nightly version) being in good shape, ideally without them having to tweak their setup. […] On the other hand, they might also pin a specific Documenter version

Thanks to @vchuravy's help on Slack, I was able to determine that the Julia documentation builds pretty much just by running `doc/make.jl`, like any other package. Just for future reference, the building and deployment of Julia's documentation is set up at https://github.com/JuliaCI/julia-buildkite, specifically in the files
[`doctest.yml`](https://github.com/JuliaCI/julia-buildkite/blob/main/pipelines/main/misc/doctest.yml) and [`deploy_docs.yml`](https://github.com/JuliaCI/julia-buildkite/blob/main/pipelines/main/misc/deploy_docs.yml).

There is a custom `DeployConfig` in `doc/make.jl`. Just from looking at that, it seems to me that `deploydocs` would set the correct inventory-version automatically during deployment.

However, there is also a `docs/Manifest.toml` file that pins the `Documenter` version, so we'll have to submit a pull request to update that. The precedent for that is https://github.com/JuliaLang/julia/pull/47105. So, @mortenpi has done this before, and we should be fine :-)

I would still include adding an `inventory_version=VERSION` to the `HTML()` options for the `makedocs` in that PR as well, though. It would be good that if people build the Julia documentation locally, they still get an `objects.inv` file that has the correct version set (and, with the design we have now, `makedocs` would *not* be able to auto-detect the version with Julia's setup).

Bottom line: Julia's peculiarities are not a blocker for merging this PR, but we will have to follow up in order to ensure that future deployments of Julia's documentation have an inventory.

goerz:
> I also wouldn't mind terribly if we waited at least until this coming Monday to actually tag a new Documenter release.

Actually, we can *tag* it right away, but maybe hold off a few days with any kind of Discourse announcement. That might even give us a chance to

> follow up in order to ensure that future deployments of Julia's documentation have an inventory.

and for a handful of other packages to make a release/build with the newest Documenter before we go public; so it's not quite "you can link to any other project, but there's basically nothing yet that actually has an inventory" (despite the handful of inventories I've collected in the [wiki](https://github.com/JuliaDocs/DocumenterInterLinks.jl/wiki/Inventory-File-Repository))

goerz:
It turns out the inventory-writing-code is more or less compatible all the way back to `Documenter v0.25`, so I put that at https://github.com/JuliaDocs/DocumenterInventoryWritingBackport.jl (currently [waiting for registration](https://github.com/JuliaRegistries/General/pull/101575)). So, no excuses for any package not to have an inventory, even if they didn't like dealing with the Documenter 1.0 update.

mortenpi (approved):
Sorry for the delay, but LGTM!

goerz:
@mortenpi Can I merge this now (manually)? Ideally, before we merge too many other PRs, since otherwise I'll have to go through the rebase and conflict resolution again.

goerz:
Ah, branch protection! 

CleanShot 2024-02-26 at 11 23 54@2x

I'd still kinda like to preserve some of those commit messages with the references to the discussion, so not sure what to do. Maybe temporarily remove the protection, or maybe you can push with higher permissions than me.

mortenpi:
Let's just squash merge this. The PRs offer a better context for the decisions here I think.

Issue #2132, opened 2023-05-25
LaTeX output requires `pages` argument to `makedocs()`


johnomotani:
When I tried to build LaTeX documentation for a project where HTML documentation was already working, I found the `.tex` file was 'empty' (just the preamble and `\end{document}`). By digging through the source code, I found that this was because `doc.user.pages` was empty here https://github.com/JuliaDocs/Documenter.jl/blob/233a7822d670b063e894f62aea00729bcb3a28dd/src/latex/LaTeXWriter.jl#L116
and could be fixed by passing a `pages` kwarg to `makedocs()`.

It would be helpful to document that the `pages` argument is required, e.g. here https://documenter.juliadocs.org/stable/man/other-formats/#pdf-output.

It would also be nice to default (when `pages` is not passed) to building all the pages in some default order, the way the HTML writer already does. The pages do seem to be present already, e.g. in `doc.blueprint.pages`, so I guess this should not be too difficult?

mortenpi:
> It would also be nice to default (when `pages` is not passed) to building all the pages in some default order, the way the HTML writer already does. The pages do seem to be present already, e.g. in `doc.blueprint.pages`, so I guess this should not be too difficult?

This seems reasonable to me. Not even sure why we don't do that already.

mortenpi:
Note: this is actually a bug -- the documented `pages` behavior is that it should pick up the `.md` files:

> If omitted, Documenter will automatically generate a flat list of pages based on the files present in the source directory.

PR #2431, opened 2024-01-29
Document `makedocs` argument `pages` for LaTeX output

ChromatinRemodeling:
Trying to fix [#2132](https://github.com/JuliaDocs/Documenter.jl/issues/2132).


mortenpi:
Actually, I'm leaning towards considering this a bug. IIRC, for HTML we do populate `pages` automatically, and I'm actually unsure why it's not also happening with LaTeX. From the `makedocs` docstrings:

> If omitted, Documenter will automatically generate a flat list of pages based on the files present in the source directory.

I would be okay merging this doc note if we make it clear that it's a bug. But it would be good to also check why the documented behavior is not working -- it might be a simple fix.

ChromatinRemodeling:
Update: I am able to build the whole Documenter.jl documents with the tweak above, by:
* commenting out the `pages` in `docs/make.jl` and the `if "pdf" in ARGS` deployment part
* passing `remotes = nothing` and `format = Documenter.LaTeX(platform = "none")` into `makedocs`
* manually fixing some broken references in the .md files
* running `julia --project docs/make.jl pdf`
* manually fixing some `\includegraphics` bugs in the .tex file
* run `latexmk -f -interaction=batchmode -halt-on-error -view=none -lualatex -shell-escape Documenter.jl.tex`

Hope this is helpful.

mortenpi:
Could you try adding another test build that looks like this, but doesn't specify `pages`?

https://github.com/JuliaDocs/Documenter.jl/blob/d521d5a902109d878556cca38afa105e170ddc61/test/examples/make.jl#L829-L847

It should ideally pick up the `pages` from https://github.com/JuliaDocs/Documenter.jl/tree/master/test/examples/src.latex_showcase

It would also need a corresponding test here, and maybe we can even reuse the `references/latex_showcase.tex` reference file:

https://github.com/JuliaDocs/Documenter.jl/blob/d521d5a902109d878556cca38afa105e170ddc61/test/examples/tests.jl#L340-L349

mortenpi:
Hmm. In `HTMLWriter` we use `doc.blueprint.pages`:

https://github.com/JuliaDocs/Documenter.jl/blob/d521d5a902109d878556cca38afa105e170ddc61/src/html/HTMLWriter.jl#L695-L699

Maybe if we use that instead of `doc.user.pages`, then we don't need to use `.navlist`?

mortenpi:
Let's merge this as is, since it can be useful for users. We can fix the behavior in a follow-up PR.

docs/src/man/other-formats.md
```diff
@@ -24,6 +24,9 @@ the version number will be appended to the title.
 The `makedocs` argument `authors` should also be specified, it will be used for the
 `\authors` field in the tex document.
 
+!!! warning "Known issue"
+    If the `makedocs` argument `pages` is not assigned, Documenter will generate tex documents without contents.
```
mortenpi, line 28:
```suggestion
    If the `makedocs` argument `pages` is not assigned, Documenter will generate tex documents without contents. ([#2132](@ref))
```

PR #2433, opened 2024-01-29
Add `version` to `HTML` for inventory

goerz:
Add the parameter `version` to `Documenter.HTMLWriter.HTML` which is used for the inventory. If the `version` is not set, the inventory falls back to searching for a `Project.toml` file that specifies a version.


goerz:
There should be some way to manually specify a `version` for the inventory, and this seems like the best source.

However, since no existing projects define this, they'll mostly end up with an empty version in the inventory, which isn't ideal. So, I also added the idea from https://github.com/JuliaDocs/Documenter.jl/issues/2385#issuecomment-1915176366 to get the `version` from the closest `Project.toml` file. That should work out of the box for pretty much all packages.

Note that this PR is on top of #2424

Project.toml
```diff
@@ -23,6 +23,7 @@ PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 RegistryInstances = "2792f1a3-b283-48e8-9a74-f99dce5104f3"
 SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
```
goerz, line 26: If we don't want to add the standard library `TOML` dependency for some reason, I can also use the same regex-based extraction as in

https://github.com/JuliaDocs/Documenter.jl/blob/deecbd7d3abc395fe31d7a393539bcda4c00d753/src/Documenter.jl#L32-L38

Using TOML seems pretty unproblematic, though. In fact, it would probably be fine to use for the definition of `DOCUMENTER_VERSION`

src/html/write_inventory.jl
```diff
@@ -72,6 +75,29 @@ function write_inventory(doc, ctx)
 end
 
 
+function _find_project_version(doc, ctx)
+    version = ctx.settings.version
+    if isempty(version)
+        current_dir = pwd()
```
goerz, line 81: I also considered starting from `doc.user.source` (the `docs/src` directory). But I think the current working directory (which I think Documenter changes to `docs`) will work just fine for pretty much any project, and saves one iteration of the loop.

goerz:
I'm going to close this one in favor of #2442

At this point, I'm really quite sure that the version in `objects.inv` should always match the version in `Project.toml`, no matter what.

For projects that don't have a `Project.toml` (some standalone documentation repository, presumably), it's okay if `version` is blank by default. The easiest solution in that case is if I add something to `DocInventories` to insert a custom version string into the `objects.inv` file that Documenter generated, after the call to `makedocs`. There's really no need to complicate the `Documenter` API with a `version`/`inventory_version` option.

PR #2436, opened 2024-01-31
See what happens with `version` in `makedocs`

goerz:
Ignore this. It's just to trigger CI and build a `preview` documentation so that I can check the behavior of `version` in `makedocs` in that context.


fingolfin:
can this be closed now?

PR #2437, opened 2024-02-02
Improve repo path error message

### LINKED ISSUE #2427: ERROR: LoadError: The repo path… should not contain the protocol
In make.jl, if 

```
deploydocs(
    repo    = "https://github.com/SomePerson/Something.jl.git"
    target  = "build"
)
```

contains `https://`, it will give the error:

```ERROR: LoadError: The repo path https://github.com/SomePerson/Something.jl.git should not contain the protocol```

If Documenter.jl can detect `https://`, then it should be robust enough to snip it out of the path go on its merry way.  If that's too hard, a more helpful error message would have been  `...should not contain the protocol 'https://'`, as  "the protocol" is a bit cryptic and non-specific.

### PR DESCRIPTION
**Pull request by Jay-sanjay**
Fixes #2427 

*edits-closes #2427*

### DIFF
--- a/src/deploydocs.jl
+++ b/src/deploydocs.jl
@@ -480,7 +480,7 @@ Disassemble repo address into user, host, and path to repo. If no user is given,
 """
 function user_host_upstream(repo)
     # If the repo path contains the protocol, throw immediately an error.
-    occursin(r"^[a-z]+://", repo) && error("The repo path $(repo) should not contain the protocol")
+    occursin(r"^[a-z]+://", repo) && error("The repo path $(repo) should not contain the protocol 'https://'")
     #= the regex has three parts:
     (?:([^@]*)@)?  matches any number of characters up to the first "@", if present,
         capturing only the characters before the "@" - this captures the username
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by goerz** [comment]
Technically, could be `http://`, too, but this seems fine to me, and probably is a clearer message for at least some people

**Review by goerz — APPROVED** [review]

File: src/deploydocs.jl [hunk 1]
```diff
@@ -480,7 +480,7 @@ Disassemble repo address into user, host, and path to repo. If no user is given,
 """
 function user_host_upstream(repo)
     # If the repo path contains the protocol, throw immediately an error.
-    occursin(r"^[a-z]+://", repo) && error("The repo path $(repo) should not contain the protocol")
+    occursin(r"^[a-z]+://", repo) && error("The repo path $(repo) should not contain the protocol 'https://'")
```
mortenpi, line 483: Nitpick, but the regex matches _any_ protocol:

```suggestion
    occursin(r"^[a-z]+://", repo) && error("The repo path $(repo) should not contain the protocol (e.g. 'https://')")
```

goerz (approved):

goerz (approved):
Note to @Jay-sanjay: We're using this PR to test some local workflows around approvals, so don't be confused by the multiple status changes

mortenpi:
Yep, sorry for the noise. But thank you for the PR!

Issue #2434, opened 2024-01-31
Wrong `\includegraphics` image path at creating LaTeX output


ChromatinRemodeling:
If I link an image in a deep directory, say, `docs/src/buggy/buggyagain/`, instead of 
```latex
\includegraphics[max width=\linewidth]{buggy/buggyagain/bugtest.png}
```
the `makedocs` function will generate
```latex
\includegraphics[max width=\linewidth]{buggy\buggyagain\bugtest.png}
```
in a .tex document, which will produce an error when compiled.

A rather simple MWE can be seen in [here](https://github.com/ChromatinRemodeling/TestDocGraphics.jl).

This issue seems to be caused by the incorrect handling of `image.path` in the function below:
https://github.com/JuliaDocs/Documenter.jl/blob/deecbd7d3abc395fe31d7a393539bcda4c00d753/src/latex/LaTeXWriter.jl#L681-L694

ChromatinRemodeling:
PS: I am using Windows. Here's my `versioninfo()`:

```julia-repl
julia> versioninfo()
Julia Version 1.10.0
Commit 3120989f39 (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 16 × AMD Ryzen 7 5800H with Radeon Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver3)
  Threads: 1 on 16 virtual cores
Environment:
  JULIA_CONDAPKG_BACKEND = Null
  JULIA_PKG_SERVER = https://pkg.julialang.org
  JULIA_PYTHONCALL_EXE = D:\Programs\mambaforge\envs\torch\python.exe
  JULIA_EDITOR = code
  JULIA_NUM_THREADS =
```

ChromatinRemodeling:
Note: `replace(image.path, '\\' => '/')` would be a quick fix in most cases.

mortenpi:
Yeah, this looks like missing handling of the windows path separator. We do such replacements in other places, so might just be an oversight.

Which `platform` are you using though? I am surprised that Windows native TeX wouldn't be able to handle the `\` separator?

ChromatinRemodeling:
I am not sure what you mean by `platform`? I am using MiKTeX on Windows, if this is the information you want to know.

[...]

PR #2440, opened 2024-02-03
Fix wrong `image.path` (#2434)

ChromatinRemodeling:


mortenpi:
Like in https://github.com/JuliaDocs/Documenter.jl/pull/2431#issuecomment-1925208280, it would be good to have a test here that would also run on Windows. Modifying the `texonly` test should be sufficient.

Note: you should be able to set the `DOCUMENTER_FIXTESTS` environment variable to automatically update the reference `.tex` files:

https://github.com/JuliaDocs/Documenter.jl/blob/c3c8754484a21834dc0120dbf5b694f729405387/test/examples/tests.jl#L55-L67

fingolfin:
@ChromatinRemodeling any interests in finishing this?

ChromatinRemodeling:
> @ChromatinRemodeling any interests in finishing this?

I would love to, but I am too busy recently. I will look into it when I have time and energy

Issue #2353, opened 2023-11-23
Simple documentation build command


mortenpi:
We probably want to move towards a more standardized and reliable docs build method. So my thinking right now is turning `docs/make.jl` just into a single Documenter command:

```julia
Documenter.build(package_directory)
```

It would then figure out some reasonable documentation build for that package environment. Key is that `package_directory` must be a valid package-type environment.

- First MVP would be something that just builds maybe the README + autogenerated list of (public) docstrings.
- Second iteration would automatically pick up `.md` files from the `docs/` directory, allowing for a proper manual to be built.
- Next step would be to pick up some (limited) config options from the project TOML files, for customization (https://github.com/JuliaDocs/Documenter.jl/issues/1350).

We should also create a GitHub Action, which instantiates the environment and calls this command. This way setting up docs for a package would become much simpler for the user (deployment side should also come in somewhere them, but `Documenter.deploy()` is available, and we can have enough metadata in `siteinfo.js` to determine the necessary `deploydocs` arguments I think).

A little bit unsure how to exactly handle environments here. Should there be a `docs/Project.toml`, that has Documenter? We definitely need to be able to `Pkg.add` Documenter together with the package though.

Note: this would not be for everyone. More complex manuals would still use custom `make.jl` setups. But I would hope that this would simplify the docs life for like 80% of packages.

Also some related discussion on Discourse: https://discourse.julialang.org/t/julias-documentation-needs-to-be-more-discoverable-and-explorable/106549

mortenpi:
Something we also discussed with @fredrikekre: it would be nice to be able to build a documentation index of a current `Project.toml` (for the user) -- i.e. single page site that has all the docs for all the packages in your current project (could also be re-used by e.g. VS Code). The docs themselves could be built with `Documenter.build`, and then maybe aggregated with MultiDocumenter.

Long-running at-example blocks would be a problem here though. So we should probably have a mode where evaluating at-blocks is not enabled.

In the automated flow, we could potentially just have a timeout. If the docs don't build within a certain amount of time, we kill Documenter and build with the at-example disabled.. and if that fails, maybe just build fallback docs.

odow:
> But I would hope that this would simplify the docs life for like 80% of packages.

Issue #1350, opened 2020-06-23
Declarative Documenter configuration


c42f:
It would be great to be able to introspect and control the documentation generation process from code outside the package itself.

* It would allow auto-configuration of CI resources, decoupling CI more from the doc build script (Related - https://github.com/JuliaDocs/Documenter.jl/issues/1087)
* For JuliaHub, we may want to link package docs to each other (https://github.com/JuliaComputing/JuliaHub/issues/55) which would require the docs for the package to be built in the *context* of a larger system.

It would seem that a good way to do this is to have declarative configuration in a script, as mentioned by @davidanthoff in https://github.com/JuliaDocs/Documenter.jl/issues/1087#issuecomment-536388827

> I think that would also imply moving away from this model of having a `make.jl` script that does different things depending on the env it is run, but instead either have a more declarative config file or something like `docs/docconfig.jl` that only populates some data structures about the structure of the docs, but doesn't actually trigger anything like a build or deployment.

I feel the ecosystem is fairly well set up for this, as the Documentation configuration for simple package documentation builds is already largely declarative.

tkf:
:+1: It'd be great to move to more declarative API. Or maybe even just a function-based API https://github.com/JuliaDocs/DocumentationGenerator.jl/issues/142 so that DocumentationGenerator.jl can be documentation framework-agnostic (e.g., maybe you'd want to use Franklin.jl instead).

MichaelHatherly:
Would definitely be good to move in this direction. I've been going through some possible options in CommonMark.jl for declarative configuration of parsing and rendering that would hopefully *just fit* into the ecosystem; tying this all together, whether it be Franklin, PkgPages, Documenter, Literate, Weave, or any other option would be ideal (though maybe a little too hopeful). Do we want to just piggyback off of `Project.toml` files and add metadata there such as

```toml
[docs]
# some docs-specific config...
```

or have another config file for this? For anyone familiar with `Pkg`, are there any issues related to adding custom data to `Project.toml` that would come back to bite us?

KristofferC:
> For anyone familiar with Pkg, are there any issues related to adding custom data to Project.toml that would come back to bite us?

Not really, unless Pkg would want to use the exact same key but I don't think we would want to use `[docs]` for anything in Pkg, especially not if Documenter itself starts using it. Although, perhaps it makes sense to have its own file in the `docs` folder.

davidanthoff:
[...]

PR #2443, opened 2024-02-03
Patch in inventory version in `deploydocs`

goerz:
Alternative to #2442, ~~#2433~~ #2448, and #2449.

~~Still work in progress (a little difficult to test).~~

I also think this is an extremely cumbersome approach, and hope we don't go with this one. My preferred alternative at this point would probably be the very simple #2442 or #2449.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -32,6 +32,7 @@ makedocs(
             highlights = ["yaml"],
             ansicolor = true,
             size_threshold_ignore = ["release-notes.md"],
+            inventory_version = Documenter.DOCUMENTER_VERSION,
         )
     end,
     build = ("pdf" in ARGS) ? "build-pdf" : "build",
--- a/src/deploydocs.jl
+++ b/src/deploydocs.jl
@@ -268,6 +268,24 @@ function deploydocs(;
                 @debug "running extra build steps."
                 make()
             end
+            objects_inv = joinpath(realpath(target), "objects.inv")
+            if isfile(objects_inv)
+                inventory_version = _get_inventory_version(objects_inv)
+                deploy_version = _get_deploy_version(deploy_subfolder)
+                if isempty(deploy_version)
+                    if isempty(inventory_version)
+                        @error "Cannot determine project version for inventory. Please set `inventory_version` in the `HTML` options in `makedocs`."
+                    end
+                else
+                    if isempty(inventory_version)
+                        _patch_inventory_version(objects_inv, deploy_version)
+                    else
+                        if inventory_version != deploy_version
+                            error("Inventory declares version `$inventory_version`, but `deploydocs` is for version `$deploy_version`")
+                        end
+                    end
+                end
+            end
             @debug "pushing new documentation to remote: '$deploy_repo:$deploy_branch'."
             mktempdir() do temp
                 git_push(
@@ -281,6 +299,45 @@ function deploydocs(;
         end
     end
 end
+
+
+function _get_inventory_version(objects_inv)
+    open(objects_inv) do input
+        for line in eachline(input)
+            if startswith(line, "# Version:")
+                return strip(line[11:end])
+            end
+        end
+        error("Invalid $objects_inv: missing or invalid version line")
+    end
+end
+
+
+function _patch_inventory_version(objects_inv, version)
+    objects_inv_patched = tempname()
+    open(objects_inv) do input
+        open(objects_inv_patched, "w") do output
+            for line in eachline(input; keep=true)
+                if startswith(line, "# Version:")
+                    @debug "Patched $objects_inv with version=$version"
+                    line = "# Version: $version\n"
+                end
+                write(output, line)
+            end
+        end
+    end
+    mv(objects_inv_patched, objects_inv; force=true)
+end
+
+
+function _get_deploy_version(foldername)
+    try
+        return string(VersionNumber(foldername))  # strips the leading "v" from foldername
+    catch
+        return ""
+    end
+end
+
 
 """
     git_push(
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -25,7 +25,7 @@ selector will be hidden. The special value `git-commit` sets the value in the ou
 
 # `HTML` `Plugin` options
 
-The [`HTML`](@ref) [`Documenter.Plugin`](@ref) provides additional customization options
+The [`HTML`](@ref) object provides additional customization options
 for the [`HTMLWriter`](@ref). For more information, see the [`HTML`](@ref) documentation.
 
 # Page outline
@@ -54,7 +54,9 @@ to link to any other project with an inventory file, see
 The [format of the `objects.inv` file](https://juliadocs.org/DocInventories.jl/stable/formats/#Sphinx-Inventory-Format)
 is borrowed from the [Sphinx project](https://www.sphinx-doc.org/en/master/). It consists
 of a plain text header that includes the project name, taken from the `sitename` argument
-to [`Documenter.makedocs`](@ref), and a project `version` (currently empty).
+to [`Documenter.makedocs`](@ref), and a project `version` taken from the
+`inventory_version` argument of the [`HTML`](@ref) options, or automatically
+determined during [deployment](@ref Documenter.deploydocs).
 The bulk of the file is a list of plain text records, compressed with gzip. See
 [Inventory Generation](http://juliadocs.org/DocumenterInterLinks.jl/stable/write_inventory/)
 for details on these records.
@@ -415,6 +417,9 @@ executable to be available in `PATH` or to be passed as the `node` keyword.
 
 **`highlightjs`** file path to custom highglight.js library to be used with prerendering.
 
+**`inventory_version`** a version string to write to the header of the
+`objects.inv` inventory file. This should not include a `v` prefix.
+
 # Default and custom assets
 
 Documenter copies all files under the source directory (e.g. `/docs/src/`) over
@@ -476,6 +481,7 @@ struct HTML <: Documenter.Writer
     size_threshold_warn :: Int
     size_threshold_ignore :: Vector{String}
     example_size_threshold :: Int
+    inventory_version ::  String
 
     function HTML(;
             prettyurls    :: Bool = true,
@@ -504,6 +510,7 @@ struct HTML <: Documenter.Writer
             # seems reasonable, and that would lead to ~80 KiB, which is still fine
             # and leaves a buffer before hitting `size_threshold_warn`.
             example_size_threshold :: Union{Integer, Nothing} = 8 * 2^10, # 8 KiB
+            inventory_version :: Union{AbstractString,VersionNumber} = "",
 
             # deprecated keywords
             edit_branch   :: Union{String, Nothing, Default} = Default(nothing),
@@ -558,7 +565,8 @@ struct HTML <: Documenter.Writer
         new(prettyurls, disable_git, edit_link, repolink, canonical, assets, analytics,
             collapselevel, sidebar_sitename, highlights, mathengine, description, footer,
             ansicolor, lang, warn_outdated, prerender, node, highlightjs,
-            size_threshold, size_threshold_warn, size_threshold_ignore, example_size_threshold,
+            size_threshold, size_threshold_warn, size_threshold_ignore,
+            example_size_threshold, string(inventory_version),
         )
     end
 end
--- a/src/html/write_inventory.jl
+++ b/src/html/write_inventory.jl
@@ -14,7 +14,8 @@ function write_inventory(doc, ctx)
 
     @info "Writing inventory file."
     project = doc.user.sitename
-    version = ""  # TODO: https://github.com/JuliaDocs/Documenter.jl/issues/2385
+    version = ctx.settings.inventory_version
+    # if `version` is empty, it will be patched during `deploydocs`
 
     io_inv_header = open(joinpath(doc.user.build, "objects.inv"), "w")
 
--- a/test/deploydocs.jl
+++ b/test/deploydocs.jl
@@ -1,4 +1,5 @@
 using Documenter: Documenter, deploydocs, git
+using DocInventories: DocInventories, Inventory
 using Test
 include("TestUtilities.jl"); using Main.TestUtilities
 
@@ -20,7 +21,13 @@ Documenter.authenticated_repo_url(c::TestDeployConfig) = c.repo_path
             full_repo_path = joinpath(pwd(), "repo")
             # Pseudo makedocs products in build/
             mkdir("build")
-            write("build/page.html", "...")
+            write(joinpath("build", "page.html"), "...")
+            inventory = Inventory(
+                project="Documenter deploydocs test",
+                version="",
+            )
+            objects_inv = joinpath("build", "objects.inv")
+            DocInventories.save(objects_inv, inventory)
             # Create gh-pages and deploy dev/
             @quietly deploydocs(
                 root = pwd(),
@@ -28,6 +35,7 @@ Documenter.authenticated_repo_url(c::TestDeployConfig) = c.repo_path
                 repo = full_repo_path,
                 devbranch = "master",
             )
+            DocInventories.save(objects_inv, inventory)  # reset `version`
             # Deploy 1.0.0 tag
             @quietly deploydocs(
                 root = pwd(),
@@ -35,6 +43,7 @@ Documenter.authenticated_repo_url(c::TestDeployConfig) = c.repo_path
                 repo = full_repo_path,
                 devbranch = "master",
             )
+            DocInventories.save(objects_inv, inventory)  # reset `version`
             # Deploy 1.1.0 tag
             @quietly deploydocs(
                 root = pwd(),
@@ -42,6 +51,7 @@ Documenter.authenticated_repo_url(c::TestDeployConfig) = c.repo_path
                 repo = full_repo_path,
                 devbranch = "master",
             )
+            DocInventories.save(objects_inv, inventory)  # reset `version`
             # Deploy 2.0.0 tag, but into an archive (so nothing pushed to gh-pages)
             @quietly deploydocs(
                 root = pwd(),
@@ -50,6 +60,7 @@ Documenter.authenticated_repo_url(c::TestDeployConfig) = c.repo_path
                 devbranch = "master",
                 archive = joinpath(pwd(), "ghpages.tar.gz"),
             )
+            DocInventories.save(objects_inv, inventory)  # reset `version`
             # Deploy 3.0.0 tag with a tag_prefix---which does not change deployment behavior
             @quietly deploydocs(;
                 root = pwd(),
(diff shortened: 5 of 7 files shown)


mortenpi:
For what it's worth, for testing, I would factor the bit in `deploydocs` into a function that takes the path to the inventory and the new version number as an argument, and then updates the file. Then we could have like 1 or 2 reference cases in the repo that we unit test against.

goerz:
I don't think I can continue with this approach in good conscience.

With the `Projects.toml` fallback, this PR is functionally equivalent to #2442, as the tag for a tagged version should always match the version in `Project.toml` (since that's where TagBot [ultimately gets it from](https://github.com/JuliaDocs/Documenter.jl/pull/2442#issuecomment-1926289407)).

Without the `Project.toml` fallback ([Morten's current preference](https://github.com/JuliaDocs/Documenter.jl/pull/2442#issuecomment-1926219443)), there's no way I can think of to meet the requirements in https://github.com/JuliaDocs/Documenter.jl/pull/2424#issuecomment-1926328218 / https://github.com/JuliaDocs/Documenter.jl/pull/2442#issuecomment-1926289407 since there's no way to obtain a `version` for any `dev` deployment.

I also really don't like that the inventory isn't complete after `makedocs`: I routinely build the docs locally, where `deploydocs` never runs. Still, I should be able to use the inventory that comes out of that for other purposes. For example, it would mess with the instructions in [What if I want to link to a project that does not provide an inventory file?](https://juliadocs.org/DocumenterInterLinks.jl/stable/howtos/#howto-manual-inventory), which in the future would be "If the project you want to link to uses an old version of Documenter that doesn't generate inventories, clone it, instantiate the `docs` environment, update Documenter, run `include("docs/make.jl")`, convert the resulting inventory to `.toml`, and copy it into your own `docs/src/inventories`." If the `objects.inv` isn't complete until after `deploydocs`, I'd have to augment that with "manually fill in the version" (which nobody is going to remember to do). Well, I guess I'd end up writing a convenience function for the entire workflow and add that to `DocInventories`. But you get my point.

I don't think it's so terrible if the inventory has its own logic on where to get its `version` metadata from, even if it's not the "authoritative source" used for the selection menu. We don't even need to expand on this in any detail in the documentation: "a plain text header that includes the project name […] and a project `version` obtained from the `Project.toml`" is all we need to say, and #2442 really just does the right thing without much fuss.

I should have started with just reading the `Project.toml`, instead of a `version`/`inventory_version` keyword from `makedocs`/`HMTL`.

Anyway, any thoughts on this from @fredrikekre or @odow? I'm still open to fresh ideas. This is certainly not to just push through my preferred solution. Maybe I'm overlooking something fundamental, but having had a few nights to sleep on this, I've only gotten more and more convinced that #2442 is the right thing to do. It also happens to be the simplest solution, with no change to Documenter's existing API.

goerz:
There is one potential advantage of setting the `version` in `deploydocs`, though: it might work better out of the box for the Julia project itself, see https://github.com/JuliaDocs/Documenter.jl/pull/2442#issuecomment-1930956278

goerz:
And just to be clear: if someone can tell me how we can get a `version` for `dev`-deployments, I'm still willing to give this PR another go.

mortenpi:
I think this was said somewhere, but to double check: does the version have to be a version number (format-wise)? It can't be a random string (e.g. `dev`, to match the version selector)?

And if not, could we just leave it empty for _dev_ docs? You'd normally link against `stable/` anyway, which _would_ have a good version number in the inventory?

Strictly speaking, `dev/` docs don't have a well-defined version (you don't know if it's going to be a patch, minor, or major bump[^1]). For packages that _are_ strict about maintaining `-DEV` versions in the `Project.toml`, how about adding a `dev_version` option to `deploydocs` that allows the user to explicitly set the version, and we provide a function to also automatically determine it from the `Project.toml` (e.g. `dev_version = Documenter.project_version()`)?

Yes, this would mean that getting a correct version in the inventory for dev versions is something the package authors would have to configure. But, for _dev_ versions, I don't feel like we have strong enough standard practice to safely assume that the `Project.toml` contains a meaningful version number.

[^1]: Well, `+DEV` on top of the last release would perhaps be systematic, but is also confusing, and not really practiced?

goerz:
> I think this was said somewhere, but to double check: does the version have to be a version number (format-wise)? It can't be a random string (e.g. dev, to match the version selector)?

It's mostly just metadata for human consumption. I don't think anything really *breaks* depending on the `version`. But in principle it's supposed to be a proper version number without a `v` prefix. The only place I'm aware of where it enters programmatically is how the tooltip is determined in Sphinx:

```
reftitle = _('(in %s v%s)') % (proj, version)
```

So with `dev`, the tooltip would read "(in Documenter vdev)".

> And if not, could we just leave it empty for dev docs?

Technically, sure, but definitely [not what I had in mind](https://github.com/JuliaDocs/Documenter.jl/pull/2424#issuecomment-1926328218).

> You'd normally link against stable/ anyway

I'm not sure that's true. For intra-org `@extrefs`, [I dynamically link dev-to-dev and stable-to-stable](https://github.com/JuliaQuantumControl/QuantumControl.jl/blob/8034e131760c06a220188105c37c5db5c0725e70/docs/make.jl#L20-L35)

> Strictly speaking, dev/ docs don't have a well-defined version

Every project will have its own policy on that – or lack of policy, in which case the `dev` version string is just going to be the previous release. But that's still ok; just as ok as `pkgversion` returning that version after a dev-install, at least. The `version` from `Project.toml`" is still the most "correct" information.

> For packages that are strict about maintaining -DEV versions in the Project.toml, how about adding a dev_version option to deploydocs that allows the user to explicitly set the version, and we provide a function to also automatically determine it from the Project.toml (e.g. dev_version = Documenter.project_version())?

I don't think that's necessary. We'd already have `inventory_version` in `makedocs`. No need for multiple options (and #2442 wouldn't need any options).

> Yes, this would mean that getting a correct version in the inventory for dev versions is something the package authors would have to configure.

Everyone deploys dev-builds and would have to set that up. Or, more likely, not set it up, so that we'd end up with a lot of inventories in the wild that lack full metadata. Which is why I was hoping to avoid everyone having to tweak their `docs/make.jl` file.

We should definitely make sure that Julia-nightly automatically gets the correct version number into their inventory. Julia has some kind of custom setup: they don't have "dev" in the versions-selection menu, but, e.g., `v1.11-dev`. With #2442 it definitely works (now that I also read the `VERSION` file). I'm 90% sure that setting it in `deploydocs` like we're trying to do here also works, but it's hard to know for sure since I can't easily run `deploydocs` locally. (One of the issues I have with setting the version in `deploydocs` instead of `makedocs`).

The most correct behavior for the inventory `version` really is "whatever is the project version", i.e., "whatever is in `Project.toml`", not "whatever the versions-selection-menu shows as a label". If you really want the version to be "whatever the versions-selection-menu shows", that's better than nothing, but then I would still keep inventory-writing features in `DocumenterInterLinks` as well that override the default Documenter behavior.

goerz:
I did the refactoring and completed the tests. This now behaves as @mortenpi requested:

* If an `inventory_version` is given as an option to `HTML`, use that. Otherwise, the inventory `version` is left blank in the `makedocs` step
* In `deploydocs`, if the inventory version is blank, patch in a version derived from deployment folder name. This will generally set the correct version for tagged releases, but result in a blank version for `dev`-deployments.

Since this does not fulfill the [minimum requirements](https://github.com/JuliaDocs/Documenter.jl/pull/2424#issuecomment-1926328218) of setting a version for `dev`-deployments, I do not consider this an acceptable solution, but at least we have a complete PR now to compare against #2442.

goerz:
@mortenpi I understand where you're coming from and wanting to avoid "multiple sources of truth", and I totally respect you having the final word on these architectural decisions.

There are technical reasons I don't like the approach in this PR very much: inventory isn't complete after `makedocs`, no `dev` versions, and just the code being a little convoluted. But I can probably get over those, or work around them somehow.

To boil down why I have a different perspective on the "multiple sources of truth" issue (reiterating some of my previous comments):

* The labels in the versions-selection-menu are just labels for different subfolders in the `gh-pages` branch. They happen to coincide with "version" because the tag names that determine the folder names are of the form "vX.Y.Z", but semantically they're not the "project version" that should go in the inventory (according to Sphinx' original semantics). We're just trying to recover the project version from the folder name by dropping the `v` prefix here.
* The *original* "source of truth" is the project's `Project.toml` file anyway (or a `VERSION` file in the case of the Julia project itself). For any tagged release, the label in the versions-menu *will* match the version in the `Project.toml`. I can't think of any normal situation where getting the inventory-`version` from the deploy folder name would give a conflicting result to getting it from `Project.toml`. Just `Project.toml` is strictly more informative, since for `dev`, the folder name no longer contains version information, so we can't recover version information from it.

I understand it's a judgement call, and no hard feelings, but I really hope I can bring you around to this perspective (and maybe @fredrikekre and @odow can still chime in to sway opinions).

mortenpi:
> The labels in the versions-selection-menu are just labels for different subfolders in the gh-pages branch.

So, I disagree with this pretty strongly. Semantically, they are versions. They just happen to be implemented via subdirectories and symlinks. The assumption that they are versions is baked in pretty strongly.

> The _original_ "source of truth" is the project's `Project.toml` file anyway (or a `VERSION` file in the case of the Julia project itself).

So, I don't actually disagree with this. But we are stuck with design decisions that predate `Project.toml`s. Arguably, we should maybe create an alternative `DeployConfig` that uses `Project.toml` to do the versioning.. but I am not sure we really want to go down the rabbit hole of `deploydocs` refactoring in the context of this PR.

The second related design decision that is causing us pain is the distinction between `makedocs` and `deploydocs`. What we want is `makedocs` to be able to know the version, but that is only determined once `deploydocs` runs, which happens after `makedocs`. Trying to duplicate or override the logic used by `deploydocs` in `makedocs` does not seem like good idea, and just adds too much complexity to the system as a whole.

So what I am trying to get to here is some solution that doesn't break the existing assumptions, but gives us a working documentation inventory that we could start using in the ecosystem. And then, at some point, we can fix it up when we overhaul the whole `make.jl` approach (my preference goes to declarative configuration; #1350, #2353).

> In deploydocs, **if the inventory version is blank**, patch in a version derived from deployment folder name.

My gut reaction is that _if `deploydocs` is deploying a tag_ (not dev version), it should unconditionally overwrite the version. Theoretically, unless the user has made a mistake in they `makedocs` call, the version should match the tag anyway, and it would ensure that any mistakes would get fixed by `deploydocs`. But also, maybe it's also okay for `deploydocs` to defer to the user here, and be like "sure, if that's what you want".

mortenpi:
One more note: I think I would nearly be okay with `deploydocs` using the version in the packages's `Project.toml` as the fallback to get the version number for dev versions.

However, I don't like the brittleness when it comes to edge cases, where Documenter is not used in the very common, but not universal `docs/Project.toml` way for a single package -- we don't really know which `Project.toml` to read:

* The docs environment (i.e. `docs/Project.toml`) doesn't have a version, so we can't use the active project.
* We could pick the version of a dependency of the environment, but we don't know which one is the "package".
* We could go up a directory and look for a `Project.toml`, and that would work like 95% of the time, but there's no particular reason to assume that the env one level up is actually the correct environment. And so having this type of automagical behavior also feels pretty icky to me -- generally I've been trying to get rid of that kind of stuff.[^1]

So just writing `vDEV` or something into the inventory seems better to me.

[^1]: Side note: if subprojects would link back to main projects, then it would be fine, but we don't have that metadata. (https://github.com/JuliaLang/Pkg.jl/issues/1233)

goerz:
> > The labels in the versions-selection-menu are just labels for different subfolders in the gh-pages branch.
> 
> So, I disagree with this pretty strongly. Semantically, they are versions. They just happen to be implemented via subdirectories and symlinks. The assumption that they are versions is baked in pretty strongly.

There's a part of this I agree with, and maybe that's the common ground we've been trying to get to: If the tag/branch/deployment folder is a valid version, then the "project version" (whether it's extracted from a `Project.toml` file or given by the user as an explicit `inventory_version` argument) *must* match it. There should never be a situation where the project version is different from the version shown in the versions-selection menu. I'd be happy to make that a hard rule and throw an `error` if there is a mismatch. I can do ([and have done](https://github.com/JuliaDocs/Documenter.jl/pull/2442#issuecomment-1937907509)) that even in #2442: Set the inventory version from `Project.toml`/`JuliaProject.toml`/`VERSION` in `makedocs`, and then *check* in `deploydocs` that that version matches the tagged version.

> > The _original_ "source of truth" is the project's `Project.toml` file anyway (or a `VERSION` file in the case of the Julia project itself).
> 
> So, I don't actually disagree with this. But we are stuck with design decisions that predate `Project.toml`s. Arguably, we should maybe create an alternative `DeployConfig` that uses `Project.toml` to do the versioning. but I am not sure we really want to go down the rabbit hole of `deploydocs` refactoring in the context of this PR.

I agree on not going down that rabbit hole right now, but I actually don't even think it would be a good idea: The versions in the version-selection menu *should* be derived from git tags, not from the project version in the `Project.toml`. Otherwise, if people forget to bump their project version after they'll make a release, they'll likely overwrite their "stable" deployment the next time they push to `master`.


> The second related design decision that is causing us pain is the distinction between `makedocs` and `deploydocs`. What we want is `makedocs` to be able to know the version, but that is only determined once `deploydocs` runs, which happens after `makedocs`. Trying to duplicate or override the logic used by `deploydocs` in `makedocs` does not seem like good idea, and just adds too much complexity to the system as a whole.

I agree with that, but I don't think the situation is that dire, if we just enforce that the version in `makedocs` must agree with the version in `deploydocs`. There isn't much duplication of logic in that.

> So what I am trying to get to here is some solution that doesn't break the existing assumptions, but gives us a working documentation inventory that we could start using in the ecosystem. And then, at some point, we can fix it up when we overhaul the whole `make.jl` approach (my preference goes to declarative configuration; #1350, #2353).
> 
> > In deploydocs, **if the inventory version is blank**, patch in a version derived from deployment folder name.
> 
> My gut reaction is that _if `deploydocs` is deploying a tag_ (not dev version), it should unconditionally overwrite the version.

As a general rule, I'm allergic to automatic values silently overwriting values that have been explicitly set by a human. But I am okay with throwing an error if `deploydocs` detects that the user gave an unambiguously wrong version.

> Theoretically, unless the user has made a mistake in they `makedocs` call, the version should match the tag anyway, and it would ensure that any mistakes would get fixed by `deploydocs`. 

I agree with that 100%, but let's just throw an error and let the user fix their mistake.

> But also, maybe it's also okay for `deploydocs` to defer to the user here, and be like "sure, if that's what you want".

Although my general rule is "a human should always have the last word", I don't think we need to allow a situation where the versions-menu shows "v1.0", but the user manually sets the inventory version to "v0.1". That's clearly a mistake.


> One more note: I think I would nearly be okay with `deploydocs` using the version in the packages's `Project.toml` as the fallback to get the version number for dev versions.

I've added another #2448 that is just #2443 with the additional fallback.

> However, I don't like the brittleness when it comes to edge cases, where Documenter is not used in the very common, but not universal `docs/Project.toml` way for a single package -- we don't really know which `Project.toml` to read:

I don't think that part is very tricky. The way I've implemented this is pretty straightforward:

* Start from the current working directory (seems to always be `docs/`, but if I overlooked something and there is some possibility that `pwd` during the build is not a subfolder of `docs`, we can change that to the `docs/src` folder, which Documenter knows the absolute path of)
* Check for files `Project.toml`, `JuliaProject.toml`, `VERSION` (what Julia itself uses), in that order. **Ignore TOML files that do not define a `version`** (Only the main `Project.toml` should define `version`, and `docs/Project.toml` specifically generally does not contain a `version`)
* If no version could be extracted from any of the files, go up one directory, and repeat until we reach the root of the file system
* If we reach the root of the file system without finding a file that specifies a project version, show a warning and return an empty string

So, basically, we identify the main `Project.toml` file as the first one that defines a `version`. It doesn't have to be in the current directory, or one level up, specifically.

This should work for at least 99% of projects. It should work for monorepos, too (although I haven't found any monorepo setups that use Documenter and deploy more than just the documentation of their "main" package). The most common situation where it might fail is for independent documentation repos (where the documentation is in a separate repo from the package). These repos currently might have a `Project.toml` in their root directory that does not define a `version`. Such setups would now see a warning when they do a dev-deployment, and an error if they deploy a tagged version. They'll have several options to fix their setup: define a `version` in the main `Project.toml` matching the version of the project that's being documented, use a `VERSION` file (like Julia does), or manually set the inventory version in their `make.jl` file before `deploydocs` is called.

I could also imagine that some repos might have a `version` in `docs/Project.toml` by accident (which would then take precedence over the main `Project.toml`), but they'll see an error message so that they can remove the stray `version`.

So in the worst case, people will get yelled at that they need to fix their setup and have a `version` somewhere that matches the tagged version they're deploying. In no case would there be a "wrong" version either in the inventory or in the menu.

So now, we have three competing PRs on top of #2424 to deal with the version setting

* #2442 – `makedocs` gets the version for the inventory from the closest `Project.toml`/`JuliaProject.toml`/`VERSION` that defines a version. Then `deploydocs` checks whether that version matches the tag for the release that's being deployed (for non-`dev` deployments). There is no need for a manual `inventory_version` option. I'd be extremely happy with this option!
* #2448 – The inventory version is set by `deploydocs`. For tagged releases, it uses the version from the tag, and for `dev`-releases, it falls back to the same logic as #2442 to get the version from the closest `Project.toml`. **The ultimate result is identical to #2442** (at least that's the intent), but the code is a bit more convoluted than necessary, and I'm not enthusiastic that the `objects.inv` file isn't generally complete after the `makedocs` step.
* #2443 – This PR, which sets the version in `deploydocs` without a fallback. As a result, `dev`-deployments would typically have an empty `version` in the inventory, and we'd have to push people to modify their setup and specify a manual `inventory_version`. I'd find this one hard to swallow.

@mortenpi Let me know if you have further questions to help you decide between these options, or if there is anything that requires tweaking or refactoring in any one of the three PRs. But besides that, tell me which one of these you decide on, and I'll merge that into #2424 and wrap everything up with a CHANGELOG entry and such.

goerz:
I turned https://github.com/JuliaDocs/Documenter.jl/pull/2442#discussion_r1485701646 into a fourth candidate PR:

* #2449 – Initially set the inventory `version` in `makedocs` based on the closest `Project.toml`/`JuliaProject.toml`/`VERSION` file that defines a `version`. Then unconditionally overwrite the `version` in `deploydocs` for a tagged release (where `deploydocs` can know the correct `version`). This overwrite shouldn't actually *change* the version for all normal setups, but it'll still let `deploydocs` have the last word.

Having slept on this variation another night, it seems like a pretty good compromise, as it takes @mortenpi comment to heart that

> My gut reaction is that if deploydocs is deploying a tag (not dev version), it should unconditionally overwrite the version. Theoretically, unless the user has made a mistake in they makedocs call, the version should match the tag anyway, and it would ensure that any mistakes would get fixed by deploydocs.

This is under the assumption that

> I think I would nearly be okay with deploydocs using the version in the packages's Project.toml as the fallback to get the version number for dev versions

is ultimately something you can bring yourself around to. Without any fallback, we'll just have to leave the `version` blank for `dev`-deployments, and I'll just have to be unhappy ;-)

goerz:
Closing in favor of https://github.com/JuliaDocs/Documenter.jl/pull/2449#issuecomment-1958774248

PR #2448, opened 2024-02-11
Patch inventory version in `deploydocs` with fallback to `Project.toml`

goerz:
This is the same as #2443, with one additional commit where I add the fallback to the closest `Project.toml` for `dev`-deployments.

Functionally, this behaves identically to ~~the [current revision](https://github.com/JuliaDocs/Documenter.jl/pull/2442#issuecomment-1937907509) of #2442~~ #2449. Both PRs set the inventory version from `Project.toml` for `dev` and guarantee that the version for a tagged release matches what is shown in the versions-selection-menu.

The only difference is this PR does everything in `deploydocs` instead of `makedocs`. I'm more enthusiastic about that, since it would be nice to be able to use an `objects.inv` file that comes out of `makedocs` for other purposes without having to worry about the `version`.

For those use cases, this PR still has the manual `inventory_version` option in `HTML`, which isn't necessary in #2442 or #2449.


goerz:
I don't think this PR has any benefits over #2449 (trying to see it from [Morten's point of view](https://github.com/JuliaDocs/Documenter.jl/pull/2443#issuecomment-1937383254), that is; certainly it has no benefits from *my* perspective). So I would tentatively close this in favor of #2449.

PR #2449, opened 2024-02-12
Overwrite inventory version in `deploydocs`

goerz:
This unconditionally overwrites an existing inventory `version` with the version determined by `deploydocs`. This only happens when `deploydocs` actually knows the version. For `dev`-deployments, an existing `version` is left unchanged.

Under normal circumstances, the existing `version` will already be the correct one, so this should do nothing. If the `deploydocs` actually has to *change* the version, that would indicate that something is set up incorrectly, and thus a warning will be printed (but no `error`)

This PR is just #2442 with the tweak discussed in https://github.com/JuliaDocs/Documenter.jl/pull/2442#discussion_r1485701646. This makes #2442 less breaking: repos that are set up to contain only the documentation for some other package but don't declare the version of that package in some `Project.toml`/`VERSION` file  won't get an `error`, but instead just a `@warn`. Their deployment will continue to work without them having to make any changes to their setup.

It satisfies @mortenpi's desire that `deploydocs` should have the last word and unconditionally patch in the `version` if it can determine a `version` from the git tags / the deployment folder name. Ultimately, this PR has the same final outcome as #2448, but the "fallback" is already in `makedocs` so that the inventory has the correct `version` already after `makedocs`. That also means there is no need for a manual `inventory_version` option to `HTML`.

I'd be quite happy with this compromise.


goerz:
Per the discussion on Slack, I've revised this so that it takes a `version` only from a `Project.toml` that is in the parent folder of the documentation root (`docs`). This is to ensure that the autodetection is restricted to projects with a standard setup. For non-standard setups, e.g., documentation-only repositories or monorepos, `makedocs` may emit a warning asking that an `inventory_version` should be set manually, but it will leave the `version` blank.

A manual inventory version can still be set by passing an explicit `inventory_version` to the `HTML()` options of `makedocs`. Moreover, an explicit `inventory_version=""` opts out of setting a version in `makedocs` altogether (and suppresses any warnings that would be generated otherwise).

In all cases, `deploydocs` will unconditionally overwrite the inventory version when deploying a tagged release. If the version in `deploydocs` contradicts a manually set / autodetected version, it will emit a warning, but still win out.

PR #2453, opened 2024-02-19
Refactor `crossref`/`xref` for resolving `@ref` links

goerz:
This is a pure refactoring of the `crossref` function and related functions used for resolving `@ref` links in `cross_references.jl`.

It only affects internal code and should have no user-facing effects, other than maybe a slightly different format of the error message when an `@ref` link cannot be resolved. The refactoring simplifies the flow of logic, which previously used a lot of backtracking (["call `xref`, then figure out that the link actually isn't an `@ref`, and backtrack to `local_links!`"](https://github.com/JuliaDocs/Documenter.jl/blob/a6cdf2038d869ef8cb8d0cb697c82dbb6c4e987e/src/cross_references.jl#L131-L136)). It makes the previous [`basicxref `](https://github.com/JuliaDocs/Documenter.jl/blob/a6cdf2038d869ef8cb8d0cb697c82dbb6c4e987e/src/cross_references.jl#L179) obsolete. The logic for "basic" and "non-basic" links is now that both `[test](@ref slug)` and `[slug](@ref)` first get normalized to a `slug`, and then that `slug` gets resolved by the same function.

The main change is that all `@ref` links are now resolved with a `Selector`-based pipeline. The PR adds [full documentation to describe the internals of this](https://documenter.juliadocs.org/previews/PR2453/lib/internals/crossref/).

The motivation for the pipeline approach (apart from just cleaning up the code) is that it allows plugins to extend that pipeline, providing the ability to resolve `@ref` links that `Documenter` ordinarily would not be able to resolve.

The obvious candidate here is for [DocumenterInterLinks](https://github.com/JuliaDocs/DocumenterInterLinks.jl) to provide a fallback for resolving `@ref` links with links to external projects. This is by no means intended as a replacement  for the `@extref` links that are the core of `DocumenterInterLinks`. But it comes in handy if e.g. in the `QuantumControl.jl` documentation I want to include a particular single docstring of a re-exported function from `QuantumControlBase.jl`, and that function has `@ref`-links that resolve in `QuantumControlBase`. Maybe the `@ref` link is even to a section header in `QuantumControlBase`, and there would be no way to make that resolvable in `QuantumControl`.

I might also want to experiment a bit with extensions to the `@ref`-resolution that get around Documenter's restrictions that an `@ref` link to another docstring must have a local binding in the module where the link is defined. I regularly brush up against this, and end up having to import symbols just so that I can link to them. Sometimes that's not possible, as it would create circular references. So, I might want to define myself a (project-local) plugin that relaxes this to "look for bindings anywhere in the project".

### DIFF
--- a/docs/src/lib/internals/crossref.md
+++ b/docs/src/lib/internals/crossref.md
@@ -1,0 +1,10 @@
+```@meta
+CollapsedDocStrings = true
+```
+
+# Crossrefs
+
+```@autodocs
+Modules = [Documenter, Documenter.XRefResolvers]
+Pages = ["cross_references.jl"]
+```
--- a/docs/src/lib/internals/deploydocs.md
+++ b/docs/src/lib/internals/deploydocs.md
@@ -2,13 +2,10 @@
 CollapsedDocStrings = true
 ```
 
-# Documenter
+# Deploydocs
 
 ```@docs
 Documenter.gitrm_copy
 Documenter.git_push
 Documenter.user_host_upstream
-Documenter.find_object
-Documenter.xrefname
-Documenter.crossref
 ```
--- a/src/cross_references.jl
+++ b/src/cross_references.jl
@@ -1,8 +1,15 @@
 """
 $(SIGNATURES)
 
-Traverses a [`Documenter.Document`](@ref) and replaces links containing `@ref` URLs with
-their real URLs.
+Traverses a [`Documenter.Document`](@ref) and process internal links and
+references.
+
+- Links containing `@ref` URLs are replaced with their real URLs. This
+  delegates to [`xref`](@ref), which in turn delegates to the
+  [`XRefResolvers.XRefResolverPipeline`](@ref).
+- Links to local (`.md`) documents are rewritten to link to the corresponding
+  path in the output build.
+- For links to local files or images, check that the linked files exist.
 """
 function crossref(doc::Documenter.Document)
     for (src, page) in doc.blueprint.pages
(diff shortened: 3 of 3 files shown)


goerz:
The documentation of the new logic is at https://documenter.juliadocs.org/previews/PR2453/lib/internals/crossref/, and in the docstring for [`Documenter.XRefResolvers.XRefResolverPipeline`](https://documenter.juliadocs.org/previews/PR2453/lib/internals/crossref/#Documenter.XRefResolvers.XRefResolverPipeline) in particular

mortenpi (approved):
For what it's worth, I am not entirely sure the pipeline logic is quite what we want to double down on for the sort of plugin architecture. But I am not sure what a better architecture would look like either right now.

So I'm happy with this, as long as we sort of keep it as "not quite official public API" (i.e. same as extending the other pipelines).

goerz:
The general concept of having processing pipelines that plugins can hook into seems fairly important. The `Selectors` implementation specifically is a bit unusual. It feels like someone (it predates *you*, right?) was very enamored with multiple dispatch and wanted to use it as the hammer for this particular nail. I don't have a problem with it, though. The `Selectors.order` could be a little iffy, since that might require that different plugins coordinate with each other to agree on a number, or else run in an undefined order (which probably isn't a real problem, either).

So while we have `Selectors`, we should probably use that consistently. It's really the only way for plugins to hook into the system, but at the same time, we haven't officially declared any *stable* API for plugins within the Documenter 1.0 series. If we want to push for a lot more people (people who aren't also involved in Documenter itself) to write plugins, we probably *should* define an official API at some point, and that would be an opportunity to maybe change the pipeline implementation. I have no objections to the ["Data-Advising"](https://tecosaur.github.io/DataToolkitDocs/base/stable/advising/#Data-Advising) in principle, although it also seemed like a fairly abstract system that I couldn't immediately wrap my head around.

If I had been the original author of Documenter, the pipeline would probably be a very boring internal list of function wrappers, and there would be `Documenter.register_plugin` or `Documenter.register_xref_resolver` functions that would update the pipeline (pretty much how Sphinx handles plugins). But I really have no strong preferences on what system we should have, and I'll adapt to anything.

Of course, if we do change the pipeline implementation, existing plugins will have no choice but to adapt. That's just the way things are right now. Plugins are based on an unstable API, and anyone who writes plugins will just have to deal with that until we decide to make it official.

Anyway, so was there anything you wanted me to change in this PR, or should I click the merge button?

mortenpi:
> The general concept of having processing pipelines that plugins can hook into seems fairly important.

I agree that we need a good system. I just have some gripes about the current implementation. Specifically, off the top of my head:

1. Using the Julia type system for extensions. It's basically relying on global state that gets set when you load packages, rather than something that you enable explicitly for each `makedocs` call. We have the interface of passing positional plugin objects to `makedocs`, and we should use that. But we need some different interface for determining what gets run and what doesn't.

2. I also don't like the hard-coded numbers that decide the order in which pipeline steps run, since they're ambiguous and can conflict. This is a tricky problem, since you sometime have to decide to run certain things in a specific order. But maybe the order of the positional arguments in `makedocs` can break the ambiguity.

> the pipeline would probably be a very boring internal list of function wrappers

I also think we should have more explicit "entry points" that the user can overload. My main question is whether there is some nice general design here that could be re-used everywhere-ish (hopefully, to make it easier to write plugins as a consequence), or do we just do ad hoc stuff for each point a plugin might need.

> It's really the only way for plugins to hook into the system, but at the same time, we haven't officially declared any stable API for plugins within the Documenter 1.0 series.

Yep. As I said, I am completely happy with this refactor. It's just that when we do design a better API, we probably also want to change up how hooking into Documenter works (which will likely break DocumenterInterlinks).

> should I click the merge button?

Let me answer that with a button press.

PR #2457, opened 2024-02-20
Allow customizing domain in deploydocs

GiggleLiu:
I find using GitHub "Custom domain" setting, which generates a `CNAME` file in the root folder, is not a reliable:
- after `forcepush`, the CNAME file generated by github pages will disappear.
- by setting `devurl=""`, the CNAME file will be removed by Documenter.

As a result, the customized domain will be reset to the default value repeatedly.

The solution could be either carefully skip removing the `CNAME` file, or generate the `CNAME` with Documenter (this PR). This PR is based on my ad hoc solution here:
https://github.com/GiggleLiu/ScientificComputingForPhysicists/blob/981e311ef7f7902bd4a77874126a1b0d3ed02975/docs/make.jl#L58
where I tried to set `devurl=""` to deploy a book.


goerz (approved):
Looks quite sensible to me!

mortenpi:
I kinda want to think about it a bit. Minimally, this is very GitHub-specific, and we should call it `github_cname` or or just `cname` or something. Importantly, it's not a URL -- it's just the domain name that goes into `CNAME` (as a user, I would intuitively set `siteurl = "https://documenter.juliadocs.org/stable"`).

Generally, Documenter shouldn't muck with the files in the top-level directory. On the other hand, it does kinda take ownership of anything that is in the directory that contains all the versions. So having Documenter (optionally) maintain this file seems also reasonable. But I want to understand the use case a bit.

> * by setting `devurl=""`, the CNAME file will be removed by Documenter.

This would indeed happen I believe, but I think we should just disallow `devurl=""`.

> * after `forcepush`, the CNAME file generated by github pages will disappear.

It shouldn't. The only case I can think of is when it's combined with `devurl=""`. Is that the use case here?

GiggleLiu:
> I kinda want to think about it a bit. Minimally, this is very GitHub-specific, and we should call it `github_cname` or or just `cname` or something. Importantly, it's not a URL -- it's just the domain name that goes into `CNAME` (as a user, I would intuitively set `siteurl = "https://documenter.juliadocs.org/stable"`).

Good point.

> > * after `forcepush`, the CNAME file generated by github pages will disappear.
> 
> It shouldn't. The only case I can think of is when it's combined with `devurl=""`. Is that the use case here?

I am also not sure, it may because I combined them.

> This would indeed happen I believe, but I think we should just disallow devurl="".

If we disallow `devurl=""`, is there an alternative way to host the documents on the top level domain name? e.g. this is what I am doing right now:
https://book.jinguo-group.science/

mortenpi:
> If we disallow `devurl=""`, is there an alternative way to host the documents on the top level domain name? e.g. this is what I am doing right now:
> https://book.jinguo-group.science/

Not really.. technically, Documenter is not designed with that use case in mind. But I think there is no reason not to keep it working.

I am also fine with `deploydocs` managing this file, let's just call the option `cname` :slightly_smiling_face:

GiggleLiu:
I see you point

> > If we disallow `devurl=""`, is there an alternative way to host the documents on the top level domain name? e.g. this is what I am doing right now:
> > https://book.jinguo-group.science/
> 
> Not really.. technically, Documenter is not designed with that use case in mind. But I think there is no reason not to keep it working.
> 
> I am also fine with `deploydocs` managing this file, let's just call the option `cname` 🙂

I see your point, then we should forbid an empty version number. I also added a sanity check for the `versions` as follows:

```julia
@assert !any(isempty ∘ first, versions) "Empty version number detected: $versions, which may cause unexpected behavior such as overwriting GitHub custom domain settings."
```
or do think it should be a warning instead?

Meanwhile, I think allowing users to specify the `CNAME` is sometimes convenient for those who are not familiar with GitHub pages settings.

docs/src/man/hosting.md
```diff
@@ -504,6 +504,10 @@ https://USER_NAME.github.io/PACKAGE_NAME.jl/vX.Y.Z
 https://USER_NAME.github.io/PACKAGE_NAME.jl/dev
 ```
 
+!!! tip
+    A custom domain can be set up with the keyword argument `cname` to
+    [`deploydocs`](@ref), which will then be used as the base URL for the documentation.
```
mortenpi, line 509:
```suggestion
!!! tip
    If you need Documenter to maintain [a `CNAME` file](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site) for you can use the `cname` argument of [`deploydocs`](@ref) to specify the domain.
```

src/deploydocs.jl
```diff
@@ -215,6 +221,9 @@ function deploydocs(;
         ispath(archive) && error("Output archive exists: $archive")
     end
 
+    # version sanity check
+    @assert !any(isempty ∘ first, versions) "Empty version number detected: $versions, which may cause unexpected behaviors such as overwriting GitHub custom domain settings."
+
```
mortenpi, line 226: We don't want this behavior change right now, right?

```suggestion
```

mortenpi (approved):
LGTM. Thank you!

Issue #2430, opened 2024-01-26
New search window does not appear [intermittent] [bug] [regression]


LilithHafner:
Steps to reproduce

1) Open a new private browsing session
2) Open https://clemens-cords.com/mousetrap/
3) Click "search docs"
4) Type "add_render_task!"
5) Hover mouse over the label "Section" to the right of the first result
6) Click
7) Wait for the page to fully load
8) Click "search docs"
9) Observe that the search pane does not appear.
10) Refresh the page
11) Repeat steps 9 & 10

I've reproduced this repeatedly on IceCat, FireFox, and Chromium browsers, but with only about a 50% success rate. This is an obnoxious kind of bug: very bad for UX and hard to reproduce.

mortenpi:
Ironically, I couldn't repro it with these instructions, but I have definitely run into the case where the search window doesn't open for some reason. cc @Hetarth02

thofma:
Some people (including me) stumbled over this, even after closing and reopening the window.

joschmitt:
I really don't know this stuff, but I could reproduce the problem in locally build docs and got the following error message in the "developer tools" of firefox:
```
Uncaught TypeError: document.querySelector(...) is null
    <anonymous> http://localhost:8000/assets/documenter.js:683
    execCb https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    check https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    enable https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    bind https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    emit https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    each https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    emit https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    check https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    enable https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    init https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
[...]

PR #2458, opened 2024-02-20
Fixes search window does not appear

Hetarth02:
This PR aims to fix Issue #2430.


goerz:
This kind of PR would really benefit from having a `preview` build of the documentation. That's only possible for PRs where the branch is directly in the main repository, though, right?

@mortenpi Is there some other way to allow `preview` builds when the PR comes from someone without write permissions to the repository? Maybe by applying some tag/label to the PR?

mortenpi:
> @mortenpi Is there some other way to allow `preview` builds when the PR comes from someone without write permissions to the repository? Maybe by applying some tag/label to the PR?

It's not obvious how to do this. PRs don't have access to secrets. Maybe we could have the workflow upload an artifact, and then a label to pull and unpack that to `gh-pages` on-demand (by a maintainer).

mortenpi (approved):
LGTM, thanks! The original issue isn't very easy to repro, but I think this should fix it.

Issue #2459, opened 2024-02-20
Console shows error when metadata is not present in certain pages


Hetarth02:
![image](https://github.com/JuliaDocs/Documenter.jl/assets/80420115/a11faefe-bc3f-4cbc-b6b8-967d05f04a34)

Hetarth02:
@mortenpi This is resolved in [2460](https://github.com/JuliaDocs/Documenter.jl/pull/2460). We should close this.

PR #2460, opened 2024-02-20
Fixes console error when metadata is not present in certain pages

Hetarth02:
This PR aims to fix Issue #2459.


assets/html/js/metadata.js
```diff
@@ -4,10 +4,13 @@
...
   let meta = $("div[data-docstringscollapsed]").data();
 
-  if (meta.docstringscollapsed) {
-    $("#documenter-article-toggle-button").trigger({
-      type: "click",
-      noToggleAnimation: true,
-    });
+  // Check if metadata object exists
+  if (meta) {
+    if (meta.docstringscollapsed) {
```
mortenpi, line 9: This might be cleaner?

```suggestion
    if (meta?.docstringscollapsed) {
```

Issue #2441, opened 2024-02-03
Typing `th` into the Documenter.jl search bar throws a javascript error


LilithHafner:
I observe this error being thrown on https://documenter.juliadocs.org/dev/ but not https://documenter.juliadocs.org/stable/. I'm guessing it is due to buggy escaping of search query results. 
![Screenshot from 2024-02-03 12-09-59](https://github.com/JuliaDocs/Documenter.jl/assets/60898866/80fc4498-472b-4d62-936d-dfdc1f6d922f)

mortenpi:
#2461

Issue #2428, opened 2024-01-25
`&`-s not escape in search results


mortenpi:
I think we directly pass the `&`-s to the generated dynamically generated HTML, which means they get interpreted as character references sometimes. We should escape them, either when constructing the HTML, or maybe even during index generation (although I am not sure this actually makes sense).

X-ref for more details for an "MWE": https://github.com/JuliaLang/julia/pull/52078

fingolfin:
I was confused by this issue at first, because I thought this was about `&` occurring in the search string entered by the user. But really this is about `&` occurring in the data being searched. The example from PR #2461 is instructive: insert
```
### &harr; &harr; &harr;
```
into the Markdown source (then `&harr;` is rendered as &harr;), then search for the text " harr" -- it will match those arrows.

fingolfin:
One way to deal with it would be to turn all those entities into Unicode first. I've implemented that recently for the Markdown stdlib. In fact perhaps that helps with the issue? See https://github.com/JuliaLang/julia/pull/60629

PR #2461, opened 2024-02-20
Fix escaping special html entites in search output

Hetarth02:
This PR aims to close Issue #2441 <s> and  Issue #2428 </s>.

**2024-02-24 Edit:** Fixes only Issue #2441.


assets/html/js/search.js
```diff
@@ -67,7 +67,10 @@ update_search
 
 function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
   importScripts(
-    "https://cdn.jsdelivr.net/npm/minisearch@6.1.0/dist/umd/index.min.js"
+    ...[
+      "https://cdn.jsdelivr.net/npm/minisearch@6.1.0/dist/umd/index.min.js",
+      "https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.21/lodash.min.js",
```
mortenpi, line 72: Hmm. This now duplicates the dependency.. so let's not include `lodash` here (see also https://github.com/JuliaDocs/Documenter.jl/pull/2415#discussion_r1477029876). The implementation is trivial, so let's just remove the comments, and put the implementation here (with appropriate attribution; lodash is MIT).

https://github.com/lodash/lodash/blob/main/src/escape.ts

mortenpi (changes requested):
I can confirm that it fixes #2441, but not #2428:

image

To repro, I just added a 

```
### &harr; &harr; &harr;
```

heading into the Markdown source.

Hetarth02:
> I can confirm that it fixes #2441, but not #2428:
> 
> image
> 
> To repro, I just added a
> 
> ```
> ### &harr; &harr; &harr;
> ```
> 
> heading into the Markdown source.

## Case 1

I looked into this a little bit and I found out that, when enter into `.md` files and render

```md
### ↔ ↔ ↔
```

I cannot find it in the search using `harr` or `↔`.
But it renders in html as the symbol.

> For ref, see below images.

### HTML Renders
image

image

### Search
image

image

## Case 2

But when I enter below into `.md` files and render

```md
### &harr; &harr; &harr;
```

I can find it in the search using `harr` but not using `↔`.
Also it does not renders in html as the symbol.

> For ref, see below images.

### HTML Renders
image

image

### Search
image

image

mortenpi (approved):

PR #2470, opened 2024-03-03
Resolve `@ref` bindings in current module and `Main`

### LINKED ISSUE #2462: Inter-Module at-refs?
Hello,

I've started using `Documenter` yesterday and I'm afraid that my experience has been most frustrating so far. I'm trying to document a package I write at work that consists of a bunch of submodules, where the docs in the submodules refer to each other at some points. I *want* these refs to be actual URLs (not just verbatim text) so users of the docs can find the relevant bits easier. I've condensed my problem into a MWE, hosted here: https://github.com/har7an/julia-documenter-inter-module-docs

Maybe I'm just using it wrong and I missed a crucial step in all of this, but I can't get the docs in module `C` to reference `B.Blarg`. I build the docs in that project with these commands:

```bash
$ cd docs
$ julia --project make.jl
```

As you will see in the output I get a lot of warnings. In the finished docs, none of the 5 at-refs I tried works and I don't understand what's wrong. Can someone please help me?

As a sidenote, I'm primarily a Rust developer and so far `Documenter` seems to introduce *a lot* of friction into the documentation process compared to e.g. `rustdoc`. Is there a way to make the docs build *entirely* from the code in `src/` only like you do in Rust? I don't need anything else, really.

goerz:
For any `@ref` link in a docstring to a code object, Documenter currently tries to find a binding for the code in the module where the docstring is defined. This often requires importing symbols only because they appear in a `@ref`. In some cases, this is not even possible, as it would create circular references. See also the very eloquent description in https://github.com/JuliaDocs/Documenter.jl/issues/2462#issuecomment-1956806650

With this change, Documenter now tries to find binding for `@ref` links in docstrings first in the local module, and then in `Main`, which is basically `docs/make.jl`. Thus, anything imported in `docs/make.jl` can be referenced anywhere.

Closes #2462

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-* A fully qualified `@ref` link now resolves in `Main` as well well as in the `CurrentModule`. For any package whose docstrings are included in the documentation, as long as that package is loaded in `make.jl`, `@ref` links to docstrings in the package will work from anywhere. This simplifies, e.g., linking between docstrings for packages that use sub-modules. ([#2470])
+* A fully qualified `@ref` link now resolves in `Main` as well, in addition to `CurrentModule`. For any package whose docstrings are included in the documentation, as long as that package is loaded in `make.jl`, fully qualified `@ref` links to docstrings in the package will work from anywhere. This simplifies, e.g., linking between docstrings for packages that use sub-modules. ([#2470])
 
 
 ## Version [v1.3.0] - 2024-03-01
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+* A fully qualified `@ref` link now resolves in `Main` as well well as in the `CurrentModule`. For any package whose docstrings are included in the documentation, as long as that package is loaded in `make.jl`, `@ref` links to docstrings in the package will work from anywhere. This simplifies, e.g., linking between docstrings for packages that use sub-modules. ([#2470])
+
+
 ## Version [v1.3.0] - 2024-03-01
 
 ### Added
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -179,7 +179,7 @@ noncanonical-block).
 ## `@ref` link
 
 Used in markdown links as the URL to tell Documenter to generate a cross-reference
-automatically. The text part of the link can be a docstring, header name, or GitHub PR/Issue
+automatically. The text part of the link can be a code object, header name, or GitHub PR/Issue
 number.
 
 ````markdown
@@ -200,13 +200,13 @@ makedocs
 
 Plain text in the "text" part of a link will either cross-reference a header, or, when it is
 a number preceded by a `#`, a GitHub issue/pull request. Text wrapped in backticks will
-cross-reference a docstring from a `@docs` block.
+cross-reference a docstring from a `@docs` or `@autodocs` block.
+
+The code enclosed in the backticks for such a reference will be evaluated first in the `CurrentModule`  given in the `@meta` block of the current page, and second in `Main`, i.e., the context of the `docs/make.jl` file. For `@ref` links inside a docstring, the `CurrentModule` is automatically set to the module containing the docstring.
 
 `@ref`s may refer to docstrings or headers on different pages as well as the current page
 using the same syntax.
 
-Note that depending on what the `CurrentModule` is set to, a docstring `@ref` may need to
-be prefixed by the module which defines it.
 
 ### Named `@ref`s
 
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -202,10 +202,15 @@ Plain text in the "text" part of a link will either cross-reference a header, or
 a number preceded by a `#`, a GitHub issue/pull request. Text wrapped in backticks will
 cross-reference a docstring from a `@docs` or `@autodocs` block.
 
-The code enclosed in the backticks for such a reference will be evaluated first in the `CurrentModule`  given in the `@meta` block of the current page, and second in `Main`, i.e., the context of the `docs/make.jl` file. For `@ref` links inside a docstring, the `CurrentModule` is automatically set to the module containing the docstring.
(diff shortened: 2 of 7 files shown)


goerz:
This is ready for feedback, although as I mentioned in https://github.com/JuliaDocs/Documenter.jl/issues/2462#issuecomment-1956901156, I might want to test this out "in production" for a little bit.

Before this can be merged, I should probably add a specific test. This PR would probably also be a good opportunity to actually document how the bindings for `@ref` links are found.

goerz:
This doesn't quite work the way I thought… stay tuned

har7an:
@goerz Just saw you picked up work on this, thanks a lot!

After our discussion in #2462 I decided to "fully qualify" all inter-module refs (using the full ref path instead of relative refs), downgrade errors to warnings and live with the messy build output. With the changes from your branch I can build the docs and the links in the resulting HTML all seem to work!

Unfortunately I can't share the code here (it's private...), but I'll happily test any additional changes, if any.

goerz:
You can play around with https://github.com/goerz/DocumenterResolveXRefInMainPrototype.jl, which is this a packaged version of this PR, so I can test it "in the wild" more easily.

This will take more work, though, because it seems like 

> Documenter currently tries to find a binding for the code in the module where the docstring is defined

isn't quite right. The code in `Documenter` for resolving docstrings is quite convoluted, and poorly documented. 

[For example](https://documenter.juliadocs.org/stable/man/syntax/#@ref-link),

> Note that depending on what the `CurrentModule` is set to, a docstring `@ref` may need to be prefixed by the module which defines it.

is quite unclear. So I'm going to have to read the code more to really understand the current code, and how it interacts with this proposed PR, and then document that.

That being said, this PR, respectively the packaged [`DocumenterResolveXRefInMainPrototype`](https://github.com/goerz/DocumenterResolveXRefInMainPrototype.jl) seems to do exactly what I want (I just don't fully understand *how* it does what I want)

mortenpi:
For the original case, doesn't 

```
[`Blarg`](@ref Main.JDIMD.B.Blarg)
```

already just work?

goerz:
I would never have thought of that trick, but it does indeed work!

So that just makes it a policy decision then, and I'm not quite sure what to push for. It might still be a good idea to decide that we always want to try resolving reference in `Main` if they cannot be resolved in `:CurrentModule`. The main effect of that would be that all fully qualified names work as long as the package is loaded in `docs/main.jl`. Realistically, I think this will result in behavior that "just works" for most people when right now they'd get an error, without them ever having to really think about where `@ref` links get resolved.

So let me know that you think, and I'll either finish up this PR with a test, or make a new PR that just does some refactoring and adds documentation. The refactoring would change the internal function `find_docref(code, meta, page)` to `find_docref(code, curmod, page)` with `curmod = get(meta, :CurrentModule, Main)` (making it more reusable for plugins) and try to improve some rather confusing error messages.

The decision is basically whether, in the documentation, we want to have the paragraph below with or without the crossed-out text:

> The code enclosed in the backticks for such a reference will be evaluated ~~first~~ in the `CurrentModule`  given in the `@meta` block of the current page, ~~and second in `Main`, i.e., the context of the `docs/make.jl` file~~. For `@ref` links inside a docstring, the `CurrentModule` is automatically set to the module containing the docstring.

That's basically what it boils down to from the user's perspective, even though the full behavior is quite complicated (see [my notes from diving into this](https://hackmd.io/@michaelgoerz/SJ4sl_j6p))


If we decide to stick to the current behavior (and then we should add a "tip" about `Main` to the documentation), I might still consider writing myself a plugin to also resolve in `Main`, If I can't get used to writing

```
… package must be set as the default AD provider using
[`QuantumControl.set_default_ad_framework`](@ref Main.QuantumControl.set_default_ad_framework).
```

instead of just ```[`QuantumControl.set_default_ad_framework`](@ref)```.

But we'll see… either decision works for me!

har7an:
Just my personal opinion, as a person that isn't primarily a Julia developer, in case anyone cares: I think that @goerz proposal is good to have from a user perspective and I'd vote for that, so to speak.

Most other languages I've written extensive documentation for so far seem to have a mechanism that resolves "full" references. Having to prepend `Main` comes unexpected with that background, but I would agree that, so long as it's clearly mentioned in the docs, it's manageable.

mortenpi:
Hmm. I am not sure we want to encourage referencing objects just because they are available `Main`. Mostly, this creates a situation where the documentation references are dependent on arbitrary global state.

That said, being able to easily reference stuff outside the current module seems also quite reasonable, and having to prepend `Main.` is a bit ugly. So.. could we do a slightly restricted middle ground and only allow referencing top-level modules as fallbacks in `Main`?

I.e. the logic would be that you first do the current logic, where you search from the current module, and then if that fails, you try to see if the reference is a `$(module).$(..)`, and see if `$module` is (1) available in `Main`, (2) a module, and (3) a top-level module (i.e. `parentmodule(module) == Main`.

This should mean that references like

```
[`set_default_ad_framework`](@ref Main.QuantumControl.set_default_ad_framework)
```
and
```
[`QuantumControl.set_default_ad_framework`](@ref)
```
would work, satisfying

> Most other languages I've written extensive documentation for so far seem to have a mechanism that resolves "full" references. Having to prepend Main comes unexpected

But you would not be able to refer to objects just because they are in `Main`.

goerz:
I agree: we should restrict the fallback to `Main` only to fully qualified `@ref` links.

> I.e. the logic would be that you first do the current logic, where you search from the current module, and then if that fails, you try to see if the reference is a `$(module).$(..)`, and see if `$module` is (1) available in `Main`, (2) a module, and (3) a top-level module (i.e. `parentmodule(module) == Main`.

Repeating the "current logic" but with `Main` instead of `CurrentModule` should take care of (1) and (2). If we get manage to get a valid `binding` out of `Main`, then all we need to do to make sure the `code` in the `@ref` is a fully qualified name is to check `startswith(code, string(binding.mod))`.

I added this fix in the latest commit. I still have to add a test and changelog entry before this is ready to go.

goerz:
@mortenpi I finally got around to adding the test and changelog.

This should be ready to merge now. As discussed, the fallback resolution to `Main` is only allowed for fully qualified names. I've been testing this in my QuantumControl packages, and it's been working very nicely: this seems like the final missing piece for being able to link between any two docstrings at all without having to jump through any crazy hoops. So having this will be extremely nice!

This PR should also make the error messages when `@ref` links fail to resolve a lot nicer and less confusing (see the test).

mortenpi (approved):
LGTM! Just a couple of phrasing suggestions -- feel free to take them or leave them, and then merge.

har7an:
I can't comment on the code-side of things, but IMO the new docs along with the examples suggested by @mortenpi make it very clear what's supported and how it works.

PR #2475, opened 2024-03-07
drop settings from `DCtx`

laborg:
PR to see if CI complains about missing setting.

### DIFF
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -881,7 +881,7 @@ function render_page(ctx, navnode)
             div[Symbol("data-docstringscollapsed") => "true"]()
         )
     end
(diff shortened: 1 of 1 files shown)


laborg:
I don't understand the linkcheck error. Is this a CI problem or has this something to do with this branch?

goerz:
It's probably fine, but `DocumenterInterLinks` might touch on these internals. So, we should run the `DocumenterInterLinks` test suite before merging this, just to double-check. That's not necessarily a blocker: I'm fine adapting `DocumenterInterLinks` if any internals change, but I'd have to know.

I'll try to get around to checking this…

goerz:
All good, as far as the plugins are concerned. I checked both `DocumenterInterLinks` and `DocumenterCitations`.

goerz (approved):

mortenpi:
We should make sure that this ends up in 1.4.0, in case it does cause some breakage somewhere.

PR #2477, opened 2024-03-08
Add `DocumenterInterLinks` to regression tests

goerz:


goerz:
@mortenpi Looks like it's good to go now: https://github.com/JuliaDocs/Documenter.jl/actions/runs/8198690317/job/22423296577

PR #2478, opened 2024-03-08
Document required GitHub Action permissions

omus:
Adds documentation for the required GitHub Actions permissions for running `deploydocs`. As I tend to reference these permissions from my CI workflows it's nice having these be in their own section.

I determined the required permissions by starting off with no permissions and slowing adding permissions until I got a successful documentation deployment. There may be additional permissions required that my particular workflow didn't encounter. We can add those to this documentation as we discover them.

- Without `contents: write` rendered documentation cannot be pushed to the repo:
   ```
   remote: Write access to repository not granted.
   fatal: unable to access 'https://github.com/...': The requested URL returned error: 403
   ```
- Without `pull-requests: read` using `push_preview=true` will fail the following check resulting in preview documentation failing to be published (the CI job will still pass though)
   ```
   ✘ PR originates from the same repository
   ```
- Without: `statuses: write` the ["Create a commit status"](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status) API call [used by Documenter.jl](https://github.com/JuliaDocs/Documenter.jl/blob/346c622bf7eb7bc70ab7de6481650285de599921/src/deployconfig.jl#L504C26-L504C54) will [silently fail](https://github.com/JuliaDocs/Documenter.jl/blob/346c622bf7eb7bc70ab7de6481650285de599921/src/deployconfig.jl#L444-L466).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added support for a `CollapsedDocStrings` key in every page's `@meta` block. Setting `CollapsedDocStrings = true` for a particular page essentially clicks the "Collapse all docstrings" in the navigation bar after the page loads, collapsing all docstrings on that page. This can make API documentation pages much more readable. ([#2282], [#2394], [#2459], [#2460])
 * Automatically write an `objects.inv` inventory file to be be included in every online deployment of the documentation. With the help of the [`DocumenterInterLinks` plugin](http://juliadocs.org/DocumenterInterLinks.jl/stable/), any other project using Documenter or [Sphinx](https://www.sphinx-doc.org/en/master/) can then externally link to any page, heading, or docstring in the documentation. ([#2366], [#2424])
 * Added a parameter `inventory_version` to the `HTML()` options that may be passed to `makedocs`. This option sets the `version` metadata field in the new `objects.inv` inventory file. In most cases, the project version for the inventory will be automatically detected from the main `Project.toml` file, respectively, during deployment in `deploydocs`, from the git tag of the release. Any project may still want to explicitly set `inventory_version` via, e.g., [`pkgversion(MyProject)`](https://docs.julialang.org/en/v1/base/base/#Base.pkgversion-Tuple{Module}) instead of relying on an auto-detected version. Projects with a non-standard setup (documentation-only-repos, monorepos) *should* modify their existing configuration to explicitly set `inventory_version`. ([#2449])
+* Added documentation on required GitHub Actions permissions. ([#2478])
 
 ### Changed
 
@@ -1810,6 +1811,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2459]: https://github.com/JuliaDocs/Documenter.jl/issues/2459
 [#2460]: https://github.com/JuliaDocs/Documenter.jl/issues/2460
 [#2461]: https://github.com/JuliaDocs/Documenter.jl/issues/2461
+[#2478]: https://github.com/JuliaDocs/Documenter.jl/issues/2478
 [JuliaLang/julia#36953]: https://github.com/JuliaLang/julia/issues/36953
 [JuliaLang/julia#38054]: https://github.com/JuliaLang/julia/issues/38054
 [JuliaLang/julia#39841]: https://github.com/JuliaLang/julia/issues/39841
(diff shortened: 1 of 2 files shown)


docs/src/man/hosting.md
```diff
@@ -192,7 +192,7 @@ jobs:
   build:
     permissions:
       contents: write
-      statuses: write
```
fredrikekre, line 195: Is this implied by `contents: write` then?

docs/src/man/hosting.md
```diff
@@ -267,6 +267,16 @@ to the configuration file, as showed in the [previous section](@ref GitHub-Actio
     of the deployment is the same as the current repository. In order to push
     elsewhere you should instead use a SSH deploy key.
 
+#### Permissions
+
+The following [GitHub Actions job or workflow permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) are required to successfully use [`deploydocs`](#the-deploydocs-function):
+
+```yaml
+permissions:
+  contents: write
```
fredrikekre, line 276: Is this required also if you have a deploy key?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 * Added a parameter `inventory_version` to the `HTML()` options that may be passed to `makedocs`. This option sets the `version` metadata field in the new `objects.inv` inventory file. In most cases, the project version for the inventory will be automatically detected from the main `Project.toml` file, respectively, during deployment in `deploydocs`, from the git tag of the release. Any project may still want to explicitly set `inventory_version` via, e.g., [`pkgversion(MyProject)`](https://docs.julialang.org/en/v1/base/base/#Base.pkgversion-Tuple{Module}) instead of relying on an auto-detected version. Projects with a non-standard setup (documentation-only-repos, monorepos) *should* modify their existing configuration to explicitly set `inventory_version`. ([#2449])
+* Added documentation on required GitHub Actions permissions. ([#2478])
```
mortenpi, line 13: I appreciate you adding this, but I don't think we want CHANGELOG entries for documentation-only PRs, since I presume the reader is someone who is trying figure out behavior changes.

```suggestion
```

CHANGELOG.md
```diff
@@ -1810,6 +1811,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2459]: https://github.com/JuliaDocs/Documenter.jl/issues/2459
 [#2460]: https://github.com/JuliaDocs/Documenter.jl/issues/2460
 [#2461]: https://github.com/JuliaDocs/Documenter.jl/issues/2461
+[#2478]: https://github.com/JuliaDocs/Documenter.jl/issues/2478
```
mortenpi, line 1814:
```suggestion
```

docs/src/man/hosting.md
mortenpi: This is in the `Authentication: GITHUB_TOKEN` section though, so I think it's appropriate?

CHANGELOG.md
omus: Sounds good. The PR status check was complaining about the missing CHANGELOG entry so I was just obliging

docs/src/man/hosting.md
omus: From the GitHub actions documentation I linked:

> You can use `permissions` to modify the default permissions granted to the `GITHUB_TOKEN`, adding or removing access as required, so that you only allow the minimum required access.

SSH deploy keys can only be used for Git repo access. Since `push_preview=true` [requires `pull-requests: true`](https://github.com/JuliaDocs/Documenter.jl/blob/346c622bf7eb7bc70ab7de6481650285de599921/src/deployconfig.jl#L526) that would also be true there

docs/src/man/hosting.md
```diff
@@ -192,7 +192,7 @@ jobs:
   build:
     permissions:
       contents: write
-      statuses: write
+      pull_requests: read
```
omus, line 195:
```suggestion
      pull-requests: read
```

docs/src/man/hosting.md
```diff
@@ -267,6 +267,16 @@ to the configuration file, as showed in the [previous section](@ref GitHub-Actio
...
     elsewhere you should instead use a SSH deploy key.
 
+#### Permissions
+
+The following [GitHub Actions job or workflow permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) are required to successfully use [`deploydocs`](#the-deploydocs-function):
+
+```yaml
+permissions:
+  contents: write
+  pull_requests: read  # Required when using `push_preview=true`
```
omus, line 277:
```suggestion
  pull-requests: read  # Required when using `push_preview=true`
```

docs/src/man/hosting.md
omus: I'll re-add this as it's still needed. However, if you don't have this permission you just get a silent failure:

> Without: statuses: write the ["Create a commit status"](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status) API call [used by Documenter.jl](https://github.com/JuliaDocs/Documenter.jl/blob/346c622bf7eb7bc70ab7de6481650285de599921/src/deployconfig.jl#L504C26-L504C54) will [silently fail](https://github.com/JuliaDocs/Documenter.jl/blob/346c622bf7eb7bc70ab7de6481650285de599921/src/deployconfig.jl#L444-L466).

docs/src/man/hosting.md
omus: I've moved this to be a section after "Authentication: SSH deploy keys" as it's relevant to both `GITHUB_TOKEN` and SSH deploy keys

mortenpi (approved):
This LGTM, thank you @omus! @fredrikekre, any last thoughts here?

PR #2480, opened 2024-03-10
Consistency on `using` and qualified access

laborg:
In this PR I tried to implement a consistent `using` approach. As can be seen, this requires qualified access for most things, with some exceptions (to not unnecessarily produce code churn).

My motivation for DocumenterEpub is that the code in HTMLWriter gets more readable/understandable (to me). This is of course a style question, so I can live with this being dropped.

### DIFF
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -862,22 +862,13 @@ end
 Constructs and writes the page referred to by the `navnode` to `.build`.
 """
 function render_page(ctx, navnode)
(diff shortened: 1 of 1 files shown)


mortenpi (approved):
I'm all in on fully qualified access. This looks like a terrific cleanup to me, thank you!

Is there anything else you'd like to add here, or is this good to go?

laborg:
No, thats good to go! Thanks for the quick approval.

PR #2481, opened 2024-03-12
[WIP] Generate a 200 redirect instead of a symlink in `git_push`

asinghvi17:
Symlinks don't work on plain github pages builds, i.e., those with a `.nojekyll` file.

### DIFF
--- a/src/deploydocs.jl
+++ b/src/deploydocs.jl
@@ -529,11 +529,23 @@ function git_push(
 end
 
 function rm_and_add_symlink(target, link)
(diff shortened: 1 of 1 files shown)


mortenpi:
Could you describe the problem this is solving a bit more?

asinghvi17:
I was having some issues deploying `gh-pages` sites with symlinks, but it seems the symlinks were not formed correctly.  Sorry for the noise!

The idea was, that if there is a webserver which does not allow symlinks, then redirects could be generated instead.

PR #2482, opened 2024-03-21
Refactor for some explicit imports and cleanups

laborg:
Using ExplicitImports.jl I found a couple of `usings` that can be improved.
Additionally there are removal of unused methods (mainly in utilities.jl)


mortenpi:
Lovely, thank you! As a side note, as a wishlist item, it would also be amazing to have a CI check for ExplicitImports.

mortenpi:
Also running upstream tests for this, just in case: https://github.com/JuliaDocs/Documenter.jl/actions/runs/8382562979

PR #2490, opened 2024-04-03
Allow renderers to override `_any_color_fmt`.

asinghvi17:
This function is used to decide whether to emit ANSI colored output to the output fields of `@example` blocks.

This has uses for DocumenterMarkdown as well as DocumenterVitepress, and will allow us to get rid of the current `@ansi` block formulation.

The PR simply changes the check in `_any_color_fmt` to calling a function `writer_supports_ansicolor` instead of a hardcoded type check; that function can then be extended.


goerz (approved):
Seems okay to me. Of course, there is still the fact that this is 100% internal. If we had an official stable plugin API it would be worth thinking about whether a function like this would be part of that plugin API. If it was, the function would have to have a non-underscore name (`format_supports_color`?) and a docstring that describes exactly what that a writer promises when it claims `format_supports_color` as `true`.

mortenpi:
I agree with Michael. I'm fine with this change as well, but in the current form, it would not have any guarantees of a public API. But I feel this could be turned into a public API as well, if we can articulate a precise docstring for it.

asinghvi17:
Thanks for the feedback!  

I would love to make this a public API - if so, it would probably go in `Documenter.Writers`, and perhaps the name `supports_ansicolor` would work?  That would also remove the need for all writers to have an `ansicolor` field if they do support it.

Perhaps a docstring like:

```julia
"""
	supports_ansicolor(writer::Documenter.Writer)::Bool

Returns `true` if the writer `writer` accepts ANSI colored input from Documenter, or `false` if it does not. 

Defaults to `false`.

This is mainly used to color the output of `@example` and `@repl` blocks.
"""
```

mortenpi:
> it would probably go in `Documenter.Writers`

We're moving away from the submodules. `Documenter.writer_supports_ansicolor` would probably be a good name.

Re the docstring: generally, looks fine. Though I think the main difference will be that the `show` and other IO methods will be called with an `color=true` IOContext?

fingolfin:
@asinghvi17 are you still interested in working on this?

src/expander_pipeline.jl
```diff
@@ -774,13 +774,18 @@ end
 # @example
 # --------
 
-# Find if there is any format with color output
+# Find if there is any format with color output.
 function _any_color_fmt(doc)
-    idx = findfirst(x -> x isa Documenter.HTML, doc.user.format)
+    idx = findfirst(_is_color_fmt, doc.user.format)
```
asinghvi17, line 779:
```suggestion
    idx = findfirst(writer_supports_ansicolor, doc.user.format)
```

src/expander_pipeline.jl
```diff
@@ -774,13 +774,18 @@ end
...
     idx === nothing && return false
     return doc.user.format[idx].ansicolor
 end
 
+# General fallback.
+_is_color_fmt(::Documenter.Writer) = false
+# HTML accepts ANSI, so this is `true`.
+_is_color_fmt(::Documenter.HTMLWriter.HTML) = true
```
asinghvi17, line 787:
```suggestion
# General fallback.
""'
	writer_supports_ansicolor(::Documenter.Writer)::Bool
	
Returns whether the writer supports ANSI-colored output (`true`) or not (`false`).

This is usually relevant in Documenter blocks that execute code, like `@example` or `@repl`.
"""
writer_supports_ansicolor(::Documenter.Writer) = false
# HTML accepts ANSI, so this is `true`.
writer_supports_ansicolor(::Documenter.HTMLWriter.HTML) = true
```

src/expander_pipeline.jl
```diff
@@ -774,13 +774,25 @@ end
...
+# Find if there is any format with color output.
 function _any_color_fmt(doc)
-    idx = findfirst(x -> x isa Documenter.HTML, doc.user.format)
+    idx = findfirst(writer_supports_ansicolor, doc.user.format)
     idx === nothing && return false
     return doc.user.format[idx].ansicolor
 end
 
+# General fallback.
+""'
```
asinghvi17, line 785:
```suggestion
"""
```

src/expander_pipeline.jl
```diff
@@ -774,13 +774,25 @@ end
...
 function _any_color_fmt(doc)
-    idx = findfirst(x -> x isa Documenter.HTML, doc.user.format)
+    idx = findfirst(writer_supports_ansicolor, doc.user.format)
     idx === nothing && return false
     return doc.user.format[idx].ansicolor
 end
 
+# General fallback.
+"""
+	writer_supports_ansicolor(::Documenter.Writer)::Bool
```
asinghvi17, line 786:
```suggestion
    writer_supports_ansicolor(::Documenter.Writer)::Bool
```

src/expander_pipeline.jl
```diff
@@ -774,13 +774,25 @@ end
...
     return doc.user.format[idx].ansicolor
 end
 
+# General fallback.
+"""
+    writer_supports_ansicolor(::Documenter.Writer)::Bool
+	
+Returns whether the writer supports ANSI-colored output (`true`) or not (`false`).
+
+This is usually relevant in Documenter blocks that execute code, like `@example` or `@repl`.
```
asinghvi17, line 790:
```suggestion
This is usually relevant in Documenter blocks that execute code, like `@example` or `@repl`.

If `true`, these blocks will call `show` on the returned Julia object with `color = true`.  If `false`, then show is called with `color=false`.  
```

src/expander_pipeline.jl
```diff
@@ -774,13 +774,27 @@ end
...
+
+If `true`, these blocks will call `show` on the returned Julia object with `color = true`.  If `false`, then show is called with `color=false`.  
+"""
+writer_supports_ansicolor(::Documenter.Writer) = false
+# HTML accepts ANSI, so this is `true`.
+writer_supports_ansicolor(::Documenter.HTMLWriter.HTML) = true
```
asinghvi17, line 796:
```suggestion
writer_supports_ansicolor(::Documenter.HTML) = true
```

src/expander_pipeline.jl
```diff
@@ -774,13 +774,27 @@ end
...
+"""
+    writer_supports_ansicolor(::Documenter.Writer)::Bool
+	
+Returns whether the writer supports ANSI-colored output (`true`) or not (`false`).
+
+This is usually relevant in Documenter blocks that execute code, like `@example` or `@repl`.
+
+If `true`, these blocks will call `show` on the returned Julia object with `color = true`.  If `false`, then show is called with `color=false`.  
+"""
+writer_supports_ansicolor(::Documenter.Writer) = false
```
asinghvi17, line 794:
```suggestion
writer_supports_ansicolor(::Writer) = false
```

src/expander_pipeline.jl
```diff
@@ -774,13 +774,27 @@ end
...
+# HTML accepts ANSI, so this is `true`.
+writer_supports_ansicolor(::Documenter.HTML) = true
```
asinghvi17, line 796:
```suggestion
writer_supports_ansicolor(::HTMLWriter.HTML) = true
```

asinghvi17 (commented):
Let's see if this works instead

asinghvi17 (commented):
Let's see if this works instead

asinghvi17:
Maybe using `HTML` is messing something up here?  But I have no clue what.

src/expander_pipeline.jl
```diff
@@ -774,13 +774,25 @@ end
...
-    idx = findfirst(x -> x isa Documenter.HTML, doc.user.format)
+    idx = findfirst(writer_supports_ansicolor, doc.user.format)
     idx === nothing && return false
     return doc.user.format[idx].ansicolor
 end
 
+# General fallback.
+"""
+    writer_supports_ansicolor(::Documenter.Writer)::Bool
+
```
goerz, line 787: There's some trailing whitespace in this line (the GitHub UI doesn't show it)

@mortenpi Don't we have automatic code formatting in place at this point? Source code formatting, including preventing trailing white space should be be flagged by CI

goerz (changes requested):
LGTM, apart from the trailing whitespace, which should be trivial to fix ;-)

src/expander_pipeline.jl
asinghvi17: There is a Runic run that seems to have passed, maybe Runic doesn't pick this up?

https://github.com/JuliaDocs/Documenter.jl/actions/runs/14786412938/job/41515527745?pr=2490

asinghvi17:
@lazarusA after this PR is released,  we can deprecate the `@ansi` block in Vitepress

src/expander_pipeline.jl
goerz: Well, now I think you deleted the blank line that makes the signature a code block.

There is also more trailing whitespace in the last line of that docstring.

Independent of `runic` picking this up: I would recommend setting up your text editor to remove trailing spaces from `.jl` files automatically, on save.

We should probably also have a `make codestyle` or something that people can use to easily ensure that all code is properly formatted when preparing PRs.

goerz: I think I was able to push a fix-commit to this PR…

mortenpi (approved):
LGTM as well, thank you! I took the liberty of putting the new function into the manual as well, under the public API section.

PR #2496, opened 2024-04-29
RFC: Catppuccin themes

fredrikekre:
This patch add the four catppuccin themes (https://github.com/catppuccin/catppuccin). Opening PR to build preview docs.


fredrikekre:
4 new themes here: https://documenter.juliadocs.org/previews/PR2496/

Hetarth02:
This is very nice!

A suggestion from my side would be to have somewhere where user can define their own color schemes through variables.

Something like what [daisyui](https://daisyui.com/theme-generator/) and other frameworks does.

mortenpi:
LGTM!

mortenpi:
In retrospect, a couple of post-merge thoughts :sweat_smile: 

* Maybe instead of using shell, we move the templated build logic into DocumenterTools as well, and use Mustache.jl?
* It would be good to have the new themes be tested too (i.e. that the CSS is up to date).

assets/html/scss/documenter/components/_admonition.scss
```diff
@@ -21,6 +21,9 @@ $admonition-border-color: () !default;
   @if not map-has-key($admonition-border-color, $name) {
     $admonition-border-color: map-merge($admonition-border-color, ($name: $color)) !global;
   }
+  @if not map-has-key($admonition-border-color, $name) {
+    $admonition-border-color: map-merge($admonition-border-color, ($name: $color)) !global;
+  }
```
mortenpi, line 26: Also, this change is unnecessary I think? It looks like the three lines just above it?

fredrikekre:
> Maybe instead of using shell, we move the templated build logic into DocumenterTools as well, and use Mustache.jl?

Yea maybe, but Makefile dependency tracking is pretty nice and if you develop Documenter themes you probably have `make`? :)

I found it pretty neat to just put `make` at the top of `docs/make.jl` in a LiveServer loop while developing this (could of course be done with Julia too).

> It would be good to have the new themes be tested too (i.e. that the CSS is up to date).

Sure, I can add a test, but maybe not necessary for the release?

assets/html/scss/documenter/components/_admonition.scss
fredrikekre: Maybe it was a bad rebase with #2499 then

PR #2497, opened 2024-04-30
Use link-hover color also for (inline) code in links.

fredrikekre:


mortenpi (approved):

fredrikekre:
Thanks for fixing the changelog

PR #2499, opened 2024-05-02
Tone down admonition coloring

fredrikekre:
This patch tones down the admonitions by removing the background color in the admonition header and the lighter color in the body. Instead the admonition header text is colored.


fredrikekre:
Increased the border thickness a bit too since otherwise it was a bit difficult to see the color of the borders. I am happy with this now.

mortenpi:
For consistency, should we increase the thickness of the docstrings blocks and pre blocks maybe?

image

fredrikekre:
Maybe? And also round the corners like admonitions then?

mortenpi:
> And also round the corners like admonitions then?

Or make admonitions rectangles? :sweat_smile:

fredrikekre:
Updated, and also the `border-bottom` of docstring headers.

fredrikekre:
> and also the border-bottom of docstring headers.

Maybe this isn't needed actually?

mortenpi:
> > and also the border-bottom of docstring headers.
> 
> Maybe this isn't needed actually?

Do you mean getting rid of the border altogether or keeping it `1px`? I think my mild preference is to just keep that `1px`.

image

image

image

fredrikekre:
> Do you mean getting rid of the border altogether or keeping it 1px?

I mean that increasing it isn't needed and agree that 1px looks nicer.

mortenpi (approved):
LGTM! I don't think we have anything else coming, so shall we tag 1.5.0 right away?

fredrikekre:
I could finish up #2496 before, but doesn't really matter.

mortenpi:
Happy to hold off until then. The two changes go together nicely.

PR #2503, opened 2024-05-12
Fix backtick typo in HTMLWriter.jl

abhro:


mortenpi:
Good catch. Thank you!

Issue #1916, opened 2022-08-28
Warning "Unable to determine HTML(edit_link = ...) from remote HEAD branch"


ufechner7:
```
julia> include("docs/make.jl")
[ Info: Precompiling KiteUtils [90980105-b163-44e5-ba9f-8b1c83bb0533]
┌ Warning: Unable to determine HTML(edit_link = ...) from remote HEAD branch, defaulting to "master".
│ Calling `git remote` failed with an exception. Set JULIA_DEBUG=Documenter to see the error.
│ Unless this is due to a configuration error, the relevant variable should be set explicitly.
└ @ Documenter.Utilities ~/.julia/packages/Documenter/yf96B/src/Utilities/Utilities.jl:822
```
This happens always if the locale is not en_EN, but also in other cases.

For me it also happens on Ubuntu 18.04 when the locale is en_EN, but not on Ubuntu 20.04.

Tested with https://github.com/ufechner7/KiteUtils.jl

ufechner7:
```
ufechner@tuxedi:~/repos/KiteUtils.jl$ JULIA_DEBUG=Documenter julia --project
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.7.3 (2022-05-06)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> include("docs/make.jl")
┌ Debug: Creating documentation metadata dictionary (META=##docmeta#291) in KiteUtils
└ @ Documenter.DocMeta ~/.julia/packages/Documenter/yf96B/src/DocMeta.jl:36
┌ Warning: Unable to determine HTML(edit_link = ...) from remote HEAD branch, defaulting to "master".
│ Failed to parse the `git remote` output. Set JULIA_DEBUG=Documenter to see the output.
│ Unless this is due to a configuration error, the relevant variable should be set explicitly.
└ @ Documenter.Utilities ~/.julia/packages/Documenter/yf96B/src/Utilities/Utilities.jl:832
[...]

PR #2509, opened 2024-05-25
Allow for remote names other than `origin`.

### LINKED ISSUE #2464: make.jl fails with git error
Hi,
I'm trying to generate documentation for my project following these instructions: https://documenter.juliadocs.org/stable/man/guide/

I created a `docs` subfolder with `make.jl` : 
```
using Documenter
push!(LOAD_PATH,"../src/")
using my_package

makedocs(sitename="my_package Documentation")
```

When I run `make.jl` I get the following error:

```
*[main][~/julia_envs/my_package/docs]$ julia --color=yes --project make.jl 
ERROR: LoadError: ArgumentError: Unable to automatically determine remote for main repo.
> `repo` is not set, and the Git repository has invalid origin.
Configure `repo` and/or `remotes` appropriately, or set `remotes = nothing` to disable remote source
links altogether (e.g. if not working in a Git repository).
  path: /Users/ralphkube/julia_envs/my_package
Stacktrace:
 [1] interpret_repo_and_remotes(; root::String, repo::String, remotes::Dict{Any, Any})
   @ Documenter ~/.julia/packages/Documenter/1HwWe/src/documents.jl:640
 [2] interpret_repo_and_remotes
   @ ~/.julia/packages/Documenter/1HwWe/src/documents.jl:499 [inlined]
 [3] Documenter.Document(; root::String, source::String, build::String, workdir::Symbol, format::Documenter.HTMLWriter.HTML, clean::Bool, doctest::Bool, linkcheck::Bool, linkcheck_ignore::Vector{Any}, linkcheck_timeout::Int64, checkdocs::Symbol, doctestfilters::Vector{Regex}, warnonly::Vector{Symbol}, modules::Vector{Module}, pages::Vector{Any}, pagesonly::Bool, expandfirst::Vector{String}, plugins::Vector{Documenter.Plugin}, repo::String, remotes::Dict{Any, Any}, sitename::String, authors::String, version::String, highlightsig::Bool, draft::Bool, others::@Kwargs{})
   @ Documenter ~/.julia/packages/Documenter/1HwWe/src/documents.jl:429
 [4] #makedocs#81
   @ Documenter ~/.julia/packages/Documenter/1HwWe/src/makedocs.jl:242 [inlined]
 [5] top-level scope
   @ ~/julia_envs/my_package/docs/make.jl:5
```

The git configuration for `my_package` works fine though and git origin is set correctly.

RomeoV:
We introduce `getremotename` that tries to check the `remote.pushDefault` option to determine the remote name, and falls back to `origin` if it's not set.

Fixes #2464 and perhaps #1916, as well as https://discourse.julialang.org/t/creating-documentation-when-remote-name-is-not-origin/110466.

### DIFF
--- a/src/utilities/utilities.jl
+++ b/src/utilities/utilities.jl
@@ -421,7 +421,31 @@ const julia_remote = Remotes.GitHub("JuliaLang", "julia")
 """
 Stores the memoized results of [`getremote`](@ref).
 """
-const GIT_REMOTE_CACHE = Dict{String,Union{Remotes.Remote,Nothing}}()
+const GIT_REMOTENAME_CACHE = Dict{String,Union{String,Nothing}}()
+const GIT_REMOTEURL_CACHE = Dict{String,Union{Remotes.Remote,Nothing}}()
+
+"""
+$(TYPEDSIGNATURES)
+
+Determines the GitHub remote of a directory by checking `remote.pushDefault` of the repository.
+Defaults to `origin`.
+
+The results for a given directory are memoized in [`GIT_REMOTENAME_CACHE`](@ref), since calling
+`git` is expensive and it is often called on the same directory over and over again.
+"""
+function getremotename(dir::AbstractString)
+    isdir(dir) || return nothing
+    return get!(GIT_REMOTENAME_CACHE, dir) do
+        try
+            remotename = readchomp(setenv(`$(git()) config --get remote.pushDefault`; dir=dir))
+            @assert !isempty(remotename)
+            remotename
+        catch e
+            @warn "git config --get remote.pushDefault failed. Assuming remote `origin`." exception=(e, catch_backtrace())
+            "origin"
+        end
+    end
+end
 
 function parse_remote_url(remote::AbstractString)
     # TODO: we only match for GitHub repositories automatically. Could we engineer a
@@ -435,24 +459,25 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Determines the GitHub remote of a directory by checking `remote.origin.url` of the
-repository. Returns a [`Remotes.GitHub`](@ref), or `nothing` is something has gone wrong
-(e.g. it's run on a directory not in a Git repo, or `origin.url` points to a non-GitHub
-remote).
+Determines the GitHub remote of a directory by checking `remote.\$(remotename).url` of the
+repository. `remotename` is determined by [`getremotename`](@ref). Returns a [`Remotes.GitHub`](@ref),
+or `nothing` is something has gone wrong (e.g. it's run on a directory not in a Git repo, or the url
+points to a non-GitHub remote).
 
-The results for a given directory are memoized in [`GIT_REMOTE_CACHE`](@ref), since calling
+The results for a given directory are memoized in [`GIT_REMOTEURL_CACHE`](@ref), since calling
 `git` is expensive and it is often called on the same directory over and over again.
 """
 function getremote(dir::AbstractString)
     isdir(dir) || return nothing
-    return get!(GIT_REMOTE_CACHE, dir) do
-        remote = try
-            readchomp(setenv(`$(git()) config --get remote.origin.url`; dir=dir))
+    return get!(GIT_REMOTEURL_CACHE, dir) do
+        remotename = getremotename(dir)
+        remoteurl = try
+            readchomp(setenv(`$(git()) config --get remote.$(remotename).url`; dir=dir))
         catch e
-            @debug "git config --get remote.origin.url failed" exception=(e, catch_backtrace())
+            @debug "git config --get remote.$(remotename).url failed" exception=(e, catch_backtrace())
             ""
         end
-        return parse_remote_url(remote)
+        return parse_remote_url(remoteurl)
     end
 end
 
@@ -616,7 +641,7 @@ it out automatically.
 `root` is the the directory where `git` gets run. `varname` is just informational and used
 to construct the warning messages.
 """
-function git_remote_head_branch(varname, root; remotename = "origin", fallback = "master")
+function git_remote_head_branch(varname, root; fallback = "master")
     gitcmd = git(nothrow = true)
     if gitcmd === nothing
         @warn """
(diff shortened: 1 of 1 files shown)


src/utilities/utilities.jl
```diff
@@ -421,7 +421,31 @@ const julia_remote = Remotes.GitHub("JuliaLang", "julia")
 """
 Stores the memoized results of [`getremote`](@ref).
 """
-const GIT_REMOTE_CACHE = Dict{String,Union{Remotes.Remote,Nothing}}()
+const GIT_REMOTENAME_CACHE = Dict{String,Union{String,Nothing}}()
```
mortenpi, line 424: I don't think we expect `nothing` values here?

```suggestion
const GIT_REMOTENAME_CACHE = Dict{String,String}()
```

mortenpi:
Generally, I think trying to use `pushDefault` makes sense. But I have a couple of concerns:

1. It will warn when `remote.pushDefault` is not set. However, it doesn't seem that `remote.pushDefault` is set in the normal case (e.g. you have just `git clone`-d a repo). So it seems that 99% of the time this code will warn (you can see that in the CI logs too), which is not great.

2. I also wonder what is the right behavior when you do have `origin`, but `pushDefault` points to a different remote? I generally don't set `pushDefault`, but the use case I can imagine is when I copy a repo that I don't have access to, then fork it on GitHub and set `pushDefault` to the fork under my user. But if I were to build docs locally, I think I would prefer if it would pick `origin` over my fork. So maybe it should use `origin` if available, and if there is not remote called `origin`, it will pick the second best on based on `pushDefault`.

   Or maybe first step could be to just pick the only remote that is available, even if it is not called `origin`? And only use `pushDefault` if you have multiple remotes, none are called `origin`, and then use `pushDefault` to disambiguate?

3. It would be good to add a note somewhere on the [`docs/src/lib/remote-links.md`](https://github.com/JuliaDocs/Documenter.jl/blob/master/docs/src/lib/remote-links.md) manual page about this behavior.
4. Ideally there would also be a test. We have some related tests in [`test/repolinks.jl`](https://github.com/JuliaDocs/Documenter.jl/blob/master/test/repolinks.jl)

RomeoV:
Thanks for the thoughts. I have changed the workflow to default to `origin`, then `remotes.pushDefault` with a warning, and then just `""` (which was the behavior before) with another warning.
I also added a related comment to the docs, and wrote some tests.

src/utilities/utilities.jl
```diff
@@ -421,7 +421,45 @@ const julia_remote = Remotes.GitHub("JuliaLang", "julia")
...
 Stores the memoized results of [`getremote`](@ref).
 """
-const GIT_REMOTE_CACHE = Dict{String,Union{Remotes.Remote,Nothing}}()
+const GIT_REMOTENAME_CACHE = Dict{String, String}()
+const GIT_REMOTEURL_CACHE = Dict{String,Union{Remotes.Remote,Nothing}}()
+
+"""
+$(TYPEDSIGNATURES)
+
+Determines the GitHub remote of a directory, returninf 'origin' if that remote  exists,
```
fingolfin, line 430:
```suggestion
Determines the GitHub remote of a directory, returning `origin` if that remote  exists,
```

src/utilities/utilities.jl
```diff
@@ -421,7 +421,45 @@ const julia_remote = Remotes.GitHub("JuliaLang", "julia")
...
+            @assert !isempty(originurl)
+            return "origin"
+        catch; end
+
+        try
+            pushdefault = readchomp(setenv(`$(git()) config --get remote.pushDefault`; dir=dir))
+            @assert !isempty(pushdefault)
+            @warn "Remote 'origin' not found. Using `git config --get remote.pushDefault` instead, which is set to '$(pushdefault)'."
+            return pushdefault
+        catch ; end
```
fingolfin, line 451: I suspect this PR needs to be reformatted with `runic`

src/utilities/utilities.jl
```diff
@@ -421,7 +421,45 @@ const julia_remote = Remotes.GitHub("JuliaLang", "julia")
...
+`git` is expensive and it is often called on the same directory over and over again.
+"""
+function getremotename(dir::AbstractString)
+    isdir(dir) || return nothing
+    return get!(GIT_REMOTENAME_CACHE, dir) do
+        try
+            originurl = readchomp(setenv(`$(git()) config --get remote.origin.url`; dir=dir))
+            @assert !isempty(originurl)
+            return "origin"
+        catch; end
```
fingolfin, line 444: You seem to be using `@assert` plus `try/catch` for control flow. That seems like a bad idea for multiple reasons (e.g. future Julia version might allow disabling `@assert` checks, which is, I think, mainly intended to enforce to check runtime invariants, which is not what this code does).

How about just e.g. this
```suggestion
        originurl = readchomp(setenv(`$(git()) config --get remote.origin.url`; dir=dir))
        if !isempty(originurl)
            return "origin"
        end
```

fingolfin: Same applies to the rest of the code, of course

src/utilities/utilities.jl
```diff
@@ -616,7 +655,7 @@ it out automatically.
 `root` is the the directory where `git` gets run. `varname` is just informational and used
 to construct the warning messages.
 """
-function git_remote_head_branch(varname, root; remotename = "origin", fallback = "master")
+function git_remote_head_branch(varname, root; fallback = "master")
```
fingolfin, line 658: The docstring still references `remotename`. So either that needs to be adjusted. Alternatively, keep the kwarg here, but with a new default value `nothing`; and then later only call `getremotename` if `remotename === nothing` ?

fingolfin:
@RomeoV are you still interested in this? If so, some conflicts need to be resolved now and a bunch of comments should be addressed.

If not, let's close it?

RomeoV:
I have no bandwidth right now for this unfortunately, and also no use at this time. So we can close the PR. I appreciate your review and thoughtful comments, though. If you would like this merged i can address the comments some time later in November.

PR #2510, opened 2024-05-28
Call new theme Makefile from main one, add CI

mortenpi:
Following up on #2496. We already have a `make themes` target. ~Critically, this re-uses the `docs/` env for the DocumenterTools dependency.~ Actually, let's use the `test/themes` environment as before, but let's just move that whole thing to `assets/html`, so that everything is in one place.

Note: it looks like the new CSS files are actually outdated, at least when building with this.


mortenpi:
Successfully failed! https://github.com/JuliaDocs/Documenter.jl/actions/runs/9277393208/job/25526482102?pr=2510
image

mortenpi:
FWIW, it looks like it was the border changes that didn't get updated.

image

assets/html/Project.toml
```diff
@@ -1,3 +1,7 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
+libsass_jll = "47bcb7c8-5119-555a-9eeb-0afcc36cd728"
+
+[compat]
+libsass_jll = "~3.6"
```
fredrikekre, line 7: I probably had another version of this locally when I built the themes then? Or why do you think they changed?

fredrikekre (approved):

assets/html/Project.toml
mortenpi: I didn't investigate it too carefully, but if it had been a libsass difference, then I would expect the `documenter-*` CSS files to be outdated as well.

PR #2513, opened 2024-05-29
Escape single paranthesis

aaruni96:
This stops a single parenthesis from crashing search. A second one will still crash it, but I don't know enough javascript to handle that cleanly.


mortenpi:
cc @LilithHafner or @Hetarth02: any thoughts here by any chance?

LilithHafner:
Ever since switching to the new search the search pane sometimes doesn't show up for me. It's flaky enough that it is inconvenient for me to debug other search-related issues.

mortenpi:
> search pane sometimes doesn't show up for me.

FWIW, I'm pretty sure that got fixed in #2458.

fingolfin:
All in all, isn't this a clear improvements with no downsides?

Can at least the workflows be run?

Hetarth02:
Since we are doing this, we should also see if there are any other characters which break it like this.

aaruni96:
> break it like this

Maybe a `/`, and `[` ?

The unescaped parenthesis is directly plugged into the regx engine, and the regx then suddenly has a wrong grouping causing the engine to crash.

AFAIK, regx only uses so many special characters in its patterns: `/` to separate, `(`, `)` to make groups, and `[`, `]` to make sets, but there could be one or two more.

fingolfin:
Depending on which type of regex, there can be many more special character, including `.` `*` `|` -- see https://www.w3schools.com/jsref/jsref_obj_regexp.asp for a more comprehensive overview relevant to JavaScript.

fingolfin:
Relevant SO post: https://codereview.stackexchange.com/questions/153691/escape-user-input-for-use-in-js-regex

mortenpi:
And SO points to MDN and to an ECMAScript proposal:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping
* https://github.com/tc39/proposal-regex-escaping

Let's just implement the MDN suggestion. It looks straightforward.

fingolfin:
Can someone please let the CI tests run? @aaruni96 pushed an update using the discussed MDN regex, so hopefully it now covers more (all?) cases.

Also if it passes and y'all are happy, I hope it can merged soon :-)

Hetarth02:
CC @mortenpi

LilithHafner:
FWIW, I prefer allowing checks to run without approval, even from first time contributors.

mortenpi (approved):
LGTM, thank you!

PR #2526, opened 2024-06-12
Add purple coloring for todo-admonitions

fredrikekre:


mortenpi:
Where is this actually used in practice? This is not a standard admonition supported by the Markdown stdlib, right? My view generally has been that we only style admonitions that are also supported by the REPL etc., and custom stuff should be a plugin.

Could we add an example of it here? https://github.com/JuliaDocs/Documenter.jl/blob/5425664e91a1dedcf59bd32139d341f9271318e6/docs/src/showcase.md?plain=1#L101-L119

fredrikekre:
> Where is this actually used

I just wanted nicely formatted todo-notes in the docs :) I can of course just add the necessary CSS locally.

mortenpi:
I think my inclination here is not to deviate from the Markdown stdlib and the Julia REPL 🙁 I only wish it was easier to customize the themes and admonitions though. The ability to add custom ones would be useful though.

fredrikekre:
I guess I just have to patch Markdown and REPL first then :)

mortenpi:
> I guess I just have to patch Markdown and REPL first then :)

That's definitely one way to go about it :rofl:

fredrikekre:
Just noticed that Github uses purple for the `Important` category; https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

mortenpi (approved):
LGTM. @fredrikekre, it's good to go, right? Sorry, this slipped through the cracks.

image

image

Issue #2254, opened 2023-09-14
Use Downloads.jl instead of `curl` binary


mortenpi:
We assume that there is a `curl` binary on the user's system if a few places. We should switch to using Downloads.jl. We already have a precedent following https://github.com/JuliaDocs/Documenter.jl/pull/2232.

odow:
We also use `curl` for checking links, x-ref https://github.com/JuliaDocs/Documenter.jl/issues/1601

PR #2532, opened 2024-06-19
HTML format: offline_version parameter

ilancoulon:
Adresses JuliaLang/julia#19354 and one of the items on #212.

The general idea is to try to add a keyword parameter to the `HTML()` struct. By putting `format = Documenter.HTML(offline_version = true)` in the parameters of `makedocs`, the HTML output should come bundled with all the CDNs.

Here is a list of what I've done:
- added the keyword as a parameter of HTML
- when a remote dependency is used in the HTMLWriter, I call a function of RD in between, called `process_remote`. This function, depending on the user-given parameter will download the dependency using curl and write it in the right file, replacing the CDN url with a relative URL.
- for CSS font files, the URL inside are relative, so those files also need to be downloaded, this is done through the function `_process_downloaded_css` (with a regex)
- for my internal use, having font files as an asset behind an authenticated portal was not working. I had to actually write them as a base64 encoded string inline the CSS file. This might not be useful for everyone but it makes docs displayable when the website needs authentication even for assets (because a font file is always fetched in an [anonymous mode](https://www.w3.org/TR/css-fonts-3/#font-fetching-requirements)).
- a limited test of the offline_version build. The rest of those functions is hard to test because this depends on curl.

I might have missed some dependencies or could improve some of the functions signature, would gladly take some feedback!

### DIFF
--- a/src/html/RD.jl
+++ b/src/html/RD.jl
@@ -144,7 +144,7 @@ module RD
             mkpath(dirname(filepath))
             open(filepath, "w") do f
                 if splitext(filepath)[end] == ".css"
-                    result = _process_downloaded_css(result, url, filepath)
+                    result = _process_downloaded_css(result, url)
                 end
                 write(f, result)
             end
@@ -153,13 +153,23 @@ module RD
         return relpath(filepath, origin_path*"/")
     end
 
-    function _process_downloaded_css(file_content, origin_url, origin_path)
+    const font_ext_to_type = Dict(
+        ".ttf" => "truetype",
+        ".eot" => "embedded-opentype",
+        ".eot?#iefix" => "embedded-opentype",
+        ".svg#webfont" => "svg",
+        ".woff" => "woff",
+        ".woff2" => "woff2",
+    )
+
+    function _process_downloaded_css(file_content, origin_url)
         url_regex = r"url\(([^)]+)\)"
         replace(file_content, url_regex => s -> begin
             rel_url = match(url_regex, s).captures[1]
             url = normpath(dirname(origin_url), rel_url)
+            font_type = font_ext_to_type[splitext(rel_url)[end]]
             encoded_file = Base64.base64encode(_download_file_content(url))
-            return "url(data:font/truetype;charset=utf-8;base64,$(encoded_file))"
+            return "url(data:font/$(font_type);charset=utf-8;base64,$(encoded_file))"
         end)
     end
 end
--- a/src/html/RD.jl
+++ b/src/html/RD.jl
@@ -1,6 +1,8 @@
 "Provides a namespace for remote dependencies."
 module RD
     using JSON: JSON
+    using Base64
+    using ....Documenter: hascurl
     using ....Documenter.JSDependencies: RemoteLibrary, Snippet, RequireJS, jsescape, json_jsescape
     using ..HTMLWriter: KaTeX, MathJax, MathJax2, MathJax3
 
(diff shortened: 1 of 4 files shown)


mortenpi:
Broadly, I think this looks pretty good, and it would be great to have this option. Thank you for taking a stab at this!

One thing that stands out to me is that we're modifying all the signatures for the different libraries in `RD.jl`. But it's really an all-or-nothing, not a per-dependency thing, right? I wonder if we could move the handling of all this into the shared [`JSDependencies.jl`](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/utilities/JSDependencies.jl)?

src/html/RD.jl
```diff
@@ -113,4 +117,59 @@ module RD
...
+    function _process_downloaded_css(file_content, origin_url)
+        url_regex = r"url\(([^)]+)\)"
+        replace(file_content, url_regex => s -> begin
+            rel_url = match(url_regex, s).captures[1]
+            url = normpath(dirname(origin_url), rel_url)
+            font_type = font_ext_to_type[splitext(rel_url)[end]]
+            encoded_file = Base64.base64encode(_download_file_content(url))
+            return "url(data:font/$(font_type);charset=utf-8;base64,$(encoded_file))"
+        end)
+    end
```
mortenpi, line 174: This is a bit non-trivial. It would be good to have a comment for it.

src/html/RD.jl
```diff
@@ -113,4 +117,59 @@ module RD
...
+            @error "$cmd failed: ", err
+            error()
+            return
+        end
+    end
+
+    function _process(url::AbstractString, output_path, origin_path)
+        if !hascurl()
+            @error "offline_version requires curl."
+            error()
```
mortenpi, line 138:
```suggestion
            error("offline_version requires curl.")
```

src/html/RD.jl
```diff
@@ -113,4 +117,59 @@ module RD
         ))
     end
     mathengine(::RequireJS, ::Nothing) = nothing
+
+    process_remote(dep, offline_version::Bool, build_path, origin_path=build_path) = offline_version ? _process(dep, build_path, origin_path) : dep
+    _process(dep::RemoteLibrary, build_path, origin_path) = RemoteLibrary(dep.name, _process(dep.url, build_path, origin_path); deps = dep.deps, exports = dep.exports)
+
+    function _download_file_content(url::AbstractString)
+        cmd = `curl $url -s --output -`
```
mortenpi, line 125: Could also use Downloads.jl here. But `curl` is what we use elsewhere right now, and while we want to move away from it (https://github.com/JuliaDocs/Documenter.jl/issues/2254), its also completely fine to keep it as is.

ilancoulon:
> Broadly, I think this looks pretty good, and it would be great to have this option. Thank you for taking a stab at this!

Thank you! I tried applying your comments. I basically added some comments, and used Downloads.jl instead of `curl`, it's much better!

> One thing that stands out to me is that we're modifying all the signatures for the different libraries in `RD.jl`. But it's really an all-or-nothing, not a per-dependency thing, right? I wonder if we could move the handling of all this into the shared [`JSDependencies.jl`](https://github.com/JuliaDocs/Documenter.jl/blob/master/src/utilities/JSDependencies.jl)?

I agree it's not nice to have to change these signatures and tried avoiding that. At some point I was indeed using JSDependencies but the thing is that all these functions in RD.jl actually convey a lot of information to JSDependencies when constructing the RemoteLibraries. And they need that build_path/output_path at this moment. I'd be happy to improve this part of the code but I don't see how yet, without modifying the whole RD.jl file, changing completely how it interacts with JSDependencies. Happy to hear a more specific idea on this if you have it.

PR #2552, opened 2024-08-05
Add video MIME types to `Documenter.display_dict`

asinghvi17:
Following from what was discussed on the community call a couple of months ago.  Still WIP and I will have to test locally.

### DIFF
--- a/src/utilities/utilities.jl
+++ b/src/utilities/utilities.jl
@@ -551,6 +551,7 @@ function display_dict(x; context = nothing)
     out[MIME"text/plain"()] = limitstringmime(MIME"text/plain"(), x, context = context)
     for m in [MIME"text/html"(), MIME"image/svg+xml"(), MIME"image/png"(),
               MIME"image/webp"(), MIME"image/gif"(), MIME"image/jpeg"(),
+              MIME"video/mp4"(), MIME"video/mpeg"(), MIME"video/webm"(),
               MIME"text/latex"(), MIME"text/markdown"()]
         showable(m, x) && (out[m] = stringmime(m, x, context = context))
     end


mortenpi:
Two thoughts here:

1. So we do this thing where we `show` every available MIME type. I'm a little concerned that we might end up unintentionally rendering heaps of videos in some doc suddenly, which I think can be quite expensive. So I wonder if we'd want this to be per-block opt-in? Maybe the concern is a bit unfounded though -- how common are the video `show` methods?

2. I wonder if we could add something really simple to the Showcase page? I guess we'd have to commit a tiny video file into the repository for that, unless we could somehow generate it.

fredrikekre:
For 1. I think it only becomes a problem if an object support all of the video formats and all three are rendered but perhaps this is common? Or is there generally just one implemented?

As a side note, some people complain about slow build times because of the display_dict so perhaps the mimes should somehow be filtered keeping in mind that not all options may end up being used.

mortenpi:
> For 1. I think it only becomes a problem if an object support all of the video formats and all three are rendered but perhaps this is common? Or is there generally just one implemented?

Rendering multiple videos is actually something I didn't even consider, but I agree, this is also something we should think about.

My concern was mostly about existing docs with many complex figures now potentially rendering heaps of videos though.

> As a side note, some people complain about slow build times because of the display_dict so perhaps the mimes should somehow be filtered keeping in mind that not all options may end up being used.

I'm not sure how we should approach this, but doing something about that wouldn't hurt, I agree.

fredrikekre:
> My concern was mostly about existing docs with many complex figures now potentially rendering heaps of videos though.

But are there objects which have both a video representation and image representations? I would expect it to be mutually exclusive?

mortenpi:
> But are there objects which have both a video representation and image representations?

I wouldn't be surprised if animations would fall back to the first frame maybe? But whether that's actually common, I don't know.

mortenpi:
> > As a side note, some people complain about slow build times because of the display_dict so perhaps the mimes should somehow be filtered keeping in mind that not all options may end up being used.
> 
> I'm not sure how we should approach this, but doing something about that wouldn't hurt, I agree.

Just a passing thought: I wonder if we could do a kind of a "look ahead" here, and query the writer for a list of MIMEs it wants/supports, and then just render the first one?

asinghvi17:
That could be cool, let's discuss on Tuesday

asinghvi17:
Closing this PR since it's horribly out of date.  But I find myself needing this again, so will circle back.

PR #2553, opened 2024-08-06
Add `actions: write` permissions to example Actions workflow

AnzhiZhang:
add `actions: write` to permissions as julia-actions/cache required

See also: https://github.com/AnzhiZhang/TypedMatrices.jl/actions/runs/10271054984/job/28420170392#step:11:18


mortenpi (approved):
LGTM, thanks!

PR #2563, opened 2024-08-27
Add changelog section to contributing docs

frankier:


mortenpi (approved):
Good idea!

docs/src/contributing.md
```diff
@@ -58,6 +58,12 @@ Follow the style of the surrounding text when making changes. When adding new fe
   * always use fenced code blocks instead of indented blocks;
   * follow the conventions outlined in the Julia documentation page on documentation.
 
+### Changelog
```
fredrikekre, line 61: Should be a level 2 heading? Or did you intend to put it under "Style Guide"?
```suggestion
## Changelog
```

frankier: I did intend that since it's a style guide for the change log. Do you definitely think it should be top level?

fredrikekre: Ah okay, I thought this was about the changelog in general.

mortenpi: I think it actually makes sense to have this on the top level --- I think it's generally a good idea to let people know that a CHANGELOG is a thing. So I'll just go ahead and make it `h2`.

PR #2564, opened 2024-08-28
Manual changes after Runic formatting

fredrikekre:


fredrikekre:
Update the PR to what I think will be Runic version 1. I skimmed through it and found one bug which is now fixed. Would be great if someone (@mortenpi ?) could have a look and see if there are anything weird too.

src/DocSystem.jl
```diff
@@ -123,16 +122,14 @@ function docstr(md::Markdown.MD; kws...)
...
+    return doc
 end
 docstr(other) = other
 
 
 ## Formatting `DocStr`s. ##
 
 
-
-
```
mortenpi, line 135: Does the formatter remove empty lines? Not opposed to it, just curious where this change came from.

src/deploydocs.jl
```diff
@@ -289,35 +295,37 @@ function deploydocs(;
...
-    open(objects_inv) do input
+    v = open(objects_inv) do input
         for line in eachline(input)
             if startswith(line, "# Version:")
                 return strip(line[11:end])
             end
         end
         error("Invalid $objects_inv: missing or invalid version line")
     end
+    return v
```
mortenpi, line 320: Did the formatter do this? :open_mouth: I would the automatic change to be just a `return` in front of `open`.

src/docchecks.jl
```diff
@@ -141,24 +143,31 @@ function footnotes(doc::Document)
...
 # ------------
 
-hascurl() = (try; success(`curl --version`); catch err; false; end)
+function hascurl()
+    try
+        return success(`curl --version`)
+    catch err
+        return false
+    end
+end
```
mortenpi, line 170: This was also by the formatter?

src/docchecks.jl
```diff
@@ -331,8 +342,8 @@ end
 
 GITHUB_ERROR_ADVICE = (
     "This means automatically finding the source URL link for this package failed. " *
-    "Please add the source URL link manually to the `remotes` field " *
-    "in `makedocs` or install the package using `Pkg.develop()``."
+        "Please add the source URL link manually to the `remotes` field " *
+        "in `makedocs` or install the package using `Pkg.develop()``."
 )
```
mortenpi, line 347: I guess I'm not a huge fan of this one.. but I supposed it's consistent with the general rules?

I wonder if this would be a better general pattern (not for the formatter, but when trying to write long single-line string literals):

```
GITHUB_ERROR_ADVICE = """
    This means automatically finding the source URL link for this package failed. \
    Please add the source URL link manually to the `remotes` field \
    in `makedocs` or install the package using `Pkg.develop()``."""
```

src/doctest.jl
```diff
@@ -17,21 +17,24 @@ will skip doctesting of manual pages altogether.
...
-         """))
+        source = normpath(joinpath(dirname(pathof(package)), "..", "docs", "src"))
+        if !isdir(source)
+            msg = """
+            Package $(package) does not have a documentation source directory at standard location.
+            Searched at: $(source)
+            If ...
+            """
+            throw(ArgumentError(msg))
+        end
```
mortenpi, line 34: This is a manual rewrite? It's more readable, so no objections.

src/documents.jl
```diff
@@ -78,50 +78,52 @@ end
...
-    codeblock   :: MarkdownAST.CodeBlock # original code block
-
-    function IndexNode(codeblock;
+    pages::Vector{String} # Which pages to include in the index? Set by user.
+    modules::Vector{Module} # Which modules to include? Set by user.
+    order::Vector{Symbol} # What order should docs be listed in? Set by user.
+    build::String         # Path to the file where this index will appear.
+    source::String         # Path to the file where this index was written.
+    elements::Vector         # (object, doc, page, mod, cat)-tuple for constructing links.
+    codeblock::MarkdownAST.CodeBlock # original code block
```
mortenpi, line 87:
```suggestion
    pages::Vector{String} # Which pages to include in the index? Set by user.
    modules::Vector{Module} # Which modules to include? Set by user.
    order::Vector{Symbol} # What order should docs be listed in? Set by user.
    build::String # Path to the file where this index will appear.
    source::String # Path to the file where this index was written.
    elements::Vector # (object, doc, page, mod, cat)-tuple for constructing links.
    codeblock::MarkdownAST.CodeBlock # original code block
```

src/documents.jl
```diff
@@ -78,50 +78,52 @@ end
...
-            Pages  = [],
-            Depth  = 1:2,
-            build  = error("missing value for `build` in `ContentsNode`."),
+    pages::Vector{String} # Which pages should be included in contents? Set by user.
+    mindepth::Int         # Minimum header level that should be displayed. Set by user.
+    depth::Int            # Down to which level should headers be displayed? Set by user.
+    build::String         # Same as for `IndexNode`s.
+    source::String        # Same as for `IndexNode`s.
+    elements::Vector      # (order, page, anchor)-tuple for constructing links.
+    codeblock::MarkdownAST.CodeBlock # original code block
```
mortenpi, line 120: Did you adjust the comments manually here?

src/html/HTMLWriter.jl
```diff
@@ -2194,11 +2234,11 @@ function domify(dctx::DCtx, node::Node, i::ImageElements)
...
 
     if occursin(r"\.(webm|mp4|ogg|ogm|ogv|avi)$", url)
-        video[:src => url, :controls => "true", :title => alt](
+        return video[:src => url, :controls => "true", :title => alt](
             a[:href => url](alt)
         )
     else
-        img[:src => url, :alt => alt]
+        return img[:src => url, :alt => alt]
     end
```
mortenpi, line 2242: Is this by the formatter? I would have expected a `return` in front of the `if`.

src/latex/LaTeXWriter.jl
```diff
@@ -391,40 +411,48 @@ function latex(io::Context, node::Node, contents::Documenter.ContentsNode)
...
     if haskey(d, MIME"image/png"())
         write("$(filename).png", base64decode(d[MIME"image/png"()]))
-        _println(io, """
-        \\begin{figure}[H]
-        \\centering
-        \\includegraphics[max width=\\linewidth]{$(filename)}
-        \\end{figure}
-        """)
+        _println(
+            io, """
```
mortenpi, line 439: I guess I would actually expect this here:

```suggestion
            io,
            """
```

src/utilities/DOM.jl
```diff
@@ -105,21 +105,27 @@ export @tags
...
-    :a, :abbr, :acronym, :b, :bdo, :big, :br, :button, :cite, :code, :dfn, :em,
-    :i, :img, :input, :kbd, :label, :map, :object, :q, :samp, :script, :select,
-    :small, :span, :strong, :sub, :sup, :textarea, :time, :tt, :var,
-])
-const VOID_ELEMENTS = Set([
-    :area, :base, :br, :col, :command, :embed, :hr, :img, :input, :keygen,
-    :link, :meta, :param, :source, :track, :wbr,
-])
+const BLOCK_ELEMENTS = Set(
+    [
```
mortenpi, line 109: I do like the conciseness of "merging" the two parentheses.

mortenpi (commented):
I skimmed through the PR and I did not see anything really weird, so generally LGTM.

I did leave some questions and thoughts.  The ones where I'm like "I don't 100% like this formatting choice" are more just me thinking out loud. I don't really want to second guess the decisions you made in Runic, and I think the formatting bikeshedding ship there has probably passed, and so I am perfectly happy with the formatter, even if it's not a formatting change I would make myself.

If this PR contains both manual and automated changes (seems like it does), and it has been open for a while with changes to `master` in the meanwhile, I am a little concerned merging it as is. Maybe we could open a separate PR that _just_ runs Runic and adds the workflow, to get all the automated changes in? And then resolve the diffs/conflicts with this PR? It would make it easier to spot any issues and bad merges etc, if there are any.

mortenpi:
The test failures

> MethodError: Cannot `convert` an object of type Nothing to an object of type String

do imply that somewhere we're maybe too aggressively `return`-ing `nothing`s now?

src/DocSystem.jl
fredrikekre: It removes empty lines so that there are at max two emtpy lines between code.

src/docchecks.jl
fredrikekre: No, formatter did:
```
hascurl() = (
    try
        success(`curl --version`)
    catch err
        false
    end
)
```
(It doesn't allow single-line block constructs).

src/docchecks.jl
fredrikekre: Yea it just falls out naturally from the "rule set", but I have grown to like it because it makes it easier to see that this is just one thing and not a three-tuple, compare
```
GITHUB_ERROR_ADVICE = (
    "This means automatically finding the source URL link for this package failed. " *
    "Please add the source URL link manually to the `remotes` field " *
    "in `makedocs` or install the package using `Pkg.develop()``."
)
```
vs
```
GITHUB_ERROR_ADVICE = (
    "This means automatically finding the source URL link for this package failed. ",
    "Please add the source URL link manually to the `remotes` field ",
    "in `makedocs` or install the package using `Pkg.develop()``."
)
```
vs
```
GITHUB_ERROR_ADVICE = (
    "This means automatically finding the source URL link for this package failed. " *
        "Please add the source URL link manually to the `remotes` field " *
        "in `makedocs` or install the package using `Pkg.develop()``."
)
```

Slightly more useful if there are more things though:
```
GITHUB_ERROR_ADVICE = (
    "This means automatically finding the source URL link for this package failed. " *
        "Please add the source URL link manually to the `remotes` field " *
        "in `makedocs` or install the package using `Pkg.develop()``.",
    "This means automatically finding the source URL link for this package failed. " *
        "Please add the source URL link manually to the `remotes` field " *
        "in `makedocs` or install the package using `Pkg.develop()``."
)
```
etc

src/doctest.jl
fredrikekre: Yea it is manual, formatter does
```
throw(
    ArgumentError(
        """
        Package $(package) does not have a documentation source directory at standard location.
        Searched at: $(source)
        If ...
        """
    )
)
```

src/html/HTMLWriter.jl
fredrikekre: Formatter adds `return if`. I considered letting the formatter recurse into branches, but then probably it should also add a `else return nothing` branch and I didn't really like that (https://github.com/fredrikekre/Runic.jl/issues/52).

After formatting a couple of code bases it is roughly 50/50 where you actually want to return things or if `return nothing` after the `if` is what is really intended.

src/utilities/DOM.jl
fredrikekre: Yea I have thought about special-casing when there is just one thing inside. This is a bit similar to the `throw(ArgumentError("msg"))` example above.

fredrikekre:
> do imply that somewhere we're maybe too aggressively return-ing nothings now?

Yea, but should be a manual change by me. I like your suggestion with first merging a "raw" formatted patch and then clean up after.

fredrikekre:
Rebased on master so this now only contains the manual changes. Added another commit to add Runic formatting to the changelog and to the contributing docs. Lets keep these commit separate so we can add the first to blame-ignore.

PR #2571, opened 2024-09-03
Fix linkcheck_useragent implementation and tests

mortenpi:
* Changes the `linkcheck_useragent` value that unsets the customization completely to `nothing`.
* Fixes an error where linkcheck would throw when `linkcheck_useragent = ""` because the `if` block returns a `nothing` then.
* Apparently, Intel already blocked the new user agent or something. So reorganizing the tests around the fact that they want the `curl` user agent.

### DIFF
--- a/test/online_linkcheck.jl
+++ b/test/online_linkcheck.jl
@@ -16,12 +16,32 @@ using Test
             [FTP (no proto) success](ftp.iana.org/tz/data/etcetera)
             [Redirect success](google.com)
             [HEAD fail GET success](https://codecov.io/gh/invenia/LibPQ.jl)
-            [Linkcheck old Chrome UA fail](https://www.intel.com/content/www/us/en/developer/tools/oneapi/mpi-library.html)
             """
         )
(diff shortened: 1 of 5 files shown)


test/online_linkcheck.jl
```diff
@@ -25,6 +24,27 @@ using Test
...
+            md"""
+            [Linkcheck Empty UA](https://www.intel.com/content/www/us/en/developer/tools/oneapi/mpi-library.html)
+            """
+        )
+
+        # The default user-agent fails (intel servers block it)
+        doc = Documenter.Document(; linkcheck=true, linkcheck_timeout=20)
+        doc.blueprint.pages["testpage"] = Documenter.Page("", "", "", [], Documenter.Globals(), src)
+        @test_logs (:error,) @test linkcheck(doc) === nothing
+        @test doc.internal.errors == Set{Symbol}([:linkcheck])
```
fredrikekre, line 39: Looks like this doesn't error anymore.

PR #2573, opened 2024-09-13
Follow symlinks when walking the docs directory

fingolfin:
This way external directories with a bunch of markdown files can be easily included.

This is not a perfect solution but it works surprisingly well in practice. We've been using it for some time in Oscar.jl via some hacky monkey-patching...

### DIFF
--- a/src/builder_pipeline.jl
+++ b/src/builder_pipeline.jl
@@ -94,7 +94,7 @@ function Selectors.runner(::Type{Builder.SetupBuildDirectory}, doc::Documenter.D
     # Markdown files, however, get added to the document and also stored into
     # `mdpages`, to be used later.
     mdpages = String[]
-    for (root, dirs, files) in walkdir(source)
+    for (root, dirs, files) in walkdir(source; follow_symlinks=true)
         for dir in dirs
             d = normpath(joinpath(build, relpath(root, source), dir))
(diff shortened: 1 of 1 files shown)


fingolfin:
@fredrikekre @mortenpi thoughts on this? I could refine it but of course if you outright say "no way", I'll rather not... :-)

mortenpi:
Sorry for the delay. This seems fine to me to be honest. I also don't expect this to be breaking in practice -- since they don't work, I wouldn't expect any real `docs/src` directories to actually contain symlinks to directories.

@fredrikekre, can you foresee any potential issues with this?

fingolfin:
@mortenpi thank you. My colleague @aaruni96 will probably take over and work on a changelog and testcase.

For the testcase, we are trying to understand the test setup. From a few minutes looking, my naive idea would be to go into https://github.com/JuliaDocs/Documenter.jl/tree/master/test/examples/src and add a new file `symlink.md` which is a symlink to one of the other `.md` files (or a simple new file, or maybe to an `.md` file in an outside directory). And then perhaps after the docs are built, check if there is a `symlink.html` and whether its content is identical (or close enough) the the output produced for whatever file the symlink pointed at.

Does that sound plausible? Or any better suggestions? Anything we should watch out for?

Issue #2576, opened 2024-09-15
Breakage: Custom theme builds fail now when compiling the dark theme.


Datseris:
Hello, the documentation builds for JuliaDynamics  are no suddenly failing with the message ERROR: LoadError: Undefined variable: `"$todo"`.. This error occurs when we try to compile our own theme. It occurs when we call Themes.compile for the dark theme. The script we run to compile the theme is this one: https://github.com/JuliaDynamics/doctheme/blob/master/build_docs_with_style.jl . A stack trace can be found here for example: https://github.com/JuliaDynamics/PeriodicOrbits.jl/actions/runs/10847716300/job/30103351759?pr=3#step:5:8 . 

Some (unknowingly) breaking change in documenter themeing triggers this now.

The reason I believe that this is a Documenter.jl problem is that there is no `$todo` in the whole repo that has the doc theme, nor in the repo that the docs are build for.
This is where the error is coming from: https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/assets/html/scss/documenter-dark.scss#L35
We don't re-color any admonitions in our theme, so I don't really understand why the `$todo` admonition is the one that errors.

fingolfin:
@Datseris there is a proposed fix in PR #2579 could you verify that it solves the issue for you?

Datseris:
Hi, I am sorry, I do not have this issue any more, because I have removed this line. I am sorry, I am swamped with work and won't be able to tryout the fix.

fingolfin:
No worries @Datseris . I actually managed to reproduce the issue in your repository and created a proper fix, see PR #2664.

Thank you for taking the time to report this in the first place!

PR #2579, opened 2024-09-17
Add 'todo' field to dark mode settings

jd-foster:
Possibly addresses: #2576 
cf. https://github.com/JuliaDynamics/PeriodicOrbits.jl/pull/3#issuecomment-2348661221

### DIFF
--- a/assets/html/scss/documenter-dark.scss
+++ b/assets/html/scss/documenter-dark.scss
@@ -37,7 +37,7 @@ $admonition-default: $grey-lighter;
 
 $admonition-background: (
   'default': $background, 'info': $background, 'success': $background, 'warning': $background,
-  'danger': $background, 'compat': $background
+  'danger': $background, 'compat': $background, 'todo', $background
 );
 // These would normally get initialized in _admonition.scss. However, since we are not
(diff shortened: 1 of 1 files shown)


assets/html/scss/documenter-dark.scss
```diff
@@ -37,7 +37,7 @@ $admonition-default: $grey-lighter;
 
 $admonition-background: (
   'default': $background, 'info': $background, 'success': $background, 'warning': $background,
-  'danger': $background, 'compat': $background
+  'danger': $background, 'compat': $background, 'todo', $background
```
jd-foster, line 40:
```suggestion
  'danger': $background, 'compat': $background, 'todo': $background
```

fredrikekre:
> Possibly addresses: https://github.com/JuliaDocs/Documenter.jl/issues/2576

Can you or @Datseris check this perhaps? I also think you need to recompile the themes (there is a makefile in the themes folder).

mortenpi:
> I also think you need to recompile the themes (there is a makefile in the themes folder).

The CI check passes, so I don't think this affects the generated CSS actually.

PR #2589, opened 2024-10-17
Suggested using julia-docdeploy workflow

arindamsaha1507:
I think this simplifies the workflow a bit?

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -202,13 +202,10 @@ jobs:
         with:
           version: '1'
       - uses: julia-actions/cache@v2
-      - name: Install dependencies
-        run: julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
-      - name: Build and deploy
+      - uses: julia-actions/julia-docdeploy@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # If authenticating with GitHub Actions token
           DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # If authenticating with SSH deploy key
-        run: julia --project=docs/ docs/make.jl
 ```
 
 This will install Julia, checkout the correct commit of your repository, and run the


fredrikekre:
I think that action is a bit opinionated and, for example, doesn't work when you commit the docs manifest (which I think everyone should do). Is this action used much in the wild?

giordano:
> I think that action is a bit opinionated and, for example, doesn't work when you commit the docs manifest

But at least it avoids doing the manual `'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'` dance.  BTW, using the `sources` section helps a lot here, now that's available in Julia v1.11.

> (which I think everyone should do).

I don't agree on this, but that's orthogonal to this discussion.

> Is this action used much in the wild?

4.3k repos according to https://github.com/julia-actions/julia-docdeploy/network/dependents

fredrikekre:
> But at least it avoids doing the manual 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()' dance. 

Yea but I don't want this dance :) Perhaps if `Pkg.instantiate` can move out to a separate step and then the if you have a manifest you can use the existgin `install-package` input to disable the `Pkg.develop`.

> 4.3k repos 

Oh wow, I guess because PkgTemplates or something adds this? Roughly 25% of all repos with gh actions CI then (using setup-julia as a counter of total number of repos).

giordano:
> Yea but I don't want this dance :)

Me neither, and I much prefer the `[sources]` approach, which is cleaner and doesn't require the Manifest 🙂

mortenpi:
That action has had some issues in the past as well, so I am a little hesitant to recommend it as the default. But I do think that, in general, we should have an action for this, to simplify the workflow files for the users.

A starting point would be to maybe just link to the action, rather than defaulting to it? There are cases where you probably need to roll your own workflow anyway (some cases not covered by the action etc), so having the full workflow also has value I think.

fingolfin:
How about adding this as a second example, with some explanatory text? "One can also use `julia-docdeploy` like in THIS example. THis has the following pros and cons: bla bla "

PR #2590, opened 2024-10-18
add PauliStrings.jl to examples.md

nicolasloizeau:


mortenpi (approved):
LGTM. The CI failures are unrelated, but would be good to sort out before merging, so it might be a little while before we merge this.

PR #2593, opened 2024-10-29
Fix the race condition causing search not to initialize properly

mortenpi:
Sometimes the search index takes longer to download than, and the `documenter.js` script already runs, in which case it will error with:

```
Uncaught ReferenceError: documenterSearchIndex is not defined
    <anonymous> https://docs.julialang.org/en/v1/assets/documenter.js:603
    execCb https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    check https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    enable https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    bind https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    emit https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    each https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    emit https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    check https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    enable https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    init https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    a https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    completeLoad https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
    onScriptLoad https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js:1
```

and the search UI doesn't work.

This effectively adds a loop with a `sleep` to the search code that will wait until the `documenterSearchIndex` is actually available, before running any of the search initialization code.


Hetarth02:
@mortenpi 

As far as the code looks, it just delays the function call by 1 second if the index is `undefined`.

Shouldn't we keep looping or something and give a bit more time for the index to load(slow networks/other problems). Then for example after 10 seconds the index doesn't load we could give an alert/console.warn something?

mortenpi:
It does loop. `setTimeout` will run `waitUntilSearchIndexAvailable` again, which will again check for the search index, and delay again if need be.

PR #2594, opened 2024-10-29
Fix doctest testing on Julia 1.11

fredrikekre:


fredrikekre:
Nightly fails because https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/test/examples/references/latex_showcase.tex#L1243 becomes `\begin{minted}{julia}` (from https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/test/examples/make.jl#L65). Probably due to a change in Markdown that considers code blocks to be julia code by default?

PR #2602, opened 2024-11-07
Improve warnings in fix_doctest

### LINKED ISSUE #2601: Improve error message `could not find input line in code block` by indicating location of that code block
I just did execute `makedocs` with `doctest = :fix` in a package to adjust docstrings to some printing tweaks I made, and got this very frustrating message:
```
┌ Warning: could not find input line in code block
└ @ Documenter ~/.julia/packages/Documenter/C1XEF/src/doctests.jl:507
```
I call it frustrating because all I know now is that "some docstring somewhere in your package has a problem, maybe". That's not a lot to go on.

If I knew which docstring it was, at least I'd have a chance.

If there is any way to improve this message to tell me more, that'd be awesome. Ideally of course the file and starting line of the code block, but even just the file would be a great start. Or a snippet of the content. Anything?

fingolfin:
Resolves #2601

TODO:
- changelog
- test case
- perhaps there is a better way to get the startline?
- perhaps the warning could be improve further by showing what content it failed to match

### DIFF
--- a/src/doctests.jl
+++ b/src/doctests.jl
@@ -490,13 +490,14 @@ function fix_doctest(result::Result, str, doc::Documenter.Document; prefix::Muta
     r = Regex(rcode)
     codeidx = findfirst(r, content)
     if codeidx === nothing
-        @warn "could not find code block in source file"
+        @warn "could not find code block in source file $filename"
         return
     end
     # use the capture group to identify indentation
     indent = match(r, content).captures[1]
     # write everything up until the code block
-    write(io, content[1:prevind(content, first(codeidx))])
+    data = content[1:prevind(content, first(codeidx))]
+    write(io, data)
     # next look for the particular input string in the given code block
     # make a regex of the input that matches leading whitespace (for multiline input)
     composed = prefix.content * result.raw_input
@@ -504,7 +505,8 @@ function fix_doctest(result::Result, str, doc::Documenter.Document; prefix::Muta
     r = Regex(rinput)
     inputidx = findfirst(r, code)
     if inputidx === nothing
-        @warn "could not find input line in code block"
+        startline = count("\n", data)
+        @warn "could not find input line in code block in $filename:$startline"
         return
     end
     # construct the new code-snippet (without indent)


fingolfin:
@mortenpi does this seem sane? If so I'll go ahead and add tests and a changelog entry.

src/doctests.jl
```diff
@@ -490,21 +490,23 @@ function fix_doctest(result::Result, str, doc::Documenter.Document; prefix::Muta
     r = Regex(rcode)
     codeidx = findfirst(r, content)
     if codeidx === nothing
-        @warn "could not find code block in source file"
+        @warn "could not find code block in source file $filename"
```
fingolfin, line 493: Actually it would be kinda helpful if one would see the code block. Any reason why I shouldn't just print it here?

src/doctests.jl
```diff
@@ -490,21 +490,23 @@ function fix_doctest(result::Result, str, doc::Documenter.Document; prefix::Muta
...
     # next look for the particular input string in the given code block
     # make a regex of the input that matches leading whitespace (for multiline input)
     composed = prefix.content * result.raw_input
     rinput = replace(Documenter.regex_escape(composed), "\\n" => "\\n\\h*")
     r = Regex(rinput)
     inputidx = findfirst(r, code)
     if inputidx === nothing
-        @warn "could not find input line in code block"
+        startline = count("\n", data)
+        @warn "could not find input line in code block in $filename:$startline"
```
fingolfin, line 509: And here the "current" code block which was found: any particular reason not to print it here? Would have been helpful to me in the past

src/doctests.jl
mortenpi: I think that would be fine. I think we do that for some other messages already.

mortenpi:
I haven't tested this, but from the looks of it, LGTM. Any extra context like this in the error messages is good.

It would be good to have a test case, yes. We have some tests here for the fixing here (https://github.com/JuliaDocs/Documenter.jl/tree/master/test/doctests/fix). Not sure how easy it would be to expand those though to somehow test the new things. Maybe a few simple `occursin` checks on the output would be enough though.

fingolfin:
(Un?)fortunately I currently don't have any tests where this warning is relevant, and I don't quite remember what caused the issue in the case that made me make this PR to track it down. So I have a hard time creating a suitable test, or improving this further.

At the same time, it really worked well and it is a pity this is not included in Documenter.

So perhaps the rules for merging could be relaxed here a bit?

Otherwise, if I ever run into this again, I surely will try to revive this. If anyone else runs into the issue and finds this useful, let me know how you triggered it.

mortenpi (approved):
Let's merge this as is, I'm happy with that! Took the liberty of adding a CHANGELOG note.

fingolfin:
Thank you @mortenpi

PR #2619, opened 2025-01-20
Support substitutions in doctestfilters

### LINKED ISSUE #2360: `doctest` `filter` keyword does not accept substitution
Is there a reason the `filter` keyword does not accept substitutions? I am trying to define a filter for all doctests (e.g. for comparing floating point numbers only consider up to 10 decimal values): 

```julia
julia> doctest(MyPackage; doctestfilters = [(r"(\d*).(\d{1,10})\d+" => s"\1.\2"),])
...
TypeError: in keyword argument doctestfilters, expected Vector{Regex}, got a value of type Vector{Pair{Regex, SubstitutionString{String}}}
...
```

passing this filter to each `jldoctest` individually works, but I would rather do it just once.

sethaxen:
As noted in #2360, `doctestfilters` currently only supports vectors of `Regex`, while `filters` supports vectors including regex/substitution pairs. This PR fixes #2360, supporting substitutions also in `doctestfilters`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


mortenpi (approved):
This LGTM, thank you! Sorry for the delay in getting this in.

Issue #2621, opened 2025-01-28
Detected access to binding `MyModule.##docmeta#274` in a world prior to its definition world.


lgoettgens:
Since https://github.com/JuliaLang/julia/pull/57133, we see a lot of these issues in our CI at https://github.com/oscar-system/GAP.jl, when running doctests with julia nightly.

I managed to break it down to a very small mwe:
```julia
julia> using Documenter

julia> module Foo
       end
Main.Foo

julia> DocMeta.setdocmeta!(Foo, :DocTestSetup, :(using Foo); recursive=true)
WARNING: Detected access to binding `Foo.##docmeta#274` in a world prior to its definition world.
  Julia 1.12 has introduced more strict world age semantics for global bindings.
  !!! This code may malfunction under Revise.
  !!! This code will error in future versions of Julia.
Hint: Add an appropriate `invokelatest` around the access to this binding.
```

cc @keno

Keno:
Yes, the documenter code needs to be fixed. Usually if have done that first, but were trying to get the first rc out, so I'm focusing on the Julia base changes first.

fingolfin:
@Keno well @lgoettgens also submitted a potential fix as PR #2622 so I hope that can be merged and released quickly

PR #2622, opened 2025-01-28
Fix world age warning

lgoettgens:
Resolves https://github.com/JuliaDocs/Documenter.jl/issues/2621.

cc @fingolfin @Keno

pinging @mortenpi @fredrikekre for review


mortenpi:
Any ideas why the Windows CI is still failing? It seems to be due to the same issue, but I'm confused why it would only affect Windows..

PR #2629, opened 2025-02-11
Replace `isexported` with `ispublic` as criterion for `@autodocs`

gdalle:


src/expander_pipeline.jl
```diff
@@ -524,8 +524,12 @@ function Selectors.runner(::Type{Expanders.AutoDocsBlocks}, node, page, doc)
         for mod in modules
             for (binding, multidoc) in Documenter.DocSystem.getmeta(mod)
                 # Which bindings should be included?
-                isexported = Base.isexported(mod, binding.var)
-                included = (isexported && public) || (!isexported && private)
+                ispublic = if isdefined(Base, :ispublic) # Julia v1.11 and later
```
asinghvi17, line 527: This could potentially be
```suggestion
                ispublic = @static if isdefined(Base, :ispublic) # Julia v1.11 and later

**Comment by asinghvi17** [comment]
Thanks!  I'm happy to approve this PR, but if you'd like to add a test I think Compat.jl has an `@public` macro you can use.  `public x` won't parse in Julia < 1.11.

File: test/examples/pages/e.jl [hunk 2]
```diff
@@ -1,6 +1,17 @@
 module E
 
-export f_1, f_2, f_3
+# https://discourse.julialang.org/t/is-compat-jl-worth-it-for-the-public-keyword/119041/
+macro public_or_export(ex)
+    args = ex isa Symbol ? (ex,) : Base.isexpr(ex, :tuple) ? ex.args : error()
+    if Base.isdefined(:ispublic)
```
asinghvi17, line 6:
```suggestion
    if Base.isdefined(Base, :ispublic)
```

gdalle:
I think this is good to go

gdalle:
However I think this could break some workflows by including docstrings which are documented elsewhere, triggering a "duplicate docstrings" error. How does Documenter handle SemVer for this sort of thing?

mortenpi:
> However I think this could break some workflows by including docstrings which are documented elsewhere, triggering a "duplicate docstrings" error. How does Documenter handle SemVer for this sort of thing?

Yes, this is my main concern right now. I think this creates a behavior change if you have a package that already uses the `public` keyword? And `public` is in 1.11, so it's very much out in the wild by now. That feels like a breaking change.

Since Documenter is normally auto-upgraded, the policy is to be conservative and not introduce breaking changes of that sort.

The easy solution is to guard this with an option to `makedocs`, but I wonder if there's something I'm missing, and there's some better solution?

asinghvi17:
We could make public-but-not-exported docstrings non canonical by default?  That would at least not be breaking, and maybe we could also issue a warning that contains all public-but-not-exported docstrings that have been documented elsewhere.

This might also need a change to the check on whether all exported docstrings are documented - maybe we need to expand that criterion to all public or exported docstrings.

gdalle:
> This might also need a change to the check on whether all exported docstrings are documented - maybe we need to expand that criterion to all public or exported docstrings.

That would be breaking for the same kind of reason I guess.

gdalle:
> We could make public-but-not-exported docstrings non canonical by default? That would at least not be breaking, and maybe we could also issue a warning that contains all public-but-not-exported docstrings that have been documented elsewhere.

Would that really be non-breaking? It may leave some symbols without any canonical docstring, whereas they had one before. Does that mean the links are dangling?

mortenpi:
> This might also need a change to the check on whether all exported docstrings are documented - maybe we need to expand that criterion to all public or exported docstrings.

We should be consistent, yes.. but it depends a bit on how we're phrasing things right now.

> We could make public-but-not-exported docstrings non canonical by default?

I think this will become confusing to users down the line.

Thinking a bit more -- maybe we can be okay with it, if we add a big warning in the CHANGELOG? It would be a shame not to have `ispublic` the default behavior. So, not supporting `public` could be considered a bug. Ideally we would have done this change during the 1.11 development cycle already.

Would it be feasible to check that, if a duplicate docstring error gets detected, we check if it's due to the "public-but-not-exported" case, and print a warning with the context? We'd might still break existing workflows, but we can immediately tell the user what's going on.

We should still do the change in a minor release though, due to it's slightly "technically breaking change" character.

mortenpi:
> This might also need a change to the check on whether all exported docstrings are documented

Looking at `checkdocs` -- we default to `:all` (which should include non-public ones), and then we have `:exported` for exports. I think we just add `:public` there, to check for all `ispublic` docstrings.

gdalle:
> Would it be feasible to check that, if a duplicate docstring error gets detected, we check if it's due to the "public-but-not-exported" case, and print a warning with the context? We'd might still break existing workflows, but we can immediately tell the user what's going on.

Done in the latest commit.

> Looking at checkdocs -- we default to :all (which should include non-public ones), and then we have :exported for exports. I think we just add :public there, to check for all ispublic docstrings. So I think we're fine here w.r.t. to any breakage.

Done in the latest commit.

---

Do we need additional tests for these two changes? I think we would want at least one for the new warning, otherwise that line won't get hit. Where should it go?

mortenpi:
And thinking even more: I think giving context on the warning is a good idea also because you can run into this scenario:

- You have a package with no `public` markings, and you run doc builds on e.g. 1.10.
- You use Compat.jl to add `public` marking, but keep running doc builds on 1.10 for now.
- You update the Julia version for the doc build, and you get errors / a behavior change.

But this is fine too, since the "problem" is actually the user [declaring things public conditionally on the Julia version](https://github.com/JuliaLang/Compat.jl/blob/8819abe252579b610cb48c5ad83d51b45a90ddba/src/compatmacro.jl#L43C3-L50). So we can argue that it's not Documenter's fault. But we can still hopefully provide a useful error message.

---

All in all, I think I'm warming up to doing the "breaking change" here. But let's leave this open for a week or so for feedback. And let me also solicit opinions from @fredrikekre, @goerz and @odow.

mortenpi:
> Do we need additional tests for these two changes? I think we would want at least one for the new warning, otherwise that line won't get hit. Where should it go?

Maybe here? https://github.com/JuliaDocs/Documenter.jl/blob/master/test/errors/make.jl

gdalle:
I think there's another place where the warning should be included in case of duplicate docstrings, namely here: https://github.com/gdalle/Documenter.jl/blob/580be25b7b0f358674770909e799da270cf26308/src/expander_pipeline.jl#L421-L433

However I can't seem to get the module in which to test whether names are public or private. Can someone help?

mortenpi:
@gdalle I took some liberties in re-arranging the code a bit as I was thinking what to do with the other error.

I _think_ what we can do is just use the `.mod` field of the binding object there. I think there might be cases where that's not quite consistent, because `ispublic` for a function depends on the module. I.e. consider this:

```
julia> module Foo; import Base: cd; end
Main.Foo

julia> module Bar; import Base: cd; public cd; end
Main.Bar

julia> Base.ispublic(Base, :cd)
true

julia> Base.ispublic(Main, :cd)
false

julia> Base.ispublic(Foo, :cd)
false

julia> Base.ispublic(Bar, :cd)
true
```

And I am not off the top of my head 100% sure what this implies about how we even include docstrings. But since this is just about printing an informative error, I think we can live with any potential inconsistencies.

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Symlinks are now followed when walking the docs directory. ([#2610])
 * PDF/LaTeX builds now throw a more informative error when `sitename` is not provided. ([#2636])
+* `@autodocs` now lists public unexported symbols by default (i.e. when `Public = true`). ([#2629])
+
+  This is **potentially breaking** as it can cause previously working builds to fail if they are being run in strict mode.
+  This can happen if there are unexported docstring marked with `public` that are being included manually with e.g. `@docs` blocks, and there is also an `@autodocs` block including all public docstrings.
+  The solution is to remove the duplicate inclusion.
+
+* `checkdocs` has a new option `:public` to check for unincluded public symbols that are marked with `public`, but are unexported. ([#2629])
```
gdalle, line 18:
```suggestion
* `checkdocs` has a new option `:public` to check that unexported symbols marked with `public` are included in the docs. ([#2629])
```

src/makedocs.jl
```diff
@@ -174,8 +174,9 @@ by setting `Draft = true` in an `@meta` block.
 **`checkdocs`** instructs [`makedocs`](@ref) to check whether all names within the modules
 defined in the `modules` keyword that have a docstring attached have the docstring also
 listed in the manual (e.g. there's a `@docs` block with that docstring). Possible values
-are `:all` (check all names; the default), `:exports` (check only exported names) and
-`:none` (no checks are performed).
+are `:all` (check all names; the default), `:exports` (check only exported names),
+`:public` (check exported names and those marked with the `public` keyword in Julia ≥),
```
gdalle, line 178:
```suggestion
`:public` (check exported names and those marked with the `public` keyword in Julia ≥ 1.11),
```

src/DocSystem.jl
```diff
@@ -315,4 +315,43 @@ function parsedoc(docstr::DocStr)
...
+
+"""
+This error message is reused in duplicate docstring warnings when we detect
+the case when a duplicate docstring in a non-explored public name.
+"""
+function public_unexported_msg(apistatus::APIStatus)
+    return if apistatus.ispublic && !apistatus.isexported
+        """\n
+        Note: this binding is marked `public`, but not exported. Starting from Documenter 1.9.0,
+        these are now included in `@autodocs` blocks that include public APIs with `Public = true`,
```
gdalle, line 346:
```suggestion
        such bindings are now included in `@autodocs` blocks that list public APIs (blocks with `Public = true`),
```

gdalle (commented):
The refactor looks good to me, I made a few suggestions on error messages. I'm okay with you merging this, but I wonder whether more tests should be added first. If so, please feel free to take over and add them wherever you feel they would be relevant.

gdalle:
Note that codecov fails to run, so we're not sure the new code is hit during tests

gdalle:
Gentle bump on this @mortenpi

mortenpi:
Thank you @gdalle! I think let's go and merge this!

> but I wonder whether more tests should be added first.

I think we're okay, let's not postpone this. We can always add more tests later.. for example together with any bugfix PRs that might be necessary :smile:

gdalle:
Awesome! I'm excited to bring the full power of the `public` keyword to the world 😎

gdalle:
What's the release process for Documenter? is it triggered automatically?

mortenpi:
It's manual, but let's aim to get the release out this week. There are a few more PRs that are probably ready to merge, so we can bundle them all together in one release.

mortenpi:
I know, sorry. There are a few PRs I'd hope we can get into the same release though: https://github.com/JuliaDocs/Documenter.jl/pull/2654

gdalle:
Sounds good!

PR #2636, opened 2025-02-20
Throw an informative error if `sitename` is missing when building LaTeX

asinghvi17:
Just lost a bunch of time tracking this down :D

### DIFF
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -107,6 +107,13 @@ const DOCUMENT_STRUCTURE = (
 )
 
 function render(doc::Documenter.Document, settings::LaTeX = LaTeX())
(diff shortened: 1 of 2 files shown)


src/latex/LaTeXWriter.jl
```diff
@@ -107,6 +107,13 @@ const DOCUMENT_STRUCTURE = (
 )
 
 function render(doc::Documenter.Document, settings::LaTeX = LaTeX())
+    if isempty(doc.users.sitename) # otherwise, the latex compiler will terminate with a cryptic "There's no line here to end" error
```
mortenpi, line 110:
```suggestion
    if isempty(doc.user.sitename) # otherwise, the latex compiler will terminate with a cryptic "There's no line here to end" error
```

asinghvi17: oops :D

mortenpi:
Thanks @asinghvi17, sorry for the delay!

PR #2642, opened 2025-03-03
When fixing doctests with errors, don't overwrite `[...]`

### LINKED ISSUE #2511: Doctest fix versus exceptions
Consider a doctest looking like this, testing a [thrown exception with omitted stacktrace](https://documenter.juliadocs.org/stable/man/doctests/#Exceptions):

````
```jldoctest
julia> error("x")
ERROR: x
```
````

It is very nice that Documenter allows me to write that, so the paths and line numbers from the stack trace are omitted.

Alas, it seems that code for [fixing outdated doctests](https://documenter.juliadocs.org/stable/man/doctests/#Fixing-Outdated-Doctests) is not quite aware of this. Running that turns the above back into
````
```jldoctest
julia> error("x")
ERROR: x
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[2]:1
```
````
One then has to carefully edit those out again, preserving any "real" fixes.

Would it be possible to teach `fix` about the exception made for exceptions, so it preserves the output if it "matched" according to the exceptions rule?

fingolfin:
Sometimes it is useful to show errors in doctests. However if those contain stacktraces, this can be problematic as those contain line numbers which will change over time (and often also when e.g. Julia or some dependencies change). And sometimes it is simply desirable to shorten the error message for other reasons.

For this reason Documenter supports replacing part of the error message with `[...]` which can be used to omit e.g. a stacktrace while still passing doctest checks.

However so far using Documenter's `doctest=:fix` feature would replace all uses of `[...]` by the full error message, making it unnecessarily tedious to update doctests in packages that use this feature.

This patch changes that by skipping doctest fixups for tests where the reference output ends with `[...]` and otherwise matches the actual error message as a prefix.

Resolves #2511

CC @thofma

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The solution is to remove the duplicate inclusion.
 
 * `checkdocs` has a new option `:public` to check that unexported symbols marked with `public` are included in the docs. ([#2629])
+* Fixing doctests that use `[...]` to hide part of an error message (such as a stacktrace) no longer replaces the `[...]` if the output otherwise matches ([#2642])
 
 ## Version [v1.8.1] - 2025-02-11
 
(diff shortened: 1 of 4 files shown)


fingolfin:
@mortenpi @fredrikekre any thoughts on this? it works well for me and saves me quite some work :-)

mortenpi (approved):
This makes a lot of sense to me, and LGTM. Thank you!

Issue #2543, opened 2024-07-02
Can we make search modify the URL


EssamWisam:
I was willing to integrate documenter's search with an external site for one package that I'm working on. 
<img width="1373" alt="image" src="https://github.com/JuliaDocs/Documenter.jl/assets/49572294/7546a07e-3bc4-4b36-9cf6-3be40eff49d6">
However, that seems impossible because the current search implementation doesn't the URL in any way. It would be nice to have search function similar to common search engines, where the url is modified based on the input.

mortenpi:
We actually used to do that, and I think it should doable with the new search as well, where we just look at and update a special query parameter.

> integrate documenter's search with an external site

Do you mean you want to be able to link to the search results? Or do you mean something else?

EssamWisam:
> We actually used to do that, and I think it should doable with the new search as well, where we just look at and update a special query parameter.
> 
> > integrate documenter's search with an external site
> 
> Do you mean you want to be able to link to the search results? Or do you mean something else?

Yes, to be able to link it with the link being meaningful. An example from Google:
```
https://www.google.com/search?q=julia+documenter
```
First, it can be linked and second, it allows integration with another site since any external site can do 
```
https://www.google.com/search?q=${custom-search-term}
```

mortenpi:
Pre-1.0 that used to be possible: https://documenter.juliadocs.org/v0.27/search/?q=deploydocs

I'd say we can bring this back if someone wants to take a stab at implementing it. The only caveat is that there wouldn't be  a dedicated search page, but it would probably work on every page. I guess the canonical choice can just be the root page though.

Hetarth02:
Happy to take a try at this.

Rahban1:
would love to work on this.

PR #2646, opened 2025-03-05
Added the feature of search sharing by adding query param to the search. Issue [#2543]

Rahban1:


Rahban1:
I apologize for this silliness, I tried to make the prettier version from 3.5.3 to 3.0.0, tried manually correcting linting issues but the prettier test is just not passing

Rahban1:
All the code changes are in first commit only, rest all are my attempts to pass prettier test

mortenpi:
For formatting, I am not sure what editor you use, but I just let the VS Code's Prettier extension do it for me: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

Rahban1:
Thank you so much, will keep this in mind for future PRs

mortenpi:
I was testing it, and this seems to work pretty well! I'd like to rope @Hetarth02 into looking over the JS.

Two thoughts that I had:

1. I think we might want to clear the `?q=...` when you close the search modal? Right now, if you exit the modal, it clears the search. I'm not sure we want to change that? Any thoughts?
2. We also have the filter badges -- should we also put those in the query parameters? Maybe something like `?filter=function&q=...`?

Rahban1:
now I think of it, I guess people only need to share the basic search query. filters are often personal preferences applied during interaction. What do you think?

Hetarth02:
> filters are often personal preferences applied during interaction.

While this is true, I think in regards to the search in documentation adding filters are useful as they arrow down the search. Moreover, when people share the search link it should reflect the same state as in the UI. So, in summary would be good if you can add the filters in the url.

Also, @mortenpi do you think we should have a copy link(share) button or something through which users can easily share the link rather than copying it from the URL bar?

Rahban1:
makes sense. I'll try implementing this.

I would second the button idea! It would make sharing much more intuitive. Happy to add this to complete the feature.

Hetarth02:
> Happy to add this to complete the feature.

If possible can you post some screenshots of the feature as to how it might look. Just UI mockups, nothing more.

Rahban1:
On it!

Rahban1:
> > Happy to add this to complete the feature.
> 
> If possible can you post some screenshots of the feature as to how it might look. Just UI mockups, nothing more.

this placement looks fine to me [
Screenshot 2025-03-06 at 11 24 53 PM
](url)

Rahban1:
Apologies for the delay, I am in the middle of my practical examinations :)

Rahban1:
we can change the font, add hover effects, and add a confirmation when the link is copied to clipboard

Rahban1:
Another idea is to add an SVG like this, IMO with the same placement, do tell me what do you think about the button placement
Screenshot 2025-03-06 at 11 29 25 PM

Hetarth02:
> Apologies for the delay, I am in the middle of my practical examinations :)

No Worries! 😄

> what do you think about the button placement

I like the button idea, but can we put the button on right hand side of the search input. Can you do a mockup for this, please?

Moreover, I was also thinking we can then hide/display button when the search input has any value in it. But this is for later.

Rahban1:
Screenshot 2025-03-07 at 9 37 50 AM

Rahban1:
thoughts???

mortenpi:
Yea, let's put it to the right of the text box. And let's make it use that chain icon (no text -- that can be a tooltip I think). Users should be somewhat familiar with the chain UI, since we use it for the heading deep links:

image

Rahban1:
It will look something like this :-
Screenshot 2025-03-07 at 11 33 45 AM

another idea which I had in mind is we can do something like this also :-
Screenshot 2025-03-07 at 11 35 19 AM

would love to know your thoughts!

assets/html/js/search.js
```diff
@@ -404,18 +404,33 @@ function runSearchMainCode() {
...
+      if (url.searchParams.has("filter")) {
+        url.searchParams.delete("filter");
+      }
+    }
+
+    // Only update history if there are parameters, otherwise use the base URL
+    if (url.search) {
+      window.history.replaceState({}, "", url);
+    } else {
+      window.history.replaceState({}, "", url.pathname + url.hash);
```
Rahban1, line 432: I am dicey about concatenating url.hash, please advise to remove it or add it??

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Whatever you search will be added to the URL as a query parameter and you can use that link to share search results. ([#2543])
+
+### Added
+
```
Hetarth02, line 13:
```suggestion
* Whatever you search will be added to the URL as a query parameter and you can use that link to share search results. ([#2543])

```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Whatever you search will be added to the URL as a query parameter and you can use that link to share search results. ([#2543])
```
Hetarth02, line 10:
```suggestion
* Whatever you search will be added to the URL as a query parameter and you can use that link to share search results. ([#2543], [#2646])
```

Once the PR is up, it's good to add it to the log.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Added
```
Hetarth02, line 12:
```suggestion
```

We can just add the change inside the existing `### Added` block. No need for separately marking it.

Also, it's good to run `make changelog` whenever you make some changes to `CHANGELOG.md` file.

CHANGELOG.md
Rahban1: Will keep this in mind. Thanks!

assets/html/js/shortcut.js
```diff
@@ -85,6 +98,13 @@ $(document).ready(function () {
       <div class="has-text-centered my-5 py-5">Type something to get started!</div>
     `;
 
+    //removing the query param when the modal is closed
+    if (window.history.replaceState) {
+      const url = new URL(window.location.href);
+      url.searchParams.delete("q");
+      window.history.replaceState({}, document.title, url.toString());
+    }
+
```
Hetarth02, line 107: I think a better implementation might be,

```js
    const url = new URL(window.location.href);
    const keys = [...url.searchParams.keys()];

    // Removing all query params
    if (keys.length) {
      keys.forEach((value) => url.searchParams.delete(value));
      window.history.replaceState(null, "", url.toString());
    }
```

Happy to take feedback, if I have misunderstood something.

assets/html/js/search.js
```diff
@@ -515,6 +548,29 @@ function runSearchMainCode() {
...
+    if (filterParam) {
+      selected_filter = filterParam.toLowerCase();
+    }
+
+    // Trigger input event if there's a search query to perform the search
+    if (searchQuery) {
+      $(".documenter-search-input").trigger("input");
+    }
+  }
+  setTimeout(checkURLForSearch, 100);
```
Hetarth02, line 572: Can we brainstorm some logic where we don't need to rely on `setTimeout`?

assets/html/js/search.js
```diff
@@ -515,6 +548,29 @@ function runSearchMainCode() {
...
+    // Set the selected filter if present in URL
+    if (filterParam) {
+      selected_filter = filterParam.toLowerCase();
+    }
+
+    // Trigger input event if there's a search query to perform the search
+    if (searchQuery) {
+      $(".documenter-search-input").trigger("input");
+    }
+  }
```
Hetarth02, line 571: We can restructure this code as follows to avoid multiple `if` checks,

```suggestion
  function checkURLForSearch() {
    const urlParams = new URLSearchParams(window.location.search);
    const searchQuery = urlParams.get("q");
    const filterParam = urlParams.get("filter");

    if (filterParam) {
      selected_filter = filterParam.toLowerCase();
    }

    if (searchQuery) {
      // Opening of modal handled in shortcut.js
      // Trigger input event to perform the search
      $(".documenter-search-input").val(searchQuery).trigger("input");
    }
  }
```

assets/html/js/search.js
Hetarth02: The url should retain whatever the information it already has, which in our case is the `hash` as it redirects to the particular section of the page.

Hetarth02:
I also found a edge case, where if you close the modal it doesn't clear the selected filter and the next time we open the modal it has pre-selected filter.

Hetarth02:
> It will look something like this :- Screenshot 2025-03-07 at 11 33 45 AM
> 
> another idea which I had in mind is we can do something like this also :- Screenshot 2025-03-07 at 11 35 19 AM
> 
> would love to know your thoughts!

I am fine with either of them...

CC @mortenpi

Rahban1:
> I also found a edge case, where if you close the modal it doesn't clear the selected filter and the next time we open the modal it has pre-selected filter.

great catch!, I forgot to clear out the selected filter when you close the modal, I'll fix it.

assets/html/js/search.js
Rahban1: makes sense, great refactoring. Thanks!

assets/html/js/shortcut.js
Rahban1: you are right, I forgot about removing the filter from the url here, I think we can do just :- 
```javascript
  if(url.searchParams.size() > 0) {
    url.search = '';
    window.history.replaceState(null, "", url.toString());
  }
```
no need to make additional array IMO, thoughts?

Rahban1:
> I also found a edge case, where if you close the modal it doesn't clear the selected filter and the next time we open the modal it has pre-selected filter.

I need some help here, I want to clear the `selected_filter` so that pre-selected filter does not come up but it is declared in search.js and the close modal function is in shortcut.js, so I have two approaches in mind : 1) make the selected_filter in search.js global and use the same in shortcut.js 2) if we are not making `selected_filter` global in search.js we can create a custom event like this
 ```javascript
document.addEventListener('documenter:reset-filter', function() {
    selected_filter = "";
    update_search();
  });
```
and dispatch it in search.js to reset the filter. which approach should we follow? or could we do anything else???

CHANGELOG.md
Rahban1: got it!

assets/html/js/search.js
Rahban1: I just did it to give some time for DOM to get ready before searching in the input value

Rahban1: I tried using without setTimeout as I am calling it at the end of runSearchMainCode() but it didn't work.

I also tried adding an even listener for dom to get loaded, apparently that also didn't work. 

couldn't think of anything else.

Rahban1: can we use :
$(document).ready(function() {checkURLForSearch()});???

mortenpi:
I'll just note that I was testing this, and this seems to work as is, except that `?q=...` does not get cleared when you close the modal (`filter=` does get removed though).

assets/html/js/search.js
Hetarth02: Hmm...alright, maybe let's keep the timeout for now and we can refactor in follow-up PRs. What are you thoughts on this @mortenpi?

Hetarth02:
> 1) make the selected_filter in search.js global and use the same in shortcut.js

I don't think we can truly make anything global, can we? Because documenter uses `requirejs` which wraps everyfile under its own function.(Correct me if I am wrong here...)

> 2) if we are not making `selected_filter` global in search.js we can create a custom event like this
> 
> ```js
> document.addEventListener('reset-filter', function() {
>    selected_filter = "";
>    update_search();
>  });
> ```
> 
> and dispatch it in shortcut.js to reset the filter. which approach should we follow? or could we do anything else???

Events have a chance of not working sometimes because not all things would be loaded into the DOM.

One other approach could be,
Maybe, we can move all the search related logic inside a single file(`search.js`) so that we can access it more easily. (Might be a large refactor/change) 

CC @mortenpi

Rahban1:
everything works well locally, this was fun!

Rahban1:
I used the event approach and it seem to work fine locally

Rahban1:
> Maybe, we can move all the search related logic inside a single file(`search.js`) so that we can access it more easily. (Might be a large refactor/change)

It would be cleaner but as you said a large refactor, maybe it can be a seperate PR in the future.

Rahban1:
> I don't think we can truly make anything global, can we? Because documenter uses `requirejs` which wraps everyfile under its own function.(Correct me if I am wrong here...)
> 
> >

you are right, global does not seem to be the correct approach here. Thank you!

assets/html/js/search.js
mortenpi: It looks like this whole function is not used anymore? In this file that is. There's a function with the same name in `shortcut.js`.

mortenpi:
Other than the one unused (?) function, this looks great to me, and it seems to work perfectly as is. Happy to take any refactorings to improve the code, but maybe as a follow-up yea -- I would love to actually go ahead an just merge this, to get it into 1.10.0 :slightly_smiling_face: Do we think it's ready to go @Rahban1 @Hetarth02 ?

assets/html/js/search.js
Rahban1: It is used. The names are same in both the files but the `search.js` one is responsible for extracting query params and triggering search, and the one in `shortcut.js` is used just to open the modal if the url has additional params.

Rahban1:
I think it's ready to go, It works great locally. It was super fun working on this 😊

mortenpi:
Thank you!

Hetarth02 (approved):

Hetarth02:
@mortenpi @Rahban1 What is the `traces.txt` file?

icweaver:
Just saw that this feature was added back in the [1.13 dev version of the Julia manual](https://docs.julialang.org/en/v1.13-dev/?q=macros&filter=section), y'all are amazing! Would it be possible to backport this to the [v1.11 stable docs](https://docs.julialang.org/en/v1.11/) before the v1.11.7 release? Totally understand if not

mortenpi:
We generally haven't backported the Documenter updates unless there's a bug. But, in principle, it would be a simple matter of preparing the PRs against the `release-1.11` and `release-1.12` updating `doc/Manifes.toml`.

PR #2657, opened 2025-03-13
Use SVG for brand icons

fingolfin:
This is a revival of PR #1475 by @friggog but without the change to use an icon for source links (not because I am strictly opposed to that, but rather because I don't want to mix two things in one PR).


src/html/HTMLWriter.jl
```diff
@@ -1405,20 +1405,76 @@ function edit_link(f, ctx, navnode)
...
+    )
+
+    return (host = "GitLab", logo = logo)
+end
+
+function host_logo_azure()
+    @tags svg path
+
+    # from SimpleIcons: https://github.com/simple-icons/simple-icons/blob/develop/icons/azuredevops.svg
+    # https://simpleicons.org/ License - https://github.com/simple-icons/simple-icons/blob/develop/LICENSE.md (CC0 1.0 Universal)
```
mortenpi, line 1448: So.. apparently the Azure Devops logo has been removed from SimpleIcons on trademark grounds: https://github.com/simple-icons/simple-icons/pull/10019

This actually makes me a bit hesitant to, in general, include that actual icons here :cold_face:  Need to think about it a bit.

fingolfin: I was not aware.

Personally I don't use Azure, I just wanted to help take care of some of the oldest PRs on this repo.

My view: if MS is hostile against users like this, then I would not bother providing them free PR by trying hard to show their branding: let's keep the generic icon for Azure links then.

friggog: Haha I added this as a MSFT employee because we were using it! That project is long gone, but it would be nice to support azure still… can’t speak for the lawyers but I’d hope this is a reasonable use case…

fingolfin: Well given that MS explicitly requested that simple icons remove that content, I'd personally not want to add this without explicit permission. In the end this is fringe enough to not warrant that headache, IMHO.

But in the end I'll leave it to Documenter maintainers to decide -- feel free to close this PR or merge it, I won't complain either way :-)

mortenpi: Yea, let's remove the Azure one. It's their loss :man_shrugging:

Vendoring GitHub (https://github.com/logos) and Bitbucket (https://www.atlassian.com/legal/trademark#our-trademarks) shouldn't be a problem according to their trademark T&Cs. And FontAwesome looks good too.

mortenpi: Also, we're currently using the Microsoft icon for Azure. It's not even clear if that's okay. So let's just fall back to the default icon (i.e. remove all Azure special casing).

fingolfin: But is there any point to switching to SVG icons left then? Other than doing it to add Azure (and then doing it for all brands for consistency)?

(Don't get me wrong, I am happy to adjust this PR as indicated, I just wonder if that's really what you want?)

mortenpi: Hmm. A fair point. It would allow us to still drop the dependency on the FontAwesome brands CSS file. But whether that's worth the complexity of maintaining the icons here, especially since we still keep the FontAwesome dep in general..

So.. I don't have super strong opinion about it to be honest. I'm also happy if we stick to the status quo, since it doesn't seem to cause any issues.

What's your preference here @fingolfin?

fingolfin: Personally I'd prefer the status quo: it works right now so there is no risk of regression and I don't see how not having "FontAwesome brands CSS" helps us (as long as we use some parts of FontAwesome anyway...)

mortenpi: Alright, then let's close these PRs -- I also lean towards the status quo. Sorry we didn't arrive at this conclusion earlier! At least this is a small case study in integrating icons and stuff directory into Documenter, so we can refer back to this if we need to.

mortenpi:
https://github.com/JuliaDocs/Documenter.jl/pull/2657#discussion_r2013034203

PR #2658, opened 2025-03-17
Fix variable confusion in doc block expansion

Keno:
The later error messages are expecting `str` to refer to the original value of the variable, but it gets overwritten with the text of the admonition giving a non-sensical error message.


fingolfin (approved):
Indeed that makes senes

PR #2659, opened 2025-03-17
Show full file paths in various error situations (by changing `locrepr`)

fingolfin:
This is useful for anyone using a terminal emulator that makes file paths with line numbers clickable (to open an editor at that file and line), such as VS Code or iTerm and many others.

I think this is what @simonbyrne wanted to achieve in PRs #2336 and #2310.

Perhaps somewhat nicer would be to print paths relative to the doc root (as suggested in the TODO comment visible in the changes for this patch). But probably best would be to show a path relative to *the current working directory* as observed by the user -- which is unfortunately *not* the actual current working directory at the time the code runs. Which I guess is the source of the problem to begin with :-). Because the terminal I know interpret relative paths as relative to the working dir (makes sense). Now usually that will be the package root directory. At least the way I invoke Documenter, but I am not sure how "standard" this is...

For this nicer solution we could record the PWD / CWD in a global place `locrepr` can access. Perhaps at the start of `makedocs` ?

Closes #2336.
Closes #2310.

### DIFF
--- a/src/utilities/utilities.jl
+++ b/src/utilities/utilities.jl
@@ -11,6 +11,9 @@ using .Remotes: Remote, repourl, repofile
 using .Remotes: RepoHost, RepoGithub, RepoBitbucket, RepoGitlab, RepoAzureDevOps,
     RepoUnknown, format_commit, format_line, repo_host_from_url, LineRangeFormatting
 
+const original_pwd = Ref{String}()  # for printing relative paths in error messages
+
+
 """
     @docerror(doc, tag, msg, exs...)
 
@@ -70,7 +73,9 @@ end
 
 # Pretty-printing locations
 function locrepr(file, line = nothing)
-    str = Base.contractuser(file) # TODO: Maybe print this relative the doc-root??
+    basedir = isassigned(original_pwd) ? original_pwd[] : currentdir()
+    file = abspath(file)
+    str = Base.contractuser(relpath(file, basedir))
     line !== nothing && (str = str * ":$(line.first)-$(line.second)")
     return str
 end
--- a/test/docsxref/make.jl
+++ b/test/docsxref/make.jl
@@ -54,17 +54,17 @@ end
     end
     @test isnothing(captured.value)
     @test contains(
-        replace(captured.output, "src\\index" => "src/index"),
+        replace(captured.output, "\\src\\index" => "/src/index"),
         """
-        ┌ Warning: Cannot resolve @ref for md"[`AbstractSelector`](@ref)" in src/index.md.
+        ┌ Warning: Cannot resolve @ref for md"[`AbstractSelector`](@ref)" in docsxref/src/index.md.
         │ - No docstring found in doc for binding `Main.DocsReferencingMain.AbstractSelector`.
         │ - Fallback resolution in Main for `AbstractSelector` -> `Documenter.Selectors.AbstractSelector` is only allowed for fully qualified names
         """
     )
     @test contains(
-        replace(captured.output, "src\\page" => "src/page"),
+        replace(captured.output, "\\src\\page" => "/src/page"),
         """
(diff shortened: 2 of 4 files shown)


fingolfin:
I've now changed it to record the initial pwd. Seems to work great for me, debugging errors and warnings when building our docs is now much more convenient.

This PR fails on Windows, but I'll not look into that before I know whether this has any chance of being merged :-)

src/utilities/utilities.jl
```diff
@@ -11,6 +11,9 @@ using .Remotes: Remote, repourl, repofile
 using .Remotes: RepoHost, RepoGithub, RepoBitbucket, RepoGitlab, RepoAzureDevOps,
     RepoUnknown, format_commit, format_line, repo_host_from_url, LineRangeFormatting
 
+const original_pwd = Ref{String}()  # for printing relative paths in error messages
```
mortenpi, line 14: I'd like this type of state to be stored in the `Document` object, and to avoid global state. I think that should be feasible here?

mortenpi: Oh.. we don't pass the context down to `locrepr` now. Would it feasible to change all the call sites...? Otherwise, ScopedValues could be an option..

mortenpi:
My initial gut reaction is that the paths should be relative to the `make.jl` file (so `docs/`). But I can see the argument that if you run `julia --project=docs/ docs/make.jl`, then maybe it should be relative to the working directory. Is there any "prior art" for this we could reference?

fingolfin:
I am not sure what kind of prior art you'd accept. In my experience printing paths relative to the *users* current working directory seems a pretty much universal norm in Unix tools? Basically the only alternative are absolute paths.

So I'd turn this around: can you name any "prior art" that behaves like Documenter does right now?

src/utilities/utilities.jl
fingolfin: And `ScopedValues` require Julia >= 1.11 ...

fingolfin: Even `ScopedValues.jl` require Julia >= 1.8

mortenpi:
The UNIX utilities argument is good. E.g. even `git status` will print paths relative to the working directory, even though there is a very canonical "root" directory. Also, thinking about it a bit more: I think my gut reaction comes mostly comes from the assumption that you'd mostly need to reference files in `docs/src`. But that's actually not true.. you also need to print paths for the `.jl` files, which would live outside of `docs/`.

So, all in all, I think I'm sold on making this "dynamic" based on the working directory :slightly_smiling_face:

src/utilities/utilities.jl
mortenpi: I guess `task_local_storage` could be another option. But.. actually.. Documenter is not async/multi-thread safe anyway (like, the `cd`s that we do, for example). So let's just stick with the global variable for now -- we have some precedent, and it is actually the simplest one to find (since it's quite explicit if you have some dynamic global variables around).

src/utilities/utilities.jl
```diff
@@ -70,7 +73,9 @@ end
 
 # Pretty-printing locations
 function locrepr(file, line = nothing)
-    str = Base.contractuser(file) # TODO: Maybe print this relative the doc-root??
+    basedir = isassigned(original_pwd) ? original_pwd[] : currentdir()
+    file = abspath(file)
```
fingolfin, line 77: This was part of the fix for Windows...

test/docsxref/make.jl
```diff
@@ -54,17 +54,17 @@ end
     end
     @test isnothing(captured.value)
     @test contains(
-        replace(captured.output, "src\\index" => "src/index"),
+        replace(captured.output, "\\src\\index" => "src/index"),
```
fingolfin, line 57: ... and this (and its sibling a few lines below) were the other part of the fix for Windows...

fingolfin:
@mortenpi ready now :-)

mortenpi (approved):
Lovely! Thank you @fingolfin!

PR #2660, opened 2025-03-17
Add test/docsxref/build and test/symlinks/build to .gitignore; then simplify the ignore rules

fingolfin:


mortenpi:
I kind of like being explicit.. but this is fine too, so let's just go with it :slightly_smiling_face: Thank you!

PR #2662, opened 2025-03-19
Reduce memory usage when using many pages with examples

### LINKED ISSUE #2640: Memory usage of example modules
This was discussed on Slack, making an issue in case others run into the same problem.

When Documenter builds examples it creates a local module for each named example: https://github.com/JuliaDocs/Documenter.jl/blob/1dce9852de811c0055e6118a34cef2090e16ad8f/src/expander_pipeline.jl#L768

The issue is that these modules persist both after all blocks of an example has been passed and also after `makedocs` finishes. If the examples only make use of a trivial amount of memory, this is no issue, but if examples allocate significant amounts of memory (e.g. some large arrays or loading of test data) the memory is lost. For free GitHub runners this can lead to jobs being cancelled, as the build process gets OOM killed.

In my case, my big examples were all generated through Literate.jl, so the solution was to add the following at the end of each example [based on a Discourse discussion](https://discourse.julialang.org/t/delete-a-module/62226/2):

```julia
function clear_module!(M::Module)        # hide
    for name ∈ names(M, all=true)        # hide
        if !isconst(M, name)             # hide
            @eval M $name = $nothing     # hide
        end                              # hide
    end                                  # hide
end                                      # hide
clear_module!(@__MODULE__)               # hide
GC.gc();                                 # hide
```

Ideal behavior would be that Documenter could parse the order of named example blocks, figure out when they are no longer needed for the build and make sure that the variables in the module are then available to GC.

fingolfin:
Resolves #2640 -- but I am not sure how to test this well. Perhaps by adding an object with a finalizer and then forcing a full GC?

Anyway, perhaps at least @moyner can try if this helps with his concrete issue?

### DIFF
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -1,5 +1,26 @@
 # Defines node "expanders" that transform nodes from the parsed markdown files.
 
+# helper for "cleaning up" content of modules to enable garbage collection.
+# See also <https://github.com/JuliaDocs/Documenter.jl/issues/2640>.
+function clear_module!(M::Module)
+    # we need `invokelatest` here for Julia >= 1.12 (or 1.13?)
+    for name in Base.invokelatest(names, M, all = true)
+        if !isconst(M, name)
+            @eval M $name = $nothing
+        end
+    end
+    return
+end
+
+function clear_modules!(d::Dict{Symbol, Any})
+    for (k, v) in d
+        startswith(String(k), "__atexample__") || continue
+        v isa Module && clear_module!(v)
+    end
+    GC.gc()
+    return
+end
+
 function expand(doc::Documenter.Document)
     priority_pages = filter(doc.user.expandfirst) do src
         if src in keys(doc.blueprint.pages)
@@ -23,6 +44,7 @@ function expand(doc::Documenter.Document)
             expand_recursively(node, page, doc)
         end
         pagecheck(page)
+        clear_modules!(page.globals.meta)
     end
     return
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Don't require custom themes to set a color for the 'todo' admonition. ([#2576])
 * Entries in `@repl` blocks that were hidden with `# hide` no longer produce erroneous empty lines ([#1521], [#2054], [#2399])
+* Reduce memory usage when there are many pages with doctests and examples, by clearing out the "sandbox" modules used for running the code, allowing the garbage collector to reclaim memory. ([#2640])
 
 ## Version [v1.9.0] - 2025-03-17
 
@@ -1984,6 +1985,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2624]: https://github.com/JuliaDocs/Documenter.jl/issues/2624
 [#2629]: https://github.com/JuliaDocs/Documenter.jl/issues/2629
 [#2636]: https://github.com/JuliaDocs/Documenter.jl/issues/2636
+[#2640]: https://github.com/JuliaDocs/Documenter.jl/issues/2640
 [#2642]: https://github.com/JuliaDocs/Documenter.jl/issues/2642
 [#2646]: https://github.com/JuliaDocs/Documenter.jl/issues/2646
 [#2648]: https://github.com/JuliaDocs/Documenter.jl/issues/2648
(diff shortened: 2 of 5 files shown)


asinghvi17:
You could just allocate two large arrays, I think, in separate example blocks?  A few 50000x50000 arrays might cause the GH Actions runner to go OOM

fingolfin:
Something like this should work:
```
Main.finalizer_count = 0

mutable struct MyMutableStruct
  bar
  function MyMutableStruct(bar)
      global finalizer_count
      x = new(bar)
      finalizer(x) do y
        finalizer_count += 1
      end
  end
end
```
then do something like this in a doctest or so
```
a = MyMutableStruct(1)
```
and then on a different (later) page do something like this
```
GC.gc(true)
@assert finalizer_count == 1
```

But where should `finalizer_count` live? It must be a "global" module that survives. E.g. in `Documenter` itself; but how would the doctest know where to find that? Or perhaps in `Main` (not nice but perhaps OK since this is "just" in a test). Better ideas?

asinghvi17:
IMO living in Main is the easiest thing to do if this is purely for testing purposes.

asinghvi17:
It may also pay off to put the first page, where the struct is instantiated, into `expandfirst` in makedocs, to make sure that that code runs first.

src/expander_pipeline.jl
```diff
@@ -1,5 +1,24 @@
...
+        end
+    end
+    return
+end
+
+function clear_modules!(d::Dict{Symbol, Any})
+    for (k, v) in d
+        startswith(String(k), "__atexample__") || continue
+        v isa Module && clear_module!(v)
+    end
```
mortenpi, line 18: Should we also immediately trigger the GC here, to ensure that things get deallocated?

```suggestion
    end
    GC.gc()
```

mortenpi:
> You could just allocate two large arrays, I think, in separate example blocks? A few 50000x50000 arrays might cause the GH Actions runner to go OOM. But that would not be great for testing, so a finalizer is probably best...

I would be a bit sceptical to try to test it with a large allocation, since we don't know where the tests will run (e.g. PkgEval). But a finalizer-based approach seems like a good idea. Putting stuff into `Main` seems completely fine.

src/expander_pipeline.jl
fingolfin: Would be OK by me. But of course this may not collect everything, for that we'd have to call `GC.gc(true)`, but that can be costly. So for the test I'd anyway insert a call to `GC.gc(true)` before checking whether the finalizer did run.

fingolfin:
@mortenpi I find it a bit difficult to decide into which of the many test files to insert such a test (or add a new file?) Do you have any advice or recommendation were to put it?

mortenpi:
Integrating it into the normal examples is annoying. Let's just do a new one maybe. This one looks like a small one that can be copy-pasted and edited I think: https://github.com/JuliaDocs/Documenter.jl/tree/master/test/workdir

src/expander_pipeline.jl
fingolfin: On second thought in the test the GC should always be able to reclaim the freshly created object, so let's try with only this `GC.gc()` call

fingolfin:
Test added, it was simpler than I expected :-)

fingolfin:
Hmm, this example "works" in the REPL even with nightly:
```
module Foo
    @eval Main finalizer_count = [0]
    
    mutable struct MyMutableStruct
      bar
      function MyMutableStruct(bar)
          x = new(bar)
          finalizer(x) do y
            Main.finalizer_count[1] += 1
          end
      end
    end
    
    a = MyMutableStruct(1)
end

@test finalizer_count[1] == 0
Foo.a = nothing
GC.gc()
@test finalizer_count[1] == 1
```


So... maybe something somewhere is retaining a reference? Or maybe there is a worldage issue (lots of binding changes recently)? Huh

fingolfin:
So I had to use `invokelatest` for Julia nightly, i.e. it was a worldage binding issue as suspected (I still find this horribly counter intuitive. Ah well).

mortenpi (approved):
LGTM! I took the liberty of editing the CHANGELOG a tiny bit.

src/expander_pipeline.jl
```diff
@@ -1,5 +1,26 @@
 # Defines node "expanders" that transform nodes from the parsed markdown files.
 
+# helper for "cleaning up" content of modules to enable garbage collection.
+# See also <https://github.com/JuliaDocs/Documenter.jl/issues/2640>.
+function clear_module!(M::Module)
+    # we need `invokelatest` here for Julia >= 1.12 (or 1.13?)
+    for name in Base.invokelatest(names, M, all = true)
+        if !isconst(M, name)
+            @eval M $name = $nothing
```
KristofferC, line 9: This can throw (https://github.com/JuliaDocs/Documenter.jl/issues/2673).

PR #2663, opened 2025-03-20
Fix codecov coverage reporting (maybe)

### LINKED ISSUE #2638: Codecov fails
Just opening this to keep track, does anyone know why codecov doesn't give reports anymore?

fingolfin:
Attempt to fix #2638 -- we'll see.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -52,10 +52,7 @@ jobs:
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v5
         with:
-          file: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
-          plugins: noop
-          disable_search: true
       - name: Submit coverage to Coveralls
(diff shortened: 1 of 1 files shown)


fingolfin:
Hmmm it is difficult to say if this helps because https://app.codecov.io/gh/JuliaDocs/Documenter.jl/pulls actually shows that coverage reporting for PRs was fine all along. It is only broken on master: https://app.codecov.io/gh/JuliaDocs/Documenter.jl/commits

So short of merging this I am not sure how test this :/

.github/workflows/CI.yml
```diff
@@ -52,10 +52,7 @@ jobs:
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v5
         with:
-          file: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
-          plugins: noop
-          disable_search: true
```
fingolfin, line 58: Actually I think this line broke it, together with the fact that `file` is gone and one has to use `files`. So in this sense I am really hopefully this will work.

What mystifies me is why PR reports were working all along?!

PR #2667, opened 2025-03-28
Set version to 1.10.0

mortenpi:
## Pre-release

 - [x] Change the version number in `Project.toml`
   * If the release is breaking, increment MAJOR
   * If the release adds a new user-visible feature, increment MINOR
   * Otherwise (bug-fixes, documentation improvements), increment PATCH
 - [x] Update `CHANGELOG.md`, following the existing style (in particular, make sure that the change log for this version has the correct version number and date).
 - [x] Run `make changelog`, to make sure that all the issue references in `CHANGELOG.md` are up to date.
 - [x] Check that the commit messages in this PR do not contain `[ci skip]`
 - [ ] Run [regression-tests workflow](https://github.com/JuliaDocs/Documenter.jl/actions/workflows/regression-tests.yml) using a `workflow_dispatch` trigger to check for any changes that broke extensions: https://github.com/JuliaDocs/Documenter.jl/actions/runs/14123340267

## The release

 - [ ] After merging the pull request, tag the release. There are two options for this:

   1. [Comment `[at]JuliaRegistrator register` on the GitHub commit.](https://github.com/JuliaRegistries/Registrator.jl#via-the-github-app)
   2. Use [JuliaHub's package registration feature](https://help.juliahub.com/juliahub/stable/contribute/#registrator) to trigger the registration.

   Either of those should automatically publish a new version to the Julia registry.
 - Once registered, the `TagBot.yml` workflow should create a tag, and rebuild the documentation for this tag.
 - These steps can take quite a bit of time (1 hour or more), so don't be surprised if the new documentation takes a while to appear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-# Unreleased
+# Version [v1.10.0] - 2025-03-28
 
 ## Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-# Version [v1.10.0] - 2025-03-28
+# Version [v1.10.0] - 2025-03-31
 
 ## Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1455,6 +1455,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [v1.8.0]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.8.0
 [v1.8.1]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.8.1
 [v1.9.0]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.9.0
+[v1.10.0]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.10.0
 [#198]: https://github.com/JuliaDocs/Documenter.jl/issues/198
 [#245]: https://github.com/JuliaDocs/Documenter.jl/issues/245
 [#487]: https://github.com/JuliaDocs/Documenter.jl/issues/487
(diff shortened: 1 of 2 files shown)


mortenpi:
@goerz Does this just require adjusting the reference output in DocumenterCitations tests?

https://github.com/JuliaDocs/Documenter.jl/actions/runs/14123340267/job/39568323831#step:6:335

goerz:
I'll have to have a closer look at this (tomorrow)… It's definitely not a blocker, but the error messages might now be showing longer file paths than necessary. A little hard to tell from the log how this would actually look to a user, or when/how/why this changed.

goerz:
Okay, so this is due to the deliberate change in how `locrepr` works in #2659.

Right now, `DocumenterCitations` follows the [old philosophy](https://github.com/JuliaDocs/Documenter.jl/pull/2659#issuecomment-2739502901) of showing file paths relative to `make.jl`. So this will require some kind of bugfix in `DocumenterCitations`. If I were to leave it as-is (i.e., simply adjust the reference output), messages emitted by `DocumenterCitations` would be inconsistent. Some would be relative to the CWD, some would be relative to `make.jl`. That would be less than ideal. So this requires some work on `DocumenterCitations` to make it consistent either way. Probably, being consistent with the new philosophy is better (since the user doesn't know/care whether warnings come from `Documenter` proper or from `DocumenterCitations`). In the context of the tests, the new way is a bit unfortunate, since the tests build documentations in various temporary folders not below the CWD. That means the file paths shown in the errors are both long and non-deterministic. But in normal use, it shouldn't be a problem.

So yeah, this shouldn't be a blocker for the 1.10 release, but it means I have to put in a little bit of work on the `DocumenterCitations` side ASAP. Yay for downstream testing doing its job! :-)

goerz:
Opened https://github.com/JuliaDocs/DocumenterCitations.jl/pull/89 to address this

goerz:
[DocumenterCitations v1.3.7](https://github.com/JuliaDocs/DocumenterCitations.jl/releases/tag/v1.3.7) has been released and fixes the [regression tests for Documenter 1.10](https://github.com/JuliaDocs/Documenter.jl/actions/runs/14123340267/job/39637169654)

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-# Unreleased
+# Version [v1.10.0] - 2025-03-28
```
```suggestion
# Version [v1.10.0] - 2025-03-31
```

mortenpi:
Thank you @goerz!

mortenpi:
https://github.com/JuliaRegistries/General/pull/128033

PR #2669, opened 2025-03-28
Add DocumenterPages.jl to regression tests

asinghvi17:
will edit this PR when DocumenterDescriptions.jl is registered.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -20,7 +20,7 @@ jobs:
           - package: 'DocumenterMermaid'
           - package: 'Asciicast'
           - package: 'DocumenterPages'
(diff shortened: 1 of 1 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -19,6 +19,8 @@ jobs:
           - package: 'DocumenterInterLinks'
           - package: 'DocumenterMermaid'
           - package: 'Asciicast'
+          - package: 'DocumenterPages'
+          # - package: 'DocumenterDescriptions' # when the package gets registered in 3 days
```
mortenpi, line 23: Looks like it has!

```suggestion
          - package: 'DocumenterDescriptions'
```

asinghvi17:
Sounds good to me!

PR #2672, opened 2025-03-31
added SearchRecord custom functions for at-blocks

### LINKED ISSUE #1929: Remove at-block source code from search index
It looks like that for the at-blocks, we just write the contents of the original code block to the search index.

A case in point: we can "find" a bunch of spurious `DocTestSetup`s in the Julia manual (https://docs.julialang.org/en/v1.8.1/search/?q=doctestsetup) which do not appear in the HTML, due to at-meta blocks on those pages.

Rahban1:
this is an attempt to solve and potentially close #1929 
my thought process was that right now the SearchRecord objects are created and added to the search index which are very general so I made custom function for adding SearchRecords and add special handling for MultiOutput which is @example block, MultiCodeBlock which is the @repl block and EvalNode which is the @eval block


Rahban1:
I can add the changelog and will fix the linting issues once we decide what is the most optimal solution :)

src/html/HTMLWriter.jl
```diff
@@ -713,6 +713,61 @@ function SearchRecord(ctx, navnode, node::Node, ::MarkdownAST.AbstractElement)
     return SearchRecord(ctx, navnode; text = mdflatten(node))
 end
 
+
+# for @example blocks
+function SearchRecord(ctx::HTMLContext, navnode::Documenter.NavNode, node::MarkdownAST.Node{Nothing}, element::Documenter.MultiOutput)
+    @info "MultiOutput SearchRecord method called!"
+    children_array = collect(node.children)
```
mortenpi, line 720: `node.children` is an iterable. There's no need to actually materialize this.

src/html/HTMLWriter.jl
```diff
@@ -713,6 +713,61 @@ function SearchRecord(ctx, navnode, node::Node, ::MarkdownAST.AbstractElement)
...
+
+    if isempty(children_array) || length(children_array) < 2
```
mortenpi, line 722:
```suggestion
    if length(children_array) < 2
```

src/html/HTMLWriter.jl
```diff
@@ -713,6 +713,61 @@ function SearchRecord(ctx, navnode, node::Node, ::MarkdownAST.AbstractElement)
...
+        if isa(child.element, Documenter.MultiOutputElement) && isa(child.element.element, Dict)
+            for mime in [MIME"text/plain"(), MIME"text/markdown"(), MIME"text/html"()]
+                if haskey(child.element.element, mime)
+                    output_text *= string(child.element.element[mime]) * " "
+                    break
+                end
+            end
+        else 
+            output_text *= mdflatten(child) * " "
+        end
```
mortenpi, line 737: So, for at-example blocks, we render both the input and the output. We probably want to put the source code into the output as well?

image

In general, though, I wonder if it's better to just add the search index entries in the actual `domify` methods, like here:

https://github.com/JuliaDocs/Documenter.jl/blob/4328aec711dd3ee04d1989c796b241f547962fcd/src/html/HTMLWriter.jl#L2394-L2527

X-ref https://github.com/JuliaDocs/Documenter.jl/pull/2672#issuecomment-2767917688 about how to remove the other entries.

mortenpi:
Just thinking out loud here. The methods you're adding, they are getting called here:

https://github.com/JuliaDocs/Documenter.jl/blob/4328aec711dd3ee04d1989c796b241f547962fcd/src/html/HTMLWriter.jl#L1683-L1684

It feels like what we want is sometimes _not_ to write any `SearchRecord` here. For example, for at-docs, we kinda get duplicate entries right now, one for the at-docs source, and one for the actual docstring.

```json
    {
      "location": "",
      "page": "Example Site",
      "title": "Example Site",
      "text": "Main.DocString",
      "category": "page"
    },
    {
      "location": "#DocString",
      "page": "Example Site",
      "title": "DocString",
      "text": "Docstring\n\n\n\n\n\n",
      "category": "constant"
    }
```

The second one (actual docstring content) gets pushed here:

https://github.com/JuliaDocs/Documenter.jl/blob/4328aec711dd3ee04d1989c796b241f547962fcd/src/html/HTMLWriter.jl#L1775-L1783

So what I think we might be able to do is change these methods to functions (`searchrecord`, to differentiate from the constructor), which sometimes just returns a `nothing`. We'll add a `searchrecord` method for e.g. `DocsNode` to return `nothing`, and then in the first loop at the top, we just exclude any `nothing`s. That should give us an easy way to exclude nodes at this level.

mortenpi:
Just one more note: as I was playing around with this PR, and to actually see what ends up in the search index, I put together a small toy build that also prettyfies the index, in case that might be useful: https://github.com/JuliaDocs/Documenter.jl/commit/becf65c8c91e6c2c774329c0618c153a5c18a2ae

In general (but probably not in this PR, to not scope creep this), I think we want some explicit tests for the search index generation, like with reference output etc.

Rahban1:
> Just one more note: as I was playing around with this PR, and to actually see what ends up in the search index, I put together a small toy build that also prettyfies the index, in case that might be useful: [becf65c](https://github.com/JuliaDocs/Documenter.jl/commit/becf65c8c91e6c2c774329c0618c153a5c18a2ae)
> 
> In general (but probably not in this PR, to not scope creep this), I think we want some explicit tests for the search index generation, like with reference output etc.

this is a great idea, I will make a new PR for tests after this PR's completion.

src/html/HTMLWriter.jl
Rahban1: got it :)

Rahban1:
this time I made a should_index_for_search function checking for at-blocks and only added the one which were not at blocks but still I don't think this issue is solved

src/html/HTMLWriter.jl
```diff
@@ -713,6 +713,91 @@ function SearchRecord(ctx, navnode, node::Node, ::MarkdownAST.AbstractElement)
...
+        return SearchRecord(ctx, navnode; text = "")
+    end
+    
+    return SearchRecord(ctx, navnode, node, element.element)
+end
+
+"""
+Filter function to prevent duplicate search entries.
+"""
+function should_index_for_search(navnode, node, element)
```
mortenpi, line 779: I think it should be sufficient to just check the type, no need to introspect into the objects? So something as simple as this should be enough probably:

```julia
exclude_from_search_index(element) = typeof(element) in (Documenter.MultiOutputElement, Documenter.MultiCodeBlock, Documenter.EvalNode)

Rahban1:
I have tested it locally it does not add duplicated to the search index, I will add changelog entries once we decide to go ahead with this approach

fingolfin:
It would be nice to get this over the finish line. I've just merged `master` into this branch to get CI to run again.

fingolfin:
Needs a rebase, now that your other PR #2762 has been merged :-/

PR #2676, opened 2025-04-06
Added anchor links to admonition blocks for direct referencing

Rahban1:
Close #2505 

I have added a new field to HTMLContext `admonition_counters` which stores counters for each admonition so each one gets a unique identifier. The main challenge was generating unique id for each admonition which is the meat of the pr, added some css for hover functionality of link 

Before : 
Screenshot 2025-04-07 at 2 22 43 AM

After : 
Screenshot 2025-04-07 at 2 23 25 AM
 
works well when tested locally

PR #2685, opened 2025-04-24
Fix performance regression in `clear_modules!`

fingolfin:
See https://github.com/oscar-system/Oscar.jl/issues/4824 for background.

Unfortunately there is a much worse performance regression on master introduced by PR #2676 (ping @Rahban1). Luckily this one has not yet been in a release of Documenter.

It would be great to have this fix in a release ASAP. Perhaps it would make sense to make a branch for a 1.10.2 release with just this fix but not the other changes, so we have time to properly deal with the other regression (other than reverting the PR in question) ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added anchor links to admonition blocks, making it possible to create direct links to specific admonitions. ([#2505], [#2676])
 * Added different banners for dev and unreleased docs ([#2382], [#2682])
 
-## Fixed
+### Fixed
 
 * `@meta`, `@setup`, and `@docs` blocks no longer generate spurious entries in the search index. ([#1929], [#2675])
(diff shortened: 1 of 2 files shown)


mortenpi:
LGTM!

> Perhaps it would make sense to make a branch for a 1.10.2 release with just this fix but not the other changes, so we have time to properly deal with the other regression (other than reverting the PR in question) ?

We haven't done those in a while, but we can definitely do backports when needed.

Side note -- do you think we could add Oscar to be part of our upstream tests? I wouldn't have noticed the performance issues, but it's a big, complex document, so it might reveal a regression sometimes.

https://github.com/JuliaDocs/Documenter.jl/blob/10c41a5dfbcac9b92bd9333f57ac566c8bb577bc/.github/workflows/regression-tests.yml#L47-L56

fingolfin:
Thank you for the quick release @mortenpi !

We should open an issue on the other performance regression. (I'd do it now but need to run now to get my son to daycare and then get to teaching. I'll try to remember it later.)

I see no reason why we couldn't/ shouldn't add Oscar to those tests.

PR #2688, opened 2025-04-25
replaced SHA.sha with Base.hash

Rahban1:
this improved the build time by a lot, thank you @fredrikekre for the suggestion :) ref : https://github.com/JuliaDocs/Documenter.jl/pull/2676#discussion_r2059308904

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* changed SHA.sha to Base.hash as it was slowing the build process
+
 ### Changed
 
 * Added anchor links to admonition blocks, making it possible to create direct links to specific admonitions. ([#2505], [#2676])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Fixed
-
-* changed SHA.sha to Base.hash as it was slowing the build process
-
 ### Changed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* changed SHA.sha to Base.hash as it was slowing the build process
+
```
fingolfin, line 11: I don't think you need to mention this here, as the change you are fixing is not yet released. To the contrary, that text would be really confusing for someone from the outside reading the changelog

fingolfin:
I verified that with this patch, building the Oscar manual is fast again :-)

src/html/HTMLWriter.jl
```diff
@@ -2396,7 +2396,7 @@ function domify(dctx::DCtx, node::Node, a::MarkdownAST.Admonition)
             isempty(cat_sanitized) ? "" : ".is-category-$(cat_sanitized)"
         end
     node_repr = sprint(io -> show(io, node))
-    content_hash = bytes2hex(SHA.sha1(node_repr))[1:8]
+    content_hash = string(hash(node), base = 16)[1:8]
```
fingolfin, line 2399: Should we just keep the full 16 characters to minimize the risk of collisions?

CHANGELOG.md
Rahban1: I also thought that, but I did it to pass the tests.

src/html/HTMLWriter.jl
Rahban1: I am fine with that, but I think the likelihood of collision is minimal with 8 characters also

fingolfin (approved):

PR #2689, opened 2025-04-25
CI: test building Oscar manual

fingolfin:
It is big and may help detect regressions early.

@mortenpi is that what you had in mind?


mortenpi (approved):
Yes, indeed! Thank you! Let's see if the CI for MathOptInterface passes on retry. But even if it looks like it has regressed, I wouldn't actually block the merge here.

PR #2690, opened 2025-04-27
Update examples.md

msainsburydale:


docs/src/man/examples.md
```diff
@@ -44,6 +44,7 @@ Packages that have tagged versions available in the general Registry:
 - [Luxor.jl](https://juliagraphics.github.io/Luxor.jl/stable/)
 - [MergedMethods.jl](https://michaelhatherly.github.io/MergedMethods.jl/stable/)
 - [Mimi.jl](https://www.mimiframework.org/Mimi.jl/stable/)
+- [NeuralEstimators.jl](https://github.com/msainsburydale/NeuralEstimators.jl)
```
fingolfin, line 47: @mortenpi do we really want to add to this list? Nowadays most Julia packages use Documenter, so it seems to me we'd be better off listing a very small curated list of packages which for some reason are "interesting" (e.g. they use some feature not used by everyone).

Or just remove this list... Right now the text before the list says:

>    Packages are listed alphabetically. If you have a package that uses Documenter then
    please open a PR that adds it to the appropriate list below; a simple way to do so
    is to navigate to
    <https://github.com/JuliaDocs/Documenter.jl/edit/master/docs/src/man/examples.md>.

which then should be removed. 


It also says this:

>    The `make.jl` file for all listed packages will be tested to check for potential
    regressions prior to tagging new Documenter releases whenever possible.


which I think is just wrong ?

mortenpi: > it seems to me we'd be better off listing a very small curated list of packages which for some reason are "interesting"

Yes, very much this. I have actually been thinking about the same thing for a long time and been meaning to remove/replace the list, but haven't gotten around to it.

Another issue with this long list is also that we often get linkcheck failures here, because packages move their docs domains around sometimes. And the more package you have, the more often we get hit by that.

> which I think is just wrong ?

Yup. We do have the regressions tests workflow nowadays, so we should probably mention that somewhere.

PR #2693, opened 2025-04-29
Fix clear_module! world semantics

vtjnash:
The `isconst` is a world-specific query, so it needs `invokelatest` too. Restructure a bit of the code to do that in a helper function. Also add a call to delete all bindings in v1.12+. Although that may be expensive and currently warns that it may be incorrect, it may be helpful to get testing of that.

### DIFF
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -1,20 +1,25 @@
 # Defines node "expanders" that transform nodes from the parsed markdown files.
 
+function clear_global!(M::Module, name::Symbol)
+    isconst(M, name) && return
+    VERSION >= v"1.9" ? setglobal!(M, name, nothing) : Core.eval(M, :($name = $nothing))
+    nothing
+end
+
 # helper for "cleaning up" content of modules to enable garbage collection.
 # See also <https://github.com/JuliaDocs/Documenter.jl/issues/2640>.
 function clear_module!(M::Module)
     # we need `invokelatest` here for Julia >= 1.12 (or 1.13?)
     for name in Base.invokelatest(names, M, all = true)::Vector{Any}
-        if !(Base.invokelatest(isconst, M, name::Symbol)::Bool)
-            # see, e.g https://github.com/JuliaDocs/Documenter.jl/issues/2673
-            # it is not possible to set `nothing` to variables, which are strongly typed
-            # still attempt to set it, but ignore any errors
-            try
-                VERSION >= v"1.9" ? setglobal!(M, name, nothing) : Core.eval(M, :($name = $nothing))
-            catch err
-                @debug "Could not clear variable `$name` by assigning `nothing`" err
-            end
+        # see, e.g https://github.com/JuliaDocs/Documenter.jl/issues/2673
+        # it is not possible to set `nothing` to variables, which are strongly typed
+        # still attempt to set it, but ignore any errors
+        try
+            invokelatest(clear_global!, M, name::Symbol)
+        catch err
+            @debug "Could not clear variable `$name` by assigning `nothing`" err
         end
+        VERSION >= v"1.12" && Base.delete_binding(M, name)
     end
     return
 end
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -2,7 +2,12 @@
 
 function clear_global!(M::Module, name::Symbol)
     isconst(M, name) && return
-    VERSION >= v"1.9" ? setglobal!(M, name, nothing) : Core.eval(M, :($name = $nothing))
+    if VERSION >= v"1.9"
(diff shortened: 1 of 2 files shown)


src/expander_pipeline.jl
```diff
@@ -1,20 +1,25 @@
...
-            # it is not possible to set `nothing` to variables, which are strongly typed
-            # still attempt to set it, but ignore any errors
-            try
-                @eval M $name = $nothing
-            catch err
-                @debug "Could not clear variable `$name` by assigning `nothing`" err
-            end
+    for name in Base.invokelatest(names, M, all = true)::Vector{Any}
+        # see, e.g https://github.com/JuliaDocs/Documenter.jl/issues/2673
+        # it is not possible to set `nothing` to variables, which are strongly typed
```
fingolfin, line 15: Unrelated, but while we are it, isn't the grammar here a bit odd? Perhaps like this?
```suggestion
        # it is not possible to set variables which are strongly typed to `nothing`
```

fingolfin: More related: is there a clean way for us to determine that a variable / binding is strongly typed?

fingolfin (approved):
Looks good to me, thank you!

src/expander_pipeline.jl
```diff
@@ -1,20 +1,25 @@
...
-        if !isconst(M, name)
-            # see, e.g https://github.com/JuliaDocs/Documenter.jl/issues/2673
-            # it is not possible to set `nothing` to variables, which are strongly typed
-            # still attempt to set it, but ignore any errors
-            try
-                @eval M $name = $nothing
-            catch err
-                @debug "Could not clear variable `$name` by assigning `nothing`" err
-            end
+    for name in Base.invokelatest(names, M, all = true)::Vector{Any}
```
fingolfin, line 13:
```suggestion
    for name in Base.invokelatest(names, M, all = true)::Vector{Symbol}
```

src/expander_pipeline.jl
```diff
@@ -1,20 +1,25 @@
...
+    for name in Base.invokelatest(names, M, all = true)::Vector{Symbol}
+        # see, e.g https://github.com/JuliaDocs/Documenter.jl/issues/2673
+        # it is not possible to set `nothing` to variables, which are strongly typed
+        # still attempt to set it, but ignore any errors
+        try
+            invokelatest(clear_global!, M, name::Symbol)
+        catch err
+            @debug "Could not clear variable `$name` by assigning `nothing`" err
         end
+        VERSION >= v"1.12" && Base.delete_binding(M, name)
```
fingolfin, line 22: If there is a chance for this to be slow, I'd definitely would want to check this in the Oscar manual were a slowdown in this function caused us major pain before.

src/expander_pipeline.jl
vtjnash: Sure. `Nothing <: Core.get_binding_type(M, name)`

fingolfin:
Hmm, fails Julia 1.6 tests (finalizer of an object in the cleared module is not called).

(and of course a changelog entry is missing)

src/expander_pipeline.jl
```diff
@@ -1,20 +1,25 @@
...
-                @eval M $name = $nothing
-            catch err
-                @debug "Could not clear variable `$name` by assigning `nothing`" err
-            end
+    for name in Base.invokelatest(names, M, all = true)::Vector{Symbol}
+        # see, e.g https://github.com/JuliaDocs/Documenter.jl/issues/2673
+        # it is not possible to set `nothing` to variables, which are strongly typed
+        # still attempt to set it, but ignore any errors
+        try
+            invokelatest(clear_global!, M, name::Symbol)
```
vtjnash, line 18:
```suggestion
            Base.invokelatest(clear_global!, M, name::Symbol)
```

fingolfin:
Building Oscar.jl docs with Documenter.jl master ...
- ... and Julia 1.10 takes 35 secs
- ... and Julia 1.11 takes 45 secs
- ... and Julia 1.12 takes 47-53 secs 

So it seems Julia 1.12 is getting slightly slower than 1.11, which is a bunch slower than 1.11 (unfortunately that's a general pattern I am seeing and a reason why I do my daily work with Julia 1.10 still).

On the bright side, `Base.delete_binding` doesn't seem to matter much here (I tried commenting that line out and measured no difference).

fingolfin:
@vtjnash would you mind adding a changelog entry?

mortenpi (approved):
Thank you!

If this is a blocking bug somewhere (Base manual?) then we could get this out in a backport to 1.10 quickly. Otherwise, there are a few more PRs we might want to get into Documenter 1.11.0.

vtjnash:
Anytime in a week or two is fine and will hold the bugfix upstream. Not urgent

PR #2694, opened 2025-04-30
CI: run regression-tests.yml on all PRs & primary branches

fingolfin:
It seems natural to run the regression tests on all PRs so that we actually notice regressions as they are introduced resp. as soon as possible (if caused by external factors), and not just when we remember to manually run these tests.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -1,6 +1,14 @@
 name: 'Regression Tests'
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - master
+      - release-*
+      # You can suffix the branch name with '-runci' to run the CI on that branch
+      # without having to open a pull request. This is useful for long-lived branches.
+      - '*/*-runci'
+    tags: '*'
   pull_request:
     paths:
       - '.github/workflows/regression-tests.yml'


fingolfin:
If there is a reason *not* to do this, then we should at least document it with a comment in the workflow yaml file

mortenpi:
I think the main concerns were that they are a bit expensive (which actually doesn't seem to be that true -- they all run in a few minutes), and that you might get red CI due to the downstream package being somehow broken. But maybe we don't need to be concerned about that -- as long as we don't choke the JuliaDocs CI runners, I think it would be fine to always run them.

goerz:
> as long as we don't choke the JuliaDocs CI runners, I think it would be fine to always run them.

This is more of general philosophical point, but I’ve started to feel a little concerned about all the energy that gets burned accumulatively and collectively on excessive CI testing at the scale of GitHub. You know, global warming and all that. I’ve been trying to be more deliberate on what CI runs give a good bang for the buck. Running downstream test in every single PR push seems excessive to me in that frame of mind, in addition to the argument that failures often may be outside of the control of the submitting author, and wouldn’t actually block the PR. Running them only on `master` (after a PR is merged) seems more appropriate, and then failures can be followed up on as necessary.

mortenpi:
> Running them only on `master` (after a PR is merged) seems more appropriate, and then failures can be followed up on as necessary.

This was also a "middle ground" that I was thinking. The other thing we could theoretically have is a merge queue, so it would only run once the PR is marked for merging, but it would still block the merge if something fails (but this might get annoying if you want to ignore the failure).

fingolfin:
Updated to only run on branches as suggested; that will still help us to narrow down when a regression was introduced

mortenpi:
Thank you!

PR #2670, opened 2025-03-28
Add DocumenterVitepress to regression tests

asinghvi17:
I'll update the other PR when this is merged.

PR #2695, opened 2025-05-02
Dispatch logic on `versions` when deploying

jkrumbiegel:
This PR is intended to allow DocumenterVitepress to deploy multiple vitepress builds of the same docs. This doesn't mesh with the current versioning behavior that Documenter allows, so I've refactored the places where `versions` is used with dispatchable functions instead of if blocks. I know that these were not intended to be public API, and maybe they can become sort of a shadow-API that only DocumenterVitepress uses, but this was the smallest change that I could think of to solve our issues without forking or reimplementing too much Documenter deploy logic.

You can see the functions created here used in this draft PR https://github.com/LuxDL/DocumenterVitepress.jl/pull/246/

I would like to get some feedback whether these changes are going to be considered or if we have to deal with things differently on the DocumenterVitepress side

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Refactored `deploydocs` internals to allow dispatch on the `versions` keyword, intended as a non-public API for DocumenterVitepress which cannot use the default versioning mechanism during deployment ([#2695]).
 * Added anchor links to admonition blocks, making it possible to create direct links to specific admonitions. ([#2505], [#2676], [#2688])
 * Added different banners for dev and unreleased docs ([#2382], [#2682])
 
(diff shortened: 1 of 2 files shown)


mortenpi:
> I know that these were not intended to be public API, and maybe they can become sort of a shadow-API that only DocumenterVitepress uses, but this was the smallest change that I could think of to solve our issues without forking or reimplementing too much Documenter deploy logic.

I would like this I think. I would prefer to not make these proper public APIs. If the DocumenterVitepress test suite could catch any breaking changes to these APIs, then we should be able to catch things before releases via regression tests (https://github.com/JuliaDocs/Documenter.jl/pull/2670).

The general approach looks great to me. The only thing I'd add is maybe a small paragraph of comment in the code explaining what functions and signatures are part of this "shadow API" and who uses it (i.e. DocumenterVitepress).

jkrumbiegel:
Ok @mortenpi I've turned the docstrings into comments so that Documenter doesn't think they should be included in the docs, I also added mentions of DocumenterVitepress in the appropriate places for people looking at that code later. With that, it's ready from my side.

The DocumenterVitepress test suite is currently nonexistent, I guess they are currently just using the docs build as a big smoke test. Although I'm also not sure how `deploydocs` can be part of the tests as it's so dependent on CI variables, branches being available, etc. Not easy to mock..

mortenpi:
> Although I'm also not sure how `deploydocs` can be part of the tests as it's so dependent on CI variables, branches being available, etc. Not easy to mock..

Yeah, fair -- end to end testing of the behavior is tricky. But one option would be to just add like `hasmethod` calls or something in the test suite, to just check that the internal functions exist and have expected signatures -- better than nothing :slightly_smiling_face:

jkrumbiegel:
> But one option would be to just add like hasmethod calls or something in the test suite, to just check that the internal functions exist and have expected signatures -- better than nothing 🙂

Oh yeah that's one way to do it of course, good point. I can add something like that, but I guess this can come at a later date and we'd then update regression tests here?

mortenpi:
> I guess this can come at a later date and we'd then update regression tests here?

Absolutely, definitely not something to block the merge here.

This is good to go then I guess?

jkrumbiegel:
> This is good to go then I guess?

Yes, I've just done another test deploy from DocumenterVitepress, all good. If you tag this, we should be able to get a DocumenterVitepress release with the new approach out pretty quickly.

PR #2696, opened 2025-05-02
Add `make clean` target

goerz:
This deletes all files generated during local development workflows. Cf.

    git ls-files --others --ignored --exclude-standard --directory

Helpful to run before switching branches. I'd find this pretty helpful when reviewing PRs locally. Obviously this is Unix only, but so is the entire `Makefile,` presumably.


mortenpi:
LGTM!

Issue #2126, opened 2023-05-21
Easier doctests of non-exported functions


tecosaur:
Hello,

I've recently started adding more doctests to unexported functions in my code, and I seem to be repeating the same boilerplate-y `setup` quite a bit, e.g.

````julia
"""
    parsechecksum(checksum::String)
    
...text...

```jldoctest; setup = :(import DataToolkitCommon.Store.parsechecksum)
julia> parsechecksum("crc32c:9c0188ee")
(:crc32c, 0x9c0188ee)
```
"""
````

Is there some way this boilerplate could be avoided?

mortenpi:
You can set up shared setup code to all docstrings in a module: https://documenter.juliadocs.org/dev/man/doctests/#Module-level-metadata

mortenpi:
I don't think we'd want to `eval` into the package scope, since that will start polluting it, and you'd lose the property that all doctests are always evaluated in a clean environment. Just dynamically generating the global `DocTestSetup` from `names` seems like a reasonable workaround here though.

I would also argue that importing everything would be the more useful behavior anyway, as opposed to just importing a single function, since doctests often reference other functions from the same module as well, and not just the one the docstring is attached to. So the sort of "evaluate in the current module context (but still in a different module)" behavior seems good (and useful) to me. So maybe it could even be a built-in option here.

tecosaur:
Particularly considering multi-step doctests which introduce variable bindings, the `Core.eval` approach seems like a dodgy idea.

I'm glad to hear that "importing everything via `names`" sounds like a viable approach to you, I think this could strike a nice balance between convenience and robustness.

tecosaur:
I don't have the time to work on this ATM (or any time soon likely), but now it's been a few years I think it may be worth bumping this and stating that I still think this would be a worthwhile QoL improvement.

I've found myself using `julia-repl` over `jldoctest` for this reason in library code, because I just cannot be fussed to add a setup block each time.

fingolfin:
But what is the plan here? Importing everything surely can't be default behavior. It'd be breaking as hell among other things, and it would also render `jldoctest` useless for *me* (because I want to ensure things work *as in the REPL*.)

All in all this seems like a valid but somewhat niche requirement (BTW we also use `jldoctest` to test internal functions but we just don't import them; this has the added bonus that I can run the tests manually by copy&pasting into the REPL. But of course I accept that this is not how you want to use it).

[...]

PR #2697, opened 2025-05-06
Allow specifying a global default for `@meta` block settings

### LINKED ISSUE #1253: `setdocmeta!` does not work for non-docstring doctests
I currently have a bunch of  `@meta` block like the one below at the top of my `md` files which I though I could get rid of by using `setdocmeta!`, but that doesn't seem to be true. From playing around, `setdocmeta!` appears to only works with docstrings in `jl` files.

```julia
```@meta
DocTestSetup = quote
    using MyPackage
end
```

As an aside, it seems like it would make a lot of sense to have the package being documented "used" by default, instead of having to specify it manually in `DocTestSetup`.

### LINKED ISSUE #2512: Allow setting a global `@meta` data (such as a `DocTestSetup` block) for all `.md` files in one place
we currently have this in every single .md file for our packages:
```
DocTestSetup = quote
  using MyPackage
end
```
so that jldoctest work as desired. For the docstrings, a single invocation
```
DocMeta.setdocmeta!(MyPackage, :DocTestSetup, quote using MyPackage end; recursive=true)
```
sufficed to take care of them.

It would be great if there was a similar way to tell Documenter once, in one place, to use some code as `DocTestSetup` for all the `.md` files.

### LINKED ISSUE #2684: Allow setting global default for `CollapsedDocStrings`
We would like to have all (or almost all) or docstring collapsed. Right now that means we have to add `CollapsedDocStrings = true` to every single .md file.

It would be nice if there was a way to set this once, globally, say in the call to `makedocs`. Perhaps like this?
```julia
    makedocs(;
        sitename = "Blah",
        format = Documenter.HTML(
            collapsed_docstrings = true,
        ),
    )
```
(This is somewhat orthogonal to the approach suggested in #2512 and strictly speaking we only need one, but it seems natural to me to have this, too?)

fingolfin:
This allows to e.g. use a common doctest setup block on all pages, thus complementing `DocMeta.setdocmeta!`.

Other nice applications include changing certain settings, such `CollapsedDocStrings = true` globally, or being able to specify `CurrentModule = YourPackageModule` globally.

Resolves #2512. Resolves #2684. Resolves #1253

CC @lgoettgens 

---

This is a first prototype. No documentation or changelog entry yet as the design is still not finalized.

This adds a new kwarg `default_meta` to `makedocs` -- contrary to the discussion in #2512 this expects a `Dict{Symbol,Any}`. The main reasons for doing it this way are:
1. this was the easiest to implement :-) which is important for an MVP
2. arguably this is also the easiest to document and explain (as in: what are the semantics of this option)
3. it is sufficient to set up a basic test, and also sufficient for my primary applications in Oscar.jl
4. it does not preclude us from accepting a quoted code block in the future -- we could simply add support for that *in addition* to what this PR implements...

I've been wondering if it wouldn't be nicer to accept a named tuple instead of (or in addition to?) a `Dict{Symbol,Any}`, the syntax then would look a bit closer to the actual content of a `@meta` block.

Of course accepting actual quoted code blocks would make the input *even closer* to an actual `@meta` block. As I wrote above, we could still add support for that if it turns out someone needs it. When we do, it then should also be easier to resolve the question "when to evaluate the code block" (once at the start? once for each page? something else?) because we'll have a concrete usecase at hand...

That said, I am totally open to change this MVP based on your comments.

### DIFF
--- a/src/doctests.jl
+++ b/src/doctests.jl
@@ -14,7 +14,7 @@ struct DocTestContext
     file::String
     doc::Documenter.Document
     meta::Dict{Symbol, Any}
-    DocTestContext(file::String, doc::Documenter.Document) = new(file, doc, Dict())
+    DocTestContext(file::String, doc::Documenter.Document) = new(file, doc, copy(doc.user.meta))
 end
 
 """
--- a/src/makedocs.jl
+++ b/src/makedocs.jl
@@ -166,6 +166,11 @@ For example, if you have `foo.md` and `bar.md`, `bar.md` would normally be evalu
 Evaluation order among the `expandfirst` pages is according to the order they appear in the
 argument.
 
+**`meta`** can be used to provide default values for the `@meta` blocks executed on every
+page. For example `meta = Dict(:DocTestSetup => :(using MyPackages))` sets `DocTestSetup`
+on every page to the code block `:(using MyPackages)`. This can still be overridden on
+each individual page.
+
 **`draft`** can be set to `true` to build a draft version of the document. In draft mode
 some potentially time-consuming steps are skipped (e.g. running `@example` blocks), which is
 useful when iterating on the documentation. This setting can also be configured per-page
--- a/test/default_meta/tests.jl
+++ b/test/default_meta/tests.jl
@@ -1,0 +1,13 @@
+using Documenter
+using Test
+
+const pages = [
+    "Home" => "index.md",
+]
+
+makedocs(
+    sitename = "Test", pages = pages, doctest = true,
+    meta = Dict(:DocTestSetup => :(x = 42))
+)
+
+# the test is passing the doctest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* Added `meta` keyword to `makedocs` which allows to specify default values for `@meta` blocks. This can be used to alter the default values for anything that can be configured via a `@meta` block for all pages of the documentation. E.g. this can be used to set `CollapsedDocStrings = true` on all pages. ([#2512], [#2697])
+
 ### Changed
 
 * Work around GitHub rate limiting of link checks by passing `GITHUB_TOKEN` when applicable and available. ([#2729])
@@ -2046,6 +2050,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2499]: https://github.com/JuliaDocs/Documenter.jl/issues/2499
 [#2505]: https://github.com/JuliaDocs/Documenter.jl/issues/2505
 [#2511]: https://github.com/JuliaDocs/Documenter.jl/issues/2511
+[#2512]: https://github.com/JuliaDocs/Documenter.jl/issues/2512
 [#2513]: https://github.com/JuliaDocs/Documenter.jl/issues/2513
 [#2514]: https://github.com/JuliaDocs/Documenter.jl/issues/2514
 [#2526]: https://github.com/JuliaDocs/Documenter.jl/issues/2526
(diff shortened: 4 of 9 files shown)


src/doctests.jl
```diff
@@ -14,7 +14,7 @@ struct DocTestContext
     file::String
     doc::Documenter.Document
     meta::Dict{Symbol, Any}
-    DocTestContext(file::String, doc::Documenter.Document) = new(file, doc, Dict())
+    DocTestContext(file::String, doc::Documenter.Document) = new(file, doc, copy(doc.user.default_meta))
```
fingolfin, line 17: Perhaps this should only copy `DocTestSetup` and `DocTestTeardown` if present?

fingolfin:
I am happy with this now. Any thought, @lgoettgens @mortenpi ?

src/doctests.jl
mortenpi: It looks like when we parse the at-meta blocks when doctesting, we put everything into the `.meta` dict. So I think it's best to stay consistent here (even though we don't use the other values).

src/makedocs.jl
```diff
@@ -166,6 +166,8 @@ For example, if you have `foo.md` and `bar.md`, `bar.md` would normally be evalu
 Evaluation order among the `expandfirst` pages is according to the order they appear in the
 argument.
 
+**`default_meta`** can be set to TODO. TODO default `@meta` block. TODO
```
mortenpi, line 169: Just bikeshedding this here. I was thinking maybe it could be `page_meta` or even just `meta`.

So, currently, this will only affect the Markdown pages, and for docstrings you'd still have to do `setdocmeta!`, right? I am now wondering if this should also apply to docstrings (with `setdocmeta!` providing a per-module override still).

Do you have any sense if the global at-meta stuff you'd like to apply for pages would also be useful for docstrings? Are they usually basically the same in your experience?

mortenpi:
> it does not preclude us from accepting a quoted code block in the future -- we could simply add support for that in addition to what this PR implements...

I'm happy if it's a `Dict`. The argument that we can always expand it in the future is convincing. Could even add a code comment to that effect somewhere, that we may want to expand it in the future.

src/makedocs.jl
fingolfin: We use the same `DocTestSetup` in `@meta` blocks in every `.md` file and also set it via `setdocmeta!`. I believe they would be basically the same, and having `default_meta`/`page_meta`/`meta` also set a default for `doctests` inside docstrings would then unify 2 of 3 places where in my experience one needs to set these

So it'd help! But maybe not quite enough, at least if one also wants to run doctests outside of building the documentation. We do that (via an explicit `doctest(MyPackage)` call in our CI) in order to get code coverage.

For details see also my comment https://github.com/JuliaDocs/Documenter.jl/issues/2126#issuecomment-2854253638

test/default_meta/tests.jl
```diff
@@ -0,0 +1,13 @@
+using Documenter
+using Test
+
+const pages = [
+    "Home" => "index.md",
+]
+
+makedocs(
+    sitename = "Test", pages = pages, doctest = true,
+    default_meta = Dict(:DocTestSetup => :(x = 42))
```
fingolfin, line 10: Just to say, one could also allow (instead or in addition) named tuples:
```suggestion
    default_meta = (DocTestSetup = :(x = 42), )
```
OK that's a bit awkward in the single argument list. But it is a tad closer to how the content of the `@meta` blocks look (no `:` in front of names, `=` instead of `=>`)

Anyway I am totally fine either way, just wanted to bring it up

fingolfin:
I've resolved the TODOs and renamed `default_meta` to `meta` as suggested. From my point of view this is ready for review.

test/default_meta/tests.jl
asinghvi17: NamedTuples would be nice...

mortenpi (approved):
@fingolfin I think this LGTM. Thank you, and sorry for the delay!

If we want, we could add the support for named tuples as well. I'd add them in addition to the `Dict`, since the latter is easier to construct dynamically. But feel free to merge if you don't want to tackle that right now.

fingolfin:
I've just rebased this, and it would be nice if we could merge this already now -- I am willing to add the named tuple support but am busy right now and would love if we could just use this feature already.

PR #2700, opened 2025-05-06
Load time optimization: download search index asynchronously; make cacheable

xal-0:
If you have not visited https://docs.julialang.org for 10 minutes, there is a very noticeable delay as all of the assets hosted by GitHub are either validated or redownloaded.  The worst offender is `search_index.js`, which is 4.4 MiB (1 MiB gzipped).

This PR does two things to make the situation a little better:
- We block rendering until the search index is fully loaded, even though `search.js` handles asynchronously loading the index just fine. This is a simple matter of making the script tag `async`.
- The search index becomes stale absurdly quickly.  Unfortunately, GitHub pages sets `Cache-Control: max-age=600` on everything it hosts, so we're mostly stuck with this situation there. We can still make it a little better for docs hosted elsewhere by putting the search index's hash into its filename, so it can stay fresh indefinitely.

Before, with slow internet and render blocking network requested boxed in red:
Screenshot 2025-05-06 at 4 23 48 PM

After:
Screenshot 2025-05-06 at 4 24 01 PM

### DIFF
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -771,7 +771,6 @@ function render(doc::Documenter.Document, settings::HTML = HTML())
     end
 
     ctx = HTMLContext(doc, settings)
-    ctx.search_index_js = "search_index.js"
     ctx.themeswap_js = copy_asset("themeswap.js", doc)
     ctx.warner_js = copy_asset("warner.js", doc)
 
@@ -800,9 +799,34 @@ function render(doc::Documenter.Document, settings::HTML = HTML())
         copy_asset("themes/$(theme).css", doc)
     end
 
-    size_limit_successes = map(collect(keys(doc.blueprint.pages))) do page
+    function page_navnode(page)
         idx = findfirst(nn -> nn.page == page, doc.internal.navlist)
         nn = (idx === nothing) ? Documenter.NavNode(page, nothing, nothing) : doc.internal.navlist[idx]
+        nn, idx
+    end
+
+    foreach(keys(doc.blueprint.pages)) do page
+        nn, idx = page_navnode(page)
+        @debug "Indexing $(page) [$(repr(idx))]"
+        generate_index!(ctx, nn)
+    end
+
+    # Make an attempt at avoiding spurious changes to the search index hash.
+    sort!(ctx.search_index; by=r -> (r.src, r.fragment, r.text))
+    search_index_buf = let io = IOBuffer()
+        println(io, "var documenterSearchIndex = {\"docs\":")
+        # convert Vector{SearchRecord} to a JSON string + do additional JS escaping
+        println(io, JSDependencies.json_jsescape(ctx.search_index), "\n}")
+        take!(io)
+    end
+    # Put the index's hash in its filename so it can have a long max-age.
+    ctx.search_index_js = "search_index_$(dataslug(search_index_buf)).js"
+    open(joinpath(doc.user.build, ctx.search_index_js), "w") do io
+        write(io, search_index_buf)
+    end
+
+    size_limit_successes = map(collect(keys(doc.blueprint.pages))) do page
+        nn, idx = page_navnode(page)
         @debug "Rendering $(page) [$(repr(idx))]"
         render_page(ctx, nn)
     end
@@ -836,12 +860,6 @@ function render(doc::Documenter.Document, settings::HTML = HTML())
     end
     # Check that all HTML files are smaller or equal to size_threshold option
     all(size_limit_successes) || throw(HTMLSizeThresholdError())
-
-    open(joinpath(doc.user.build, ctx.search_index_js), "w") do io
-        println(io, "var documenterSearchIndex = {\"docs\":")
-        # convert Vector{SearchRecord} to a JSON string + do additional JS escaping
-        println(io, JSDependencies.json_jsescape(ctx.search_index), "\n}")
-    end
 
     write_inventory(doc, ctx)
 
@@ -1027,7 +1045,7 @@ function render_head(ctx, navnode)
             :src => RD.requirejs_cdn,
             Symbol("data-main") => relhref(src, ctx.documenter_js),
         ],
-        script[:src => relhref(src, ctx.search_index_js)],
+        script[:src => relhref(src, ctx.search_index_js), :async],
 
         script[:src => relhref(src, "siteinfo.js")],
         script[:src => relhref(src, "../versions.js")],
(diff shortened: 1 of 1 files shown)


fingolfin:
Thank you, sounds useful!

Prettier fails (note that we use Prettier v2 not v3)

Also needs a changelog entry

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added anchor links to admonition blocks, making it possible to create direct links to specific admonitions. ([#2505], [#2676], [#2688])
 * Added different banners for dev and unreleased docs ([#2382], [#2682])
 * Added an API function `writer_supports_ansicolor` for a Documenter writer to indicate that it supports rendering ANSI-colored strings.  This is useful for new backends that may be implemented. ([#2490])
+* The search index now loads asynchronously, and can be cached indefninitely. ([#2700])
```
mortenpi, line 13:
```suggestion
* The search index now loads asynchronously, and can be cached indefinitely. ([#2702], [#2700])
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Refactored `deploydocs` internals to allow dispatch on the `versions` keyword, intended as a non-public API for DocumenterVitepress which cannot use the default versioning mechanism during deployment ([#2695]).
 * Use different banners for dev and unreleased docs ([#2382], [#2682])
+* The search index now loads asynchronously, and can be cached indefinitely. ([#2702], [#2700])
```
fingolfin, line 18: @xal-0 those PRs also need to be added to the list at the end of the file (make sure to pull our changes first :-) )

src/html/HTMLWriter.jl
```diff
@@ -800,9 +799,34 @@ function render(doc::Documenter.Document, settings::HTML = HTML())
...
+        return nn, idx
+    end
+
+    foreach(keys(doc.blueprint.pages)) do page
+        nn, idx = page_navnode(page)
+        @debug "Indexing $(page) [$(repr(idx))]"
+        generate_index!(ctx, nn)
+    end
+
+    # Make an attempt at avoiding spurious changes to the search index hash.
```
fingolfin, line 814: Would this be accurate?
```suggestion
    # Sort the index to avoid spurious changes to the search index hash caused
    # by difference in the order of entries
```

fingolfin (approved):
Looks good to me, thank you

src/html/HTMLWriter.jl
```diff
@@ -800,9 +799,34 @@ function render(doc::Documenter.Document, settings::HTML = HTML())
         copy_asset("themes/$(theme).css", doc)
     end
 
-    size_limit_successes = map(collect(keys(doc.blueprint.pages))) do page
+    function page_navnode(page)
```
mortenpi, line 802: Bikeshed, but let's move this out of the function body perhaps. I am not a huge fan of declaring functions within functions. I think it just needs to take `doc::Document` as the first argument?

CHANGELOG.md
mortenpi: Note: one can run `make changelog`, which will update the references in `CHANGELOG.md` automatically.

src/html/HTMLWriter.jl
```diff
@@ -1848,6 +1862,14 @@ function domify(::DCtx, ::Node, rawnode::Documenter.RawNode)
     return rawnode.name === :html ? DOM.Tag(Symbol("#RAW#"))(rawnode.text) : DOM.Node[]
 end
 
+function generate_index!(ctx::HTMLContext, navnode::Documenter.NavNode)
+    return foreach(getpage(ctx, navnode).mdast.children) do node
+        rec = searchrecord(ctx, navnode, node)
+        if !isnothing(rec)
+            push!(ctx.search_index, rec)
+        end
+    end
```
mortenpi, line 1871: `foreach` doesn't return anything, so it looks a bit weird to me.

```suggestion
    foreach(getpage(ctx, navnode).mdast.children) do node
        rec = searchrecord(ctx, navnode, node)
        if !isnothing(rec)
            push!(ctx.search_index, rec)
        end
    end
    return nothing
```

Also, as a side note: I don't mind having `foreach`-es, but I don't really find them particularly idiomatic. I'd just write this as `for node in getpage(ctx, navnode).mdast.children ... end`.

src/html/HTMLWriter.jl
```diff
@@ -800,9 +799,34 @@ function render(doc::Documenter.Document, settings::HTML = HTML())
...
         idx = findfirst(nn -> nn.page == page, doc.internal.navlist)
         nn = (idx === nothing) ? Documenter.NavNode(page, nothing, nothing) : doc.internal.navlist[idx]
+        return nn, idx
+    end
+
+    foreach(keys(doc.blueprint.pages)) do page
+        nn, idx = page_navnode(page)
+        @debug "Indexing $(page) [$(repr(idx))]"
+        generate_index!(ctx, nn)
+    end
```
mortenpi, line 812: Mmm. I don't think this works. A lot of the pushing into `.search_index` happens during the DOM generation deep in the call tree, i.e. downstream from `render_page`. So right now a bunch of stuff is missing (e.g. docstrings).

But yea, we have a chicken-and-egg situation here, where in order to get the full search index, we need to render everything, but we don't know how to render it because we don't know the search index hash yet, since we haven't generated the index yet.

I do think that with some reorganization of the code, we could somehow generate all the DOM first, and then inject the search index file name later. Although one nice property of the current structure is that we don't need to keep the DOM of all the pages in memory at once, which I am not sure how we could keep.

What I might suggest (unless you have a good simple idea of how to tackle this) is to undo the hashing related changes for now. There's also another issue with it -- MultiDocumenter, for example, relies on the `search_index.js` file being where it is, so we may need to think about this change a bit more anyway. But we should definitely get the front end loading changes in.

mortenpi (changes requested):
The hashing logic needs changes: https://github.com/JuliaDocs/Documenter.jl/pull/2700#discussion_r2078839979

fingolfin:
I see the value for inserting the hash but I agree with @mortenpi here, it is better to do this separately, and then differently.

A simpler solution than refactoring the rendering pipeline etc. would be to use simple post-processing: after the HTML docs have been generated, compute the checksum of the (by now complete) `search_index.js`. Then inject the hash into each HTML file using a regex to pick out the relevant places. Yes, it is a bit of a hack, but it is very easy to implement and understand.

I believe you don't need to rename `search_index.js` -- using an URL of the form `search_index.js?HASH` should do the trick as well.

Hetarth02:
> asynchronously loading the index

If we do this, then wouldn't user face a problem when trying to search while the index is loading asynchronously?

For example, as soon as the page loads I as a user try to search something. User presses `ctrl + /` there are 2 scenarios,

1. The modal UI doesn't trigger(which is unlikely think it's pretty independent from the logic of search) and user has to wait or try repeatedly.

2. The modal UI triggers but it doesn't do anything because the search index is getting loaded in the background.

PR #2692, opened 2025-04-29
Separate decision making repo from deploy repo

vchuravy:
When deploying the documentation to an different repository we try to make decisions based on the `repo` argument.

Naively changing `repo` to the target repository leads to:

```
┌ Info: Deployment criteria for deploying preview build from GitHub Actions:
│ - ✘ ENV["GITHUB_REPOSITORY"]="trixi-framework/Trixi.jl" occurs in repo="github.com/trixi-framework/TrixiDocumentation"
│ - ✔ ENV["GITHUB_REF"] corresponds to a PR number
│ - ✔ PR originates from the same repository
│ - ✔ `push_preview` keyword argument to deploydocs is `true`
│ - ✔ ENV["GITHUB_ACTOR"] exists and is non-empty
│ - ✔ ENV["DOCUMENTER_KEY"] exists and is non-empty
└ Deploying: ✘
```

Since `ENV[GITHUB_REPOSITORY]` and `repo` disagree.
Currently the docs in https://documenter.juliadocs.org/stable/man/hosting/#Out-of-repo-deployment recommend:

```
# You have to override the corresponding environment variable that
# deplodocs uses to determine if it is deploying to the correct repository.
# For GitHub, it's the GITHUB_REPOSITORY variable:
withenv("GITHUB_REPOSITORY" => repo) do
  deploydocs(repo=repo)
end
```

Yet this leads to:

```
┌ Info: Deployment criteria for deploying preview build from GitHub Actions:
│ - ✔ ENV["GITHUB_REPOSITORY"]="github.com/trixi-framework/TrixiDocumentation" occurs in repo="github.com/trixi-framework/TrixiDocumentation"
│ - ✔ ENV["GITHUB_REF"] corresponds to a PR number
│ - ✘ PR originates from the same repository
│ - ✔ `push_preview` keyword argument to deploydocs is `true`
│ - ✔ ENV["GITHUB_ACTOR"] exists and is non-empty
│ - ✔ ENV["DOCUMENTER_KEY"] exists and is non-empty
└ Deploying: ✘
```

Since we use the `ENV[GITHUB_REPOSITORY]` to check that a PR originates from the same repository.

My proposed fix is to disentagle the `repo` from the deployment target. `repo` remains the default and always describes the source (which may or may not be the target),
but we introduce a new keyword `deploy_depo` that can be used to override the target repository.

This allows us to remove the `with_env` hack.

PR #2706, opened 2025-05-07
Adapt wording regarding `deploy_repo` and `repo_preview`

lgoettgens:
A small docs follow-up to https://github.com/JuliaDocs/Documenter.jl/pull/2692. I haven't seen this before that got merged, but I think that currently the docstring differs from the functionality. (Apart from that I am really looking forward for a release with this change!)

https://github.com/JuliaDocs/Documenter.jl/pull/2692 set the default for both `deploy_repo` and `repo_previews` to `nothing` (instead of `repo_previews = repo` previously), but the explanation of `repo_previews` didn't get adapted accordingly.
Furthermore, it should be documented that `repo_previews` first falls back to `deploy_repo` and only after that to `repo` (cf https://github.com/JuliaDocs/Documenter.jl/blob/fcee2bfb2ff1ea8c88858fee784caa7b113efe82/src/deployconfig.jl#L232)

cc @goerz @mortenpi @vchuravy

### DIFF
--- a/src/deploydocs.jl
+++ b/src/deploydocs.jl
@@ -146,7 +146,7 @@ forks.
 It defaults to the value of `branch`.
 
 **`repo_previews`** can be used to override the remote repository to which pull request previews are
-deployed. If this is unset, it will be the same as `deploy_repo` (if that is set) or `repo` otherwise.
+deployed. If this is not set, it will be the same as `deploy_repo` (if that is set) or `repo` otherwise.
 The expected format of the argument is the same as for `repo`.
 
 !!! note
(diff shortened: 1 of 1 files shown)


src/deploydocs.jl
```diff
@@ -145,9 +145,9 @@ forks.
 **`branch_previews`** is the branch to which pull request previews are deployed.
 It defaults to the value of `branch`.
 
-**`repo_previews`** is the remote repository to which pull request previews are
-deployed. It defaults to the value of `repo`, and when specified manually, must
-follow its formatting scheme.
+**`repo_previews`** can be used to override the remote repository to which pull request previews are
+deployed. If this is unset, it will be the same as `deploy_repo` (if that is set) or `repo` otherwise.
```
fingolfin, line 149:
```suggestion
deployed. If this is not set, it will be the same as `deploy_repo` (if that is set) or `repo` otherwise.
```

PR #2708, opened 2025-05-08
Fix LaTeXWriter with local images on Windows

### LINKED ISSUE #2434: Wrong `\includegraphics` image path at creating LaTeX output

FriesischScott:
LaTeX expects paths to be separated with `/` on all platforms. Using `normpath` changes the separators to `\\` on Windows which then need to be replaced for `includegraphics`.

I've also updated the CI workflow to run the PDF/LaTeX backend on `ubuntu-latest` and `windows-latest`.

Fixes #2434, closes #2440.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -71,6 +71,10 @@ jobs:
           - windows-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: fontist/setup-fontist@v2
+        if: matrix.os == 'windows-latest'
+      - run: fontist install "DejaVu"
+        if: matrix.os == 'windows-latest'
       - uses: julia-actions/setup-julia@v2
         with:
           version: '1'
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,8 +61,14 @@ jobs:
           path-to-lcov: lcov.info
 
   latex:
-    name: "PDF/LaTeX backend"
-    runs-on: ubuntu-latest
+    name: "PDF/LaTeX backend - ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -71,13 +71,15 @@ jobs:
           - windows-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: "3.0.0"
-        if: matrix.os == 'windows-latest'
-      - uses: fontist/setup-fontist@v2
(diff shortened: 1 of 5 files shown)


FriesischScott:
I have now disabled all latex test except for `latex_simple_tctonic" when running on windows in github actions.

Would you be open to an alternative where we setup a native `texlive` for compilation with https://github.com/teatimeguest/setup-texlive-action?

FriesischScott:
The tectonic build fails because windows doesn't have the DejaVu fonts. 

The simplest fix for this would probably be using the fontist action to install them: https://github.com/marketplace/actions/setup-fontist

FriesischScott:
I've enabled the `nondocker` test for native compilation on both platforms and it works.

On windows in addition to `texlive` and the fonts I had to install python for `minted` and then `pygments` to get tectonic to work. These days `minted` works by calling the `latexminted` executable it supplies but `tectonic` seems to depend on the system `pygments`. I think it's using an outdated version of `texlive`  from before `minted` 3.0 was released.

.github/workflows/CI.yml
```diff
@@ -61,10 +61,44 @@ jobs:
...
+        with:
+          packages: |
+            scheme-medium
+            hyperref
+            minted
+            newunicodechar
+            transparent
+            upquote
+            tabulary
+            tcolorbox
```
mortenpi, line 101: Can we also only do this step when running in Windows?

```suggestion
            tcolorbox
        if: matrix.os == 'windows-latest'
```

mortenpi: @FriesischScott I did the change in [bd3359e](https://github.com/JuliaDocs/Documenter.jl/pull/2708/commits/bd3359eecb86a8f4371959c639a47586afa5e3b8), but still would like to double check that this seems correct to you too?

FriesischScott: Only installing `texlive` on Windows breaks the native test on `ubuntu-latest`. I think it's worth doing it to ensure the native build runs on both Windows and Linux.

FriesischScott: I went ahead and reverted the commit. If this is about the time it takes to install texlive I could find out what the smallest set of necessary packages is starting from `scheme-basic` instead of `scheme-medium`.

mortenpi: Sounds good, thank you! I just wasn't sure if it was necessary and intentionally enabled for Linux or not.

Issue #2699, opened 2025-05-06
Standard `makedocs` call fails if you aren't in a git repo


asinghvi17:
```
┌ Warning: Unable to determine HTML(edit_link = ...) from remote HEAD branch, defaulting to "master".
│ Calling `git remote` failed with an exception. Set JULIA_DEBUG=Documenter to see the error.
│ Unless this is due to a configuration error, the relevant variable should be set explicitly.
└ @ Documenter ~/.julia/packages/Documenter/ZlqFj/src/utilities/utilities.jl:660
ERROR: LoadError: ArgumentError: Unable to automatically determine remote for main repo.
> `repo` is not set, and makedocs is not in a Git repository.
Configure `repo` and/or `remotes` appropriately, or set `remotes = nothing` to disable remote source
links altogether (e.g. if not working in a Git repository).
  repo: nothing
Stacktrace:
 [1] interpret_repo_and_remotes(; root::String, repo::String, remotes::Dict{Any, Any})
   @ Documenter ~/.julia/packages/Documenter/ZlqFj/src/documents.jl:667
 [2] interpret_repo_and_remotes
   @ ~/.julia/packages/Documenter/ZlqFj/src/documents.jl:504 [inlined]
 [3] Documenter.Document(; root::String, source::String, build::String, workdir::Symbol, format::Documenter.HTMLWriter.HTML, clean::Bool, doctest::Bool, linkcheck::Bool, linkcheck_ignore::Vector{…}, linkcheck_timeout::Int64, linkcheck_useragent::String, checkdocs::Symbol, checkdocs_ignored_modules::Vector{…}, doctestfilters::Vector{…}, warnonly::Vector{…}, modules::Vector{…}, pages::Vector{…}, pagesonly::Bool, expandfirst::Vector{…}, plugins::Vector{…}, repo::String, remotes::Dict{…}, sitename::String, authors::String, version::String, highlightsig::Bool, draft::Bool, others::@Kwargs{})
   @ Documenter ~/.julia/packages/Documenter/ZlqFj/src/documents.jl:432
 [4] makedocs(; debug::Bool, format::Documenter.HTMLWriter.HTML, kwargs::@Kwargs{…})
   @ Documenter ~/.julia/packages/Documenter/ZlqFj/src/makedocs.jl:267
```

We should make this simply default to what happens when `remotes = nothing`.

asinghvi17:
It seems to me that the first branch of this should not error.

[...]

PR #2709, opened 2025-05-08
Don't error in `makedocs`, if not a git repo AND not on CI

asinghvi17:
Fix https://github.com/JuliaDocs/Documenter.jl/issues/2699

### DIFF
--- a/src/documents.jl
+++ b/src/documents.jl
@@ -664,7 +664,16 @@ function interpret_repo_and_remotes(; root, repo, remotes)
         end
         err *= '\n'
         err *= "  repo: $(repo_normalized)"
-        throw(ArgumentError(err))
+        # xref https://github.com/JuliaDocs/Documenter.jl/issues/2699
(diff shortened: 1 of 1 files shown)


mortenpi:
Sorry, but I don't think I'd want to accept this: https://github.com/JuliaDocs/Documenter.jl/issues/2699#issuecomment-2864550290

I _can_ offer one workaround if you are very set on having some solution for this. We can use a Documenter-specific environment variable that you can set on _your_ system (e.g. in `~/.bash_aliases`) that will turn the error into a warning. Something like `DOCUMENTER_IGNORE_BROKEN_LINKS=true` (happy to bikeshed). We can document it as a "power user" feature that you can use if you often get bitten by this issue. I just don't want it to be the default behavior.

asinghvi17:
From discussion we settled on having something like `Documenter.OptionalRepo` that will trigger ignoring this behaviour if not in a repo.  Closing this in favour of that kind of solution.

PR #2710, opened 2025-05-09
Serialize MarkdownAST nodes via Markdown.jl when getting hash for admonitions

asinghvi17:
This was an online only PR so it might fail CI :D 

But the idea here is to do a best effort render to Markdown.jl, assuming all the crazy magic has already been expanded by ExpandTemplates.  This might potentially help with issues hashing admonition nodes as reported in https://github.com/JuliaDocs/Documenter.jl/pull/2676#discussion_r2082477911.

Still needs a changelog and tests to make sure this doesn't happen again, but let's see if this approach is viable first

### DIFF
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -2369,6 +2369,26 @@ function domify(dctx::DCtx, node::Node, f::MarkdownAST.FootnoteDefinition)
     return DOM.Node[]
 end
 
+# This function provided by Michael Goerz in https://github.com/JuliaDocs/MarkdownAST.jl/issues/18
+function _markdownast_to_str(node::MarkdownAST.Node)
+    if node.element isa MarkdownAST.Document
+        document = node
+    elseif node.element isa MarkdownAST.AbstractBlock
+        document = MarkdownAST.@ast MarkdownAST.Document() do
+            MarkdownAST.copy_tree(node)
+        end
+    else
+        @assert node.element isa MarkdownAST.AbstractInline
+        document = MarkdownAST.@ast MarkdownAST.Document() do
+            MarkdownAST.Paragraph() do
+                MarkdownAST.copy_tree(node)
+            end
+        end
+    end
+    text = Markdown.plain(convert(Markdown.MD, document))
+    return strip(text)
+end
+
 function domify(dctx::DCtx, node::Node, a::MarkdownAST.Admonition)
     @tags header div details summary
     colorclass =
@@ -2399,7 +2419,7 @@ function domify(dctx::DCtx, node::Node, a::MarkdownAST.Admonition)
             # apply a class
             isempty(cat_sanitized) ? "" : ".is-category-$(cat_sanitized)"
         end
-    node_repr = sprint(io -> show(io, node))
+    node_repr = _markdownast_to_str(node)
     content_hash = string(hash(node_repr), base = 16)
     admonition_id = if !isempty(a.title)
         base_id = Documenter.slugify(a.title)


vtjnash:
It probably would work, but internally `MarkdownAST.copy_tree` is implemented with `deepcopy`, so that function crashes if there's anything in the tree that isn't simple Markdown nodes. We'd first have to fix that bug in `MarkdownAST.copy_tree`, or reimplement a non-broken version of that function here :/

asinghvi17:
Oops - will look into that at some point.  An MWE is probably not too hard to create on the MarkdownAST end from the tree you posted.

vtjnash:
Turns out to be not too bad to avoid the bad methods, just undocumented. This version of the function fixes the build of Base docs for me:
```diff
diff --git a/src/html/HTMLWriter.jl b/src/html/HTMLWriter.jl
index 21cb8f6d1..05abd6c0b 100644
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -2369,6 +2369,20 @@ function domify(dctx::DCtx, node::Node, f::MarkdownAST.FootnoteDefinition)
     return DOM.Node[]
 end
 
+# This function inspired by Michael Goerz in https://github.com/JuliaDocs/MarkdownAST.jl/issues/18
+function _markdownast_to_str(node::MarkdownAST.Node)
+    md = MarkdownAST._convert_element(node)
+    text = Markdown.plain(md)
+    return strip(text)
+end
+# uncomment the following to debug mystery hangs:
+#MarkdownAST._convert_element(::MarkdownAST.Node, e::MarkdownAST.AbstractElement) =
+#   error("Unable to convert element of type $(typeof(e))")
+MarkdownAST._convert_element(::MarkdownAST.Node, e::Documenter.AbstractDocumenterBlock) = "TODO"
+MarkdownAST._convert_element(::MarkdownAST.Node, e::Documenter.PageLink) = "TODO"
+MarkdownAST._convert_element(::MarkdownAST.Node, e::Documenter.LocalLink) = "TODO"
+MarkdownAST._convert_element(::MarkdownAST.Node, e::Documenter.LocalImage) = "TODO"
+
 function domify(dctx::DCtx, node::Node, a::MarkdownAST.Admonition)
     @tags header div details summary
     colorclass =
```

vtjnash:
FWIW, that also points to the source of the failure also: there is no show method for `PageLink`, and so it falls back to printing everything reachable from that page instead. This would usually not run into too many issues (other than getting a incorrectly computed hash here), except that `MarkdownAST` declares that all subtypes are expected to be `mutable struct`, but here there are custom AST elements created, and those are declared just `struct`, and show ends up not liking that they are just declared `struct` when `MarkdownAST` expected them to be mutable.

src/html/HTMLWriter.jl
```diff
@@ -2369,6 +2369,26 @@ function domify(dctx::DCtx, node::Node, f::MarkdownAST.FootnoteDefinition)
...
+    else
+        @assert node.element isa MarkdownAST.AbstractInline
+        document = MarkdownAST.@ast MarkdownAST.Document() do
+            MarkdownAST.Paragraph() do
+                MarkdownAST.copy_tree(node)
+            end
+        end
+    end
+    text = Markdown.plain(convert(Markdown.MD, document))
+    return strip(text)
```
asinghvi17, line 2389:
```suggestion
    text = Documenter.MDFlatten.mdflatten(node)
    return strip(text)
```

mortenpi:
> that `MarkdownAST` declares that all subtypes are expected to be `mutable struct`, but here there are custom AST elements created, and those are declared just `struct`, and show ends up not liking that they are just declared `struct` when `MarkdownAST` expected them to be mutable.

@vtjnash I'm confused by this. `AbstractElement`s do not have to be mutable and MarkdownAST makes no such declaration (most of the "native" elements in MarkdownAST.jl are immutable).

---

As for the bug itself -- I think the correct solution here is to add the necessary `show` methods for the custom types that are problematic? That seems simpler and probably faster than jumping through the mdflatten/Markdown.jl hoop (where you also have to define convert methods for all custom elements).

vtjnash:
I have the impression now that `mdflatten` is the expected way this would be implemented, based upon the impression that is the way we usually create other id's. Defining show seems like a good bugfix anyways too.

mortenpi:
> all of the native elements have all their fields declared mutable

That's fair actually. But yea, the intent there is two fold: (1) you should be able to change the `.element` field of a `Node` object, to change the element, and (2) you should also be able to edit the `.element` objects themselves (i.e. their fields) so that you could update them in place (e.g. if you want to change the code in a code block, or a URL of a link). The AST tree as a whole is meant to be mutable.

But there is no hard requirement for the `.elements` to be mutable per se. It's just something that you probably want in most cases. But there might be cases where you want immutable structs, or partially mutable (i.e. mutable, but const fields).

> I have the impression now that `mdflatten` is the expected way this would be implemented

Hmm. Actually, maybe, yes, it would make sense (and be safer?). And potentially produce more stable hashes (it wouldn't be dependent on internal Julia representation etc). But it shouldn't be strictly necessary for the bugfix, so let's leave that as a change for the next minor release.

fingolfin:
This now uses `mdflatten`. Other than a missing changelog entry, is it OK to merge this now, @mortenpi @vtjnash ?

I guess the important question is: does this fix the issue? I'll try to use it to build the Julia manual now.

fingolfin:
So I grabbed Julia master, did `cd doc`, then made sure to `dev` Documenter.jl from my branch (I double checked that it really used the right Documenter version by inserting `@show pkgdir(Documenter)` into Julia's `doc/make.jl`).

Turns out the manual builds fine (well, at least `make html`) even with Documenter.jl `master`, so without this PR?

asinghvi17:
It was probably https://github.com/JuliaDocs/Documenter.jl/pull/2714 that fixed it.  This is probably not a bad thing to add though.

src/html/HTMLWriter.jl
```diff
@@ -2484,6 +2484,12 @@ function domify(dctx::DCtx, node::Node, f::MarkdownAST.FootnoteDefinition)
     return DOM.Node[]
 end
 
+# This function provided by Michael Goerz in https://github.com/JuliaDocs/MarkdownAST.jl/issues/18
+function _markdownast_to_str(node::MarkdownAST.Node)
+    text = Documenter.MDFlatten.mdflatten(node)
```
fingolfin: This file has `using ...MDFlatten: mdflatten` so you can write
```suggestion
    text = mdflatten(node)
```
but at that point I wonder: why even bother with this helper?

src/html/HTMLWriter.jl
```diff
@@ -2514,7 +2520,7 @@ function domify(dctx::DCtx, node::Node, a::MarkdownAST.Admonition)
             # apply a class
             isempty(cat_sanitized) ? "" : ".is-category-$(cat_sanitized)"
         end
-    node_repr = sprint(io -> show(io, node))
+    node_repr = _markdownast_to_str(node)
```
fingolfin: ... just do
```suggestion
    node_repr = strip(mdflatten(node))
```

src/html/HTMLWriter.jl
```diff
@@ -2484,6 +2484,12 @@ function domify(dctx::DCtx, node::Node, f::MarkdownAST.FootnoteDefinition)
     return DOM.Node[]
 end
 
+# This function provided by Michael Goerz in https://github.com/JuliaDocs/MarkdownAST.jl/issues/18
+function _markdownast_to_str(node::MarkdownAST.Node)
+    text = Documenter.MDFlatten.mdflatten(node)
+    return strip(text)
+end
```
asinghvi17:
```suggestion
```

src/html/HTMLWriter.jl
```diff
@@ -2484,6 +2484,8 @@ function domify(dctx::DCtx, node::Node, f::MarkdownAST.FootnoteDefinition)
     return DOM.Node[]
 end
 
+# This function provided by Michael Goerz in https://github.com/JuliaDocs/MarkdownAST.jl/issues/18
+
```
asinghvi17:
```suggestion
```

fingolfin:
@asinghvi17 Looks good to me now, but could you please add a changelog entry?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Improved several warning/error messages to (more accurately) report the location (filename, line range) in which the warning/error originated. ([#2426], [#2752], [#2793], [#2803], [#2809])
 * Warn/error if jldoctest starts with an empty line; or lacks empty line between two REPL prompts. ([#2031], [#2083], [#2679], [#2808])
 * The `forcepush=true` option to `deploydocs` now uses `--force-with-lease` instead of `--force`. ([#2817])
+* Improved hashing in `HTMLWriter` to use `mdflatten` instead of trying to compute the hash of the Julia native node structure. ([#2710])
```
fingolfin: I am not sure this helpful for end users, as they don't know what is being hashed. Better to explicitly mention admonitions, and also what the hash is being used for? For example
```suggestion
* Improved the generation of anchors for admonitions to be more stable and e.g. not change due to changes of the internal representation across different Julia versions. ([#2710])
```

asinghvi17:
```suggestion
* Improved the generation of anchors for admonitions in HTML output to be more stable and e.g. not change due to changes of the internal representation across different Julia versions. ([#2710])
```

asinghvi17: Thanks, that makes sense.  Just changed it a bit to specify that this is only for HTML output, not latex or any other backend.

PR #2714, opened 2025-05-12
Use appropriate show method for PageLink

### LINKED ISSUE #2711: Build and deploy with GitHub Actions
With the release of version 1.11.0, GitHub Actions consistently shows "The operation was canceled" without providing a clear explanation. The process halts right after the first warning about exceeding the size threshold, as shown in [this run](https://github.com/mcosovic/JuliaGrid.jl/actions/runs/14946702760/job/41990946499).

I also tested the new version locally and encountered the same issue. With version 1.10.2, everything works as expected.

mortenpi:
Competitor for #2710, to resolve https://github.com/JuliaDocs/Documenter.jl/pull/2676#discussion_r2082477911. Also organizes the Documenter `AbstractInline`s into an abstract supertype, with a fallback `show` method.

Fix #2711, might supersede #2710.

cc @asinghvi17 @vtjnash


vtjnash (commented):
SGTM

asinghvi17 (approved):
This looks good to me as well, a bit cleaner than my solution :D 

Should we have some defined interface for AbstractDocumenterBlock and AbstractDocumenterInline, that defines what methods they need to implement?

PR #2716, opened 2025-05-12
Remove misleading remark about ./ paths in showcase.md

### LINKED ISSUE #2559: ./ is relative, not an alias for src
The [Showcase Images](https://documenter.juliadocs.org/stable/showcase/#Images) section states:
> Alternatively, use ./ to begin a path relative to the src of the documents, e.g., ./assets/logo.png.

However, this is not actually a feature.

`./assets/logo.png` and `assets/logo.png` both work in this file because it's within `docs/src/`

But if you try `./` in a subdirectory's markdown file, such as `docs/src/man/doctests.md`
```markdown
![Enter a descriptive caption for the image](./assets/logo.png)
```

You get this error:
```
┌ Error: invalid local link/image: file does not exist in src\man\doctests.md
│   link =
│    @ast MarkdownAST.Image("./assets/logo.png", "") do
│      MarkdownAST.Text("Enter a descriptive caption for the image")
│    end
```

fingolfin:
Resolves #2559 by removing the incorrect sentence. I thought about rephrasing it but could not think of anything that just restates the previous sentence with more complicated wording.


mortenpi (approved):
LGTM!

One thing that I would like to have working is that if you have absolute paths (e.g. `/assets/logo.png`), _then_ it would resolve that relative to `src/`. But that doesn't seem to work right now. And I agree that `./` is definitely not correct. So just removing the extra sentence seems to be the right way to go.

PR #2717, opened 2025-05-12
Preemptively support JSON 1.0 release

quinnj:
As [noted](https://github.com/JuliaIO/JSON.jl/pull/374#issuecomment-2870385813), the JSON 1.0 release is currently blocked on using Documenter due to the circular dependency of needing Documenter to build docs. The usage of JSON.jl in Documenter.jl is very vanilla, so this PR proposed "preemptive" support for JSON.jl 1.0 since the usage of JSON in Documenter is known to not rely on any breaking changes proposed.


mortenpi:
LGTM! We should have a 1.11.1 out soon too which will include this.

PR #2718, opened 2025-05-12
add patch file option

matthias314:
Imagine that you have a package with an extensive, up-to-date documentation. Then you modify `show` for some type defined in your package, one that appears in almost all docstrings. (Alternatively, someone else did it in a package you use.) Now many doctests will fail, and you have to edit all them manually. It is for this scenario (that of course has happened to me) that I have added a `patchfile` option to `makedocs`. If one uses the options
```
warnonly = true, patchfile = "doctest.patch"
```
then after running Documenter one gets a patch file (in the `docs` directory) that one can apply to the source code. Assuming that all doctests could be located in the source, this will bring the documentation completely up-to-date.

The patch still needs some polishing, but I first want to inquire if you are interested. One thing one may want to change is how paths are stored in the patch file. Currently I'm using absolute paths.

### DIFF
--- a/src/doctests.jl
+++ b/src/doctests.jl
@@ -17,6 +17,33 @@ struct DocTestContext
     DocTestContext(file::String, doc::Documenter.Document) = new(file, doc, Dict())
 end
 
+struct Hunk
+    line::Int
+    old::String
+    new::String
+end
+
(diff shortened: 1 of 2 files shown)


matthias314:
OK, just discovered the `fix` option ...

PR #2721, opened 2025-05-13
Use symbol :display instead of function

quinnj:
I believe these Dict keys are meant to be Symbols rather than the `display` function. Caught in https://github.com/JuliaIO/JSON.jl/actions/runs/14942383442/job/42145192605?pr=374 where JSON.jl will no longer automatically call string on non-string-like Dict keys.


mortenpi (approved):
Good catch, thank you! I'll tag it right away.

PR #2722, opened 2025-05-15
Fix non-standard versions in deploydocs

mortenpi:
The Julia manual passes a `struct` as `versions. So we should make sure this keeps working. It's pretty undocumented, so let's add a warning.

X-ref: https://github.com/JuliaLang/julia/pull/58404#issuecomment-2882061362


mortenpi:
Note: regression test failure is unrelated (looks like random linkcheck failure).

mortenpi:
cc @jkrumbiegel just as an FYI. I think this should be fine for the DocumenterVitepress, so I'll just go ahead and tag this.

mortenpi:
https://github.com/JuliaRegistries/General/pull/131023

PR #2723, opened 2025-05-16
Fix non-standard versions in deploydocs (v2)

mortenpi:
Debugging: https://github.com/JuliaLang/julia/pull/58421#issuecomment-2883941876


mortenpi:
The preview build went through in [57a0a60](https://github.com/JuliaDocs/Documenter.jl/pull/2723/commits/57a0a606b265d73bddf1c15fea750f279dafd568) with the same setup as the Julia manual. So this should properly fix this.

mortenpi:
https://github.com/JuliaRegistries/General/pull/131108

PR #2724, opened 2025-05-16
Build Julia HTML manual in regression tests

mortenpi:


.github/workflows/regression-tests.yml
```diff
@@ -94,3 +95,24 @@ jobs:
...
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - name: "Regressions test"
+        run: ./.github/scripts/julia-manual-regression-test.sh
```
fingolfin, line 118:
```suggestion
        run: julia ./.github/scripts/julia-manual-regression-test.jl
```

.github/workflows/regression-tests.yml
```diff
@@ -94,3 +95,24 @@ jobs:
...
+      - uses: actions/cache@v4
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
```
fingolfin, line 116: Can't this just use `julia-actions/cache` ?
```suggestion
      - uses: julia-actions/cache@v2
```

.github/scripts/julia-manual-regression-test.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+#
+# This builds the Julia manual for the latest Julia nightly build with
+# the current Documenter, acting as a regression test.
+
+# Terminate the script early if we see any non-zero error codes
+set -e
```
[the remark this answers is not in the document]
mortenpi: Actually, I think this is a very fair point, and I also generally tend follow that philosophy. I think what just happened here was that at first I was like "this should be like 3 shell commands", but then the complexity silently crept in :smile:

.github/workflows/regression-tests.yml
mortenpi: This makes sense. I think we should update the other places in the regression tests workflow too.

.github/scripts/julia-manual-regression-test.sh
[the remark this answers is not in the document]
fingolfin: Hehe I have been in *exactly that* situation soooo many times ;-)

.github/workflows/regression-tests.yml
fingolfin: It has been updated everywhere else by now

.github/workflows/regression-tests.yml
```diff
@@ -94,3 +95,24 @@ jobs:
...
+      - name: "Regressions test"
```
fingolfin, line 95: Another request: upload the result as an artifact. That would be super useful for actually testing the result.

fingolfin:
I think this would be *really* useful to have.

.github/scripts/julia-manual-regression-test.sh
```diff
@@ -0,0 +1,44 @@
...
+# .. which we also want to clean up when the script exits
+trap "rm -r $TMP" EXIT
+
+# Get the date one week ago, which we'll use for shallow-cloning the Git repo
+one_week_ago=$($JULIA -e 'using Dates; print(Dates.format(now() - Week(1), dateformat"yyyy-mm-dd"))')
+julia_commit=$($JULIA -e 'print(Base.GIT_VERSION_INFO.commit)')
+
+# Clone the repo
+echo "Cloning JuliaLang/julia.git (--shallow-since=${one_week_ago}), checkout ${julia_commit}"
+git clone --branch master --shallow-since="${one_week_ago}" https://github.com/JuliaLang/julia.git
```
fingolfin:
```suggestion
git clone --branch master --shallow-since="1 week ago" https://github.com/JuliaLang/julia.git
```
You don't need to muck around with commits or computing dates and can just do the above.

But *why* is this desirable, as opposed to just a regular shallow clone?

fingolfin:
I've simplified the script quite a bit.

Running into this error:
```
ERROR: LoadError: IOError: readdir("/home/runner/work/Documenter.jl/Documenter.jl/julia.git/usr/share/julia/stdlib/v1.14/"): no such file or directory (ENOENT)
```

.github/scripts/julia-manual-regression-test.sh
mortenpi: I didn't know you can neither specify the date this way, nor about `git clone --revision` :slightly_smiling_face:

mortenpi:
@fingolfin I missed your changes to the bash script initially. But I looked through your cleanup there, and I was able to incorporate some of those into the Julia script as well, so they were still very useful!

PR #2725, opened 2025-05-16
Add Documenter version to default footer

IanButterworth:
Also add the datetime?

### DIFF
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -502,7 +502,7 @@ struct HTML <: Documenter.Writer
             highlights::Vector{String} = String[],
             mathengine::Union{MathEngine, Nothing} = KaTeX(),
             description::Union{String, Nothing} = nothing,
-            footer::Union{String, Nothing} = "Powered by [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) and the [Julia Programming Language](https://julialang.org/).",
+            footer::Union{String, Nothing} = "Powered by [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) v$(pkgversion(Documenter)) and the [Julia Programming Language](https://julialang.org/).",
             ansicolor::Bool = true,
             lang::String = "en",
             warn_outdated::Bool = true,


IanButterworth:
1.6 failure because `pkgversion` wasn't available back then. Time to update compat to 1.10?

IanButterworth:
After this I realized the info is in the settings modal. Perhaps the footer just needs a "more info" link to the settings page?

For the package version there's the version selector?

goerz:
> After this I realized the info is in the settings modal.

Oh, I didn't realize that, but that's certainly useful information

> Perhaps the footer just needs a "more info" link to the settings page?

I guess my question is: why do you think that prominently showing the version of Documenter is important at all? Not that I object to us doing a little self-advertising, but even the "powered by Documenter" doesn't seem like the most critical information to include in the footer. Much less, which version of Documenter. I wouldn't _mind_ adding the version of Documenter to that default footer, as long as that involves no trade-offs like changing the Julia compatibility of Documenter. So I'm just curious what issue you ran into that prompted this PR (maybe there _is_ a good reason to include the Documenter version that would justify dropping compatibility).


> For the package version there's the version selector?

Not for `dev`, though, which is a long-standing pet peeve I have. I've long advocated for the version to also be shown between the project name and the search bar, but @mortenpi and I have a disagreement about the role of the "version" that is defined in `Project.toml` (what I show in the footer, and what I think should be shown in the sidebar), respectively the "version" derived from git tags (shown in the selector). [Long story](https://github.com/JuliaDocs/Documenter.jl/pull/2442). Me manually putting the project version in the footer of all my projects is my work-around for that pet peeve. But anyway, that's a total side discussion. My general point was: I would encourage people setting the footer manually with whatever information they feel is relevant in a footer, which may very well be the Documenter-version. Changing the default footer in any non-trivial way would be something that would require reasonably strong justification from my point of view, especially if that involves any tradeoffs… But that's just my 2 cents; that's really @mortenpi's decision 😉

goerz:
Separately, though: we have `Documenter.DOCUMENTER_VERSION` as an existing constant that contains the version. So that could presumably be used in the footer, and I don’t think we actually need to drop compatibility with older Julia versions just to modify the footer.

PR #2726, opened 2025-05-19
Assets: allow passing direct HTML content into `asset` kwarg (for Pluto support)

fonsp:
Hi! 👋

I am working on adding easy Pluto support to Documenter via a plugin 😊 I am not yet sure about the user-facing API, but I got an implementation working where Pluto is embedded inside Documeter pages (alongside existing content). 

# Demo
Here is a WIP demo of the Pluto notebook embedded in Documenter.jl. Note that you can "Edit or run this notebook" with binder!

https://dashing-custard-e9c5cc.netlify.app/

The code for this demo is here: https://github.com/fonsp/Documenter.jl-Pluto-embed-test

# No `<iframe>`
I currently do the embedding **without** `<iframe>`, but with a Custom Element. This is also used in other Pluto-embedding web sites like [Computational Thinking](https://computationalthinking.mit.edu/Fall24/homework/hw1/). The benefit is better performance, better scrolling, accessibility and other UA features, and your notebook content can be indexed by search engines.

For this embedding to work, I need to place content in the `<head>` of the Documenter.jl output HTML. I saw that there is an existing `HTMLAsset` system, but unfortunately this is hard to use in our case: the `<head>` content of Pluto notebooks contains more asset types than just JS, CSS and ICO. We also don't have the `<head>` content in a Julia format: the `<head>` is generated dynamically by our bundler. So to use `HTMLAsset`, we would first need a DOM parser to parse the head content, then convert it into `HTMLAsset` objects...

# Changes
This PR adds the ability to add raw HTML content as an element of the `assets` kwarg, using the new type `RawHTMLHeadContent`. This gives plugins more flexibility in loading `<head>` content.

Take a look at the code changes for the implemented new types.

I'm also happy to hear other ideas for implementing this goal!

# Existing solutions
There are some existing solutions for Pluto in Documenter:
- PlutoStaticHTML.jl (maintained by another Pluto developer). The idea here is different: output as-simple-as-possible static HTML, to make the content match the Documenter.jl style. In some cases, this is exactly what users wants. In other cases, having a true Pluto embed would add consistent styling (WYSIWYG publishing), interactivity, run-with-binder, and more. 
- DemoCards.jl uses PlutoStaticHTML.jl
- Iframe embeds: I believe @adrhill once made a website that embeds Pluto notebooks using iframes. The goal of the new plugin is to have the direct embedding technical benefits described above, plus an easier workflow for adding notebooks to documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Version [v1.13.0] - not yet released
+## Version [v1.13.0] - 2025-06-19
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1544,6 +1544,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [v1.11.3]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.11.3
 [v1.11.4]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.11.4
 [v1.12.0]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.12.0
+[v1.13.0]: https://github.com/JuliaDocs/Documenter.jl/releases/tag/v1.13.0
 [#198]: https://github.com/JuliaDocs/Documenter.jl/issues/198
 [#245]: https://github.com/JuliaDocs/Documenter.jl/issues/245
 [#487]: https://github.com/JuliaDocs/Documenter.jl/issues/487
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Version [v1.13.0] - not yet released
+
+### Added
+
+* Added new type `RawHTMLHeadContent` to `HTML` format object, which allows to add raw HTML to the head of the HTML output, by passing it as a element in the `assets` keyword argument. ([#2726])
+
 ## Version [v1.12.0] - 2025-06-06
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2111,6 +2111,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2721]: https://github.com/JuliaDocs/Documenter.jl/issues/2721
 [#2722]: https://github.com/JuliaDocs/Documenter.jl/issues/2722
 [#2723]: https://github.com/JuliaDocs/Documenter.jl/issues/2723
+[#2726]: https://github.com/JuliaDocs/Documenter.jl/issues/2726
 [#2729]: https://github.com/JuliaDocs/Documenter.jl/issues/2729
 [#2737]: https://github.com/JuliaDocs/Documenter.jl/issues/2737
 [JuliaLang/julia#36953]: https://github.com/JuliaLang/julia/issues/36953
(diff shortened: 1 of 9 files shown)


mortenpi:
Allowing raw HTML head content seems fine to me. Regardless of the result of the discussion below, I think it's worth having this -- this seems like a useful thing other people can use to hack stuff together.

For the Pluto case, I have one question though -- do you want it to be included in every page? I guess ideally it would get added into the `<head>` "on-demand"?

I think the at-raw block is a good MVP. In the longer term, my two cents would be to try to turning this into a plugin package (DocumenterPluto?) that then defines a custom block. Something like this:

````
```@pluto
Path="docs/notebooks/notebook1.jl"
```
````

That might take care of the unnecessary duplication and the need to manually pass the asset -- the at-pluto block is present, then we add the Pluto stuff into `<head>`.

We don't have a really great plugin API (contributions very welcome!), but in DocumenterMermaid I hacked something similar together in this way (here: https://github.com/JuliaDocs/DocumenterMermaid.jl/blob/main/src/DocumenterMermaid.jl):

1. There's an at-mermaid block (`MermaidBlock`) that inserts the diagram itself.
2. It also adds a build step that loops over all pages and pushes a `MermaidScriptBlock` to the beginning of each page, which then loads the JS dependency.

In the Pluto case, you want this to be in the HEAD. What might work is, instead of pushing a block to the page AST, you just push the `RawHTMLHeadContent` into the `HTML`'s `.assets` field in a build step.

fonsp:
Thanks! I added docs, but I'm not sure how to write the test. I thought to check if `Documenter.HTML(assets = [RawHTMLHeadContent("<!-- hello -->")])` writes the comment to the output HTML, but I need a `Document` to test it on. Any suggestion?

fonsp:
Thanks for the ideas! The `@pluto` block looks good :)

> In the Pluto case, you want this to be in the HEAD. What might work is, instead of pushing a block to the page AST, you just push the RawHTMLHeadContent into the HTML's .assets field in a build step.

This sounds like a good approach! How can I do this selectively for different pages, so that only the pages with Pluto notebooks load the assets?

asinghvi17:
> but I need a Document to test it on.

`debug = true` on `makedocs` _should_ return a document....

> do this selectively for different pages

I don't think HTMLWriter allows for per-page assets?  It seems like this rendering happens here:

https://github.com/JuliaDocs/Documenter.jl/blob/90a1a0ea739768fc3db96b55bc5357d63397df65/src/html/HTMLWriter.jl#L1048-L1049

and the nav node doesn't actually hold the page struct, only its key, so it's not possible to do it that way either.  What solution were you thinking of for dynamic HEAD injection @mortenpi?

Of course the way you select the page is a bit simpler: see e.g [here](https://github.com/asinghvi17/OhMyCards.jl/blob/87e1a14656ec5a8524dbee1c3f835267a67c6556/src/build_step.jl#L17-L27), but you can replace `filter` with `findfirst`.  If the result is `nothing`, then no block met that criterion and you can ignore that particular page.

fonsp:
> debug = true on makedocs should return a document....

I mean specifically for writing tests for this PR: I'm wondering what to write in [test/htmlwriter.jl](https://github.com/JuliaDocs/Documenter.jl/pull/2726/files#diff-745e5768bc7fc9cbb28484764f0c5a5ed3626375138d38cd4aeb05f1bb6be278) so that I can check the output HTML contents.

mortenpi:
> Thanks! I added docs, but I'm not sure how to write the test.

The best option would probably to test this as part of the examples tests.

- https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/make.jl
- https://github.com/JuliaDocs/Documenter.jl/blob/master/test/examples/tests.jl

I'd add it to one example build, and then grep-check if it appears in the generated HTML.

> > do this selectively for different pages
> 
> I don't think HTMLWriter allows for per-page assets?

No. We'd have to add support for that :slightly_smiling_face: 

Just thinking out loud: `Page` should be writer-independent, so it might be better to do this somehow in the writer directly. E.g. add some facility for extensions to return extra assets, but in a way that they can return different stuff for different pages.

fonsp:
This PR should be ready now 👍

fonsp:
Didn't see the CI failure, should be good now :) I also added it to the changelog

mortenpi (approved):
LGTM, thank you @fonsp! Also, sorry for the delay in reviewing!

PR #2734, opened 2025-05-31
Allow Documenter.RawNode through `@eval`

asinghvi17:
Markdown backends may sanitize text to avoid HTML detection in Markdown objects, permitting raw nodes may prevent that.

### DIFF
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -727,6 +727,8 @@ function Selectors.runner(::Type{Expanders.EvalBlocks}, node, page, doc)
             nothing
         elseif isa(result, Markdown.MD)
             convert(Node, result)
+        elseif isa(result, Documenter.RawNode)
+            Node(result) # wrap directly in a Node
         else
             # TODO: we could handle the cases where the user provides some of the Markdown library
             # objects, like Paragraph.


mortenpi:
Can you explain the use case here a bit? You have an at-eval block that tries to interpolate a `Documenter.RawNode`?

In general, I'd like the at-eval blocks to be a bit more powerful in this respect (being able to eval raw HTML into a document can often be handy). But I am not 100% sure exposing `RawNode` is the right approach.

asinghvi17:
Specifically I want to interpolate a video tag but only if the page is not in draft mode (otherwise it breaks the Vitepress build)

mortenpi:
Should we maybe use the standard `Docs.HTML` for that? I.e. if the block returns that, we'll just convert it to an `@raw html` node?

asinghvi17:
That loses the ability to return other forms of raw content, like Markdown, though (Docs doesn't seem to have a markdown thing).  Specifically the thing of interest there is to avoid automated markdown sanitization, for example...

Plus, RawNode is sufficiently simple that I don't think it will change much...

fingolfin:
Could you perhaps show an example of how you'd use this?

Even better, could we perhaps add a test case demonstrating (and testing) this?

PR #2740, opened 2025-06-11
Adding the complete architecture for search benchmarking

### LINKED ISSUE #2417: Add search benchmarks
It would be nice to have benchmarks to track search performance (i.e. ensuring that good results are ranked highly)

For example

Base.push! should rank above LibGit2.push! for the search query "push!" on docs.julialang.org. (it currently does not)

Feel free to contribute more candidate benchmarks to this thread. Ideally we could get them to run automatically in CI when folks propose changes to the searching/indexing functionality.

Rahban1:
this closes #2417, this still has a lot of work to do, I am just putting it out there so everybody can see and give their suggestions on it. this is nowhere completed. I am going to work on it, maybe end up changing a substantial amount but right now it is just a proof of concept. would love to get some feedback on it. what do you think could be improved and what do you guys have in mind???
Thank you


test/search/real_search.jl
```diff
@@ -0,0 +1,91 @@
...
+    return parsed["docs"]  # Return just the docs array
+end
+
+# Simple function that uses the existing search.js with real search data
+function real_search(query::String)
+    # Load the real search index automatically
+    search_index_data = load_real_search_index()
+    
+    # Create a Node.js script that recreates the search logic from search.js
+    wrapper_js = """
```
mortenpi: This one is pretty big. Might be better to have this as a `.js` file that we `read(..., String)` in.

test/search/real_search.jl
```diff
@@ -0,0 +1,50 @@
...
+    # Ensure minisearch is available
+    node_modules_path = joinpath(@__DIR__, "node_modules")
+    if !isdir(node_modules_path)
+        cd(@__DIR__) do
+            run(`npm install minisearch`)
+        end
+    end
+
+    # Run the wrapper
+    result = read(`node -e $wrapper_js`, String)
```
Hetarth02: Since we already have a file, can't we just do `node file.js`?

test/search/real_search.jl
```diff
@@ -0,0 +1,50 @@
...
+
+# Simple function that uses the existing search.js with real search data
+function real_search(query::String)
+    # Load the real search index automatically
+    search_index_data = load_real_search_index()
+
+    # Read the JS wrapper and inject data
+    wrapper_js = read(joinpath(@__DIR__, "wrapper.js"), String)
+    wrapper_js = replace(wrapper_js, "__SEARCH_INDEX__" => JSON.json(search_index_data))
+    wrapper_js = replace(wrapper_js, "__QUERY__" => "\"" * query * "\"")
```
Hetarth02, line 39: Am not sure, but do we need to close the file since we used `read`?

test/search/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "documenter-search-benchmark",
+  "version": "1.0.0",
+  "description": "Search benchmarking for Documenter.jl",
+  "dependencies": {
+    "minisearch": "^6.1.0"
```
Hetarth02: We should fix the version to what we use in the repo.

test/search/real_search.jl
```diff
@@ -0,0 +1,50 @@
...
+
+    # Ensure minisearch is available
+    node_modules_path = joinpath(@__DIR__, "node_modules")
+    if !isdir(node_modules_path)
+        cd(@__DIR__) do
+            run(`npm install minisearch`)
```
Hetarth02: Should install exact version which we use in the repo

test/search/real_search.jl
Rahban1: but we're replacing the placeholders with their values here
```julia
wrapper_js = replace(wrapper_js, "__SEARCH_INDEX__" => JSON.json(search_index_data))
wrapper_js = replace(wrapper_js, "__QUERY__" => "\"" * query * "\"")

```

test/search/real_search.jl
Rahban1: the read function in Julia automatically closes the file when done!

test/search/package.json
Rahban1: I guess we're using 6.1.0 version only :
```julia
importScripts(
    "https://cdn.jsdelivr.net/npm/minisearch@6.1.0/dist/umd/index.min.js"
  );
```

test/search/wrapper.js
mortenpi, line 1: This duplicates the `search.js`, so it would be good to somehow avoid the duplication here by reusing the `search.js` in the tests. It would likely require some changes on how we handle the `search.js` file.

https://github.com/JuliaDocs/Documenter.jl/blob/022c013b36ecb328c217db9e977e0339ec30b5b9/assets/html/js/search.js#L189-L216

test/search/run_benchmarks.jl
```diff
@@ -0,0 +1,74 @@
...
+    all_results = evaluate_all(real_search, all_test_queries)
+    
+    # Show only overall results in terminal
+    println("\n== Overall Results ==")
+    println("Average Precision: $(round(all_results.average_precision * 100, digits=1))%")
+    println("Average Recall: $(round(all_results.average_recall * 100, digits=1))%")
+    println("Average F1 Score: $(round(all_results.average_f1_score * 100, digits=1))%")
+    println("Total Relevant Found: $(all_results.total_relevant_found)")
+    println("Total Documents Retrieved: $(all_results.total_documents_retrieved)")
+    println("Total Relevant Documents: $(all_results.total_relevant_documents)")
```
mortenpi: It would be good if we'd compare these values against some known reference values (that we update as the search gets better).

mortenpi (commented):
A couple of thoughts:

- We could add a `make search-benchmarks` command: https://github.com/JuliaDocs/Documenter.jl/blob/master/Makefile
- We could add a simple CI job that runs these benchmarks on CI. It should be fine to just copy-paste a thing here for now: https://github.com/JuliaDocs/Documenter.jl/blob/master/.github/workflows/CI.yml
- It would be good to have a more realistic reference manual / search index we test against. Options: Documenter's manual or Julia manual.

.github/workflows/Benchmarks.yml
```diff
@@ -0,0 +1,52 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Generate search index
+        run: julia --color=yes --project=test/examples --code-coverage test/examples/make.jl
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DOCUMENTER_TEST_EXAMPLES: html
+      - name: Run search benchmarks
+        run: make search-benchmarks
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mortenpi: Just a thought: we could also upload the full benchmark output as an artifact if we want.

See https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#uploading-build-and-test-artifacts

.github/workflows/Benchmarks.yml
```diff
@@ -0,0 +1,52 @@
...
+          show-versioninfo: true
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-buildpkg@v1
+      - name: Build test examples
+        shell: julia --color=yes --project=test/examples {0}
+        run: |
+          using Pkg
+          Pkg.instantiate()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mortenpi: Actually, on the artifacts note: what we could do is to put this CI into the `CI.yml` and make it depend on the main test suite. We build the manuals anyway, so we could upload the search index from the CI run, and then download it into this CI benchmark job. Then we don't have to rebuild the test manual here.

Makefile
```diff
@@ -32,6 +32,9 @@ install-runic:
 test:
 	${JULIA} --project -e 'using Pkg; Pkg.test()'
 
+search-benchmarks:
+	cd test/search && ${JULIA} --project=../.. run_benchmarks.jl
```
mortenpi: Not sure if this messes up any path stuff in the script (since this will change the working directory). But if it does, it should probably be changed to use `@__DIR__` anyway when it tries to find files with relative paths.

```suggestion
	${JULIA} --project test/search/run_benchmarks.jl
```

test/search/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "documenter-search-benchmark",
+  "version": "1.0.0",
+  "description": "Search benchmarking for Documenter.jl",
+  "dependencies": {
+    "minisearch": "6.1.0"
```
mortenpi: We should figure out a way to avoid duplicating the version number in the codebase.

Rahban1: I am thinking of making the package.json as a source of truth and in real_search.jl read the version from package.json only. what do you think?

.github/workflows/Benchmarks.yml
Rahban1: I think this is a genius idea! worked beautifully

test/search/package.json
Rahban1: fixed it

test/search/run_benchmarks.jl
Rahban1: done

Project.toml
mortenpi, line 1: We don't want `Crayons` and `PrettyTables` as deps of the package. Instead, they should be test dependencies. Just need to be moved into `[extras]` and `[targets]` should be updated. Alternatively, we can have a separate environment for the search benchmark tests.

https://pkgdocs.julialang.org/v1/creating-packages/#Test-specific-dependencies

test/search/package.json
mortenpi: Sorry, I missed this. I think I would prefer if the version number is closer to the feature, not the tests. Also, are you proposing we copy `packages.json` too when we build the HTML site?

I can see that parsing it out from `search.js` could be annoying possibly.. but maybe could be a variable that we just regex for in the tests? Or it could be in the Julia source code in HTMLWriter, and we just template it into `search.js` when we build the HTML site?

test/search/real_search.jl
```diff
@@ -0,0 +1,49 @@
...
+    wrapper_js = replace(wrapper_js, "__SEARCH_INDEX__" => JSON.json(search_index_data))
+    wrapper_js = replace(wrapper_js, "__QUERY__" => "\"" * query * "\"")
+
+
+    # Write the wrapper to a temporary file and run it
+    return mktemp(@__DIR__) do path, io
+        write(io, wrapper_js)
+        close(io)
+        cd(@__DIR__) do
+            result = read(`node $path`, String)
```
mortenpi: Could we make this use the Node JLL? This means that the user wouldn't have to install Node just to run the benchmarks -- we can pull it in via the Julia package manager.

- https://juliahub.com/ui/Packages/General/NodeJS_22_jll

There's some example usage here for example: https://github.com/JuliaComputing/MultiDocumenter.jl/blob/e112da8b744f3393d037a7380e544797c2a41953/src/search/pagefind.jl#L31-L55

.github/workflows/CI.yml
```diff
@@ -213,6 +213,62 @@ jobs:
...
+          using Pkg
+          Pkg.instantiate()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+      - name: Install Node.js dependencies
+        run: npm install
+        working-directory: test/search
```
mortenpi: If we switch to the Node JLL, we can remove these.

```suggestion
```

test/search/wrapper.js
mortenpi, line 1: I would really like us to also get rid of this duplication somehow (i.e. that we'd have one place where we configure minisearch).

.github/workflows/CI.yml
```diff
@@ -213,6 +213,62 @@ jobs:
...
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
```
mortenpi: I don't think we need this to run in a matrix? I wouldn't expect this to change between different OSs.

```suggestion
    name: Search Benchmarks
    runs-on: ubuntu-latest
    needs: docs
    steps:
      - uses: actions/checkout@v4
      - uses: julia-actions/setup-julia@v2
        with:
          version: '1'
```

mortenpi (commented):
I think I would be happy to merge this as the first iteration if we could get the duplication & version number sorted as well!

Project.toml
Rahban1: done

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Created benchmarks for the search functionality ([#2740])
+
```
mortenpi: It's not a user-facing change, so doesn't actually warrant a CHANGELOG. I also added the `Skip Changelog` label to suppress the CI check.

```suggestion
```

.github/workflows/CI.yml
```diff
@@ -213,6 +213,59 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: 1
+          arch: x64
+          show-versioninfo: true
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-buildpkg@v1
+      - name: Instantiate main project environment
+        run: julia --project=. -e 'using Pkg; Pkg.instantiate()'
```
mortenpi, line 237: We need to instantiate the test environment here. There is [TestEnv](https://github.com/JuliaTesting/TestEnv.jl), but a better idea might be to use `[sources]` actually.

mortenpi: https://github.com/JuliaDocs/Documenter.jl/pull/2740/commits/3f3200105c55dcf8b5b3f84b8c6be3d009afb081

Rahban1:
how's this implementation @mortenpi ?

mortenpi (approved):
@Rahban1 LGTM! I think we should work a bit on getting more test cases in etc. But I link this infra, so let's get this merged!

Issue #2114, opened 2023-04-25
Search can't find e.g. `__init__`


BioTurboNick:
https://docs.julialang.org/en/v1/search/?q=__init__

![image](https://user-images.githubusercontent.com/1438610/234297738-a59a825a-a888-4f2f-8cf0-fd99bf2691ff.png)

mortenpi:
It's not surprising that this might fail. But how is `__init__` actually documented? I presume it's not a docstrings..

BioTurboNick:
Ah, complicating matters is that `__init__` has some new documentation in 1.9. But it does have a docstring, in `base/docs/basedocs.jl`.

There are a few longer pages it shows up in that I'd expect to see, though, such as here: https://docs.julialang.org/en/v1/manual/modules/

mortenpi:
Is the docstring included in the manual?

KristofferC:
On dev it is at least: https://docs.julialang.org/en/v1.10-dev/base/base/#__init__

BioTurboNick:
And that aside, `__module__` has an entry but isn't found by search either. It can be found by searching for `module`. So the underscores are maybe tripping up the index?

![image](https://user-images.githubusercontent.com/1438610/234593606-44588c5a-d889-48d5-8182-0185e8e78878.png)
![image](https://user-images.githubusercontent.com/1438610/234593725-a2c9a764-a35c-493c-953d-a373cc9503fa.png)

mortenpi:
> So the underscores are maybe tripping up the index?

Yeah, we are probably stripping those characters somewhere, but then forget to do it in another place or something along those line.

fredrikekre:
https://github.com/JuliaLang/julia/issues/45518

timholy:
There's a package example that's much smaller than Julia's own docs, in case that helps debugging: https://github.com/JuliaLang/PrecompileTools.jl/issues/34#issuecomment-1925275043. You'll see I created a gist with the contents of the search index, in case it helps debug this.

nsajko:
On https://docs.julialang.org/en/v1.12-dev/, the search also can't find `.=`, `->` or `::`, even though they're now included in the docs.

xref JuliaLang/julia#36290

timholy:
Note that in https://github.com/JuliaLang/PrecompileTools.jl/issues/34#issuecomment-1925275043 I've found that:
- the target string *does* appear in the `search_index.js` file
[...]

PR #2744, opened 2025-06-26
Improving the tokenizer

### LINKED ISSUE #1457: Search often gives insufficient results
The search function on https://docs.julialang.org is IMHO not very good. Some examples:
- searching for operators like `^` or `&&`  simply does not give any useful results at all
- OK, maybe I have to search for `Base.^` or `Base.:^`? Nope, does not work either
- Interestingly, searching for `Base.!=` gives me a correct hit, but at about position 15 or so of the search results (number 1 is for some reason `Base.fd`?)
- but what I really wanted to find is `Base.literal_pow`, which is mentioned in the documentation for `Base.:^`, so let's search for that; oh yeah, that gives the right result -- but again only at position 10 or so of the search results
- though searching for `literal_pow` *does* work right, at least something
- searching for `@info` gives no hit at all
- searching for `...` gives useless results (see https://github.com/JuliaLang/julia/issues/36496)

I'll try to add more concrete examples as I come across them; but I've really struggled many times to find stuff in the Julia manual, even when I *know* it is there.

(This was https://github.com/JuliaLang/julia/issues/38238)

Rahban1:
I have updated the tokenizer and improved on the custom trimmer. This aim to close #1457 and #2114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Version [v1.13.0] - 2025-06-19
 
+### Changed
+
+* improved the tokenizer and the custom trimmer to improve search results. ([#2741])
+
 ### Added
 
 * Added new type `RawHTMLHeadContent` to `HTML` format object, which allows to add raw HTML to the head of the HTML output, by passing it as a element in the `assets` keyword argument. ([#2726])
--- a/assets/html/js/search.js
+++ b/assets/html/js/search.js
@@ -204,7 +204,30 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
     },
     // add . as a separator, because otherwise "title": "Documenter.Anchors.add!", would not
     // find anything if searching for "add!", only for the entire qualification
-    tokenize: (string) => string.split(/[\s\-\.]+/),
+    //updated tokenizer
+    tokenize: (string) => {
+      const tokens = [];
+      let remaining = string;
+
+      const patterns = [
+        // regular identifiers and function names
+        /\b[A-Za-z_][A-Za-z0-9_!]*\b/g
+      ];
+
+      for (const pattern of patterns) {
+        pattern.lastIndex = 0;
+        let match;
+        while ((match = pattern.exec(remaining)) != null) {
+          const token = match[0].trim();
+          if (token && !tokens.includes(token)) {
+            tokens.push(token);
+          }
+        }
+      }
+
+      return tokens.filter(token => token.length > 0);
+
+    },
     // options which will be applied during the search
     searchOptions: {
       prefix: true,
--- a/assets/html/js/search.js
+++ b/assets/html/js/search.js
@@ -192,11 +192,11 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
     processTerm: (term) => {
       let word = stopWords.has(term) ? null : term;
       if (word) {
-        // custom trimmer that doesn't strip @ and !, which are used in julia macro and function names
+        // custom trimmer that doesn't strip (@,!,+, -, *,/,^,&, |, %,<, >, =, :, .) which are used in julia macro,function names and identifiers
         word = word
-          .replace(/^[^a-zA-Z0-9@!]+/, "")
-          .replace(/[^a-zA-Z0-9@!]+$/, "");
-
+          .replace(/^[^a-zA-Z0-9@!+\-/*^&%|<>._=:]+/, "")
+          .replace(/[^a-zA-Z0-9@!+\-/*^&%|<>._=:]+$/, "");
+        
         word = word.toLowerCase();
       }
 
--- a/assets/html/js/search.js
+++ b/assets/html/js/search.js
@@ -209,19 +209,41 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
       const tokens = [];
       let remaining = string;
 
+      // julia specific patterns
       const patterns = [
-        // regular identifiers and function names
-        /\b[A-Za-z_][A-Za-z0-9_!]*\b/g
+        // Module qualified names (e.g., Base.sort, Module.Submodule. function)
+        /\b[A-Za-z0-9_1*(?:\.[A-Z][A-Za-z0-9_1*)*\.[a-z_][A-Za-z0-9_!]*\b/g,
+        // Macro calls (e.g., @time, @async)
+        /@[A-Za-z0-9_]*/g,
+        // Type parameters (e.g., Array{T,N}, Vector{Int})
+        /\b[A-Za-z0-9_]*\{[^}]+\}/g,
+        // Function names with module qualification (e.g., Base.+, Base.:^)
+        /\b[A-Za-z0-9_]*\.:[A-Za-z0-9_!+\-*/^&|%<>=.]+/g,
+        // Operators as complete tokens (e.g., !=, aã, ||, ^, .=, →)
+        /[!<>=+\-*/^&|%:.]+/g,
+        // Function signatures with type annotations (e.g., f(x::Int))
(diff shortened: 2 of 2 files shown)


Rahban1:
These are the benchmarks results after these changes in the tokenizer : 
Screenshot 2025-07-16 at 3 26 33 PM

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Version [v1.13.0] - 2025-06-19
 
+### Changed
+
+* improved the tokenizer and the custom trimmer to improve search results. ([#2744])
```
asinghvi17:
```suggestion
* Improved the search tokenizer and custom trimmer to improve search results. ([#2744])
```

This also needs a definition of the link 2744 down at the bottom of the file.

assets/html/js/search.js
```diff
@@ -192,10 +192,10 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
     processTerm: (term) => {
       let word = stopWords.has(term) ? null : term;
       if (word) {
-        // custom trimmer that doesn't strip @ and !, which are used in julia macro and function names
+        // custom trimmer that doesn't strip (@,!,+, -, *,/,^,&, |, %,<, >, =, :, .) which are used in julia macro,function names and identifiers
```
fingolfin:
```suggestion
        // custom trimmer that doesn't strip (@,!,+, -, *, /, ^, &, |, %, <, >, =, :, .) which are used in julia macro names, function names and identifiers
```
Why the addition of `identifiers` here -- which identifiers are relevant here that aren't function or macro names?

Might be nicer to make the list more compact like so:
```suggestion
        // custom trimmer that doesn't strip special characters `@!+-*/^&|%<>=:.` which are used in julia macro and function names.
```

assets/html/js/search.js
```diff
@@ -204,7 +204,53 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
     },
     // add . as a separator, because otherwise "title": "Documenter.Anchors.add!", would not
     // find anything if searching for "add!", only for the entire qualification
-    tokenize: (string) => string.split(/[\s\-\.]+/),
+    //updated tokenizer
```
fingolfin:
```suggestion
```

assets/html/js/search.js
```diff
@@ -337,6 +412,15 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
       combineWith: "AND",
     });
 
+    //calculate custom scores for all results
```
fingolfin:
```suggestion
    // calculate custom scores for all results
```

test/search/wrapper.js
```diff
@@ -21,12 +21,62 @@ const index = new MiniSearch({
...
         }
+  
         return word ?? null;
     },
-    tokenize: (string) => string.split(/[\s\-\.]+/),
+    tokenize: (string) => {
+        const tokens = [];
+        let remaining = string;
+  
+        // julia specific patterns
```
fingolfin, line 38: Huh, it seems a lot of code is duplicated here (and hence might get out of sync) ? I wonder if there is a way we could avoid that, e.g. putting shared code into a separate file that is included in both places?

(Of course this goes way beyond the scope of this PR, i.e., I am not asking you to take care of this here, I just wanted to put the thought out there.

Rahban1:
@fingolfin Thank you for reviewing this, you have a great point about testing the updates and that is why I created the benchmarks #2740 to test these changes. 


> These are the benchmarks results after these changes in the tokenizer : Screenshot 2025-07-16 at 3 26 33 PM

this picture is from these benchmarks only and they are also in the ci/cd pipeline to check if any new change cause regression to these tests

and added more test cases here #2757

fingolfin:
@Rahban1 ahhh thank you for your explanation, I wasn't aware of the benchmarking setup, that's fantastic!

So, does that mean this PR makes it so that search in the Julia manual, when built with your patch, will be able to find entries for `^` or `&&` or `...`?

Rahban1:
Yessss 😄

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Version [v1.13.0] - 2025-06-19
 
+### Changed
```
fingolfin: I think the `Changed` section should be after the "Added" section.

But in any case, this is in the wrong place here, we are already beyond 1.15.0 :-)

assets/html/js/search.js
```diff
@@ -204,7 +204,52 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
...
+        /@[A-Za-z0-9_]*/g,
+        // Type parameters (e.g., Array{T,N}, Vector{Int})
+        /\b[A-Za-z0-9_]*\{[^}]+\}/g,
+        // Function names with module qualification (e.g., Base.+, Base.:^)
+        /\b[A-Za-z0-9_]*\.:[A-Za-z0-9_!+\-*/^&|%<>=.]+/g,
+        // Operators as complete tokens (e.g., !=, aã, ||, ^, .=, →)
+        /[!<>=+\-*/^&|%:.]+/g,
+        // Function signatures with type annotations (e.g., f(x::Int))
+        /\b[A-Za-z0-9_!]*\([^)]*::[^)]*\)/g,
+        // Numbers (integers, floats,scientific notation)
```
fingolfin:
```suggestion
        // Numbers (integers, floats, scientific notation)
```

test/search/wrapper.js
```diff
@@ -21,12 +21,62 @@ const index = new MiniSearch({
...
+          /@[A-Za-z0-9_]*/g,
+          // Type parameters (e.g., Array{T,N}, Vector{Int})
+          /\b[A-Za-z0-9_]*\{[^}]+\}/g,
+          // Function names with module qualification (e.g., Base.+, Base.:^)
+          /\b[A-Za-z0-9_]*\.:[A-Za-z0-9_!+\-*/^&|%<>=.]+/g,
+          // Operators as complete tokens (e.g., !=, aã, ||, ^, .=, →)
+          /[!<>=+\-*/^&|%:.]+/g,
+          // Function signatures with type annotations (e.g., f(x::Int))
+          /\b[A-Za-z0-9_!]*\([^)]*::[^)]*\)/g,
+          // Numbers (integers, floats,scientific notation)
```
fingolfin, line 52:
```suggestion
          // Numbers (integers, floats, scientific notation)
```

test/search/wrapper.js
```diff
@@ -21,12 +21,62 @@ const index = new MiniSearch({
...
         return word ?? null;
     },
-    tokenize: (string) => string.split(/[\s\-\.]+/),
+    tokenize: (string) => {
+        const tokens = [];
+        let remaining = string;
+  
+        // julia specific patterns
+        const patterns = [
+          // Module qualified names (e.g., Base.sort, Module.Submodule. function)
```
fingolfin:
```suggestion
          // Module qualified names (e.g., Base.sort or more generally Module.Submodule.function)
```

assets/html/js/search.js
```diff
@@ -204,7 +204,52 @@ function worker_function(documenterSearchIndex, documenterBaseURL, filters) {
...
     // add . as a separator, because otherwise "title": "Documenter.Anchors.add!", would not
     // find anything if searching for "add!", only for the entire qualification
-    tokenize: (string) => string.split(/[\s\-\.]+/),
+    tokenize: (string) => {
+      const tokens = [];
+      let remaining = string;
+
+      // julia specific patterns
+      const patterns = [
+        // Module qualified names (e.g., Base.sort, Module.Submodule. function)
```
fingolfin, line 213:
```suggestion
        // Module qualified names (e.g., Base.sort or more generally Module.Submodule.function)
```

CHANGELOG.md
mortenpi: I took the liberty of moving this to the right place.

PR #2745, opened 2025-06-26
latest prettier version in CI

Rahban1:
This uses node to install the most recent prettier version and then lint according to it, so in future if prettier version 4 comes we don't have to change this again. This close #2195

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -264,9 +264,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actionsx/prettier@v2
+      - uses: actions/setup-node@v4
         with:
-          args: --check .
+          node-version: '20'
+      - run: npm install --global prettier
+      - run : prettier --check . 
 
   runic:
     name: "Runic (Julia code formatting)"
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Version [v1.13.0] - 2025-06-19
 
+### Changed
+
+* changed CI so that it always fetches the most recent version of prettier for linting
+
```
mortenpi: This is not a substantial change that a user cares about, so it doesn't warrant a CHANGELOG entry.

```suggestion
```

.github/workflows/CI.yml
```diff
@@ -264,9 +264,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actionsx/prettier@v2
+      - uses: actions/setup-node@v4
         with:
-          args: --check .
+          node-version: '20'
+      - run: npm install --global prettier
+      - run : prettier --check .
```
mortenpi: I'd prefer we stick to Prettier v3 and also reuse the action, rather than roll our own. Using the latest like this means that we'll get hit by CI failures should there be another breaking Prettier release.

fingolfin:
I guess if this PR here gets merged, then PR #2195 can be closed?

Rahban1:
I didn't notice there was an issue open for it. Thanks for pointing it out :)

Rahban1:
after conversation with @mortenpi I have just changed the version from v2 to v3, as we don't want random CI errors if prettier upgrades

Issue #2747, opened 2025-07-07
deploy on `GITHUB_EVENT_NAME` set to `release`


t-bltg:
I'm wondering why the rationale is for **not** allowing deployment on `GITHUB_EVENT_NAME` set to `release` ?

This occurs when creating a new release from github.

E.g. https://github.com/JuliaPlots/Plots.jl/actions/runs/16108223109/job/45447292552#step:7:9975:
```julia
│ - ✘ ENV["GITHUB_EVENT_NAME"]="release" is "push", "workflow_dispatch" or "schedule"
```

Would the following changes be acceptable ?
```diff
diff --git a/src/deployconfig.jl b/src/deployconfig.jl
index e0f5e0875..c2cf0a719 100644
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -357,7 +357,7 @@ function deploy_folder(
     println(io, "- $(marker(repo_ok)) ENV[\"GITHUB_REPOSITORY\"]=\"$(cfg.github_repository)\" occurs in repo=\"$(repo)\"")
     if build_type === :release
         ## Do not deploy for PRs
-        event_ok = in(cfg.github_event_name, ["push", "workflow_dispatch", "schedule"])
+        event_ok = in(cfg.github_event_name, ["push", "workflow_dispatch", "schedule", "release"])
         all_ok &= event_ok
         println(io, "- $(marker(event_ok)) ENV[\"GITHUB_EVENT_NAME\"]=\"$(cfg.github_event_name)\" is \"push\", \"workflow_dispatch\" or \"schedule\"")
         ## If a tag exist it should be a valid VersionNumber
```

Instead I have to manually trigger `workflow_dispatch`, and this is a pain.
I have this is my `docs.yml`:
```yaml
on:
  workflow_dispatch:
  push:
    branches: [v2]
    tags: '*'
  release:
    types: [published]
  pull_request:
```
, so I expect `Documenter` to deploy my docs when I publish a new release.

mortenpi:
This seems fine to me. I'm pretty sure the reason it was never added to that list is just because no-one has really needed it so far, since the standard flow is to trigger on the `push` event for the tag (usually created by TagBot).

Just out of curiosity: when you create a release, and it creates the tag, it won't trigger the `push` event, just the `release` one? Or is the `push` just excluded in your case because it doesn't match `on.push.branches`?

t-bltg:
Thanks !

[...]

PR #2748, opened 2025-07-07
Add `release` to allowed vals for `GITHUB_EVENT_NAME`

t-bltg:
Fix https://github.com/JuliaDocs/Documenter.jl/issues/2747.


mortenpi:
Thank you! Could use a CHANGELOG note. But the other CI failures are unrelated.

mortenpi (approved):
Thank you @t-bltg! Should be in a release shortly.

PR #2750, opened 2025-07-09
fix info message for `release`

t-bltg:
Minor fix, following https://github.com/JuliaDocs/Documenter.jl/pull/2748.
I should have caught that, my bad :cry: !!

The `PR` works well (so this fix can wait), but the information message might be confusing, see the most recent run :
https://github.com/JuliaPlots/Plots.jl/actions/runs/16168125338/job/45634742499#step:7:9919


mortenpi (approved):
All good, thanks for catching it!

PR #2753, opened 2025-07-22
Warn when the search index size is  too big

### LINKED ISSUE #2423: Warn/error when search index is too big
We now error if the generated HTML files are too big. But it is also possible to generate very large search indexes, which we probably also want to avoid. Should we should probably also warn/error for those.

Making it an error is probably a breaking change (?), so the default probably should be a warning for this file. Although an argument could be made that the file size limits are about any generated HTML/JS/CSS content, and hence it's just a bug that the search index is currently not being checked.

Rahban1:
this close #2423 
right now the threshold decided is 500Kib, anything bigger than that a warning will be displayed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Version [v1.14.1] - 2025-07-09
 
+### Changed
+
+* Created a warning for when the search index size is too big (500Kib). ([#2423])
+
 ### Fixed
 
 * Fixed the `GITHUB_EVENT_NAME` message to include `release` as one of the options. ([#2750])
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -502,6 +502,8 @@ struct HTML <: Documenter.Writer
     size_threshold_ignore::Vector{String}
     example_size_threshold::Int
     inventory_version::Union{String, Nothing}
+    search_size_threshold::Int
+    search_size_threshold_warn::Int
 
     function HTML(;
             prettyurls::Bool = true,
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Version [v1.14.1] - 2025-07-09
 
+### Changed
+
+* Created a warning for when the search index size is too big (500Kib). ([#2423], [#2753])
+
```
mortenpi, line 11: This needs to go under a new `## Unreleased` section. 1.14.1 is already released.

src/html/HTMLWriter.jl
```diff
@@ -1941,6 +1952,22 @@ function format_units(size)
...
+function check_search_index_size(path, settings)
+    file_size = filesize(path)
+    file_size_format_results = format_units(file_size)
+    size_threshold_warn_format_results = format_units(settings.search_size_threshold_warn)
+
+    size_threshold_msg(var::Symbol) = """
+    Generated search index over $(var) limit:
+        Generated file size: $(file_size_format_results)
+        search_size_threshold_warn: $(size_threshold_warn_format_results)
+        Search index file:   $(path)"""
```
mortenpi: This could all be within the `if` in principle. Also `size_threshold_msg` is not reused, so no need to make it a function.

src/html/HTMLWriter.jl
```diff
@@ -502,6 +502,7 @@ struct HTML <: Documenter.Writer
     size_threshold_ignore::Vector{String}
     example_size_threshold::Int
     inventory_version::Union{String, Nothing}
+    search_size_threshold_warn::Int
```
mortenpi: I'd group the fields together.

```suggestion
    search_size_threshold_warn::Int
    inventory_version::Union{String, Nothing}
```

Would also need a change to the `new()` call below.

mortenpi (commented):
This LGTM, thank you @Rahban1! The comment are purely bikeshedding the code. Only the CHANGELOG entry should be fixed.

src/html/HTMLWriter.jl
Rahban1: this makes a lot of sense, the whole purpose of function is reusability which we are not doing here, Thank you for the refactor :)

src/html/HTMLWriter.jl
Rahban1: I don't understand what do you mean here by grouping them together

asinghvi17: Size threshold parameters should be consecutive

PR #2755, opened 2025-07-26
Support self-hosted github installations

tz-lom:
This PR enables support for self-hosted GitHub installations

I've tested it with the following setup
```
repo = Remotes.GitHub("XXX/YYY.git";host="AAA.com")

deploydocs(;
    deploy_config = Documenter.GitHubActions(
        "gh.asml.com",
        "https://pages.AAA.com/XXX/YYY"
    ),
    repo = repo,
    push_preview = true,
```

Both deployment and notification worked well

Was made during JuliaCon 2025 hackathon!


mortenpi:
This looks fine to me! Could we also make sure that it's documented in the relevant docstrings please?

* https://github.com/JuliaDocs/Documenter.jl/blob/e972ea249269d2ffcf979b8d9848b2f1d516e520/src/deployconfig.jl#L295-L316
* https://github.com/JuliaDocs/Documenter.jl/blob/e972ea249269d2ffcf979b8d9848b2f1d516e520/src/utilities/Remotes.jl#L105-L120

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Version [v1.14.1] - 2025-07-09
 
+### Added
+
+* Support self-hosted GitHub instances.
+
```
mortenpi, line 87:
```suggestion
## Unreleased

### Added

* Support self-hosted GitHub instances.

## Version [v1.14.1] - 2025-07-09

```

tz-lom:
> This looks fine to me! Could we also make sure that it's documented in the relevant docstrings please?

I've updated documentation and hopefully - coverage.
But I want to mention that documentation is miss leading at this moment:
```
makedocs(
    repo = GitHub("JuliaDocs", "Documenter.jl")
)
```
The one above will probably work, but the one bellow will not (not because of my changes)
```
deploydocs(
    repo = GitHub("JuliaDocs", "Documenter.jl")
)
```

Is it just me reading documentation in a wrong way?

mortenpi:
Do we have somewhere where we advertise passing e.g. `GitHub()` to `deploydocs`'s `repo=`? I think right now we just accept a string there. I agree that it's a bit unfortunate that you can't just pass the object form there (we should probably expand that), but I am not sure there's anything misleading anywhere?

tz-lom:
> Do we have somewhere where we advertise passing e.g. `GitHub()` to `deploydocs`'s `repo=`? I think right now we just accept a string there. I agree that it's a bit unfortunate that you can't just pass the object form there (we should probably expand that), but I am not sure there's anything misleading anywhere?

No, I can't find anything in documentation, but in output of the build I see this:
image
and with `repo` parameter in `deploydocs` this is totally misleading.

Anyway, I've double checked that this PR works in self-hosted environment, do you need anything else or you can merge it?

tz-lom:
@mortenpi , can you review this please?

src/utilities/Remotes.jl
```diff
@@ -102,8 +102,11 @@ function repofile(remote::Remote, ref, filename, linerange = nothing)
     return fileurl(remote, ref, filename, isnothing(linerange) ? nothing : Int(first(linerange)):Int(last(linerange)))
 end
 
+
+const GITHUB_HOST = "github.com"
+
 """
-    GitHub(user :: AbstractString, repo :: AbstractString)
+    GitHub(user :: AbstractString, repo :: AbstractString, [host :: AbstractString])
     GitHub(remote :: AbstractString)
```
mortenpi, line 145: I took the liberty of removing `host` for the single-argument method. I think it would be better if it stays single-argument. If we want to add support for overriding the `host`, then let's figure out some syntax for the string, like `github.selfhosted:Org/Repo.jl`, which we can regex.

src/deployconfig.jl
```diff
@@ -311,19 +314,48 @@ when using the `GitHubActions` configuration:
...
+    # Compute GitHub Pages URL from repository
+    parts = split(github_repository, "/")
+    github_pages_url = if length(parts) == 2
+        owner, repo = parts
+        "https://$(owner).github.io/$(repo)/"
+    else
+        ""
+    end
+
+    return GitHubActions(github_repository, github_event_name, github_ref, "github.com", github_api, github_pages_url)
```
mortenpi: If we hard-code `github_host` here, it feels like we should also hard-code the API URL? I also have a minor security concern here, where this could allow someone redirect the API calls somewhere (including the token) by somehow attacking the `GITHUB_API_URL` environment variable (for normal GitHub-hosted repos).

```suggestion
    # Compute GitHub Pages URL from repository
    parts = split(github_repository, "/")
    github_pages_url = if length(parts) == 2
        owner, repo = parts
        "https://$(owner).github.io/$(repo)/"
    else
        ""
    end

    return GitHubActions(github_repository, github_event_name, github_ref, "github.com", "https://api.github.com", github_pages_url)
```

src/deployconfig.jl
```diff
@@ -311,19 +314,48 @@ when using the `GitHubActions` configuration:
...
+end
+
+function GitHubActions(host, pages_url)
+    github_repository = get(ENV, "GITHUB_REPOSITORY", "") # "JuliaDocs/Documenter.jl"
+    github_event_name = get(ENV, "GITHUB_EVENT_NAME", "") # "push", "pull_request" or "cron" (?)
+    github_ref = get(ENV, "GITHUB_REF", "") # "refs/heads/$(branchname)" for branch, "refs/tags/$(tagname)" for tags
+    github_api = get(ENV, "GITHUB_API_URL", "") # https://api.github.com
```
mortenpi, line 370: Is `GITHUB_API_URL` something that is configurable for a self-hosted instance, or can we assume it's `api.$(host)`?

There's also `GITHUB_SERVER_URL`, which could potentially be used for determining `host` automatically? Or would that not be reliable and/or two automagical?

src/deployconfig.jl
```diff
@@ -489,17 +521,18 @@ function post_status(::GitHubActions; type, repo::String, subfolder = nothing, k
...
 end
 
-function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
+
+function post_github_status(cfg::GitHubActions, type::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
     try
         Sys.which("curl") === nothing && return
         ## Extract owner and repository name
-        m = match(r"^github.com\/(.+?)\/(.+?)(.git)?$", deploydocs_repo)
+        m = match(Regex("^(?:https?://)?$(cfg.github_host)\\/(.+?)\\/(.+?)(.git)?\$"), deploydocs_repo)
```
mortenpi: I feel like that for clarity, it would be better to capture the host and string-compare later, rather than dynamically constructing the regex.

tz-lom: You right, I will change that

src/deployconfig.jl
tz-lom: I don't think we can assume anything about it, I don't know exactly if it is configurable but I would expect that it is as some orgs have strict rules about domain names

I need to check GITHUB_SERVER_URL , for some reason I haven't used that, don't know if it is because we don't have it or because I've missed it

if something like that exists I would prefer to used it indeed instead of specifying it manually

tz-lom:
@mortenpi I have number of issues in testing which I will address, but can you provide initial feedback about url parsing thing that I did based on your ideas, what do you think about it ?

src/deployconfig.jl
```diff
@@ -297,6 +297,9 @@ end
 
 Implementation of `DeployConfig` for deploying from GitHub Actions.
 
+For self-hosted GitHub installation use `GitHubActions(host, pages_url)` constructor
```
fingolfin:
```suggestion
For self-hosted GitHub installation use the `GitHubActions(host, pages_url)` constructor
```

src/deployconfig.jl
```diff
@@ -297,6 +297,9 @@ end
...
+to specify the host name and a **full path** to the GitHub pages location.
```
fingolfin: Is the "full path" referring to `pages_url` ? I find that confusing, a path is not an URL; maybe clarify the text?

src/deployconfig.jl
```diff
@@ -311,19 +314,45 @@ when using the `GitHubActions` configuration:
  - `GITHUB_TOKEN` or `DOCUMENTER_KEY`: used for authentication with GitHub,
    see the manual section for [GitHub Actions](@ref) for more information.
 
+ - `GITHUB_API_URL`: specifies the GitHub API URL, which generally is `https://api.github.com`,
+   but may be different for self-hosted GitHub instances.
+
+ - `GITHUB_ACTOR`: name of the person or app that initiated the workflow; this is used to construct
+   API calls.
```
fingolfin: That's too vague to actually be useful, unless the term "GitHub actor" is already defined in the documentation for self-hosted GitHub installations. If the latter is the case, I'd suggest state so explicitly, possibly even add a link to relevant docs.

Ah, I did some more googling and finally found <https://docs.github.com/en/actions/reference/workflows-and-actions/variables> which briefly mentions this and defines it thus:
> The name of the person or app that initiated the workflow. For example, `octocat`.

That last sentence with the example is super helpful. So how about
```suggestion
 - `GITHUB_ACTOR`: name of the person or app that initiated the workflow. For
   example, `octocat`. This is used to construct API calls.
```

src/deployconfig.jl
fingolfin: Note that `GITHUB_API_URL` is [mentioned in the GitHub manual](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) as a variable they set in GitHub Action workflows.

And in the corresponding enterprise docs at <https://docs.github.com/en/enterprise-server@3.18/actions/reference/workflows-and-actions/variables> the example value given is `http(s)://HOSTNAME/api/v3`  -- so no, we can't just assume it is `api.$(host)` (I assume that's the URL one gest if one enables "subdomain isolation")

src/deployconfig.jl
```diff
@@ -311,19 +314,45 @@ when using the `GitHubActions` configuration:
...
 end
+
 function GitHubActions()
     github_repository = get(ENV, "GITHUB_REPOSITORY", "") # "JuliaDocs/Documenter.jl"
+
+    # Compute GitHub Pages URL from repository
+    parts = split(github_repository, "/")
+    github_pages_url = if length(parts) == 2
+        owner, repo = parts
+        "https://$(owner).github.io/$(repo)/"
```
fingolfin: Of course you are just moving this around, and that's already a good step; so what I say below is not a change request, but a general thought (perhaps we should record it it into an issue)

Instead of hardcoding the pages URL like that (which won't if e.g. a custom domain name is used, as e.g. `Documenter.jl` does) one could also query the pages URL from the GitHub API. E.g. if the `gh` tool is installed:
```console
$ gh api repos/JuliaDocs/Documenter.jl --jq '.homepage'
https://documenter.juliadocs.org
```
But of course one also do that without, using just the `github_api`:
```
curl -s https://api.github.com/repos/JuliaDocs/Documenter.jl | jq -r '.homepage'
```
Obviously in Julia we'd use the `JSON` module, not `jq`, to parse this data.

src/deployconfig.jl
```diff
@@ -489,20 +518,25 @@ function post_status(::GitHubActions; type, repo::String, subfolder = nothing, k
...
 end
 
-function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
+
+function post_github_status(cfg::GitHubActions, type::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
     try
         Sys.which("curl") === nothing && return
         ## Extract owner and repository name
-        m = match(r"^github.com\/(.+?)\/(.+?)(.git)?$", deploydocs_repo)
+        m = match(Regex("^(?:https?://)?([^/]+)\\/(.+?)\\/(.+?)(.git)?\$"), deploydocs_repo)
```
fingolfin: Why not keep using an `r`-string?
```suggestion
        m = match(r"^(?:https?://)?([^/]+)\\/(.+?)\\/(.+?)(.git)?$", deploydocs_repo)
```

BTW, the original string assumed `github.com` at the start, no `https://` -- why the difference?

src/deployconfig.jl
```diff
@@ -489,20 +518,25 @@ function post_status(::GitHubActions; type, repo::String, subfolder = nothing, k
...
         m === nothing && return
-        owner = String(m.captures[1])
-        repo = String(m.captures[2])
+        host = String(m.captures[1])
+        owner = String(m.captures[2])
+        repo = String(m.captures[3])
+
+        host != cfg.github_host && return
```
fingolfin: perhaps this should be a warning?

src/deployconfig.jl
```diff
@@ -489,20 +518,25 @@ function post_status(::GitHubActions; type, repo::String, subfolder = nothing, k
             sha = get(ENV, "GITHUB_SHA", nothing)
         end
         sha === nothing && return
-        return post_github_status(type, repo, sha, subfolder)
-    catch
-        @debug "Failed to post status"
+        return post_github_status(cfg, type, repo, sha, subfolder)
+    catch e
+        @debug "Failed to post status" e
```
fingolfin:
```suggestion
        @debug "Failed to post status" exception = e
```

src/utilities/Remotes.jl
```diff
@@ -103,7 +103,30 @@ function repofile(remote::Remote, ref, filename, linerange = nothing)
...
 
 """
-    GitHub(user :: AbstractString, repo :: AbstractString)
+    parse_url(url)
+
+Return named tuple with values `scheme`, `authority`, `userinfo`, `username`, `password`, `hostname`, `port`, `path`, `query`, `fragment`.
+If certain value is not present in URL then it is omitted in return tuple.
+"""
+function parse_url(url)
+    m = match(r"^(?:(?P<scheme>https?\:)\/\/)?(?P<authority>(?:(?:(?P<userinfo>(?P<username>[^:]+)(?::(?P<password>[^@]+)?)?)@)?)(?P<hostname>[^:\/?#]*)(?:\:(?P<port>[0-9]+))?)(?P<path>[\/]{0,1}[^?#]*)(?P<query>\?[^#]*|)(?P<fragment>#.*|)$", url)
```
fingolfin: URI parsing is a bit tricky, though [rfc3986 officially suggests using a regex](https://datatracker.ietf.org/doc/html/rfc3986#appendix-B).

Is your regex based on that? Perhaps using a multiline regex would help to make this more understandable.

Or perhaps it would be better to instead just use the [`URIs` package](https://juliaweb.github.io/URIs.jl/dev/) to perform the parsing?

src/utilities/Remotes.jl
```diff
@@ -103,7 +103,30 @@ function repofile(remote::Remote, ref, filename, linerange = nothing)
...
+    return NamedTuple((name => m[name] for name in (:scheme, :authority, :userinfo, :username, :password, :hostname, :port, :path, :query, :fragment)))
+end
+
+"""
+    github_host()
+
+Returns hostname of the GitHub installation it is running on.
```
fingolfin: What is "it"? Perhaps "we" or "this code" or so?

src/utilities/Remotes.jl
```diff
@@ -103,7 +103,30 @@ function repofile(remote::Remote, ref, filename, linerange = nothing)
...
+Is determined by `ENV[GITHUB_SERVER_URL]` variable which is set by GitHub Actions runtime.
```
fingolfin: Some grammar issues; maybe this is a bit better?
```suggestion
This is derived from the `ENV[GITHUB_SERVER_URL]` variable which is set in every GitHub Actions workflow.
```

src/utilities/Remotes.jl
```diff
@@ -103,7 +103,30 @@ function repofile(remote::Remote, ref, filename, linerange = nothing)
...
+In case of missing variable default is "github.com"
```
fingolfin:
```suggestion
If this variable is not set, return "github.com".
```

src/utilities/Remotes.jl
```diff
@@ -103,7 +103,30 @@ function repofile(remote::Remote, ref, filename, linerange = nothing)
...
+"""
+function github_host()
+    url = get(ENV, "GITHUB_SERVER_URL", "github.com")
```
fingolfin: `github.com` is not a valid URL, is it? It may work here, but I think it would be better to do that:
```suggestion
    haskey(ENV, "GITHUB_SERVER_URL") || return "github.com"
    url = ENV["GITHUB_SERVER_URL"]
```
or
```suggestion
    url = get(ENV, "GITHUB_SERVER_URL", nothing)
    url === nothing && return "github.com"
```

test/deployconfig.jl
```diff
@@ -140,7 +140,7 @@ end
     with_logger(NullLogger()) do
         # Regular tag build with GITHUB_TOKEN
         withenv(
-            "GITHUB_EVENT_NAME" => "push",
+            "GITHUB_EVENT_NAME" => "push", "GITHUB_SERVER_URL" => "github.com",
```
fingolfin: Let's keep this to one per line?
```suggestion
            "GITHUB_EVENT_NAME" => "push",
            "GITHUB_SERVER_URL" => "github.com",
```
and then the same everywhere below.

test/deployconfig.jl
```diff
@@ -1494,3 +1517,91 @@ end
...
+@testset "post_status" begin
+    if Sys.which("curl") === nothing
+        @warn "'curl' binary not found, skipping related tests."
+    else
+        @testset "Default GitHubActions push" begin
+            buffer = IOBuffer()
+            logger = SimpleLogger(buffer, Logging.Debug)
+            with_logger(logger) do
+                withenv(
+                    "GITHUB_EVENT_NAME" => "push", "GITHUB_SERVER_URL" => "github.com",
```
fingolfin:
```suggestion
                    "GITHUB_EVENT_NAME" => "push",
                    "GITHUB_SERVER_URL" => "github.com",
```

fingolfin (commented):
Thank you, I think this is quite useful and sensible. I still have  a bunch of comments :-)

src/deployconfig.jl
tz-lom: That would be great, however I can't find any public API that will tell me where github pages are located.
homepage is not reliable and can point somewhere else 
There is private api `/repos/{owner}/{repo}/pages` but it requires authentication.

src/deployconfig.jl
tz-lom: * [x] Add warning

src/utilities/Remotes.jl
tz-lom: * [x] switch to URI package

src/utilities/Remotes.jl
tz-lom:
```suggestion
Returns hostname of the GitHub installation where this code is running on at the moment.
```

test/deployconfig.jl
tz-lom: * [x] check that this is not because of auto format and do the change uniformly

src/deployconfig.jl
tz-lom: * [x] review this description, seems that API update is not reflected here (is there constructor that takes `host` as argument?)

src/deployconfig.jl
asinghvi17: If this is running in CI, the `GITHUB_TOKEN` _might_ have sufficient permissions?

tz-lom: that's an interesting idea, I need to try this on my setup if that will work

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Support self-hosted GitHub instances. ([#2755])
+
```
fingolfin, line 11: This now needs to be moved up (it currently is in the 1.16.0 section but that version has already been released)

fingolfin:
This now has a merge conflict, and the changelog entry still needs to be moved.

It would be a shame to let this go to waste, it's so close to finished.

tz-lom:
@fingolfin 
Believe it or not, I was thinking to come back to this PR on Tuesday.
I've solved conflict and updated CHANGELOG

CodeCov say I am missing few lines in my change while I improved overall coverage 🤔
100% of uncovered lines is either changes in previously uncovered lines or exception handling, both of them are very tricky to test with tools that I have now.
I may add just a few more lines of coverage and maybe it will make the difference, but if not, will you consider this PR with codecov/patch failing?

tz-lom:
Just to mention - I've tested this PR on my self-hosted GH installation

tz-lom:
I believe Julia Manual failure is an infra issue, not code change (looks like LeX live do not install upquote anymore)

tz-lom:
Ah, looks like it was an issue with the pipeline in the end.

src/deployconfig.jl
```diff
@@ -469,7 +514,7 @@ function version_tag_strip_build(tag; tag_prefix = "")
     return "$s0$s1$s2$s3$s4"
 end
 
-function post_status(gha::GitHubActions; type, repo::String, subfolder = nothing, kwargs...)
+function post_status(gha::GitHubActions; type, subfolder = nothing, kwargs...)
```
giordano, line 517: The `repo` argument is [used](https://github.com/LuxDL/DocumenterVitepress.jl/blob/05209bf8d0a7c3a7c67690e3d562b6477a5030ad/src/DocumenterVitepress.jl#L184) ([incorrectly](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1008#issuecomment-4059057717)) by `DocumenterVitepress`, so removing it would break deployment with `DocumenterVitepress`.  ***Edit***: actually, this shouldn't fail the deployment because the extra `repo` argument will be swallowed by `kwargs`, but I'm pretty sure the status will still be pointing to the wrong repo, because they set `deploy_config = Documenter.auto_detect_deploy_system()`, and that wouldn't set the pages URL correctly.

PR #2761, opened 2025-08-19
Navigate search modal using up and down keys

Rahban1:
as discussed in the biweekly meeting @mortenpi @Hetarth02 

this is how it looks like locally for me : 

https://github.com/user-attachments/assets/f07d1803-65d7-4b45-a2a9-49841046f7b5


Hetarth02:
The UI looks pretty good, should we add a border while keyboard navigation?

CC @mortenpi

assets/html/js/shortcut.js
Hetarth02, line 1: Can't we just use, `element.focus()`?

Would eliminate all this extra logic from both js and scss...?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus

Rahban1: If I am using focus, I am getting this blue white border like : 
image
which I am not being able to remove I have done border: none, outline:none still it is there

Rahban1: most probably because these are < a > tags so there will always be an unavoidable outline

Hetarth02: You can remove the border using the `<element_classname>:focus {border:none;}`.

Also, I like the idea of the border it's not subtle(maybe we can make it) since it kinda highlights the current selection more than just the hover styles.

Rahban1: oh it works now, I forgot to run make themes so the changes were not reflecting, now the outline's gone

Rahban1: there's a subtle border idea or one more thing I was thinking of was we can also create a shadow effect for the selected result. What do you think? I am fine with both or neither whatever you recommend

Hetarth02: Ia it possible to get a small demo recording for both options, subtle border and shadow?

Rahban1: This is with the border

https://github.com/user-attachments/assets/7720ba1f-c364-41bf-9973-5a4c272d0427

This is with the shadow

https://github.com/user-attachments/assets/7706e69f-3771-4da4-ab47-7fe3a189243a

mortenpi: I kinda like the border FWIW, but @Hetarth02 feel free to make the call here.

mortenpi:
This looks really nice! @Hetarth02, if you're happy with the UI, we can merge whenever as far as I am concerned.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added 
+
+* Navigate search modal using up and down keys ([#2761])
```
```suggestion
* In the HTML output, the search modal can now be navigated using using up and down keys. ([#2761])
```

assets/html/scss/documenter/layout/_search.scss
```diff
@@ -33,6 +33,8 @@ kbd.search-modal-key-hints {
 
 .search-result-link:hover, .search-result-link:focus {
   background-color: $search-result-link-hover;
+  outline : none;
+  border: 2px solid #1DD2AF;
```
Hetarth02: Can we use a variable instead of hard-coded values for colors?

Also, would be good to add relevant color values for different themes.(but not super important)

assets/html/js/shortcut.js
Hetarth02: We went ahead with the border one.

assets/html/scss/documenter/layout/_search.scss
Rahban1: replaced the hard-coded value with a variable in _variables.scss

Rahban1: also checked every theme, the colors are looking good

Rahban1:
Thank you for the commit @mortenpi 😁

Hetarth02 (approved):

Hetarth02 (approved):

mortenpi:
The nightly failures look unrelated. But the search benchmark one is also still failing.

Rahban1:
PrettyTables.jl released a new version 4 days ago, that caused the ci issue :)

Rahban1:
https://discourse.julialang.org/t/ann-prettytables-v3-0-0/131821

mortenpi (approved):

mortenpi:
Thanks a lot for this @Rahban1 and @Hetarth02. Sorry for the delay in merging.

Rahban1:
Thank you @mortenpi 👍

PR #2773, opened 2025-10-14
Turn docstring accordions from article-header into details-summary

### LINKED ISSUE #2772: Usability issue: click target for collapsing and expanding docstrings is "suboptimal"
This was originally reported to use by @johnabbott at https://github.com/oscar-system/Oscar.jl/issues/5273 and I've asked him to turn it into an issue here, but he seems to be busy with other stuff, so I am doing it now.

To reproduce, go to e.g. <https://docs.julialang.org/en/v1/base/base/#Getting-Around> and try to collapse and then expand the docstring for `Base.exit`. Try to figure out where you can click and where you can't.

It turns out clicking on the text does not work -- for "Base.exit" itself, I understand it (it's a link anchor), but I find it surprising that clicking on "— Function" also has no effect.

(Personally I also find the animation sluggish and distracting but that's probably just me, I prefer no or minimal animations for such things).

Looking at the implementation, it seems to rely on JavaScript, which may explain some other issues we have. E.g. on https://docs.oscar-system.org/stable/ we have all docstrings *collapsed* by default. This interacts badly with search: try searching e.g. for `is_quasisimple`. This leads to [this link](https://docs.oscar-system.org/stable/Groups/group_characters/#is_quasisimple-Tuple%7BOscar.GAPGroupCharacterTable%7D) -- at least for me, if I follow that link the scroll position ends up at the completely wrong spot.

---

Possible solution: have you considered using [`<details>` + `summary`](https://www.w3schools.com/tags/tag_details.asp) tags instead? That way one could eliminate the JavaScript (animations can still be done via CSS), and possibly solve both issues at once.

If there is general interest in trying such an approach out, I could ask @aaruni96 to look into doing experimentally (at least the "easy" part of changing the HTML generating code. CSS styling is another matter... but one step after the other).

aaruni96:
Fixes #2772

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,11 +34,15 @@ jobs:
           - macos-latest
           - windows-latest
         arch:
-          - x64
+          - default
         include:
           - os: ubuntu-latest
             version: '1'
             arch: x86
+        exclude:
+          - os: macos-latest
+            version: '1.6'
+            arch: default
     steps:
       - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2
@@ -229,7 +233,6 @@ jobs:
       - uses: julia-actions/setup-julia@v2
         with:
           version: 1
-          arch: x64
           show-versioninfo: true
       - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-buildpkg@v1
--- a/docs/src/lib/internals/search.md
+++ b/docs/src/lib/internals/search.md
@@ -1,0 +1,206 @@
+# Search 
+
+## Overview
+
+The search system provides full-text search functionality for documentation sites through a two-phase architecture. 
+During build time, a Julia-based indexer processes all documentation content and generates a searchable index. 
+At runtime, a JavaScript client-side interface performs real-time search operations against this pre-built index using a Web Worker for performance optimization.
+
+## Architecture
+
+The search implementation consists of three primary components operating in sequence:
+
+1. **Build-time Index Generation** - Julia code in `src/html/HTMLWriter.jl` processes documentation content during site generation.
+2. **Client-side Search Interface** - JavaScript code in `assets/html/js/search.js` handles user interactions and search execution. 
+3. **Web Worker Processing** - Background thread execution prevents UI blocking during search operations.
+
+## Index Generation Process
+
+### 1. SearchRecord Structure
+
+The core data structure is the `SearchRecord` struct defined in `src/html/HTMLWriter.jl`:
+
+```julia
+struct SearchRecord
+    src::String          # URL/path to the document
+    page::Documenter.Page # Reference to the page object
+    fragment::String     # URL fragment (for anchored content)
+    category::String     # Content category (page, section, docstring, etc.)
+    title::String        # Display title for search results
+    page_title::String   # Title of the containing page
+    text::String         # Searchable text content
+end
+```
+
+### 2. Index Generation Pipeline
+
+The indexer processes documentation content through a multi-stage pipeline during HTML generation:
+
+1. **AST Traversal** - The system walks each page's markdown abstract syntax tree structure at `src/html/HTMLWriter.jl` in the function `function domify(dctx::DCtx)`
+2. **Record Instantiation** - Each content node generates a `SearchRecord` via the `searchrecord()` function at `src/html/HTMLWriter.jl`
+3. **Content Classification** - The categorization system assigns content types
+4. **Text Normalization** - The `mdflatten()` function extracts plain text from markdown structures for indexing.
+5. **Deduplication Pass** - Records sharing identical locations undergo merging to optimize index size.
+6. **JavaScript Serialization** - The processed index outputs as JavaScript object notation for client consumption.
+
+### 3. Index Output
+
+The search index is written to `search_index.js` in the following format:
+
+```javascript
+var documenterSearchIndex = {"docs": [
+  {
+    "location": "page.html#fragment",
+    "page": "Page Title", 
+    "title": "Content Title",
+    "category": "section",
+    "text": "Searchable content text..."
+  }
+  // ... more records
+]}
+```
+
+### 4. Content Filtering
+
+The indexer excludes specific node types from search index generation (`src/html/HTMLWriter.jl`):
+- `MetaNode` - Metadata annotation blocks containing non-searchable directives
+- `DocsNodesBlock` - Internal documentation node structures  
+- `SetupNode` - Configuration and setup directive blocks
+
+## Client-Side Search Implementation
+
+### 1. Search Architecture
+
+The client-side implementation employs a multi-threaded Web Worker architecture for computational isolation:
+
+- **Main Thread** - Manages user interface event handling, result filtering, and DOM manipulation operations
+- **Web Worker Thread** - Executes search algorithms using the [MiniSearch library](https://lucaong.github.io/minisearch/) without blocking the user interface
+
+### 2. MiniSearch Configuration
+
+The search system uses MiniSearch with the following configuration (`assets/html/js/search.js`):
+
+```javascript
+let index = new MiniSearch({
+  fields: ["title", "text"],           // Fields to index
+  storeFields: ["location", "title", "text", "category", "page"], // Fields to return
+  processTerm: (term) => {
+    // Custom term processing with stop words removal
+    // Preserves Julia-specific symbols (@, !)
+  },
+  tokenize: (string) => string.split(/[\s\-\.]+/), // Custom tokenizer
+  searchOptions: {
+    prefix: true,       // Enable prefix matching
+    boost: { title: 100 }, // Boost title matches
+    fuzzy: 2           // Enable fuzzy matching
+  }
+});
+```
+
+### 3. Stop Words
+
+The search engine implements a stop words filter (`assets/html/js/search.js`) derived from the Lunr 2.1.3 library, with Julia-language-specific modifications that preserve semantically important Julia symbols and keywords from filtration.
+
+### 4. Search Workflow
+
+#### Main Thread Execution Flow:
+1. **Input Event Processing** - User keystrokes in search input trigger `input` event listeners
+2. **Worker Thread Communication** - Available worker threads receive search requests via `postMessage` API
+3. **Result Set Processing** - Worker thread responses undergo filtering and DOM rendering
+4. **Browser State Management** - Search queries and active filters update browser URL parameters
+
+#### Web Worker Execution Flow:
+1. **Query Reception** - Main thread search requests arrive through message passing interface
+2. **Search Algorithm Execution** - MiniSearch performs full-text search with minimum score threshold of 1
+3. **Result Set Generation** - Search matches generate HTML markup limited to 200 results per content category
+4. **Response Transmission** - Formatted search results return to main thread via message passing
+
+### 5. Result Rendering
+
+The search result rendering system generates structured output elements (`assets/html/js/search.js`):
+- **Title Component** - Content titles with syntax highlighting and category classification badges
+- **Text Snippet Component** - Extracted text excerpts with search term highlighting via HTML markup
+- **Navigation Link Component** - Direct URL references to specific content locations within documentation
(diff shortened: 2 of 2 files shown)


src/html/HTMLWriter.jl
```diff
@@ -1854,14 +1854,13 @@ function domify(dctx::DCtx, mdast_node::Node, docsnode::Documenter.DocsNode)
     )
     push!(ctx.search_index, rec)
 
-    return article[".docstring"](
-        header(
-            a[".docstring-article-toggle-button.fa-solid.fa-chevron-down", :href => "javascript:;", :title => "Collapse docstring"],
-            a[".docstring-binding", :id => docsnode.anchor.id, :href => "#$(docsnode.anchor.id)"](code("$(docsnode.object.binding)")),
```
fingolfin, line 1907: (First off: thanks for working on this!)

I think the `href` was here so that people could click on the name and the URL changes to include the href, thus making it possible to get a bookmarkable URL pointing directly at a specific docstring. Of course the downside is that clicking the text then does not expand/collapse the docstring; but perhaps that's not too bad?

Anyway, that's fine tuning, I fully understand this is a first prototype, and I do not mean to interfere with that, just point something out for consideration before a potential merge.

aaruni96: There were two `hrefs` in the original code, one for the anchor (which is preserved), and another which did weird and wonderful javascript things to make the `<article>` behave like an accordion. This is the one which I get rid of.

aaruni96:
Current behavior:

All collapsible sections are collapsed by default. They toggle their state between expand/collapse if you click anywhere on the header, except if you click on the link, which expands the section and scrolls it to the top.

If you share a link with an anchor pointing to a specific section, that section is expanded and scrolled to the top.


I need help with both the javascript and the stylesheets.

The javascript I've added works, but I need advise on whether it should be the only thing in that file, and if to rename the file, and whether to edit some other thing if the file is renamed.

The CSS I've come up with more or less mimics the current style, but I can't figure out how to correctly include it in Documenter. For my tests, I have been manually modifying the CSS file generated by Documenter.

Possible todo: adding animation to the expand/collapse.

Rahban1:
> The CSS I've come up with more or less mimics the current style, but I can't figure out how to correctly include it in Documenter

you can just run `make themes` on your terminal and the css will be included in Documenter

fingolfin:
@Rahban1 I think that's not the core of @aaruni96's problem; it is where and how to properly put these changes into the various SCSS files?

@aaruni96 make sure to have a look at <assets/html/scss/documenter/components/_docstring.scss>

aaruni96:
Thanks for the help. New question:

Should I `git add` the resulting files after running `make themes`  ?

aaruni96:
Got rid of merge conflicts v/s the new master.

Rahban1:
> Thanks for the help. New question:
> 
> 
> 
> Should I `git add` the resulting files after running `make themes`  ?

Yes

assets/html/js/metadata.js
```diff
@@ -1,6 +1,6 @@
 // libraries: jquery
 // arguments: $
-
+/*
 $(document).ready(function () {
   let meta = $("div[data-docstringscollapsed]").data();
```
fingolfin: My understanding based on what you just explained to me off-band is that this function dealt with scrolling the view port (which was meant to fix certain issues but never really worked well), and which is now unnecessary with your changes as the browser just does the right thing.

So I would suggest we just delete it!

goerz:
In terms of how this looks visually, the animation speed, and the click targets for collapsing/uncollapsing, I think this is perfect.

I appreciate that a direct link to a docstring will always expand it. That seems like very desirable behavior.

But I think this will have to support the `CollapsedDocStrings` settings in the meta-block to allow to choose whether docstrings on a particular page are collapsed or expanded by default. I definitely write documentation that depends on particular docstrings being expanded by default.

Also, the global collapse/expand button in top right corner of the page will have to be made to work again.

I would consider breaking either one of those behaviors a semver-breaking change, so I think we'll have to find a way to get that to work again, with appropriate JS.

Other than that, this looks great!

fingolfin:
I also just tried this on the OSCAR docs, and it is soooo much better, thank you @aaruni96 ! But of course @goerz also raises two relevant issues that need to be addressed. But I hope this won't too hard:
1. honoring `CollapsedDocStrings`: assuming this setting is known for each docstring at the time we generate the HTML, this could be handled by adding an `open="false"` or `'open="true"` attribute to the `<details>` element.
2. adding back some JavaScript to open or close all is also no major issue, see e.g. <https://stackoverflow.com/a/59742798> (and I am sure AI tools will also readily produce something).

assets/html/js/metadata.js
fingolfin: Aha this code is what drives the `documenter-article-toggle-button` which @goers mentioned. So this actually should be reactivated and modified suitably

assets/html/js/collapse.js
```diff
@@ -1,78 +1,14 @@
...
-
-      parent
-        .children(".docstring-article-toggle-button")
-        .prop("title", articleToggleTitle);
-      parent.siblings("section").slideToggle();
-    });
-  },
-);
-
-$(document).on("click", ".docs-article-toggle-button", function (event) {
```
fingolfin, line 37: This is the other part driving the button which collapses / expands all sections. So this would need to be retained in modified form to expand/collapse all

src/html/HTMLWriter.jl
```diff
@@ -1854,13 +1854,13 @@ function domify(dctx::DCtx, mdast_node::Node, docsnode::Documenter.DocsNode)
     )
     push!(ctx.search_index, rec)
 
-    return article[".docstring"](
-        header(
-            a[".docstring-article-toggle-button.fa-solid.fa-chevron-down", :href => "javascript:;", :title => "Collapse docstring"],
```
mortenpi, line 1906: Visually, it would be nice if we could retain the `fs-chevron-*` symbols for the triangle. Not sure if that's easy to do with CSS or not.

src/html/HTMLWriter.jl
```diff
@@ -1854,13 +1854,13 @@ function domify(dctx::DCtx, mdast_node::Node, docsnode::Documenter.DocsNode)
     )
     push!(ctx.search_index, rec)
 
-    return article[".docstring"](
```
mortenpi, line 1904: I wouldn't block on this, but once small concern I have is that `<details>` is probably not the right element here semantically. Docstrings have been wrapped in `<article>`-s, since they're generally important/main content, as opposed to a "widget from which the user can obtain additional information or controls".

I don't think we can really have a non-JS `<details>`-like behavior, but maybe we should at least retain an `<article>` element around the `<details>`?

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+* Changed the docstring block accordions from a custom implementation to HTML details+summary tag. ([#2773])
```
```suggestion
* Changed the docstring block accordions from a custom implementation to HTML details+summary tag. ([#2772], [#2773])
```

mortenpi (changes requested):
Agreed with @goerz's comments, plus a couple of small things from me. But generally I think this is great and the approach seems good to me.

I also think that getting rid of the animation is an improvement, since it makes the UX feel more snappy.

src/html/HTMLWriter.jl
fingolfin: What purpose would the `<article>` serve? As a hint to screen readers? Wrapping the `<details>` in an `<article>` should be fine I guess, implementation-wise?

src/html/HTMLWriter.jl
```diff
@@ -1854,16 +1854,16 @@ function domify(dctx::DCtx, mdast_node::Node, docsnode::Documenter.DocsNode)
...
-        header(
-            a[".docstring-article-toggle-button.fa-solid.fa-chevron-down", :href => "javascript:;", :title => "Collapse docstring"],
-            a[".docstring-binding", :id => docsnode.anchor.id, :href => "#$(docsnode.anchor.id)"](code("$(docsnode.object.binding)")),
+    return article(details[".docstring"](
```
fingolfin: Needs runic reformatting.

assets/html/js/metadata.js
```diff
@@ -3,25 +3,10 @@
 
 $(document).ready(function () {
   let meta = $("div[data-docstringscollapsed]").data();
-
-  if (meta?.docstringscollapsed) {
+  console.log(meta);
```
fingolfin:
```suggestion
```

assets/html/js/metadata.js
```diff
@@ -3,25 +3,10 @@
...
+  if (!meta.docstringscollapsed) {
```
fingolfin: Could you please explain why `meta?.docstringscollapsed` changed to `!meta.docstringscollapsed` ? It's not at all clear to me :-(

src/html/HTMLWriter.jl
```diff
@@ -1901,16 +1901,28 @@ function domify(dctx::DCtx, mdast_node::Node, docsnode::Documenter.DocsNode)
...
+            domify_doc(dctx, mdast_node)
+        )
     )
+    if get(getpage(ctx, navnode).globals.meta, :CollapsedDocStrings, false)
+        # if DocStringsCollapse = true in `@meta`
+        # shouldn't the check then be false?
+        # see line 1530 for original comment?
+        # collapse everything!
+        # do nothing, that is
+    else
```
fingolfin:
```suggestion
    if !get(getpage(ctx, navnode).globals.meta, :CollapsedDocStrings, false)
        # if CollapsedDocStrings = true in `@meta`, do nothing, as the default
        # state is collapsed.
        # But if it is false, then set the `open` attribute to expand the
        # details+summary blocks.
    else
```

assets/html/js/metadata.js
aaruni96: I needed to add the negation because in the old way, everything was expanded by default, and would collapse if this property was true, and in the new way, everything is collapsed by default, and will be expanded if the property is set to false (or does not exist).

The `?.` went away as an accident and needs to be added back in.

src/html/HTMLWriter.jl
aaruni96: I don't actually understand what this `get()` does, but, in the behavior before my changes, we have (line 1570 after the rebase)

```julia
    if get(getpage(ctx, navnode).globals.meta, :CollapsedDocStrings, false)
        # if DocStringsCollapse = true in `@meta`, we let JavaScript click the
        # collapse button after that page has loaded.
        push!(meta_divs, div[Symbol("data-docstringscollapsed") => "true"]())
    end
```

So, if the `get()` thing is called with `false`, then somehow, that corresponds to `DocStringsCollapse` being set to true?

So, we really need to do nothing in the `if get(getpage(ctx, navnode).globals.meta, :CollapsedDocStrings, false)` clause, and need to expand all docstrings in the `else` case.

fingolfin: The `get` method performs a dictionary lookup: the first argument `BLA.meta` is the dictionary, we look for the key `:CollapsedDocStrings`; if the key is absent, the default value `false` is substituted.

We then check the return value of `get`: if it returns `true`, then someone explicitly set `CollapsedDocStrings` to `true` and we should collapse things. If it returns `false`, then either it was set explicitly, or this is the default; in any case, we should explicitly expand things.

fingolfin: We talked about it and part of the confusion was caused by me forgetting to delete the final `else` in my proposed code change 🤦 after we figured that out all became clear

fingolfin (approved):
Lovely. I just tried it with the Documenter documentation and as far as I could tell, everything works perfectly. Great job, @aaruni96 !

fingolfin:
@goerz @mortenpi as far as I can tell all your concerns were addressed, but I'll wait a day or two in case  you want to complain (or let me know if you need more time).

goerz:
Looks fantastic!

PR #2777, opened 2025-10-21
Set version to 1.15.0

mortenpi:
## Pre-release

 - [x] Change the version number in `Project.toml`
   * If the release is breaking, increment MAJOR
   * If the release adds a new user-visible feature, increment MINOR
   * Otherwise (bug-fixes, documentation improvements), increment PATCH
 - [x] Update `CHANGELOG.md`, following the existing style (in particular, make sure that the change log for this version has the correct version number and date).
 - [x] Run `make changelog`, to make sure that all the issue references in `CHANGELOG.md` are up to date.
 - [x] Check that the commit messages in this PR do not contain `[ci skip]`
 - [x] Check that the [regression-tests workflow](https://github.com/JuliaDocs/Documenter.jl/actions/workflows/regression-tests.yml) did not reveal any changes that broke extensions (this should run as part of the CI suite, but can also be triggered manually).

## The release

 - [x] After merging the pull request, tag the release. There are two options for this:

   1. [Comment `[at]JuliaRegistrator register` on the GitHub commit.](https://github.com/JuliaRegistries/Registrator.jl#via-the-github-app)
   2. Use [JuliaHub's package registration feature](https://help.juliahub.com/juliahub/stable/contribute/#registrator) to trigger the registration.

   Either of those should automatically publish a new version to the Julia registry.
 - Once registered, the `TagBot.yml` workflow should create a tag, and rebuild the documentation for this tag.
 - These steps can take quite a bit of time (1 hour or more), so don't be surprised if the new documentation takes a while to appear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,15 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Version [v1.15.0] - 2025-10-21
-
-### Added 
-
-* In the HTML output, the search modal can now be navigated using using up and down keys. ([#2761])
+## Version [v1.15.0] - 2025-10-22
 
 ### Changed
 
 * The HTML output shows a warning now when the search index size is too big (500Kib). ([#2423], [#2753])
+* In the HTML output, the search modal can now be navigated using using up and down keys. ([#2761])
 
 ### Other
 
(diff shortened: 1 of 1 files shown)


lgoettgens:
Regarding the failure of the Oscar build test: It is not possible to see the git rev of Oscar that was used in CI, to try to reproduce it, right? Or am I just blind?

mortenpi:
I think it should be pulling just the latest `master`/`main`: https://github.com/JuliaDocs/Documenter.jl/blob/d01718d8300a60975cc8fb66c22e2973b8e9283d/.github/workflows/regression-tests.yml#L85-L86

lgoettgens:
> I think it should be pulling just the latest `master`/`main`:
> 
> https://github.com/JuliaDocs/Documenter.jl/blob/d01718d8300a60975cc8fb66c22e2973b8e9283d/.github/workflows/regression-tests.yml#L85-L86

Yeah, I've seen that. But that does not map to a git rev trivially. Maybe we could adapt the CI script to print the git rev to make such debugging easier?

lgoettgens:
The issue observed in CI is not due to Documenter, but due to a julia 1.12 change. https://github.com/oscar-system/Oscar.jl/pull/5484 will take care of this.

mortenpi:
> Maybe we could adapt the CI script to print the git rev to make such debugging easier?

I think that's a good idea.

mortenpi:
@lgoettgens By the way, I'm happy to release this with the failing regression CI check, since we know the issue (and that it's not directly due to Documenter). So no need to jump through any hoops in Oscar.

mortenpi:
https://github.com/JuliaRegistries/General/pull/140978

PR #2781, opened 2025-10-25
Make missingbindings and allbindings deterministic

IanButterworth:
See https://github.com/JuliaLang/julia/pull/55267

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,11 +34,15 @@ jobs:
           - macos-latest
           - windows-latest
         arch:
-          - x64
+          - default
         include:
           - os: ubuntu-latest
             version: '1'
             arch: x86
+        exclude:
+          - os: macos-latest
+            version: '1.6'
+            arch: default
     steps:
       - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2
@@ -129,7 +133,7 @@ jobs:
           path-to-lcov: lcov.info
           flag-name: latex-${{ join(matrix.*, '-') }}
           parallel: true
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v5
         with:
           name: "PDFs-${{ matrix.os }}"
           path: test/examples/builds/*/*.pdf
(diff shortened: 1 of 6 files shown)


mortenpi:
I'm a bit confused by the changes / fix.

1. We now pass the `checkdocs_ignored_modules` when we gather the bindings.. which is probably fine, but it should be redundant, since `doc.blueprint.modules` should already be filtered?
2. I don't really understand why the sorting of keys would make a difference?

IanButterworth:
1. Julia base docs only partially cover Pkg. Pkg has its own docs. So we need to pass Pkg for docs but exclude from the docs check

2. I don't either really. That part was Claude thinking random ordering was the reason for the random results

src/docchecks.jl
```diff
@@ -69,21 +70,39 @@ function missingbindings(doc::Document)
...
+                current == ignored && return true
+                parent = parentmodule(current)
+                # Guard against infinite loops if parentmodule returns itself
+                parent == current && break
+                current = parent
+            end
+            return false
+        end
+            continue
+        end
```
fingolfin: Something is off here.

IanButterworth: The `any` `do` above is confusing. I guess double indenting it would be clearer?

fingolfin: Or doing
```julia
res = any(ignored_modules) do ignored
   ...
end
res && continue
```

Or the `do` block could be moved into a local helper function.

IanButterworth:
```suggestion
        end && continue
```

src/docchecks.jl
```diff
@@ -69,21 +70,37 @@ function missingbindings(doc::Document)
...
+    metadata = collect(meta(mod))
+    sort!(metadata; by = entry -> string(entry[1]))
+    for (binding, doc) in metadata
         # The keys of the docs meta dictionary should always be Docs.Binding objects in
         # practice. However, the key type is Any, so it is theoretically possible that
         # some non-binding metadata gets added to the dict. So on the off-chance that has
         # happened, we simply ignore those entries.
         isa(binding, Docs.Binding) || continue
+        # Skip bindings from ignored modules or their submodules
+        if any(ignored_modules) do ignored
```
IanButterworth:
```suggestion
        any(ignored_modules) do ignored
```

src/docchecks.jl
```diff
@@ -69,21 +70,37 @@ function missingbindings(doc::Document)
...
+            current = binding.mod
+            while current != Main && current != Base
+                current == ignored && return true
+                parent = parentmodule(current)
+                # Guard against infinite loops if parentmodule returns itself
+                parent == current && break
+                current = parent
+            end
+            return false
+        end && continue
```
fingolfin, line 103: Yes that's nicer, thanks

mortenpi:
> 1. Julia base docs only partially cover Pkg. Pkg has its own docs. So we need to pass Pkg for docs but exclude from the docs check

But they should already be excluded by here, no?

https://github.com/JuliaDocs/Documenter.jl/blob/f887f34d5d8d80503e6bcf90918d0e30d7fe9616/src/documents.jl#L521-L524

IanButterworth:
@mortenpi I don't know. It might be worth investigating this with https://github.com/JuliaLang/julia/pull/55267 without my guesses. This PR was necessary for me to get Pkg excluded

IanButterworth:
To be clearer, I don't know if these are the correct fixes, and I'm asking for help from devs more familiar with Documenter

IanButterworth:
Claude figured this out

PR #2782, opened 2025-10-27
CI: don't test on macOS-x64

fingolfin:
The macOS runners all use ARM.

When merging this PR, the list of "required" CI checks will
have to be adjusted.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -133,7 +133,7 @@ jobs:
           path-to-lcov: lcov.info
           flag-name: latex-${{ join(matrix.*, '-') }}
           parallel: true
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v5
         with:
           name: "PDFs-${{ matrix.os }}"
           path: test/examples/builds/*/*.pdf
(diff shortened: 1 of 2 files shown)


fingolfin:
Perhaps we should also test against the Julia LTS release?

.github/workflows/CI.yml
```diff
@@ -34,11 +34,15 @@ jobs:
           - macos-latest
           - windows-latest
         arch:
-          - x64
+          - default
         include:
           - os: ubuntu-latest
             version: '1'
             arch: x86
```
giordano, line 41: The `arch` key is now unused, so this job is actually duplicating the other `ubuntu-latest` with version `1`.

fingolfin: That doesn't seem to match what the CI logs here report, one Julia identifies as 32bit the other as 64bit, as expected.

giordano: Aaaah, the deleted line 232 is from another job

fingolfin: Ahh indeed now I understand your point, and where the confusion comes from (I thought you meant that `setup-julia` would generally ignore `arch` now, but I see now that of course you referred to its usage in the workflow).

fingolfin:
This requires action by @mortenpi or someone else with the right access levels to adjust the branch protection rules for the changed names of the CI jobs.

mortenpi:
I updated the branch protection rules. LGTM!

Issue #2668, opened 2025-03-28
`"Header with slug '$slug' is not unique"` error should always be thrown


lgoettgens:
Please see the mwe example code in the end of this issue.

We ran into this issue in https://github.com/oscar-system/Oscar.jl/pull/4763, and it needed quite a bit of debugging the Documenter internals to reduce the codebase to the mwe and find the issue.

This is about where the reference `[Some header](@ref)` in `index.md` points to. I would expect to to point to either of the same-named headers on `other_page_1.md` or `other_page_2.md`, but instead it points to the single docstring for a `-` method.

### Explanation:
The reference gets tried to get resolved in three steps: `Header`, `Issue`, `Docs`.
In the `Header` step, the `Selectors.matcher` call returns `true`, but the `Selectors.runner` call does not resolve the reference due to two headers matching, and adds an appropriate error message to the `errors` variable (x-ref https://github.com/JuliaDocs/Documenter.jl/blob/8dc39054746fe702d209506223a501ac0cbaf6c2/src/cross_references.jl#L372
This error is however never shown to the user, not even in debug mode.

In the `Issue` step, nothing happens.

In the `Docs` step, `find_docref` does its best to parse the slug `Some-header` as a method signature. And this results in `-(::MyStruct)` due to some weird eval logic.

### Possible solutions I can think of:
1. If the `Header` step fails due to a non-unique anchor, immediately print the above error message to the user.
2. Somehow change `find_docref` to avoid parsing all types of slugs as methods of `-`. In this case, we would run through all resolving steps unsuccessfully, thus printing an error in https://github.com/JuliaDocs/Documenter.jl/blob/8dc39054746fe702d209506223a501ac0cbaf6c2/src/cross_references.jl#L311

````julia
$ find ./ -type f | xargs tail -n +1

==> ./docs/make.jl <==
using Documenter, MWE

makedocs(;
  sitename="MWE",
  modules=[MWE],
  pages=[
    "Home" => "index.md",
    "Other page 1" => "other_page_1.md",
    "Other page 2" => "other_page_2.md",
  ],
  remotes=nothing,
)

==> ./docs/src/other_page_2.md <==
# Some header

==> ./docs/src/other_page_1.md <==
# Some header

==> ./docs/src/index.md <==
# MWE

```@docs
-(::MyStruct)
```

We want to link to [Some header](@ref).

==> ./Project.toml <==
name = "MWE"
[...]

PR #2787, opened 2025-10-27
Set `Selectors.strict(::Type{XRefResolvers.Header}) = true`

lgoettgens:
Resolves https://github.com/JuliaDocs/Documenter.jl/issues/2668 by the first approach outlined in the issue.

The mwe from the issue now fails with
```julia
[...]
┌ Error: Cannot resolve @ref for md"[Some header](@ref)" in docs/src/index.md.
│ - Header with slug 'Some-header' is not unique in docs/src/index.md.
└ @ Documenter ~/code/julia/Documenter.jl/src/utilities/utilities.jl:47
```

I assume that this will make Documenter fail on some packages as we now actually show the error, instead of falling through and choosing a useless xref to a docstring of `Base.:-`. Thus, I would rather consider this a bugfix instead of a breaking change.

ping @fingolfin @mortenpi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Changed the header crossref step to eagerly fail when encountering a non-unique header slug. This may cause some documentation builds to fail, that previously passed but generated incorrect references. This can be fixed by having distinct headers in your markdown pages, or by not referencing headers directly but rather their `@id`. ([#2668], [#2787])
+* Changed the header crossref step to eagerly fail when encountering a non-unique header slug. ([#2668], [#2787])
+
+  This is **potentially breaking** and may cause some documentation builds to fail.
+  Those previously passed but generated incorrect cross-references.
+  You can fix this by ensuring that you have distinct headers across your markdown pages, or by using the `@id` syntax to give the headers unique slugs.
 
 ## Version [v1.15.0] - 2025-10-22
 


goerz:
> I would rather consider this a bugfix instead of a breaking change.

I agree

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+* Changed the header crossref step to eagerly fail when encountering a non-unique header slug. ([#2668], [#2787])
```
fingolfin, line 14: Maybe we should be helpful and say something like this:
```suggestion
* Changed the header crossref step to eagerly fail when encountering a non-unique header slug. This may cause some documentation build to fail that previously passed but generated incorrect references. This can then be fixed by making the header slugs unique. ([#2668], [#2787])
```

lgoettgens:
```suggestion
* Changed the header crossref step to eagerly fail when encountering a non-unique header slug. This may cause some documentation builds to fail, that previously passed but generated incorrect references. This can be fixed by having distinct headers in your markdown pages, or by not referencing headers directly but rather their `@id`.([#2668], [#2787])
```
Good point. I changed your suggestion slightly

mortenpi:
Making this an error seems reasonable to be. We'll have this in a minor release (as opposed to a patch), so I think it's okay if this is slightly breaking.

mortenpi (approved):
I took the liberty of editing the CHANGELOG entry a bit.

Issue #1660, opened 2021-07-25
Examples should be updated


hyrodium:
We have examples [here](https://juliadocs.github.io/Documenter.jl/stable/man/examples/#Registered), but some of the links seem not good for beginner.

* The `stable` link is not deployed correctly (e.g. [ControlSystems.jl](https://juliacontrol.github.io/ControlSystems.jl/stable/))
* The site is no longer generated by Documenter.jl (e.g. [Bio.jl](https://biojulia.net/Bio.jl/stable/), [MergedMethods.jl](https://michaelhatherly.github.io/MergedMethods.jl/stable/))
* That package doesn't seem to be actively being developed. (e.g. [BanditOpt.jl](https://v-i-s-h.github.io/BanditOpt.jl/stable/))
* The site is not generated with latest Documenter.jl (e.g. [EvolvingGraphs.jl](https://etymoio.github.io/EvolvingGraphs.jl/stable/))
* The package is not registered (e.g. [NLOptControl.jl](https://huckl3b3rry87.github.io/MPCDocs.jl/stable/))

Should we remove these links? If so, I'll fix them in #1659.

fredrikekre:
Now that almost all packages use Documenter I feel like the list should either be removed or be curated to only include links to packages that does something extra (custom CSS, examples with Literate, etc).

hyrodium:
Is there any good place to find example documentation?
I thought [JuliaHub](https://juliahub.com/ui/Packages) would be a good place to find packages and documentations, but the most of documentations are deployed on juliahub, and there's no "Edit on GitHub" button.

mortenpi:
> Now that almost all packages use Documenter I feel like the list should either be removed or be curated to only include links to packages that does something extra (custom CSS, examples with Literate, etc).

I agree with this. We could perhaps point to one or two well-maintained packages that do a good job with the docs where people could go and check the basic setup (in addition to the links to packages that are doing something special).

mortenpi:
That said, in the meanwhile, we can also just make sure that the links we have there are up to date.

PR #2788, opened 2025-10-27
Update examples list

fingolfin:
Remove an outdated and now incorrect paragraph.

Remove packages that haven't been updated in the last couple years.

See also #1660

### DIFF
--- a/docs/src/man/examples.md
+++ b/docs/src/man/examples.md
@@ -6,61 +6,38 @@ examples of what others have already built with it.
 The following packages use Documenter to build their documentation and so
 should give a good overview of what this package is currently able to do.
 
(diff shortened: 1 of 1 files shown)


mortenpi:
LGTM, thanks! Just one note: in general, I think it would be better if we'd have a curated list of a few packages, rather than this long list.

fingolfin:
@mortenpi indeed, and we agreed on that before in discussion at PR #2690, and this is also what we say on issue #1660, which hence stays open.

But #1660 has been open since 2021... figuring out what the "right" set of package is seems non-trivial... so I figured we should at least remove the worst offenders from the list, and the outdated paragraph... :-)

PR #2789, opened 2025-10-27
regression-tests.yml: use julia-actions/cache

fingolfin:


fingolfin:
@thofma @lgoettgens any idea why the `docs-build-Oscar` step might fail in the Hecke tests?
````
┌ Error: failed to run `@setup` block in src/Hecke/tutorials/quaternion.md
│ ```@setup bla
│ using Hecke
│ nothing
│ ```
│   exception =
│    LoadError: ArgumentError: Package Hecke not found in current path, maybe you meant `import/using ..Hecke`.
│    - Otherwise, run `import Pkg; Pkg.add("Hecke")` to install the Hecke package.
│    Stacktrace:
│      [1] macro expansion
````

thofma:
Not sure. What am I looking at?

fingolfin:
So here in Documenter.jl, as part of the test suite we try to build the manuals of various Julia packages that are known to show case some feature / are big / for some other reason are considered helpful for noticing regressions early.

OSCAR is on the list 😉 

Last night this test started to fail, even on this PR, which does not change anything in Documenter other than some inconsequential bits. See [this log](https://github.com/JuliaDocs/Documenter.jl/actions/runs/18857240920/job/53807992676?pr=2789)

The GitHub Workflow driving that is set up in <https://github.com/JuliaDocs/Documenter.jl/blob/master/.github/workflows/regression-tests.yml> so there you can see how precisely it builds the documentation.

Anyway, I am guessing something changed in OSCAR / Hecke / Nemo / AA. Indeed, I see there was a Nemo release shortly before the CI failure. So I wonder if there might be a connection

lgoettgens:
> Last night this test started to fail, even on this PR, which does not change anything in Documenter other than some inconsequential bits. See [this log](https://github.com/JuliaDocs/Documenter.jl/actions/runs/18857240920/job/53807992676?pr=2789)

Can you point me to a successful run before that? Because I see the same failure in https://github.com/JuliaDocs/Documenter.jl/actions/runs/18773296931/job/53562083506#step:5:906 from 4 days ago, and before that there were other failures due to the `Meta.parse` changes in julia.
The latest run that succeeded was https://github.com/JuliaDocs/Documenter.jl/actions/runs/17545774048/job/49826803043, which was more than a month ago. My best guess thus is that one of the following PRs in Oscar broke this: https://github.com/oscar-system/Oscar.jl/pull/5381, https://github.com/oscar-system/Oscar.jl/pull/5335

thofma:
I also don't remember any changes in the documentation setup. I also don't understand the CI script. Doesn't Oscar need some fancy setup to make the documentation build work?

fingolfin:
Ahhh, now I see my mistake: I hadn't realized that the regression test only runs on PRs that modify `Project.toml` or `.github/workflows/regression-tests.yml` and since I didn't see it report an error on other PRs, I assumed it passed -- but instead it simply didn't run there at all. 

So, there is a problem, but it's not new. I'll look into it later then. My apologies for having bothered you guys!

PR #2793, opened 2025-10-29
Report file names in which "Unexpected Julia interpolation" occurred

### LINKED ISSUE #2752: `Unexpected Julia interpolation in the Markdown` could give some location info
The warning prints as:
```julia
┌ Warning: Unexpected Julia interpolation in the Markdown. This probably means that you
│ have an unbalanced or un-escaped $ in the text.
│ 
│ To write the dollar sign, escape it with `\$`
│ 
│ We don't have the file or line number available, but we got given the value:
│ 
│ `$` which is of type `Symbol`
└ @ Documenter.HTMLWriter ~/.julia/packages/Documenter/eoWm2/src/html/HTMLWriter.jl:2350
```

In our current situation at Oscar.jl, I had to bisect multiple weeks of activity on master to find the commit which introduced this (due to https://github.com/JuliaDocs/Documenter.jl/issues/2751). But even then, this only pointed to a PR of a few thousand lines of changes...

It would be **incredibly** helpful, if the warning could give any pointers to the offending code. IMO any of the following would already help:
- name of the html file currently being written
- name of the md file currently being converted to html
- a line number in any of the above two

aaruni96:
Fixes #2752

Depends on #2792

### DIFF
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -852,15 +852,15 @@ latex(io::Context, node::Node, ::Documenter.SetupNode) = nothing
 function latex(io::Context, node::Node, value::MarkdownAST.JuliaValue)
     message =
     """
-        Unexpected Julia interpolation in the Markdown. This probably means that you have an
-        unbalanced or un-escaped \$ in the text.
+    Unexpected Julia interpolation in the Markdown. This probably means that you have an
+    unbalanced or un-escaped \$ in the text.
 
-        To write the dollar sign, escape it with `\\\$`
+    To write the dollar sign, escape it with `\\\$`
 
-        This is in file $(io.filename), and we were given the value:
+    This is in file $(io.filename), and we were given the value:
 
-        `$(value.ref)` which is of type `$(typeof(value.ref))`
-        """
+    `$(value.ref)` which is of type `$(typeof(value.ref))`
+    """
     if io.doc.user.treat_markdown_warnings_as_error
         error(message)
     else


aaruni96:
In my testing, when I introduce a deliberate error in Oscar's `experimental/FTheoryTools/docs/src/literature.md`, I get the output 

```
ERROR:     Unexpected Julia interpolation in the Markdown. This probably means that you have an
    unbalanced or un-escaped $ in the text.

    To write the dollar sign, escape it with `\$`

    Navnode is currently set to Experimental/FTheoryTools/literature.md, with parent NavNode(nothing, "F-Theory Tools", NavNode(nothing, ...)), and we
    got given the value:

    `$` which is of type `Symbol`
```

asinghvi17:
> Navnode is currently set to Experimental/FTheoryTools/literature.md,

could probably be better rephrased to

> This is on the page Experimental/FTheoryTools/literature.md,

src/html/HTMLWriter.jl
```diff
@@ -2427,19 +2427,24 @@ function domify(dctx::DCtx, node::Node, t::MarkdownAST.Table)
...
-        have an unbalanced or un-escaped \$ in the text.
+function domify(dctx::DCtx, ::Node, e::MarkdownAST.JuliaValue)
+    message = """
+        Unexpected Julia interpolation in the Markdown. This probably means that you have an
+        unbalanced or un-escaped \$ in the text.
 
         To write the dollar sign, escape it with `\\\$`
 
-        We don't have the file or line number available, but we got given the value:
+        Navnode is currently set to $(dctx.navnode.page), with parent $(dctx.navnode.parent), and we
```
fingolfin: Suggestion by @asinghvi17 :
```suggestion
        This is on the page $(dctx.navnode.page), with parent $(dctx.navnode.parent), and we
```

src/html/HTMLWriter.jl
```diff
@@ -2427,19 +2427,24 @@ function domify(dctx::DCtx, node::Node, t::MarkdownAST.Table)
...
+        got given the value:
```
fingolfin:
```suggestion
        were given the value:
```
I realize this was already there, but: "got" here is informal, in written language I think "were" is more appropriate.

src/latex/LaTeXWriter.jl
```diff
@@ -850,18 +850,21 @@ latex(io::Context, node::Node, ::Documenter.MetaNode) = _println(io, "\n")
...
-        have an unbalanced or un-escaped \$ in the text.
+    message = """
+    Unexpected Julia interpolation in the Markdown. This probably means that you have an unbalanced
+    or un-escaped \$ in the text.
 
-        To write the dollar sign, escape it with `\\\$`
+    To write the dollar sign, escape it with `\\\$`
 
-        We don't have the file or line number available, but we got given the value:
+    We have the file name $(io.filename), and we got given the value:
```
fingolfin:
```suggestion
    We have the file name $(io.filename), and we were given the value:
```

src/html/HTMLWriter.jl
aaruni96: I didn't want to write "This is on the page" because it may not really be, its just a "helpful hint" to look in that area / direction (The actual page is `experimental/FTheoryTools/docs/src/literature.md`, but the reported page is `Experimental/FTheoryTools/literature.md`. Or, it might be that building Oscar's documentation does some copying business which results in this change?)

asinghvi17: Yeah Documenter preserves the structure of whatever ends up in `src/` in the nav nodes, I believe.  So whatever the user does before invoking makedocs you have no idea of.

It may then be worth making the language a little more circumspect?

```suggestion
        This may be on the page $(dctx.navnode.page), with parent $(dctx.navnode.parent), and we
```

fingolfin: It should be possible to improve those paths to be relative to the current working directory (cwd); I've done this for other warnings/errors before. I can dig out a relevant commit/PR later

fingolfin: The changes I had in mind are in facec83c69e996ed513ed2309876564484ffdb82 resp. PR #2659 -- that would suggest
```suggestion
        Navnode is currently set to $(locrepr(dctx.navnode.page)), with parent $(dctx.navnode.parent), and we
```

In the OSCAR example @aaruni96 brought up, the reported path is `Experimental/FTheoryTools/literature.md` because of how we merge the documentation for the "experimental" sub-projects into the main manual: Documenter does not directly see the file `experimental/FTheoryTools/docs/src/literature.md,`. Rather, when building the OSCAR manual, we create a symlink `docs/src/Experimental/FTheoryTools` pointing to `experimental/FTheoryTools/docs/src` (both paths are relative to the `Oscar.jl` directory).

So with the change I propose, the user would be told the location `src/docs/Experimental/FTheoryTools/literature.md` and I think that's fine because that's actually a valid path. Of course we could also use `realpath` to get rid of the symlink, but I don't think that's necessary given how OSCAR-specific the "problem" here is (and we don't use `realpath` anywhere else were we use `locrepr`)

aaruni96: As of 8462744, using `locrepr`, I still get told `Experimental/FTheoryTools/literature.md`

```julia
ERROR:     Unexpected Julia interpolation in the Markdown. This probably means that you have an
    unbalanced or un-escaped $ in the text.

    To write the dollar sign, escape it with `\$`

    This is on the page Experimental/FTheoryTools/literature.md, with parent NavNode(nothing, "F-Theory Tools", NavNode(nothing, ...)), and we
    were given the value:

    `$` which is of type `Symbol`
```

fingolfin: @aaruni96 thanks. I am in the middle of working on some other `Documenter` stuff, perhaps if it is not too much work for you, you could modify your `function locrepr` temporarily to print some things and tell us what it prints right before that ERROR. Specifically, maybe you can insert right at the top
```
@show original_pwd[]
@show currentdir()
@show Base.pwd()
@show file
@show abspath(file)
```

aaruni96: Here's the output

```
original_pwd[] = "/home/aaruni/Programs/git/GitHub/Oscar.jl/docs"
currentdir() = "/home/aaruni/Programs/git/GitHub/Oscar.jl/docs"
Base.pwd() = "/home/aaruni/Programs/git/GitHub/Oscar.jl/docs"
file = "Experimental/FTheoryTools/literature.md"
abspath(file) = "/home/aaruni/Programs/git/GitHub/Oscar.jl/docs/Experimental/FTheoryTools/literature.md"
ERROR:     Unexpected Julia interpolation in the Markdown. This probably means that you have an
    unbalanced or un-escaped $ in the text.

    To write the dollar sign, escape it with `\$`

    This is on the page Experimental/FTheoryTools/literature.md, with parent NavNode(nothing, "F-Theory Tools", NavNode(nothing, ...)), and we
    were given the value:

    `$` which is of type `Symbol`
```

fingolfin: So for some reason the `src` part is missing. I guess in contrast to other uses of `locrepr`, this one is different in that it is in the `HTMLWriter.jl` code.

Looking at the code comments in `NavNode`, the `dctx.navnode.page` value "should be a valid key in `doc.blueprint.pages`". And looking at that struct, we see
```julia
struct Page
    source::String
    build::String
    workdir::Union{Symbol, String}
...
```
So maybe what we should really be using is `dctx.ctx.doc.blueprint.pages[dctx.navnode.page].source` or so ?

aaruni96: That still gives us only `src/Experimental/FTheoryTools/literature.md` (so, a bonus `src/`), instead of the `src/docs/Experimental/FTheoryTools/literature.md` that we want / expect.

fingolfin: `src/docs/Experimental/FTheoryTools/literature.md` does not exist. But `docs/src/Experimental/FTheoryTools/literature.md` does. 

Anyway, I suggest to wait for PR #2803 and then we can trivially handle it here.

src/html/HTMLWriter.jl
```diff
@@ -2443,7 +2443,8 @@ function domify(dctx::DCtx, ::Node, e::MarkdownAST.JuliaValue)
 
         To write the dollar sign, escape it with `\\\$`
 
-        We don't have the file or line number available, but we got given the value:
+        This is on the page $(locrepr(dctx.navnode.page)), with parent $(dctx.navnode.parent), and we
+        were given the value:
```
fingolfin: @aaruni96 PR #2803 is now merged. Can you please resolve the conflicts in `CHANGELOG.md` and then try this:
```suggestion
        This is in file $(locrepr(dctx)), and we were given the value:
```

src/latex/LaTeXWriter.jl
```diff
@@ -856,7 +856,7 @@ function latex(io::Context, node::Node, value::MarkdownAST.JuliaValue)
 
     To write the dollar sign, escape it with `\\\$`
 
-    We don't have the file or line number available, but we got given the value:
+    We have the file name $(io.filename), and we were given the value:
```
fingolfin:
```suggestion
    This is in file $(io.filename), and we were given the value:
```

src/html/HTMLWriter.jl
fingolfin: The missing `docs` before `src` is to be expected, given that Oscar's `docs/make.jl` does this:
```
  cd(joinpath(oscardir, "docs")) do
    ...
    makedocs(;
    ...
  end
```
and the file names/paths we print are relative to the current working directory.

I am not sure why Oscar does that, it would be nice if we could get rid of it, but that's for another day.

aaruni96: > and then try this:

This doesn't work, complains about wrong data types. I tried monkey patching a few things hoping its a trivial error, but I couldn't make any headway (just more and different errors).

The error I get is 

```
ERROR: FieldError: type Documenter.HTMLWriter.HTMLContext has no field `page`, available fields: `doc`, `settings`, `scripts`, `documenter_js`, `themeswap_js`, `warner_js`, `search_index`, `search_index_js`, `search_navnode`, `atexample_warnings`
Stacktrace:
  [1] getproperty
    @ ./Base_compiler.jl:54 [inlined]
  [2] locrepr(dctx::Documenter.HTMLWriter.DCtx, lines::Nothing)
    @ Documenter.HTMLWriter ~/Programs/git/GitHub/Documenter.jl/src/html/HTMLWriter.jl:715
  [3] locrepr(dctx::Documenter.HTMLWriter.DCtx)
    @ Documenter.HTMLWriter ~/Programs/git/GitHub/Documenter.jl/src/html/HTMLWriter.jl:713
  [4] domify(dctx::Documenter.HTMLWriter.DCtx, ::MarkdownAST.Node{Nothing}, e::MarkdownAST.JuliaValue)
    @ Documenter.HTMLWriter ~/Programs/git/GitHub/Documenter.jl/src/html/HTMLWriter.jl:2446
  [5] domify(dctx::Documenter.HTMLWriter.DCtx, node::MarkdownAST.Node{Nothing})
    @ Documenter.HTMLWriter ~/Programs/git/GitHub/Documenter.jl/src/html/HTMLWriter.jl:1813
  [6] #domify##2
    @ ~/Programs/git/GitHub/Documenter.jl/src/html/HTMLWriter.jl:1816 [inlined]
...
```

fingolfin: Argh. Seems my previous PR did not have enough coverage to reveal that I messed up... Try applying this patch in addition:
```patch
diff --git a/src/html/HTMLWriter.jl b/src/html/HTMLWriter.jl
index 6868a874c..2803165fd 100644
--- a/src/html/HTMLWriter.jl
+++ b/src/html/HTMLWriter.jl
@@ -710,9 +710,7 @@ struct DCtx
 end
 
 function Documenter.locrepr(dctx::DCtx, lines = nothing)
-    doc = dctx.ctx.doc
-    page = dctx.navnode.page
-    return Documenter.locrepr(dctx.ctx.doc, dctx.ctx.page, lines)
+    return Documenter.locrepr(dctx.ctx.doc, dctx.navnode.page, lines)
 end
 
 function SearchRecord(ctx::HTMLContext, navnode; fragment = "", title = nothing, category = "page", text = "")
```

aaruni96: Without trying it now, relying on my memory of attempting something like this during my "monkey patching", I expect this to not work because `dctx.navnode.page` is a string, but `locrepr` expects something of type `::Page`.

fingolfin: No, my patch is patching the locrepr method I added earlier today. It had a merge conflict which I resolved incorrectly.

aaruni96: After applying your patch, I still get the error I predicted : 

```
ERROR: FieldError: type String has no field `source`; String has no fields at all.
Stacktrace:
  [1] getproperty
    @ ./Base_compiler.jl:54 [inlined]
  [2] locrepr(doc::Documenter.Document, page::String, lines::Nothing)
    @ Documenter ~/Programs/git/GitHub/Documenter.jl/src/utilities/utilities.jl:84
  [3] locrepr(dctx::Documenter.HTMLWriter.DCtx, lines::Nothing)
    @ Documenter.HTMLWriter ~/Programs/git/GitHub/Documenter.jl/src/html/HTMLWriter.jl:713
```

The locrepr method in utilities.jl doesn't specify needing the second argument as a `::Page`, but assumes it, and tries accessing the field `source`.

fingolfin:
I've applied the adjustments and fixed the errors. At least in local tests it works fine (it's a bit hard to tell exactly without #2804 merged, and without the dedicated tests this really deserves, but we'll get there)

aaruni96:
In my testing, I get

```
ERROR:     Unexpected Julia interpolation in the Markdown. This probably means that you have an
    unbalanced or un-escaped $ in the text.

    To write the dollar sign, escape it with `\$`

    This is in file src/Experimental/FTheoryTools/literature.md, and we were given the value:

    `$` which is of type `Symbol`
```

With your earlier comment,

> The missing docs before src is to be expected

I think, this is ready to be merged.

fingolfin:
Taking a step back and thinking about this again, I wonder: perhaps we should simply perform this check earlier. Namely when we ingest docstrings (at which point we have precise location information), we could run a quick check on what we got and look for this issue, and report it then. This way, not only do we get better location info, but also we don't have to implement this for every writer again.

Thoughts @mortenpi ?

(To be clear, I'd still want to merge this PR, as it is an improvement *now*)

PR #2794, opened 2025-10-29
Do not force push in doc preview cleanup

mortenpi:
It's a protected branch. Let's just add a commit, so we keep the history.

image

FWIW, at some point it would still be good to get rid of the history too. But actually automatically doing it it a bit fishy, since we have no backup.

### DIFF
--- a/docs/src/lib/internals/search.md
+++ b/docs/src/lib/internals/search.md
@@ -1,0 +1,206 @@
+# Search 
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


fingolfin:
Weird CI failure -- it also pops up in in some of the other PRs where I used the "Update branch" button to re-run CI now that the set of required PRs changed

mortenpi:
Hope it's intermittent / gone. It looks like some upstream website that the LaTeX stack relied on had issues.

PR #2796, opened 2025-11-01
Try to print line ranges in a few more errors

fingolfin:
... but this is just a drop in the bucket. There are sooo many more places that could benefit from showing the lines. And is the best we can do really to search the markdown files via a regex? Line ranges should be kept track of during parsing!

Would it seem realistic to either adjust the Markdown stdlib to do that? Just had a quick look at the code, it seems relatively easy to at least keep track of the name of the originating file (if any) plus the position in the file were an node starts/ends (going from there to line ranges shouldn't be hard either). The main counterargument I can think of would be "this would cause overhead in Markdown parsing even for people who don't need it".

### DIFF
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -715,7 +715,7 @@ function Selectors.runner(::Type{Expanders.EvalBlocks}, node, page, doc)
                 @docerror(
                     doc, :eval_block,
                     """
-                    failed to evaluate `@eval` block in $(Documenter.locrepr(page.source))
+                    failed to evaluate `@eval` block in $(Documenter.locrepr(page.source, lines))
                     ```$(x.info)
                     $(x.code)
                     ```
@@ -732,7 +732,7 @@ function Selectors.runner(::Type{Expanders.EvalBlocks}, node, page, doc)
             # objects, like Paragraph.
             @docerror(
                 doc, :eval_block, """
-                Invalid type of object in @eval in $(Documenter.locrepr(page.source))
+                Invalid type of object in @eval in $(Documenter.locrepr(page.source, lines))
                 ```$(x.info)
                 $(x.code)
                 ```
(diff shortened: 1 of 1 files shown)


fingolfin:
Places using `locrepr` that still don't print line ranges (so perhaps it's not quite as bad as I thought in terms of how many do not do it):
```
src/cross_references.jl:83:                "invalid local image: path missing in $(Documenter.locrepr(page.source))",
src/cross_references.jl:93:            "invalid local link/image: colons not allowed in paths on Windows in $(Documenter.locrepr(page.source))",
src/cross_references.jl:104:            "invalid local link/image: path pointing to a file outside of build directory in $(Documenter.locrepr(page.source))",
src/cross_references.jl:119:                    "invalid local image: path contains a fragment in $(Documenter.locrepr(page.source))",
src/cross_references.jl:131:            "invalid local link/image: file does not exist in $(Documenter.locrepr(page.source))",
src/cross_references.jl:317:        msg = "Cannot resolve @ref for md$(repr(md_str)) in $(Documenter.locrepr(page.source))."
src/cross_references.jl:378:        push!(errors, "Header with slug '$slug' is not unique in $(Documenter.locrepr(page.source)).")
src/cross_references.jl:446:            return (error = "unable to parse the reference `$code` in $(Documenter.locrepr(page.source)).", exception = nothing)
src/cross_references.jl:467:            error = "unable to evaluate the type signature for `$code` in $(Documenter.locrepr(page.source)) in module $(mod)",
src/cross_references.jl:537:        push!(errors, "unable to generate issue reference for '[`#$num`](@ref)' in $(Documenter.locrepr(page.source)).")
src/expander_pipeline.jl:91:        @warn "code from a continued @example block unused in $(Documenter.locrepr(page.source))."
```

fingolfin:
I am folding this into PR #2803

PR #2797, opened 2025-11-01
Fix Oscar.jl regression tests

fingolfin:
For the MathOptInterface.jl failure, see https://github.com/jump-dev/MathOptInterface.jl/issues/2877  (CC @odow)

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -47,8 +47,6 @@ jobs:
   test-documentation-build:
     name: doc-build-${{ matrix.package }}
     runs-on: ubuntu-latest
-    env:
-      PACKAGE: ${{ matrix.package }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


fingolfin:
Part of the problem is that Oscar's manual actually contains parts of the Hecke/Nemo/AbstractAlgebra manuals, which we currently have to copy over in a somewhat hackish process (we are still looking for better solutions). And the packages are tightly interlinked: so when people work on stuff, they often need to have devel'ed versions of all packages together in an environment. When building the docs those versions must be matched, so the build script `develop`s them all into the `docs` environment. 

Overall, the `docs/Project.toml` should IMO really have `Hecke`, `Nemo` etc. in it, but it currently hasn't; this is because we run into some annoying corner cases in how `Pkg.jl` works in the past that caused pain, so we removed them.

All that said, I hope we can revise this setup in Oscar in the not-so-distant future to be a bit less hackish, and then the change in this PR can be reverted. But for the currently released version, it is necessary.

Issue #2799, opened 2025-11-01
Feature request: stay on page when changing version


langestefan:
Often when I am reading documentation, I want to change the version of the docs I am looking at.

For example, I am currently reading: https://docs.sciml.ai/ModelingToolkit/stable/basics/MTKLanguage/

When I want to read the `dev` version of this page, I go to the bottom left and change the version to `dev`. 

Instead of forwarding me to `https://docs.sciml.ai/ModelingToolkit/dev/basics/MTKLanguage/`, this forwards me to the homepage at `https://docs.sciml.ai/ModelingToolkit/dev/`. 

Ofcourse sometimes that particular page is not available in different versions of the documentation, but in that case we can fall back to the homepage. Can we make it so that, if that page is available, changing the version leads to the page I'm currently viewing?

fingolfin:
I'd like this as well, but it's not so easy. Right now we have no good way to determine whether the target page exists. So a naive approach would end up on a 404 error page.

And of course pages built with an older Documenter version won't have this (so you can't switch back) unless we implement some tool that updates old page's version.

But of course all of that could be overcome, I simply wanted to explain why it isn't that easy.

langestefan:
Got it, thanks for responding! This is available in sphinx (the python docs framework), so maybe we can take inspiration from that.

fingolfin:
Sure. If anyone knows how sphinx does it, please leave a comment here explaining it.

langestefan:
Isn't it as basic as this?

1. Check if page exists. If it returns a `200 OK`, forward to it.
2. If it doesn't exist, forward to home page.

fingolfin:
Yeah, OK, but how do you get the browser to do that? By inserting some JavaScript?

goerz:
I'm not a frontend expert, but this sounds like something that should be relatively straightforward. Claude tells me this (untested, but looks reasonable, as a starting point):

```html
<!DOCTYPE html>
<html>
<head>
  <title>Smart Links</title>
</head>
<body>

<a href="/primary-page" 
   data-fallback="/backup-page"
   class="smart-link">
   Click me
</a>

<script>
document.querySelectorAll('.smart-link').forEach(link => {
  link.addEventListener('click', async (e) => {
    e.preventDefault();
    
    const primaryUrl = link.href;
    const fallbackUrl = link.dataset.fallback;
    
    try {
      const response = await fetch(primaryUrl, { 
[...]

Issue #2529, opened 2024-06-14
Add `make` command to run prettier


mortenpi:
1. I think we could add a npm `package.json` into the root of the repo, that would install prettier.
2. We could add `make` targets to (1) instantiate the `package.json` and (2) to actually run prettier

https://github.com/JuliaDocs/Documenter.jl/pull/2514#issuecomment-2165219642, https://github.com/JuliaDocs/Documenter.jl/issues/2528

fingolfin:
We have `make format-js` and `make install-biome` so I think this is resolved.

Issue #2528, opened 2024-06-14
Add a note to contributor docs about how to format code


mortenpi:
https://github.com/JuliaDocs/Documenter.jl/pull/2514#issuecomment-2165219642

We use prettier for js, and also JuliaFormatter as some point hopefully (#2404).

fingolfin:
So for code formatting, `Runic` is now mentioned in https://documenter.juliadocs.org/stable/contributing/ but prettier is not.

PR #2801, opened 2025-11-03
Update navigation logic to maintain page context

langestefan:
Context: #2799 

Tested locally, seems to work correctly. 

- Stay on page when changing versions from e.g. `/stable/page.html` to  `/dev/page.html`.
- Fallback to home page (e.g. ` /dev`) when ` /dev/page.html` is not available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed the docstring block accordions from a custom implementation to HTML details+summary tag. ([#2772], [#2773])
 * Improved the search tokenizer and custom trimmer to improve search results. ([#1457], [#2114], [#2744])
 * Improved several warning/error messages to (more accurately) report the location (filename, line range) in which the warning/error originated. ([#2426], [#2752], [#2793], [#2803], [#2809])
+* The `forcepush=true` option to `deploydocs` now uses `--force-with-lease` instead of `--force`. ([#2817])
 
 ### Fixed
 
@@ -28,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Those previously passed but generated incorrect cross-references.
   You can fix this by ensuring that you have distinct headers across your markdown pages, or by using the `@id` syntax to give the headers unique slugs.
 * Changed CSS for the navigation "breadcrumbs" at the top of each page to avoid an issue where `code` blocks in there were missing spacing around them. ([#1614], [#2813])
+* Commit status for pull request previews publishing in GitHub Actions for out-of-repo deployment now works correctly. ([#2814], [#2816])
 
 ## Version [v1.15.0] - 2025-10-22
 
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -469,7 +469,7 @@ function version_tag_strip_build(tag; tag_prefix = "")
     return "$s0$s1$s2$s3$s4"
 end
 
-function post_status(::GitHubActions; type, repo::String, subfolder = nothing, kwargs...)
+function post_status(gha::GitHubActions; type, repo::String, subfolder = nothing, kwargs...)
     try # make this non-fatal and silent
         # If we got this far it usually means everything is in
         # order so no need to check everything again.
@@ -489,20 +489,21 @@ function post_status(::GitHubActions; type, repo::String, subfolder = nothing, k
             sha = get(ENV, "GITHUB_SHA", nothing)
         end
         sha === nothing && return
-        return post_github_status(type, repo, sha, subfolder)
+        return post_github_status(type, gha.github_repository, repo, sha, subfolder)
     catch
         @debug "Failed to post status"
     end
 end
 
-function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
+function post_github_status(type::S, source::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
     try
         Sys.which("curl") === nothing && return
-        ## Extract owner and repository name
+        ## Extract owner and repository names
+        source_owner, source_repo = split(source, '/')
         m = match(r"^github.com\/(.+?)\/(.+?)(.git)?$", deploydocs_repo)
         m === nothing && return
-        owner = String(m.captures[1])
-        repo = String(m.captures[2])
+        deploy_owner = String(m.captures[1])
+        deploy_repo = String(m.captures[2])
 
         ## Need an access token for this
         auth = get(ENV, "GITHUB_TOKEN", nothing)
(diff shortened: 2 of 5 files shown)


fingolfin:
Nice, thank you!

It needs a bit more work (not all by you):
- some thorough testing on different computers / browsers (it'd be super useful if we could find a way to make a preview build available somewhere)
- a changelog entry (can be added once this is "done" and agreed to merge)
- code formatting with prettier
- on the long run (after this PR is merged) a way to "upgrade" existing documentation versions to also support this
  - i.e. I guess one needs to iterate over all `assets/documenter.js` files in a `gh-pages` branch and "patch" them suitably, which might be a matter of searching&replacing a fixed substring -- or multiple, dunno how often this segment of the code changed over the years)

langestefan:
> some thorough testing on different computers / browsers

I'm on Linux Fedora 43, I tested using firefox. I could also test chrome, and I have a windows laptop I could test firefox and chrome on as well. 

> it'd be super useful if we could find a way to make a preview build available somewhere

You mean as in a webpage? I could just make the github pages endpoint available on my fork.

fingolfin:
> You mean as in a webpage? I could just make the github pages endpoint available on my fork.

That would be super helpful!

langestefan:
> That would be super helpful!

Should work now: https://langestefan.github.io/Documenter.jl

To test:

## Redirect to home page

The following page is not available on the `v1.16` version: [https://langestefan.github.io/Documenter.jl/dev/test-another-feature/](https://langestefan.github.io/Documenter.jl/dev/test-another-feature/)

Switching to `v1.16` from that page should therefore redirect to 

[https://langestefan.github.io/Documenter.jl/v1.16](https://langestefan.github.io/Documenter.jl/v1.16).

## Stay on page when switching version

Links that are available on all versions, for example [https://langestefan.github.io/Documenter.jl/dev/lib/public/](https://langestefan.github.io/Documenter.jl/dev/lib/public/)

So switching to `v1.16` should link back to [https://langestefan.github.io/Documenter.jl/v1.16/lib/public/](https://langestefan.github.io/Documenter.jl/v1.16/lib/public/)

## Still need to update the banner

This banner that appears at the top still links to the home page and not the current page:

> This documentation is for an older version that may be missing recent changes.
[Click here to go to the documentation for the latest stable release.](https://langestefan.github.io/Documenter.jl/dev)

## Direct links do not forward to the right page

A direct link to https://langestefan.github.io/Documenter.jl/v1.16/test-another-feature/  still yields a 404. So this only works if you use the version drop down menu. I'm personally okay with keeping it that way.

asinghvi17:
> on the long run (after this PR is merged) a way to "upgrade" existing documentation versions to also support this
i.e. I guess one needs to iterate over all assets/documenter.js files in a gh-pages branch and "patch" them suitably, which might be a matter of searching&replacing a fixed substring -- or multiple, dunno how often this segment of the code changed over the years)

It would be nice if this could be directly automated in deploydocs, which has access to the whole gh-pages branch.  But we should make sure any link rewriting isn't going to mess with Vitepress.

langestefan:
I'm not sure that belongs in this PR. I'd like to limit it to fixing the version button

fingolfin:
Very good. The JavaScript code looks plausible to me and is clearly documented.

So let's try to wrap it up then?
- [ ] format code with prettier 1.x
- [x] add changelog entry
- [ ] fix the version banner to also use this feature

Then I hope we can merge this soon. The "upgrade existing installations feature" can be implemented separately.

langestefan:
@fingolfin The banner doesn't seem to be functioning correctly, no matter what forwarding logic I try it will not bring me back to the same page. After trying for a few hours, I conclude that I unfortunately don't have the JS skills to fix it (or even understand what's going on). I suggest we leave the banner what it is for now.

Regarding prettier. I ran the prettier formatter locally and pushed the changes, but CI prettier check still fails. There is no debug output at all on the CI so there is nothing for me to go on. It would be really helpful for future devs if you could add maybe a pre-commit file or something so I can replicate what the CI is doing locally. If the CI can't tell me what's wrong then I can't fix it.

fingolfin:
I've not used prettier myself, all I know is what the Documenter manual says about it:
> We use [Prettier](https://github.com/prettier/prettier) v2 for formatting the HTML/web front end related code.

Well and also that https://prettier.io/docs/install has some instructions on using it, but that's for v3 and for some reason Documenter uses v2; I am sure someone who knows npm will easily be able to adapt those instructions to run v2 (and my guess is that ChatGPT and friends will also help doing that; but I am not going to go down that rabbit hole, I already have enough on my plate...)


Regarding making this more convenient for contributors: I agree! See also issues #2528 and #2529 (in short: "someone" should indeed make this nicer, now that "someone" just needs to show up ... :-/ )

(BTW that someone may also have to look into forking [`actionsx/prettier`](https://github.com/actionsx/prettier) as that repo has been dormant for 2 years, and not even dependabot PRs are being merged; as GH constantly evolves the containers for its actions, that means that an action may eventually become unusable if it is not update to work with the latest Node.js etc....)

mortenpi:
> but that's for v3 and for some reason Documenter uses v2

I've been keeping this at v2 because the (official) Prettier VS Code extension is stuck on v2 :disappointed:

langestefan:
>I've been keeping this at v2 because the (official) Prettier VS Code extension is stuck on v2

I think [Biome](https://biomejs.dev/) would be a good replacement. It has both a [github CI action](https://biomejs.dev/recipes/continuous-integration/#github-actions) and a [pre-commit hook](https://biomejs.dev/recipes/git-hooks/#pre-commit). That way you can  tell devs to run pre-commit with your own provided config file, and it will produce identical results to the github CI.

fingolfin:
Seems Biome also has first party VS Code support. Sounds good. Would you be interested in working on that, @langestefan ? Or, alternatively, to open an issue suggesting this?

fingolfin:
It is somewhat unfortunate that the `warner.js` code is not updated, because that's actually a pretty common place where I miss that functionality.

But the perfect shouldn't be enemy of the good: this PR is already a great improvement. Perhaps we can open an issue after merging it which requests that a similar change be implemented in `warner.js`, and perhaps someone else will come along to figure it out.

(Likewise an issue could be opened for adding tooling which updates the JS of older documentation versions so that they also benefit from this change).

langestefan:
> Would you be interested in working on that, @langestefan 

Sure! I will open an issue first to discuss.

> It is somewhat unfortunate that the warner.js code is not updated, because that's actually a pretty common place where I miss that functionality.

I will take another look at it again this weekend. If I can fix it I will make a PR.

fingolfin:
I figured out that I can run prettier v2 via `npx prettier@v2 . --write` and that seems to be it. Of course that requires `node` to be installed (which I did on macos via `brew install node`).

I've pushed an update to this PR.

fingolfin:
Hm, weird, prettier check still fails. Locally I see this:
```
$ npx prettier@v2 . --write
Need to install the following packages:
prettier@2.8.8
Ok to proceed? (y)

.prettierrc 31ms
assets/html/js/collapse.js 14ms
assets/html/js/copy.js 6ms
assets/html/js/footnote.js 6ms
assets/html/js/headroom.js 1ms
assets/html/js/metadata.js 2ms
assets/html/js/search.js 25ms
assets/html/js/settings.js 2ms
assets/html/js/shortcut.js 5ms
assets/html/js/sidebar.js 3ms
assets/html/js/themes.js 2ms
assets/html/js/versions.js 3ms
assets/html/themeswap.js 2ms
assets/html/warner.js 2ms
coverage/.codecov.yml 2ms
docs/versions.js 0ms

$ npx prettier@v2 . --check
Need to install the following packages:
prettier@2.8.8
Ok to proceed? (y)

Checking formatting...
All matched files use Prettier code style!
```

mortenpi:
LGTM as far as I can tell. Let's merge this and iterate if need be. Thank you @langestefan!

PR #2803, opened 2025-11-03
Add locrepr methods, improve path generation

fingolfin:
First, add a `locrepr(::Document, ::Page) method. This receives higher level info and thus has an easier time getting paths right.

Next, improve that method to ensure the source path is computed correctly relative to the actual document root.

Finally, add a `locrepr(::DCtx)` method which ensures we get correct file paths for those as well.

Should help with some bad file paths I've seen in some cases, but also with the problem @aaruni96 has been seeing in PR #2793.


mortenpi:
Let's add a brief changelog maybe, since I believe this improves the error messages? Otherwise, LGTM.

fingolfin:
OK, I folded PR #2796 into this one (as it also is about improving error messages) and added a changelog entry.

mortenpi (approved):

PR #2804, opened 2025-11-03
Adjust test outputs in test/examples to avoid warnings; fix bug in `expandfirst` handling on Windows

### LINKED ISSUE #2798: Hidden warnings / errors (?) due to `@quietly`
Due to confusing CI failures I got in PRs #2349 and #2683 I patched `test/examples/make.jl` to remove all `@quietly` statements (I picked that file because it is were the problems I was confused about showed up, but it may a good idea to look at other files. The resulting output shows a bunch of IMHO issues that show certain test cases are not working at all how they should be.

The problem with this is that when a new actual *error* happens in the affected files, anyone debugging those is confused by a ton of unrelated errors popping up (but it's hard to know they are unrelated).

Perhaps we should use `@quietly` less? Or perhaps it should *not* suppress warnings?

Here is the output; I am going to interrupt it for comments at points
```
     Testing Running tests...
[ Info: Building example/make.jl
[ Info: Building mock package docs: HTMLWriter / deployment build
[ Info: Doctest: skipped.
┌ Warning: Invalid type of object in @eval in examples/builds/html/src/issue491.md
```
Is this supposed to happen in that particular test (I think it's the one in lines 30-32

Also, the source file really is `test/examples/src/issue491.md`.

(But perhaps it is being copied by the test mechanism? I did not dig deeper yet. Anyway, observing this made me look and at least resulted in PR #2796)

```
│ ```@eval
│ "expanded_"*"eval"
│ ```
│ Evaluated to `String`, but should be one of
│  - Nothing
│  - Markdown.MD
│ Falling back to textual code block representation.
│
│ If you are seeing this warning/error after upgrading Documenter and this used to work,
│ please open an issue on the Documenter issue tracker.
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
```
There is a bunch of this and similar ones. Intentional, accident, ...?
```
┌ Warning: failed to evaluate `@eval` block in examples/builds/html/man/src/man/tutorial.md
│ ```@eval
│ import Markdown
│ code = string(sprint(Base.banner), "julia>")
│ Markdown.MD([Markdown.Code(code)])
│ ```
│   exception =
│    UndefVarError: `banner` not defined in `Base`
│    Suggestion: check for spelling errors or missing imports.
│    Stacktrace:
│     [1] getproperty(x::Module, f::Symbol)
│       @ Base ./Base_compiler.jl:47
│     [2] top-level scope
│       @ tutorial.md:69
│     [3] eval(m::Module, e::Any)
│       @ Core ./boot.jl:489
│     [4] (::Documenter.var"#55#56"{MarkdownAST.Node{Nothing}, Documenter.Page, Documenter.Document, LineNumberNode, Module, MarkdownAST.CodeBlock})()
│       @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/expander_pipeline.jl:712
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
```
The `Base.banner` function is gone in Julia 1.11 and later.

So this test is now not testing what it was supposed to test.

But why is this just a warning, not an error?
```
┌ Warning: failed to evaluate `@eval` block in examples/builds/html/man/src/man/tutorial.md
│ ```@eval
│ import Markdown
│ code = string(sprint(Base.banner), "julia>")
│ Markdown.MD([Markdown.Code(code)])
│ ```
│   exception =
│    UndefVarError: `code` not defined in `Main.EvalBlockSandbox`
│    Suggestion: add an appropriate import or assignment. This global was declared but not assigned.
│    Stacktrace:
│     [1] top-level scope
│       @ tutorial.md:70
│     [2] eval(m::Module, e::Any)
│       @ Core ./boot.jl:489
│     [3] (::Documenter.var"#55#56"{MarkdownAST.Node{Nothing}, Documenter.Page, Documenter.Document, LineNumberNode, Module, MarkdownAST.CodeBlock})()
│       @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/expander_pipeline.jl:712
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
```
I don't get this at all
```
┌ Warning: Invalid type of object in @eval in examples/builds/html/man/src/man/tutorial.md
│ ```@eval
│ rand(20, 20)
│ ```
│ Evaluated to `Matrix{Float64}`, but should be one of
│  - Nothing
│  - Markdown.MD
│ Falling back to textual code block representation.
│
│ If you are seeing this warning/error after upgrading Documenter and this used to work,
│ please open an issue on the Documenter issue tracker.
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
```
What's going on with this test failure? Intentional, accident...?
```
┌ Warning: Cannot resolve @ref for md"[`unsafe_convert`](@ref Base.unsafe_convert)" in examples/src/lib/functions.md.
│ - No docstring found in doc for binding `Base.unsafe_convert`.
│ - No docstring found in doc for binding `Base.unsafe_convert`.
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
┌ Warning: Cannot resolve @ref for md"[`cconvert`](@ref Base.cconvert)" in examples/src/lib/functions.md.
│ - No docstring found in doc for binding `Base.cconvert`.
│ - No docstring found in doc for binding `Base.cconvert`.
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
```
There are a ton more of these warnings which I am dropping here to not exceed GitHub's limits.
```
┌ Warning: invalid local link/image: file does not exist in examples/src/index.md
│   link =
│    @ast MarkdownAST.Link("some:path", "") do
│      MarkdownAST.Text("Colons not allowed on Windows ")
│      MarkdownAST.Text("–")
│      MarkdownAST.Text(" ")
│      MarkdownAST.Code("some:path")
│    end
│
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
┌ Warning: invalid local link/image: file does not exist in examples/src/index.md
│   link =
│    @ast MarkdownAST.Link(":path", "") do
│      MarkdownAST.Text("No \"drive\" ")
│      MarkdownAST.Text("–")
│      MarkdownAST.Text(" ")
│      MarkdownAST.Code(":path")
│    end
│
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
┌ Warning: invalid local link/image: file does not exist in examples/src/index.md
│   link =
│    @ast MarkdownAST.Link("X:\\some\\path", "") do
│      MarkdownAST.Text("Absolute Windows paths ")
│      MarkdownAST.Text("–")
│      MarkdownAST.Text(" ")
│      MarkdownAST.Code("X:\\some\\path")
│    end
│
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
```
Are those warning part of the test? If not, can we suppress them? If yes, how
can we make sure these warnings are "OK", but still are informed about other unintended warnings?
```
┌ Warning: no footnotes found for 'footnote' in examples/src/lib/functions.md.
└ @ Documenter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/utilities/utilities.jl:49
```
???
```
┌ Warning: Generated HTML over size_threshold_warn limit: example-output.md
│     Generated file size: 123.63 (KiB)
│     size_threshold_warn: 100.0 (KiB)
│     size_threshold:      200.0 (KiB)
│     HTML file:           example-output/index.html
└ @ Documenter.HTMLWriter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/html/HTMLWriter.jl:1996
```
Lots of these warnings about size thresholds being exceeded are shown. Surely they
are not all there to test this feature? Perhaps some `.md` files should be split up,
or that threshold be increased for the test?
```
┌ Warning:     Unexpected Julia interpolation in the Markdown. This probably means that you have an
│     unbalanced or un-escaped $ in the text.
│
│     To write the dollar sign, escape it with `\$`
│
│     We don't have the file or line number available, but we got given the value:
│
│     `foo` which is of type `Symbol`
└ @ Documenter.HTMLWriter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/html/HTMLWriter.jl:2454
┌ Warning:     Unexpected Julia interpolation in the Markdown. This probably means that you have an
│     unbalanced or un-escaped $ in the text.
│
│     To write the dollar sign, escape it with `\$`
│
│     We don't have the file or line number available, but we got given the value:
│
│     `[1 2 3; 4 5 6]` which is of type `Expr`
└ @ Documenter.HTMLWriter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/html/HTMLWriter.jl:2454
```
No idea where this is coming from, and thus no idea if it is "intentional".

Presumably PR #2793 will help narrowing this down.
```
┌ Warning: For 2 @example blocks, the 'text/html' representation of the resulting
│ object is above the threshold (example_size_threshold: 8192 bytes).
│ - 1 blocks had no image MIME show() method representation as an alternative.
│   Sticking to the 'text/html' representation (largest block is 113087 bytes).
│   On pages: example-output.md
│ - 1 blocks had 'image/svg+xml' fallback image representation available, using that.
│   On pages: example-output.md
└ @ Documenter.HTMLWriter ~/Projekte/Julia/packages/JuliaDocs/Documenter.jl/src/html/HTMLWriter.jl:908
```
Ditto: where is this from? is it part of the test or not?

There's more, but a lot seems to be repetitions of the issues mentioned above, so I assume if we can resolve them, it will fix many of the repetitions.

fingolfin:
... so that we can run those tests with `warnonly = false`, at least in some specific Julia versions.

Resolves #2798.


test/examples/pages/a.jl
```diff
@@ -3,9 +3,9 @@
 
 Links:
 
-- [`ccall`](@ref)
+- [`deepcopy`](@ref)
 - [`while`](@ref)
-- [`@time(x)`](@ref)
```
fingolfin, line 8: Explanation: `ccall` and `@time` docstrings reference other docstrings, and some of those have further ref's; that then leads to "missing reference" warnings, unless one pulls in all of those. So I replaced them with alternatives that (currently) don't reference anything in their docstrings

test/examples/src/index.md
```diff
@@ -45,7 +45,7 @@ end
 
 ```jldoctest
 julia> [1.0, 2.0, 3.0]
-3-element Array{Float64,1}:
+3-element Vector{Float64}:
```
fingolfin, line 48: Adjust to changed output in... Julia 1.6 at least? Which is the minimum version supported by Documenter, so...

test/examples/make.jl
```diff
@@ -261,7 +261,7 @@ function html_doc(
         build_directory, mathengine;
         htmlkwargs = (;),
         image_assets = ("images/logo.png", "images/logo.jpg", "images/logo.gif"),
-        warnonly = true,
+        warnonly = (VERSION.major == 1 && VERSION.minor >= 10),
```
fingolfin:
```suggestion
        warnonly = (VERSION.major == 1 && VERSION.minor <= 10),
```

test/examples/src/index.md
```diff
@@ -199,7 +199,7 @@ julia> g(2)
 ```
 ```jldoctest
 julia> f(2)
-ERROR: UndefVarError: f not defined
+ERROR: UndefVarError: `f` not defined in `Main`
```
fingolfin, line 202: This version of the message is shown in Julia >= 1.11; this is one of the reasons why we have `warnonly = (VERSION.major == 1 && VERSION.minor <= 10)` further above

test/examples/src/index.md
```diff
@@ -264,7 +264,7 @@ julia> t = T()
 T()
 
 julia> fullname(@__MODULE__)
-(:Main,)
+(:Main, :Main)
```
fingolfin, line 267: I am unsure how this was *ever* correct... I mean, yeah, ideally it would print `(:Main,)` because that's what is printed in `Main`. But we are in a sandbox module, and we are replacing the module name in output by `Main`, so I think this is kinda the best we can hope for?

C.f.:
```julia-repl
julia> fullname(@__MODULE__)
(:Main,)

julia> module Foo println(fullname(@__MODULE__)) end
(:Main, :Foo)
Main.Foo
```

test/examples/src/index.md
```diff
@@ -273,7 +273,7 @@ julia> @__MODULE__
 Main
 ```
 
-# Issue398
+# Issue #398
```
fingolfin, line 276: That's just for consistency (I am also tempted to sort these sections by issue number... or maybe each issue test should be in a file of its own?

test/examples/src/index.md
```diff
@@ -331,16 +331,19 @@ julia> throw(MyException("test exception"))
...
-# Issue418
+# Issue #418
 
 ```jldoctest
 julia> f(x::Float64) = x
 
 julia> f("")
 ERROR: MethodError: no method matching f(::String)
+The function `f` exists, but no method is defined for this combination of argument types.
```
fingolfin, line 342: This line is new in Julia 1.11

test/examples/src/index.md
```diff
@@ -434,19 +438,24 @@ julia> ans
...
+We filter the string `Main.` in outputs to make the outputs
+look more like they would in the REPL.
 
 ```jldoctest
 julia> struct Point end;
 
 julia> println(Point)
 Point
 
+julia> import Base: sqrt
```
fingolfin, line 452: This is necessary to make this test pass in Julia >= 1.12

test/examples/src/index.md
```diff
@@ -483,12 +497,6 @@ julia> x->x # ignore error on 0.7
 r = :a
 ```
 
-# Bad links (Windows)
-
-* [Colons not allowed on Windows -- `some:path`](some:path)
-* [No "drive" -- `:path`](:path)
-* [Absolute Windows paths -- `X:\some\path`](X:\some\path)
-
```
fingolfin, line 491: This test caused a bunch of warnings. And it is unclear to me what it even is supposed to test. It was added in 80721fdf1b5707307b35b902a22f71d766e53cbb from PR #485.

My best guess is in commit 80721fdf1b5707307b35b902a22f71d766e53cbb which was part of PR #485 before squashing. It introduces a function `fixlinks!` which has some Windows specific code:
```julia
    if is_windows() && ':' in img.url
        Utilities.warn("Invalid local image: colons not allowed in paths on Windows\n    '$(img.url)' in $(get(navnode.page))")
        return
    end
```
That function is long gone, but a descendant of the code lives on in `local_links!`:
```julia
    elseif Sys.iswindows() && ':' in path
        @docerror(
            doc, :cross_references,
            "invalid local link/image: colons not allowed in paths on Windows in $(Documenter.locrepr(page.source))",
            link = node
        )
        return
    end
```

So perhaps this line is about getting coverage for that `@docerror`

But despite this, I think it is the right move to remove this here: because right now, the test results in "fake coverage": we wouldn't notice if the test for colons was broken because we ignore the output via `@quietly` and even if we didn't, a failure here would most likely not trigger a CI failure.

So instead, we need a new test which triggers this and then *specifically* watches out for the result, and checks that it shows a warning/error with the appropriate message. (And that should probably be three tests, one for each instance).

As I mentioned elsewhere, I'd like such a setup also for testing the quality of the `locrepr` output, i.e. to ensure it shows the "right" file paths.

test/examples/src/issue491.md
```diff
@@ -28,7 +28,8 @@ Tests that code-like blocks are expanded in nested Markdown contexts.
     Test the different types of code-like blocks:
 
     ```@eval
-    "expanded_"*"eval"
+    import Markdown
+    Markdown.parse("expanded_"*"eval")
```
fingolfin, line 32: Without this we get a warning because `@eval` should return `nothing` or a `Markdown.MD`. We don't want that here. Instead there should be a dedicated test for this warning, see issue #2805

test/examples/src/man/tutorial.md
```diff
@@ -66,12 +66,7 @@ julia> a / b
 
 ```@eval
 import Markdown
-code = string(sprint(Base.banner), "julia>")
-Markdown.MD([Markdown.Code(code)])
-```
-
-```@eval
-rand(20, 20)
+Markdown.MD("Hello")
```
fingolfin, line 69: `Base.banner` was moved to `REPL.banner` in Julia 1.11. But anyway, there seems no point in having a complex test here, is there?

test/examples/tests.jl
```diff
@@ -272,15 +272,15 @@ end
                             @test item.priority == 1
                             @test DocInventories.uri(item) == "lib/functions/#Main.AutoDocs.K"
                         end
-                        @test length(inv(":jl:keyword:`for`")) == 1
-                        if length(inv(":jl:keyword:`for`")) == 1
-                            item = inv[":jl:keyword:`for`"]
```
fingolfin, line 277: I had to replace `for` by `while` because in recent Julia versions, the `for` docstring as a `@ref` to `outer`.

fingolfin:
@mortenpi I tried to explain the reasoning behind all changes in PR review comments.

But tests fail on Windows, any help would be appreciated:
```
===================== @quietly: output from the expression =====================
[ Info: Doctest: skipped.
┌ Warning: expandorder/AA.md in expandfirst does not exist
└ @ Documenter D:\a\Documenter.jl\Documenter.jl\src\expander_pipeline.jl:45
┌ Error: failed to run `@example` block in examples\src\expandorder\01.md:9-11
│ ```@example
│ isfile("AA.txt") || error("AA.txt missing")
│ ```
│   exception =
│    AA.txt missing
│    Stacktrace:
│      [1] error(s::String)
│        @ Base .\error.jl:44
│      [2] top-level scope
│        @ 01.md:10
│      [3] eval(m::Module, e::Any)
│        @ Core .\boot.jl:489
│      [4] #61
│        @ D:\a\Documenter.jl\Documenter.jl\src\expander_pipeline.jl:856 [inlined]
│      [5] cd(f::Documenter.var"#61#62"{Module, Expr}, dir::String)
│        @ Base.Filesystem .\file.jl:101
```

test/examples/src/index.md
fingolfin: See PR #2822

test/examples/src/issue491.md
fingolfin: Such a test is added in PR #2823

fingolfin:
Ah of course it's right there in the error:
```
┌ Warning: expandorder/AA.md in expandfirst does not exist
└ @ Documenter D:\a\Documenter.jl\Documenter.jl\src\expander_pipeline.jl:45
```
My guess would be this is due to a different in path separators, combined with dumb string comparision on the paths. Indeed, the warning is produced here:
```julia
function expand(doc::Documenter.Document)
    priority_pages = filter(doc.user.expandfirst) do src
        if src in keys(doc.blueprint.pages)
            return true
        else
            @warn "$(src) in expandfirst does not exist"
            return false
        end
    end
```
I'll see if using `joinpath` will help.

But of course other kwargs for `mkdocs` such as `pages` use slashes in path just fine. Perhaps those arguments are processed somewhere, and the same processing should be applied to `expandfirst` ? Perhaps @mortenpi knows.

CHANGELOG.md
```diff
@@ -20,7 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Modules for `@example` environments are now generated by `eval`'ing an expression, rather than invoking the `Module` constructor, which is not recommended. (#2683)
+* Modules for `@example` environments are now generated by `eval`'ing an expression, rather than invoking the `Module` constructor, which is not recommended. ([#2683])
+* Fix handling of `expandfirst` entries containing slashes as path separators on Windows. ([#2804])
```
fingolfin, line 24: So this uncovered a genuine bug in `expandfirst` handling on Windows, yay. I've thus added a changelog entry. (And fixed the link in the entry above)

test/examples/src/index.md
mortenpi: We try to fix some of these things up though. I wonder if this regressed at some point..

fingolfin: Possibly but I see nothing in the code that would fix this. 

Anyway, if this ever worked (and I am not convinced it did, but I am also too lazy to try to reconstruct the exact setup from when this test was added), then it must have been quite some time ago. That's part of the problem this PR was meant to fix: the tests in there *did not actually act as tests* because except for the gravest kinds of failures (crashes, hard exceptions) any regressions were ignored :-(.

I think the current behavior is OK -- if someone *really* wants to print the `fullname` of a module, and this causes a problem, they can file an issue and we can worry about the possibility of a fix then.

mortenpi: Yep, agreed.

PR #2808, opened 2025-11-04
Warn/error if jldoctest starts with an empty line; or lacks empty line between two REPL prompts

### LINKED ISSUE #2031: Empty lines before first prompt not handled in doctests
````markdown
```jldoctest

julia> 2 + 2
4

julia> 2 + 3
5
```
````

leads to

```
┌ Error: doctest failure in src/index.md:3-10
│ 
│ ```jldoctest
│ 
│ julia> 2 + 2
│ 4
│ 
│ julia> 2 + 3
│ 5
│ ```
│ 
│ Subexpression:
│ 
│ 2 + 2
│ 
│ Evaluated output:
│ 
│ 4
│ 
│ Expected output:
│ 
│ 
│ 
│   diff = 4
└ @ Documenter.DocTests src/index.md:3
┌ Error: doctest failure in src/index.md:3-10
│ 
│ ```jldoctest
│ 
│ julia> 2 + 2
│ 4
│ 
│ julia> 2 + 3
│ 5
│ ```
│ 
│ Subexpression:
│ 
│ 2 + 3
│ 
│ Evaluated output:
│ 
│ 5
│ 
│ Expected output:
│ 
│ 4
│ 
│   diff = 45
└ @ Documenter.DocTests src/index.md:3
```

It looks like that somehow this makes the `inputs` and `outputs` arrays shifted in `repl_splitter` (we're probably pushing an empty output): https://github.com/JuliaDocs/Documenter.jl/blob/71e9f40d51ba855f69e21a46401610891f282d7e/src/DocTests.jl#L532-L560

This also confuses doctest fixing (which starts just appends the same output every time). I assume if we fix `repl_splitter` it will fix itself too, but we should add a test for that, just in case.

### LINKED ISSUE #2083: Unhelpful error if consecutive lines start with `julia> `
The following REPL doctest is missing an empty line between the two prompts:
```jldoctest
julia> a = 1;
julia> a + 1
2

```
Running it gives a not very helpful error message:
```
│ Subexpression:
│ 
│ a + 1
│ 
│ Evaluated output:
│ 
│ ERROR: UndefVarError: a not defined
│ Stacktrace:
│  [1] top-level scope
│    @ none:1
│ 
│ Expected output:
│ 
│ a = 1;
```

Note that the first line `a = 1;` is parsed as output, so `a` is not defined in the second line and the expected output of the second line is `a = 1;`, which is quite unexpected.

I traced this down to https://github.com/JuliaDocs/Documenter.jl/blob/579d478b7f66bf83a83fcb1ca8c2aa75b21f450a/src/DocTests.jl#L532-L560
This code seems to expect any two lines starting with `julia> ` to be separated by a non-input line. This is of course true for correct examples, but maybe the broken example above could be handled more helpfully.

If you think this is a non-issue, feel free to close it of course.

### LINKED ISSUE #2679: empty line after jldoctest invocation causes errors
I noticed in a recent PR review that an empty line after ```jldoctest causes issues. 
``````
```jldoctest

julia> do something here 
```
``````
will cause an error but removing the newline fixes it:

``````
```jldoctest
julia> do something here 
```
``````

Here are two live examples: 
https://github.com/JuliaStats/GLM.jl/actions/runs/14439279742/job/40485771112?pr=487#step:5:318

https://github.com/JuliaStats/GLM.jl/actions/runs/14439279742/job/40485771112?pr=487#step:5:391

I could reproduce these errors on my local machine, and can confirm that removing the new line indeed fixes this issue.

fingolfin:
These are minor usability errors, "paper cuts" if you will, that have annoyed me repeatedly in the past, and clearly also other people:

Resolves #2031
Resolves #2083
Resolves #2679
Alternative to and hence closes #2810

Alternative in PR #2810: we could also just *allow* all of this. (For the leading empty line: Should  it then be rendered, or hidden?)

Downside: such a doctest would not work with older Documenter versions. I therefore am inclined to prefer the strict version implemented here.

But in the end, either is preferable over the current errors.

(I am waiting with changelog and test cases until I know which direction we'll take)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed the docstring block accordions from a custom implementation to HTML details+summary tag. ([#2772], [#2773])
 * Improved the search tokenizer and custom trimmer to improve search results. ([#1457], [#2114], [#2744])
 * Improved several warning/error messages to (more accurately) report the location (filename, line range) in which the warning/error originated. ([#2426], [#2752], [#2793], [#2803], [#2809])
+* Warn/error if jldoctest starts with an empty line; or lacks empty line between two REPL prompts. ([#2031], [#2083], [#2679], [#2808])
 * The `forcepush=true` option to `deploydocs` now uses `--force-with-lease` instead of `--force`. ([#2817])
 
 ### Fixed
@@ -1999,6 +2000,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2018]: https://github.com/JuliaDocs/Documenter.jl/issues/2018
 [#2019]: https://github.com/JuliaDocs/Documenter.jl/issues/2019
 [#2027]: https://github.com/JuliaDocs/Documenter.jl/issues/2027
+[#2031]: https://github.com/JuliaDocs/Documenter.jl/issues/2031
 [#2051]: https://github.com/JuliaDocs/Documenter.jl/issues/2051
 [#2054]: https://github.com/JuliaDocs/Documenter.jl/issues/2054
 [#2058]: https://github.com/JuliaDocs/Documenter.jl/issues/2058
@@ -2011,6 +2013,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2078]: https://github.com/JuliaDocs/Documenter.jl/issues/2078
 [#2080]: https://github.com/JuliaDocs/Documenter.jl/issues/2080
 [#2081]: https://github.com/JuliaDocs/Documenter.jl/issues/2081
+[#2083]: https://github.com/JuliaDocs/Documenter.jl/issues/2083
 [#2085]: https://github.com/JuliaDocs/Documenter.jl/issues/2085
 [#2100]: https://github.com/JuliaDocs/Documenter.jl/issues/2100
 [#2103]: https://github.com/JuliaDocs/Documenter.jl/issues/2103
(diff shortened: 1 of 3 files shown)


fingolfin:
Advantages of a loose version:
- `repl_splitter` can stay being a function with one argument, which might be interesting for external projects
- "it just works" experience for users (instead of forcing the user to adapt to the code, let the code adapt to the user)
  - note that doctests failing the tests being added here can be copy&pasted into a Julia REPL just fine, they work there, so why not in Documenter
- ...

mortenpi:
I am inclined to turn these into errors/warnings (i.e. doctest failures):

- Empty line at the top just looks weird. Not sure why anyone would want that. So let's "lint" it away.
- Between the prompts, the Julia REPL always forces an empty line between prompts. So I think it's reasonable to require this. It also feels like it should help with the heuristic a bit..

Requiring empty lines would mean that this case should be fine:

image

But it also shouldn't error. Also I feel like it _should_ parse fine right now (but maybe it doesn't?).

So maybe the correct thing to do here is to parse. And when a doctest fails, then check if there is a `^julia>` match somewhere in the output part (i.e. a missing empty line), and add to the error/warning we print to the user that "hey, maybe this is meant to be two separate prompts"? This would resolve #2083 I think?

fingolfin:
Thinking some more about this, my conclusion is that the argument "this code can be pasted fine into the REPL, so we should accept it" isn't so great after all. I think the converse which @mortenpi brought up at least implicitly carries more weight: what does the *output* of the REPL look like? I think having the rule "the doctest content should match the REPL output exactly is more important.

Besides, if we reject it now, we can always switch to accepting it in the future. The other direction is much harder.

So I'd say, let's indeed go with this PR over the other.

I'll add some tests and a changelog entry.

fingolfin:
Add changelog entry, added a test -- I split this into two commits: the first commit adds the test with the current output. The second commit has the fix and adapts the test to the new output. This way one nicely sees the difference in quality between old and new message.

PR #2809, opened 2025-11-04
Report location of @docs blocks with parsing errors

fingolfin:
In particular, show the precise location where the error occurred.

Resolves #2426

I tested it locally but it still needs a test. Will add one later.

UPDATE: tests added. More so, this now resolves #2805 in so far as it adds a framework for adding tests for warning messages. I've used it here to test the warnings I am improving here, and have a couple more ready for a follow-up PR. That said, it's not "done" in so far as it does not yet cover *all* kinds of warnings; but that seems to me more like an open-ended project anyway.

PR #2812, opened 2025-11-04
Add named `@eval` blocks

### LINKED ISSUE #2074: Named at-eval blocks
Apparently we don't support named at-eval blocks. It can sometimes be useful to be able to use at-setup blocks with at-evals.

fingolfin:
Resolves #2074

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Added option `treat_markdown_warnings_as_error` which throws an error when encountering a markdown/interpolation warning ([#2792], [#2751])
 * Footnotes can now be previewed by hovering over the link. ([#2080])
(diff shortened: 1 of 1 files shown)


src/expander_pipeline.jl
```diff
@@ -691,13 +691,20 @@ function Selectors.runner(::Type{Expanders.EvalBlocks}, node, page, doc)
     @assert node.element isa MarkdownAST.CodeBlock
     x = node.element
 
+    matched = match(r"^@eval(?:\s+([^\s;]+))?\s*$", x.info)
+    matched === nothing && error("invalid '@eval <name>' syntax: $(x.info)")
+    name = matched[1]
```
mortenpi, line 696: One thing I've really wanted us to have is a reusable function for parsing the `.info` field, so that we'd ensure consistent syntax for those (including keywords). But that's orthogonal to this PR.

fingolfin: Agreed. In fact I have another PR in the pipeline doing that, and applying it also to jldoctest. But I first would like to get at least #2809 merged so that I can use the machinery I set up there for tests.

PR #2813, opened 2025-11-04
Fix missing space around code element in breadcrumb

### LINKED ISSUE #1614: Missing space between code in breadcrumb
e.g. from https://julialogging.github.io/previews/PR2/how-to/enable-debug/ you see `How to enable@debugmessages` and not `How to enable @debug messages`

fingolfin:
Resolves #1614

I verified locally that it fixes the issue, but have no idea how to add a test for this...

Should this get a changelog?


mortenpi:
LGTM!

> Should this get a changelog?

I know it's a tiny change, but my policy has generally been to always err on the side of a small changelog.

> I verified locally that it fixes the issue, but have no idea how to add a test for this...

We don't have UI tests. So local testing is how we usually do it :slightly_smiling_face:

fingolfin:
Actually it definitely should get a changelog entry because there is a chance that it breaks stuff, too, and then it's helpful for people tracking down the cause of the breakage to see it in the changelog.

fingolfin:
Added a changelog entry.

PR #2815, opened 2025-11-06
[docs] Clarify use of `DocumenterTools.genkeys` for out-of-repo deployment

giordano:
I believe the documentation as written now is quite ambiguous.

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -596,7 +596,7 @@ In those cases, it can be useful to deploy the docs in the `gh-pages` of a separ
 The following steps can be used to deploy the documentation of a "source"
 repository on a "target" repo:
 
-1. Run `DocumenterTools.genkeys()` to generate a pair of keys
+1. Run `DocumenterTools.genkeys(; user="TargetRepoOrg", repo="TargetRepo")` to generate a pair of keys
 2. Add the **deploy key** to the **"target"** repository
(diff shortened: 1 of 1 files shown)


mortenpi:
I don't mind this, but personally I never follow the URLs, so implying that you _have_ to fill those two values in can be annoying -- I really just want the key values. Could we easily accommodate both cases?

fingolfin:
@mortenpi I am confused by your comment (I might be parsing it wrong): which URLs do you mean that you "never follow", and how does this relate to the change @giordano is proposing"?

Are you suggesting something `DocumenterTools.genkeys()` for keys generated for out-of-repo deployment *is* fine? If so, what's the context?

fingolfin (approved):
This looks sensible to me, but I'll delay merging to see if @mortenpi clarifies what he meant

mortenpi:
> which URLs do you mean that you "never follow"

The only difference between calling `genkeys` with or without arguments is the URLs it generates in the output.

image

It's meant to just let you click on the URL to go to the right page, but I never actually use them myself. So I am not convinced it's good to imply here that you seemingly _have_ to fill those arguments out, since it's a bunch of annoying typing for the user which they likely do not care about.

But I don't have a strong opinion -- if we feel that this is an improvement, we can merge this.

fingolfin:
Thank you for the explanation @mortenpi now I understand what you meant.

So yeah, without any arguments, it prints URLs like `https://github.com/$USER/$REPO/settings/keys` with literatlly `$USER` in there. Not sure what is misleading about that...

@giordano your thoughts?

PR #2816, opened 2025-11-06
Correctly post GitHub commit status for out-of-repo deployment

### LINKED ISSUE #2814: `post_status` is broken for out-of-repo deployment
I'm trying to do an [out-of-repo deployment](https://documenter.juliadocs.org/stable/man/hosting/#Out-of-repo-deployment), but the status isn't being reported.

As far as I understand by reading the code, https://github.com/JuliaDocs/Documenter.jl/blob/5da6aa36126c71a7b3a4ffa408984828576aa59e/src/deployconfig.jl#L498 is receiving the wrong `deploydocs_repo` argument because there's no disentanglement anywhere between the source repo and the deployment repo.  That function is called by https://github.com/JuliaDocs/Documenter.jl/blob/5da6aa36126c71a7b3a4ffa408984828576aa59e/src/deployconfig.jl#L472 but I don't see anywhere a distinction between the two repos.  Am I missing something.

giordano:
At the moment `post_status` assumes source and deploy repos are the same, it's even completely ignoring the argument which holds the source repo.

I don't know how to test this in CI, but I did test it in https://github.com/NumericalEarth/Breeze.jl/pull/77#issuecomment-3499258742.

Fix #2814.

### DIFF
--- a/src/deployconfig.jl
+++ b/src/deployconfig.jl
@@ -534,7 +534,6 @@ function post_github_status(type::S, source::S, deploydocs_repo::S, sha::S, subf
         push!(cmd.exec, "https://api.github.com/repos/$(source_owner)/$(source_repo)/statuses/$(sha)")
         # Run the command (silently)
         io = IOBuffer()
-        @debug "About to run curl command" cmd
         res = run(pipeline(cmd; stdout = io, stderr = devnull))
         @debug "Response of curl POST request" response = String(take!(io))
     catch


src/deployconfig.jl
```diff
@@ -489,20 +489,21 @@ function post_status(::GitHubActions; type, repo::String, subfolder = nothing, k
...
     end
 end
 
-function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
+function post_github_status(type::S, source::S, deploydocs_repo::S, sha::S, subfolder = nothing) where {S <: String}
     try
         Sys.which("curl") === nothing && return
-        ## Extract owner and repository name
+        ## Extract owner and repository names
+        source_owner, source_repo = split(source, '/')
```
giordano, line 502: Is there any chance `gha.github_repository` includes the `.git` extension?  In that case I'd need to drop it with a regex like done below.

src/deployconfig.jl
```diff
@@ -530,9 +531,10 @@ function post_github_status(type::S, deploydocs_repo::S, sha::S, subfolder = not
             error("unsupported type: $type")
         end
         push!(cmd.exec, "-d", JSON.json(json))
-        push!(cmd.exec, "https://api.github.com/repos/$(owner)/$(repo)/statuses/$(sha)")
+        push!(cmd.exec, "https://api.github.com/repos/$(source_owner)/$(source_repo)/statuses/$(sha)")
         # Run the command (silently)
         io = IOBuffer()
+        @debug "About to run curl command" cmd
```
giordano: This was helpful for me for debugging: https://github.com/JuliaDocs/Documenter.jl/issues/2814#issuecomment-3499152348

src/deployconfig.jl
giordano: Hopefully not https://github.com/JuliaDocs/Documenter.jl/blob/5da6aa36126c71a7b3a4ffa408984828576aa59e/src/deployconfig.jl#L323 ? Unless someone really calls this constructor directly with the trailing `.git`.

navidcy:
@milankl, with this PR Documenter should show the deploy-docs status even when Documenter pushes on a different repo. 

https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/720#issuecomment-2848758623

src/deployconfig.jl
mortenpi: I think it's fine to assume no `.git`.

src/deployconfig.jl
mortenpi: This would print the token though, so I am a little bit hesitant to add this.

mortenpi: Let's maybe remove this? Or refactor, so that we `push!` the token after we `@debug` it? Otherwise, I have no reservations and the PR LGTM.

```suggestion
```

giordano: > This would print the token though

Secrets are never printed to screen, they're replaced with asterisks, as in the example I linked above.

mortenpi: Only if you're running things in the context where that is true (like on GHA, where the token is a secret). I'd just prefer to be defensive here and never print the secret.

giordano: Ok, done.

giordano:
> I don't claim to understand the details

The `post_github_status` function adds a custom commit status to GitHub, which has a link to the deployed documentation.  It's all good and simple if source repo and target repos are the same (which is what the current implementation implicitly assumed), but for out-of-repo deployments the two repos are different, so we need to add the custom status to the source repo, but the link should point to the target repo.  This is what this PR does (or tries to!) :slightly_smiling_face:

milankl:
This is great, thanks @giordano for your work on this: To clarify, would this deploy docs check automatically pop up when a new version of Documenter is released and used in a repo that deploys docs to another repo? Happy to report back here whether this works but do I need to do anything 😉 ?

giordano:
> would this deploy docs check automatically pop up when a new version of Documenter is released and used in a repo that deploys docs to another repo?

Yes.

> do I need to do anything 😉 ?

No, just wait for next release.

giordano:
Can a new version including this PR be tagged please? 🙂

fingolfin:
It would be  nice if some of the other read PRs would be merged first, IMHO

fingolfin:
I've merged my changes, I'll now pester @mortenpi for a new release ;-)

PR #2817, opened 2025-11-07
Push `--force-with-lease` instead of `--force`

giordano:
`git push --force` is really a bad practice, `git push --force-with-lease` prevents issues in case of race conditions.  Of course this can cause some deployments to fail and arguably users should orchestrate the docs publishing jobs to avoid conflicts (e.g. using concurrency groups), but I think it's better for a `git push` to fail loudly rather than catastrophically and silently erase data if all other barriers fail.


giordano:
Can a new version including this PR be tagged please? 🙂

Issue #2819, opened 2025-11-07
Cleanup tests: "merge" `test/warnings` into `test/doctests` (possibly renaming the latter), and more


fingolfin:
As discussed with @mortenpi on Slack. Excerpts:

> But it certainly would make sense to move my new warning tests under the `test/doctests` "test framework" -- but perhaps renamed (I thought `test/doctests` was purely for testing doctests so didn't even look there). Perhaps `test/output_tests` and then have subdirs like `doctests`, `warnings` ? and perhaps a small `test/README.md` that gives some hints on where to add certain kinds of new tests?

> it might be good if we can identify a few other "groups" of "similar" tests, and move them into other `test/...` subdirs. (anything we can not categorize / which seems like it should be redone, could go into a `test/legacy_tests` dir -- just to make clear to newcomers "don't imitate those tests"

> We also have those example manual tests that are messy.. but they are more integration-test stuff.

PR #2822, opened 2025-11-07
Move Windows path warning tests (for path containing `:`) and actually take their results into account

fingolfin:
This adds back one of the tests removed in #2804 but in a form that's less useless. (See <https://github.com/JuliaDocs/Documenter.jl/pull/2804#discussion_r2489443251> for more information about this test.)

However, this will fail CI in the current fork, because the warning (about paths containing `:`) is actually different on Windows (were our custom code triggers) versus Unix (where a path with a `:` is fine, but Documenter will complain because the specified files don't actually exist).

So how should we deal with this? Some options:
1. we could just not bother and drop this test (honestly, I think this would be fine, except that it'd drop test coverage a bit)
2. we could extend the check which forbid filenames with a `:` to trigger independently of the OS
3. we could have two different "expected texts" for this test: one for windows, one for everyone else (similar to how the `test/doctests` tests allow different versions of a test based on the Julia version)

Thoughts?


mortenpi:
In general, I'd prefer we didn't delete tests that exist and exercise lines that otherwise would not be covered. Minimally, they will catch things that throw, like missing variable errors etc.

As for this test.. I think that, in general, we do need to be able to have separate references per Julia version and per OS. So (3) seems like the simple approach, even though a bit annoying.

fingolfin:
> In general, I'd prefer we didn't delete tests that exist and exercise lines that otherwise would not be covered. Minimally, they will catch things that throw, like missing variable errors etc.

Counter point: by having a test that exercises a line of code but only catches the most severe cases of failures, we obscure the fact that the code in there is in fact not well-tested.

I'd rather have a proper test than an incomplete test whose failures are hidden by default.

This is why I did not just delete the tests, but instead also opened a PR to re-add it, but in a more helpful way.

> 
> As for this test.. I think that, in general, we do need to be able to have separate references per Julia version and per OS. So (3) seems like the simple approach, even though a bit annoying.

OK, that seems fair enough to me. Having the ability to do so may come in handy in other instances anyway.

I can't promise to implement this here right away, but I'll get to it eventually -- but probably after addressing #2819 (or perhaps even as part of it)

mortenpi:
> I'd rather have a proper test than an incomplete test whose failures are hidden by default.

I completely agree that a better test is better :smile:  I was just responding, more in general, to

> 1. we could just not bother and drop this test (honestly, I think this would be fine, except that it'd drop test coverage a bit)

fingolfin:
Rebased and implemented (3). The reference output now exists twice: once for Windows, once for everyone else. Only the matching one is attached as a doctest:

```julia
module WindowsPathWarningTests end

const windows_paths_reference_unix = raw"""..."""
const windows_paths_reference_windows = raw"""..."""

@doc(
    Sys.iswindows() ? windows_paths_reference_windows : windows_paths_reference_unix,
    WindowsPathWarningTests,
)
```

Both references are literal strings in the file, so `DOCUMENTER_FIXTESTS` still locates and rewrites the active block. The only difference between them is the warning text: `colons not allowed in paths on Windows` vs `file does not exist`.

fingolfin:
@mortenpi OK to merge?

goerz (approved):
LGTM (genuinely, not just superficially)

PR #2830, opened 2025-11-07
Guide users to use workspaces when setting up Documenter

staticfloat:
Now that Julia v1.12 is out, we can all have the nice things that the Pkg folks have been cooking up for us.  This is an initial draft at guiding users to use `[workspace]` in their `Project.toml`.  I can't help but feel that having a one-shot command to setup Documenter for this would be nicer, but c'est la vie.


docs/src/man/guide.md
```diff
@@ -14,20 +14,24 @@ From the Julia REPL, type `]` to enter the Pkg REPL mode and run
 pkg> add Documenter
 ```
 
-For package documentation, the standard approach is to install Documenter into a documentation-specific project stored in the `docs/` subdirectory of your package.
-To do this, navigate to your package's root folder and do
-
+For package documentation, the standard approach is to add a documentation-specific [workspace](https://pkgdocs.julialang.org/v1/toml-files/#The-[workspace]-section) to your package's `Project.toml` and install Documenter into it.
```
fingolfin: Thank you for working on this, much appreciated!

Alas, I have some questions. You write this is the "standard approach", but I know zero packages that use this (I am sure they exist, just none I interact with do). In fact this is the first time I hear about this feature, and I'll make sure to learn about it. 

But I still find it misleading to call it "standard approach". Perhaps "recommended approach" ?

Also, doesn't this only exist in Julia 1.12? If so, I think that would be kinda important to mention? Or can one use it in 1.10 and 1.11, too?

goerz: Yeah, I feel like 1.12 is just a little too young to make this the only setup that gets mentioned in the documentation (and yes: this is 1.12 only)

mortenpi: I think it would also be good to keep a description of the old approach around, because there is going to be a lot of packages out there that do it the old way, and people will start wondering why. So, basically "here's how we used to do it, don't do it anymore, but it's still good to know that this used to be a thing".

But I do think we should recommend `[workspace]`.

staticfloat:
How does this sound for guiding users appropriately?

docs/src/man/guide.md
```diff
@@ -14,8 +14,30 @@ From the Julia REPL, type `]` to enter the Pkg REPL mode and run
...
+```
+$ julia --project=docs
+(PkgName/docs) pkg> add Documenter PkgName
+```
+
+This will create `Project.toml` file in the `docs/` subdirectory and add `Documenter` and the top-level package as available packages.
+See also the [Pkg.jl](https://github.com/JuliaLang/Pkg.jl/) documentation on working with [project workspaces](https://pkgdocs.julialang.org/v1/toml-files/#The-[workspace]-section) and [environments](https://pkgdocs.julialang.org/v1/environments/).
+
+### Julia v1.11 and earlier
+For Julia versions v1.11 and earlier, creating a separate project in `docs/` is the standard approach:
```
fingolfin:
```suggestion
For Julia versions v1.11 and earlier, creating a separate project in `docs/` is the standard approach.
To do this, navigate to your package's root folder and do
```

fingolfin:
Needs a changelog entry, otherwise seems fine

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+* For Julia v1.12+, encourages users to use the new Pkg `[workspaces]` feature. ([#2830])
+
```
mortenpi: Sorry for the flip-flop here, but I've generally avoided changelog entries for documentation-only changes (since they don't change behavior).

```suggestion
```

PR #2837, opened 2025-11-11
Restore `cd` inside `@eval`, and add a test for it

fingolfin:
Since this fixes a regression that wasn't in any release, I believe there is no need for a changelog entry.

The new test should ensure that this won't be broken again.

### DIFF
--- a/test/warnings/src/at-eval.md
+++ b/test/warnings/src/at-eval.md
@@ -19,3 +19,15 @@ or `Markdown.MD`).
 ```@eval
 "expanded_"*"eval"
 ```
+
+# Test other `@eval` block behavior
+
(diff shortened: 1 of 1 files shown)


test/warnings/src/at-eval.md
```diff
@@ -19,3 +19,15 @@ or `Markdown.MD`).
 ```@eval
 "expanded_"*"eval"
 ```
+
+# Test other `@eval` block behavior
```
fingolfin, line 23: Putting the test into this file is kinda wrong, because it is not testing a warning. But it was the quickest and easiest way for me to write this test. I still intend to merge this with `test/doctests`, though, that will then resolve this "issue"

PR #2847, opened 2025-11-20
Fix spacing in `@docs` blocks; don't expand `details` admonition by default

### LINKED ISSUE #2845: Weird spacing in HTML output `@docs` block
In the last stable version of Documenter.jl there is an "issue" in the generated HTML output. Not sure if it's intended or not.

The spacing in the `@docs` block is weird, there is no whitespace between the paragraph and between the sections (e.g. Arguments, Examples, etc:

Image

### LINKED ISSUE #2846: `details` admonition are expanded by default
In the last stable release of `Documenter.jl`, the details admonitions are expanded by default:

Image

I don't know if it's intended, I feel like it defeats the purpose of the details admonition, since the main reason to use it is to hide something that we think is too long and irrelevant for most cases.

fingolfin:
Fixes #2845 

Would still be good to know whether @aaruni96 remembers why this was added in PR #2773 in the first place -- so that we can check if my fix for one issue doesn't break something else...

CC @franckgaga

UPDATE: also fixes #2846

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Fixed insufficient paragraph spacing in HTML output for `@docs` blocks. ([#2845])
+* Fixed insufficient paragraph spacing in HTML output for `@docs` blocks. ([#2845, #2847])
+* Don't expand `details` admonition by default. ([#2846, #2847])
 
 ## Version [v1.16.0] - 2025-11-14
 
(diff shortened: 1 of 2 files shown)


fingolfin:
@franckgaga OK sorry for first asking you to split your issue into two, and now addressing both of them in one PR 😓 I did not realize how closely related they were after all. (Well, strictly speaking, the fixes are unrelated, but it seemed easiest to just put this into one PR.)

franckgaga:
No worries, thanks for the fix 😄

franckgaga:
image
Way better, thanks a lot @fingolfin !

mortenpi (approved):

PR #2849, opened 2025-11-21
Removed superfluous vertical space at end of `@docs` blocks

fingolfin:
CC @aaruni96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed insufficient paragraph spacing in HTML output for `@docs` blocks. ([#2845, #2847])
 * Don't expand `details` admonition by default. ([#2846, #2847])
-* Removed superfluous vertical space at end of `@docs` blocks.
+* Removed superfluous vertical space at end of `@docs` blocks. ([#2849])
 
 ## Version [v1.16.0] - 2025-11-14
 


src/html/HTMLWriter.jl
```diff
@@ -1937,7 +1937,7 @@ function domify_doc(dctx::DCtx, node::Node)
     # each markdown object. The `DocStr` contains data such as file and line info that
     # we need for generating correct source links.
     return map(zip(node.element.mdasts, node.element.results)) do (markdown, result)
-        ret = div(domify(dctx, markdown))
+        ret = section(div(domify(dctx, markdown)))
```
fingolfin, line 1940: This was the key source of the problem: the `source` link got inserted as last child of the `div` block. But our CSS has rules that add extra bottom margins to all paragraphs, except for the last one, implemented using the `not(:last-child)` CSS selector. But the last paragraph wasn't the last child anymore, the source link was! By reintroducing the `section` around the `div` this problem goes away.

Link this line in the patch, the corresponding CSS changes effectively revert some more of the changes of PR #2773 -- so in a sense the diff to Documenter 1.15.0 gets smaller, which gives me some confidence in the correctness of this fix (well, that and I also of course tested it and it works fine :-) )

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed insufficient paragraph spacing in HTML output for `@docs` blocks. ([#2845, #2847])
 * Don't expand `details` admonition by default. ([#2846, #2847])
+* Removed superfluous vertical space at end of `@docs` blocks.
```
```suggestion
* Removed superfluous vertical space at end of `@docs` blocks. ([#2849])
```

mortenpi (approved):

PR #2854, opened 2025-12-22
Turn off git's message about default branch name in deploydocs.jl

abhro:
Remove the "`hint: Using 'master' as the name for the initial branch. ...`" message from documentation CI logs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+* Added `Remotes.Forgejo` for specifying a `Remote` hosted on a Forgejo instance (such as codeberg.org). ([#2857])
+
 ### Changed
 
 * Git no longer displays a message about the default branch name when calling `deploydocs()`. ([#2854])
--- a/docs/src/lib/remote-links.md
+++ b/docs/src/lib/remote-links.md
@@ -53,7 +53,7 @@ Note that enabling this option can cause documentation builds to fail due to net
 
 !!! note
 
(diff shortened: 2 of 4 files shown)


src/deploydocs.jl
```diff
@@ -379,6 +379,7 @@ function git_push(
     # Generate a closure with common commands for ssh and https
     function git_commands(sshconfig = nothing)
         # Setup git.
+        run(`$(git()) config --global advice.defaultBranchName false`)
```
asinghvi17: This sets a user-level setting, if someone is running `deploydocs` locally or on self-hosted CI we may not want that.

Is there any way this could be made a local setting?  Perhaps running after `git init`?

abhro: Oh yes. I forgot this function isn't part of workflow actions :facepalm:. But wouldn't running `git init` be the thing that generates the message?

src/deploydocs.jl
```diff
@@ -379,7 +379,7 @@ function git_push(
     # Generate a closure with common commands for ssh and https
     function git_commands(sshconfig = nothing)
         # Setup git.
-        run(`$(git()) init`)
+        run(`$(git()) init --initial-branch=master`) # specify initial branch just to turn off the advice in the deploy logs
```
fingolfin: Off topic, but why "master", hasn't the default been "main" for several years now?

abhro: The log messages say that the new branch was called master, so I was just keeping with that. But I'm more than happy to change it to main!

src/deploydocs.jl
abhro: Resolved by manually specifying branch name in 959856d

src/deploydocs.jl
abhro: Hi! Just to confirm, should I change the branch name to main in the `git init` here?

asinghvi17: Yes let's do that

asinghvi17:
```suggestion
        run(`$(git()) init --initial-branch=main`) # specify initial branch just to turn off the advice in the deploy logs
```

PR #2855, opened 2025-12-28
Full-featured Typst Backend

lucifer1004:
Hi @mortenpi and Documenter maintainers,

I'd like to submit this as a new PR following up on #2102 from 2023. The technical landscape has evolved, and I believe this is now ready for consideration.

## **Technical Blockers: Resolved**

The three original issues have been addressed:
1. ✅ **Footnotes**: Now supported natively in Typst
2. ✅ **Julia syntax highlighting**: Official Typst support available
3. ✅ **LaTeX math compatibility**: [mitex](https://typst.app/universe/package/mitex) enables inline LaTeX-to-Typst translation

The mitex integration is particularly important—it means **existing LaTeX math syntax continues to work**, while users can opt into native Typst math with ` typst` blocks. No migration required.

## **What This Solves**

From practical usage:
- **Build speed**: Complete Julia manual (available [here](https://drive.google.com/file/d/1yL746tiQALcRH8IIEtQ_VOiUUnEMkisv/view?usp=sharing)) in **< 90 seconds** (M4 Max) vs. significantly longer LaTeX compilation
- **Source clarity**: Native Typst math syntax is more readable when targeting PDF output:
  - LaTeX math (default, works everywhere)
  `\sum_{i=1}^n \frac{1}{i}`
  - Typst math (optional, Typst backend only)
  `sum_(i=1)^n 1/i`
- **Modern tooling**: Native error reporting, incremental compilation, simpler syntax

## **Implementation Overview**

Current status:
- Full `TypstWriter` module
- Comprehensive test coverage (math, builds, format options)
- Successfully builds Julia manual and Documenter.jl's own documentation
- Dual math support: LaTeX (via mitex) + native Typst
- Multiple compilation modes: native, docker, Typst_jll, or `.typ` output

**No breaking changes**: This adds a new backend option. HTML and LaTeX workflows remain unchanged.

## **Regarding Maintenance Concerns**

I understand the concern from 2023 about maintenance burden. Here's why I think this is different:

**TypstWriter validates the writer API design**. It uses the same internal abstractions as LaTeXWriter, demonstrating that the backend interface works for different output formats. Rather than adding complexity, it proves the existing architecture is sound.

As you mentioned then:
> the APIs [HTML/LaTeX writers] use should indeed all be exposed as public APIs that other output formats could depend on

This implementation shows those APIs in action.

## **Suggested Approach**

I propose:
1. Merge as an **experimental feature** (clearly documented)
2. Keep LaTeX as the default PDF backend
3. Allow community testing in real projects
4. Based on feedback, potentially promote it in a future release

If you'd prefer, I can also:
- Maintain this as a separate plugin package (though main repo testing would help)
- Help document the internal writer APIs
- Gather more community input first

**This isn't about replacing LaTeX**—it's about offering an alternative where it makes sense. Several community members expressed interest in the original thread, and the technical foundation is now solid.

Happy to discuss implementation details or adjust the approach. Thanks for considering this.

cc: @tpapp @asinghvi17 @andreeco @KronosTheLate who expressed interest in this feature.

### DIFF
--- a/test/typstwriter.jl
+++ b/test/typstwriter.jl
@@ -135,112 +135,117 @@ end
 
     @testset "Precise AST Rendering" begin
         @testset "Inline Formatting" begin
+            # Bold
             output = render_to_typst("**bold text**")
-            @test contains(output, "#strong([bold text])")
+            @test strip(output) == "#strong([bold text])"
 
+            # Italic
             output = render_to_typst("*italic text*")
-            @test contains(output, "#emph([italic text])")
+            @test strip(output) == "#emph([italic text])"
 
+            # Inline code
             output = render_to_typst("`inline code`")
-            @test contains(output, "#raw(\"inline code\"")
-            @test contains(output, "block: false")
+            @test strip(output) == "#raw(\"inline code\", block: false)"
 
-            # Combined formatting
+            # Combined formatting - exact structure verification
             output = render_to_typst("**bold** and *italic* and `code`")
-            @test contains(output, "#strong([bold])")
-            @test contains(output, "#emph([italic])")
-            @test contains(output, "#raw(\"code\"")
+            @test strip(output) == "#strong([bold]) and #emph([italic]) and  #raw(\"code\", block: false)"
         end
 
         @testset "Headings" begin
-            output = render_to_typst("# Level 1\n\n## Level 2\n\n### Level 3")
-            @test contains(output, "extended_heading")
-            @test contains(output, "Level 1")
-            @test contains(output, "Level 2")
-            @test contains(output, "Level 3")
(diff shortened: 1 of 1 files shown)


asinghvi17:
Off the top of my head, I'd say this would be much easier to test if it's a separate package, at least initially.

Documenter already has downstream tests for DocumenterVitepress so it should not be hard to add for DocumenterTypst.

Plus it'd be a lot easier to merge fixes and monkeypatches if this is not in the main repo :)

andreeco:
I’ll try out your branch and appreciate the work you did!

lucifer1004:
See https://github.com/lucifer1004/DocumenterTypst.jl

PR #2856, opened 2026-01-08
Add Julia LTS version to CI testing matrix

abhro:


abhro:
Really confused as to why tests on LTS (v1.10) are failing, but 1.6 and 1.12 are fine.

fingolfin:
The tests that fail are testing the output of Documenter and Julia, which can change in different Julia versions. May need a `testfilter`, or perhaps the tests should only be run on a specific Julia version.

fingolfin:
I've add a fix for the CI tests.

mortenpi (approved):

Issue #834, opened 2018-09-17
Missing docstring source links


mortenpi:
They've disappeared somehow in Documenter's deployed `dev/` build..

fredrikekre:
#807

fredrikekre:
Seems to be back. Would be good to understand why this sometimes happens though.

Roger-luo:
any updates on this issue, it seems we got this report as well. https://github.com/QuantumBFS/Yao.jl/issues/198

amontoison:
any news about about this issue? All packages in https://github.com/JuliaSmoothOptimizers have this problem.

fredrikekre:
> All packages in https://github.com/JuliaSmoothOptimizers have this problem.

Probably because they use `Pkg.add` instead of the recommended `Pkg.develop`, e.g. https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/584da46bbd202c67ae3649dc5b0b3e4978e5e71c/.travis.yml#L26

amontoison:
Thanks @fredrikekre, it worked ! We've been looking for a solution for a long time.

fingolfin:
Is there anything actionable here, or can this be closed?

Perhaps adding a "Troubleshooting" or "Common problems", or simply a "Best practices" section which discusses this problem and its caused & fix?

rainerrodrigues:
Hi! I'm interested in diving into this. I see some troubleshooting docs were recently added. Are we currently looking for a code-level fix to better handle macro-generated docstrings (like preserving LineNumberNodes), or is the focus on improving path resolution for Pkg.added dependencies?

PR #2864, opened 2026-01-24
Issue (#834): add troubleshooting section for missing source links

ayushactiveat:
Related Issue #834
added a new Troubleshooting section to the user guide (docs/src/man/guide.md) that explains why documenter needs git metadata to generate links.

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -464,3 +464,26 @@ There's a `sidebar_sitename` keyword option for
 [`Documenter.HTML`](@ref) that lets you hide the sitename
 that's usually displayed below a logo. This is useful if the
 logo already contains the name.
+
+## Troubleshooting
+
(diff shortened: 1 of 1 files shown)


docs/src/man/guide.md
```diff
@@ -464,3 +464,26 @@ There's a `sidebar_sitename` keyword option for
...
+Ensure your Continuous Integration (CI) script or `docs/make.jl` file uses
+`Pkg.develop` instead of `Pkg.add`.
+
+```@example
+# This block is for illustration and is not executed during tests.
+# In a real setup, MyPackage would be your actual package name.
+using Pkg
+Pkg.activate("..")
+# Pkg.develop(PackageSpec(path=pwd())) 
+# using MyPackage
```
fingolfin, line 489: This looks incomplete? At least the closing triple backticks are missing.

PR #2874, opened 2026-02-03
Make Documenter Syntax-versioning aware

Keno:
Default to the syntax version of the module that contains the doctest (so that upgrading julia version doesn't break doctests if they use old syntax) but provide an explicit override so that in principle one could document previous syntax (e.g. to demonstrate how to use something if the user syntax is older than the package syntax). I don't think that'll come up all that often (except for possibly in base), but might as well have the facility.

### DIFF
--- a/src/DocMeta.jl
+++ b/src/DocMeta.jl
@@ -32,7 +32,7 @@ const METATYPE = Dict{Symbol, Any}
 const VALIDMETA = Dict{Symbol, Type}(
     :DocTestSetup => Union{Expr, Symbol},
     :DocTestTeardown => Union{Expr, Symbol},
-    :DocTestSyntax => VersionNumber
+    :DocTestSyntax => VersionNumber,
 )
 
 """
--- a/src/doctests.jl
+++ b/src/doctests.jl
@@ -8,8 +8,8 @@ struct DocTestContext
     file::String
     doc::Documenter.Document
     meta::Dict{Symbol, Any}
-    mod::Union{Module, Nothing}  # Source module for parser selection
-    DocTestContext(file::String, doc::Documenter.Document, mod::Union{Module, Nothing}=nothing) = new(file, doc, copy(doc.user.meta), mod)
+    mod::Union{Module, Nothing}
+    DocTestContext(file::String, doc::Documenter.Document, mod::Union{Module, Nothing} = nothing) = new(file, doc, copy(doc.user.meta), mod)
 end
 
 """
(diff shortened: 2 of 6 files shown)


src/DocMeta.jl
```diff
@@ -30,7 +31,8 @@ const METATYPE = Dict{Symbol, Any}
 "Dictionary of all valid metadata keys and their types."
 const VALIDMETA = Dict{Symbol, Type}(
     :DocTestSetup => Union{Expr, Symbol},
-    :DocTestTeardown => Union{Expr, Symbol}
+    :DocTestTeardown => Union{Expr, Symbol},
+    :DocTestSyntax => VersionNumber
```
fingolfin:
```suggestion
    :DocTestSyntax => VersionNumber,
```

src/doctests.jl
```diff
@@ -8,7 +8,8 @@ struct DocTestContext
     file::String
     doc::Documenter.Document
     meta::Dict{Symbol, Any}
-    DocTestContext(file::String, doc::Documenter.Document) = new(file, doc, copy(doc.user.meta))
+    mod::Union{Module, Nothing}  # Source module for parser selection
+    DocTestContext(file::String, doc::Documenter.Document, mod::Union{Module, Nothing}=nothing) = new(file, doc, copy(doc.user.meta), mod)
```
fingolfin:
```suggestion
    DocTestContext(file::String, doc::Documenter.Document, mod::Union{Module, Nothing}=nothing) = new(file, doc, copy(doc.user.meta), mod)
```
Runic doesn't like that. Just run it once over all files?

test/doctests/src/syntax_versioning.md
```diff
@@ -0,0 +1,28 @@
...
+               end
+           end
+           0
+       end
+40
+```
+
+## Verifying syntax version with @VERSION
+
+```jldoctest; syntax = v"1.14"
```
fingolfin, line 23: Maybe also have a "negative" test which sets the syntax version to e.g. 1.13 and then verifies that worked (e.g. by using a 1.14 feature which results in an error) ?

fingolfin (commented):
Thanks that looks pretty good. Some minor nitpicks in comments. Also, this needs a changelog entry.

Keno:
@fingolfin Good to go?

Keno:
Gonna merge this since it's blocking base PRs. If there is further changes, we can do them in a separate PR.

PR #2875, opened 2026-02-03
remove .includes and add a set to make it O(n)

### LINKED ISSUE #2872: Regression: search gets unusably slow for large search index
This is a regression introduced by PR #2744 (CC @Rahban1) which caused the [Plots.jl documentation search to break](https://github.com/JuliaPlots/Plots.jl/issues/5219) (well... it does work if you are willing to wait 30-40 seconds for your first search.

The problem is this code in `assets/html/js/search.js` (and repeated in `test/search/wrapper.js`):

```js
      // apply patterns in order of specificity
      for (const pattern of patterns) {
        pattern.lastIndex = 0; //reset regex state
        let match;
        while ((match = pattern.exec(remaining)) != null) {
          const token = match[0].trim();
          if (token && !tokens.includes(token)) {
            tokens.push(token);
          }
        }
      }

      // splitting the content if something remains
      const basicTokens = remaining
        .split(/[\s\-,;()[\]{}]+/)
        .filter((t) => t.trim());
      for (const token of basicTokens) {
        if (token && !tokens.includes(token)) {
          tokens.push(token);
        }
      }

      return tokens.filter((token) => token.length > 0);
```

Note the `tokens.includes(token)` bit: this performs a *linear* search. As a result, this code become $O(n^2)$ in the number of tokens, when it could (and should) be linear.

A simple fix should be to replace `tokens` by a hashmap/hashset/dictionary/set/whatever this is called in JavaScript.
(In the end this may the have to be turned again into an array/vector/list -- I don't know the minisearch API well enough to know. If the order of the tokens is important, another option would be to leave `tokens` as it is, but add a second `tokens_set` which uses a hashtable; then the `includes` test is performed using `tokens_set`, and new entries are pushed into both `tokens` and `tokens_set`)
If the order of th

Rahban1:
I apologize for the slowdown in search, I should have thought about this O(n^2) time complexity earlier and should have optimized it. 
This closes #2872

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* reduced time complexity from O(n^2) to O(n) to improve the initial load time for search ([#2875])
 * Git no longer displays a message about the default branch name when calling `deploydocs()`. ([#2854])
 * Don't escape characters (such as ~) in URLs when writing LaTeX. ([#2210])
 


Rahban1:
don't think it needs a changelog since it's an optimization and not a user facing change

fingolfin:
Thanks a lot @Rahban1. Don't worry, to err is human, and e.g. I didn't catch this in review either -- and it works fine for e.g. the Julia website.

Regarding the changelog, I'd still like to have an entry so that people wondering "should I bother to use this" will know that this issue was there and has been fixed.

Rahban1:
I guess this is ready to merge from my side

Issue #1631, opened 2021-07-06
Doctests not catching exceptions from show


mortenpi:
This type of doctest, where `show` throws an error, does not work currently:

```
julia> gd
Error showing value of type GroupedDataFrame{DataFrame}:
ERROR: AssertionError: The current number of rows in the parent data frame is 3 and it does not match the number of rows it contained when GroupedDataFrame was created which was 2. The number of rows in the parent data frame has likely been changed unintentionally (e.g. using subset!, filter!, delete!, push!, or append! functions).
```

I believe it's because Documenter does not handle exceptions from the `show` calls which happen here: https://github.com/JuliaDocs/Documenter.jl/blob/6c8679bac445ab1c656744075b13a57835081cc3/src/DocTests.jl#L319

X-ref https://github.com/JuliaData/DataFrames.jl/pull/2811

PR #2876, opened 2026-02-10
Add error handling for result display

asinghvi17:
Handle errors that happen when a result gets displayed in an example block (via `display_dict`).

This often crops up when trying to latexify things, and the idea is that it should cause an error but not crash the doc build.  I'm not sure how you are supposed to log the error such that `warnonly` specifiers will catch it, though, so some help there would be great.

### DIFF
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -905,7 +905,12 @@ function Selectors.runner(::Type{Expanders.ExampleBlocks}, node, page, doc)
     input = droplines(x.code)
 
     # Generate different  in different formats and let each writer select
-    output = Base.invokelatest(Documenter.display_dict, result, context = :color => ansicolor)
+   output = try
+        Base.invokelatest(Documenter.display_dict, result, context = :color => ansicolor)
+    catch err
+        @error "Problem displaying result on page $(page.source): $(err)"
+        Dict(MIME"text/plain"(), "Error displaying result")
+    end
     # Remove references to gensym'd module from text/plain
     m = MIME"text/plain"()
     if haskey(output, m)
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -905,7 +905,7 @@ function Selectors.runner(::Type{Expanders.ExampleBlocks}, node, page, doc)
     input = droplines(x.code)
 
     # Generate different  in different formats and let each writer select
-   output = try
+    output = try
         Base.invokelatest(Documenter.display_dict, result, context = :color => ansicolor)
     catch err
         @error "Problem displaying result on page $(page.source): $(err)"
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -909,7 +909,7 @@ function Selectors.runner(::Type{Expanders.ExampleBlocks}, node, page, doc)
         Base.invokelatest(Documenter.display_dict, result, context = :color => ansicolor)
     catch err
         @error "Problem displaying result on page $(page.source): $(err)"
-        Dict(MIME"text/plain"(), "Error displaying result")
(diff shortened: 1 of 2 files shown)


asinghvi17:
Looks like CI errors were only on nightly (maybe due to Markdown stdlib changes)

fingolfin (dismissed):

fingolfin:
Do you have any example that reproduces the error, which could be used as basis for a test case?

fingolfin:
This seems very similar to issue #1631

mortenpi:
I feel that the right place for the error handling would actually be inside the `for` loop in `display_dict`. Right now, this catch-all `catch` is a bit too broad.

And I guess the second question is whether we want the Documenter build succeed still if we hit this or fail? I'd always lean towards the latter, even if it is due to a buggy `show` method. But then we should probably push this into the usual error list and potentially allow the user to opt out of crashing with `warnonly`?

src/expander_pipeline.jl
```diff
@@ -905,7 +905,12 @@ function Selectors.runner(::Type{Expanders.ExampleBlocks}, node, page, doc)
     input = droplines(x.code)
 
     # Generate different  in different formats and let each writer select
-    output = Base.invokelatest(Documenter.display_dict, result, context = :color => ansicolor)
+    output = try
+        Base.invokelatest(Documenter.display_dict, result, context = :color => ansicolor)
+    catch err
+        @error "Problem displaying result on page $(page.source): $(err)"
+        Dict(MIME"text/plain"(), "Error displaying result")
```
asinghvi17:
```suggestion
        Dict(MIME"text/plain"() => "Error displaying result")
```

fingolfin:
Sorry for the slow follow-up, and sorry that my earlier approval was premature, so I've dismissed it. I finally sat down with a reproducer, and it turns out the patch does not do what it says.

Here is the reproducer, for the `@example` case (the Latexify-shaped one, where `text/plain` is fine and only `text/latex` throws):

```julia
struct BadLatex end
Base.showable(::MIME"text/latex", ::BadLatex) = true
Base.show(::IO, ::MIME"text/latex", ::BadLatex) = error("boom latex")
BadLatex()
```

On `master` this aborts the build with `ErrorException("boom latex")`, even under `warnonly = true`. So the bug is real.

On this branch it aborts too, just later and with a different message:

```
ERROR: MethodError: no method matching domify(::Documenter.HTMLWriter.DCtx, ::MarkdownAST.Node{Nothing}, ::Dict{MIME{Symbol("text/plain")}, String})
```

`Dict(MIME"text/plain"() => "Error displaying result")` has type `Dict{MIME{Symbol("text/plain")}, String}`, but the writers dispatch on the invariant `Dict{MIME, Any}` (`HTMLWriter.jl:2568`, `LaTeXWriter.jl:494`), so the recovery dict never dispatches. It needs the explicit `Dict{MIME, Any}(...)` that the other construction sites in `expander_pipeline.jl` use.

Three more things while I was in there:

- The bare `@error` does not push to `doc.internal.errors`, so `makedocs` still reports success and `warnonly` has no effect. That is @mortenpi's point. Using `@docerror(doc, :example_block, ...)` gives both, and no new entry in `ERROR_NAMES` is needed.
- The `catch` also swallows `InterruptException`, which the surrounding code deliberately passes to `IOCapture.capture` as `rethrow = InterruptException`. The backtrace is dropped too.
- Also @mortenpi's point about granularity: one throwing `text/latex` currently discards a perfectly good `text/plain` and replaces the whole output with `"Error displaying result"`.

I have a rework ready that addresses all of these, and I'd rather hand it to you than open a competing PR: happy to push it onto this branch if you're fine with that. But of course if you prefer, I can also open a new PR.

PR #2880, opened 2026-02-15
Preserve page anchor when moving between versions

### LINKED ISSUE #2799: Feature request: stay on page when changing version

langestefan:
Closes #2799

Now we:
- Stay on the right page *and* the right position when switching versions.
- This now also works when navigating via the outdated version warning banner.

*Created with the help of Claude. *

Video to demonstrate that it works:
[Screencast_20260215_024650.webm](https://github.com/user-attachments/assets/be4b4805-c42b-48dc-8ffa-eb12ed978120)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* The version selector now also preserves the anchor (hash) when switching between documentation versions. Additionally, the outdated/dev version warning banner now also tries to keep you on the same page (and position) when linking to the latest stable release. ([#2880])
 * Added `Remotes.Forgejo` for specifying a `Remote` hosted on a Forgejo instance (such as codeberg.org). ([#2857])
 * Doctests now default to the `parser_for_module` of the module that the docstring appears in, allowing modules that set their syntax version via `Base.Experimental.@set_syntax_version` to have their doctests parsed with the correct syntax automatically. Also added support for `DocTestSyntax` metadata and per-block `syntax=` attributes to explicitly specify a syntax version. ([#2874])
 


assets/html/warner.js
```diff
@@ -39,6 +39,32 @@ function maybeAddWarning() {
...
+    }
+    if (current_page.startsWith(base_url_absolute)) {
+      page_path = current_page.substring(base_url_absolute.length);
+    }
+    if (page_path && page_path !== "" && page_path !== "index.html") {
+      var stable_base = href;
+      if (stable_base.endsWith("/")) {
+        stable_base = stable_base.slice(0, -1);
+      }
+      target_url = stable_base + "/" + page_path;
```
fingolfin: This removes a slash just to add it back again. How about
```suggestion
      if (!stable_base.endsWith("/")) {
        stable_base = stable_base + "/";
      }
      target_url = stable_base + page_path;
```

assets/html/js/versions.js
```diff
@@ -54,11 +54,14 @@ $(document).ready(function () {
       target_url = target_url + "/" + page_path;
```
fingolfin: I guess the same comment applies here as to the new code

assets/html/warner.js
```diff
@@ -51,12 +77,32 @@ function maybeAddWarning() {
...
+      fetch(target_url, { method: "HEAD" })
+        .then(function (response) {
+          if (response.ok) {
+            window.location.href = target_url + current_hash;
+          } else {
+            window.location.href = href;
+          }
+        })
+        .catch(function () {
+          window.location.href = href;
```
fingolfin: Could we try to keep this code as similar to the code in `versions.js` as possible, to ease future adjustments? To that end, there are two comments in selected range corresponding to the version of the code in `versions.js`, which could be replicated here perhaps?

And similar for the code which computes `base_url_absolute`.

Also, how about renaming `href` to `target_href` -- that will make the diff slightly larger but really just by one extra line, but will further reduce difference between the two sets of code.

fingolfin (approved):
Fantastic, thanks for working on this, the changes look sane and you confirmed they work.

We do need a changelog entry, though!

Also, some nitpick remarks. Those are not blockers for merging but perhaps you are willing to consider them.

langestefan:
thanks @fingolfin ! I think the comments have been addressed, let me know. I also added a changelog entry.

PR #2883, opened 2026-02-15
Fix operator binding rendering for doc headers

### LINKED ISSUE #2844: Docstrings for `Base.:(:)` is shown as `Base.::`
The docstring for the `:` range operator is shown in the Julia manual as `Base.::` (see [here](https://docs.julialang.org/en/v1.14-dev/base/math/#Base.::-Tuple{Any,%20Any,%20Any})) which is confusing in multiple ways: a reader might think this is `::`, and it also is syntactically invalid.

It should instead be shown as `Base.:(:)`.

fingolfin:
Normalize binding display to canonical Julia syntax for operator bindings by using show(Symbol) formatting, and use it in Object printing plus HTML/LaTeX doc rendering paths. Add regression tests for :, ==, +, and non-operator bindings.

Fixes #2844

Created by Codex

---

Question is: Do we want this? I've also submitted https://github.com/JuliaLang/julia/pull/61043 to fix the printing of Bindings in Julia, and assuming this is approved and backported, it would also take care of the issue.

This PR is more like a measure to make things good in the meantime. 

It also was another good training exercise for using Codex ;-). So I did not invest much into this, and won't be upset if we end up not merging it.

### DIFF
--- a/src/utilities/utilities.jl
+++ b/src/utilities/utilities.jl
@@ -128,7 +128,7 @@ slugify(object) = string(object) # Non-string slugifying doesn't do anything.
 # has been fixed in <https://github.com/JuliaLang/julia/pull/61043> but until
 # this fix is available in Julia versions, we replace calls to `string` on
 # bindings by `Documenter.bindingstring`. In the future, we may be able to
-# get rid of this again. 
+# get rid of this again.
 function bindingstring(binding::Binding)
     if VERSION < v"1.14.0-DEV.1731"
         s = string(binding)


mortenpi:
This seems fine to me actually. I don't mind if we add the `bindingstring` abstraction, rather than relying on `string` directly.

Implementation-wise, we could have a version check in there (once the upstream PR is merged), so that it would use Base.Docs directly still, for Julia versions where it is fixed?

fingolfin:
Since I just merged https://github.com/JuliaLang/julia/pull/61043, we can argue whether we still need this. I'd say yes, because people will still be building their manuals with Julia versions older than 1.14 for some time. Of course that PR might get backported... But still, that will take time. With this PR, we can get a fix into the hands of our users now. If it becomes unnecessary one day, we still remove it. (For that, it'd be good to add a comment why it is necessary in the first place. On it.)

src/utilities/utilities.jl
```diff
@@ -122,6 +122,13 @@ function slugify(s::AbstractString)
 end
 slugify(object) = string(object) # Non-string slugifying doesn't do anything.
 
+# Render doc bindings with syntax that can be copy-pasted as valid Julia code.
```
fingolfin, line 125:
```suggestion
# Render doc bindings with syntax that can be copy-pasted as valid Julia code.
# Workaround for <https://github.com/JuliaDocs/Documenter.jl/issues/2844>: in
# the Julia documentation, `Base.:(:)` was being displayed as `Base.::`. This
# has been fixed in <https://github.com/JuliaLang/julia/pull/61043> but until
# this fix is available in Julia versions, we replace calls to `string` on
# bindings by `Documenter.bindingstring`. In the future, we may be able to
# get rid of this again. 
```

src/utilities/utilities.jl
```diff
@@ -122,6 +122,13 @@ function slugify(s::AbstractString)
...
+function bindingstring(binding::Binding)
+    s = string(binding)
+    op_prefix = string(binding.mod, ".:")
+    return startswith(s, op_prefix) ? string(binding.mod, ".", sprint(show, binding.var)) : s
```
mortenpi: I think the comment is good too, but FWIW, I meant an proper version guard in code that we can then drop once we drop Julia < 1.14, for the https://github.com/JuliaLang/julia/commit/d42abc644ba4773fd48c383204f8fc1f53f55b59 commit:

```suggestion
    if VERSION < v"1.14.0-DEV.1731"
        s = string(binding)
        op_prefix = string(binding.mod, ".:")
        return startswith(s, op_prefix) ? string(binding.mod, ".", sprint(show, binding.var)) : s
    else
        return string(binding)
    end
```

PR #2887, opened 2026-02-16
Support more MarkdownAST nodes

fingolfin:
In Julia 1.14, the Markdown stdlib has learned some new tricks: support for strikethrough, HTML blocks and inline HTML. And MarkdownAST 0.1.3 has learned to handle them. Make sure we can handle all of them in Documenter, too.

This PR deliberately does *not* update `docs/src/showcase.md` nor anything in `test/examples/` with the new features. Here is my rationale:
1. the new features require Julia nightly, so it should not (yet!) be in the "Showcase" section in the manual
2. however, we really should support these new features in a Documenter release *before* Julia 1.14 is released, otherwise it may be painful for some package authors...
3. ... and in fact a tiny part of the Julia manual would otherwise be broken in an ugly fashion, as it [uses inline HTML](https://docs.julialang.org/en/v1.14-dev/manual/noteworthy-differences/#Julia-C/C:-Quick-reference) which right now is rendered as `<br>` there, but would be rendered worse (something like `MarkdownAST.HTMLInline("<br>")` or so. (Of course it is also [questionable that `<br>` is used there are all](https://github.com/JuliaLang/julia/issues/49991)).
4. ... and this is also why my PR <https://github.com/JuliaLang/julia/pull/60732> has not yet been merged -- I was waiting for a Documenter release with support first...

So this is a bit of a chicken-and-egg situation. To get out of this, I (or rather, Codex) made this initial PR. Once everything has landed in Julia nightly, I can follow up with a PR that adds a few more tests that will only be run in Julia >= 1.14. And once Julia 1.14 is out, we can think about including examples in the "Showcase" section.

---

By the way, there is also one design issue I am ignoring right now: what to do with HTML blocks and inline HTML when rendering to LaTeX? Some options:
1. Right now, we print the HTML with some escaping.
2. But perhaps it should instead be put into a `verbatim` environment
3. Or an error message should be printed
4. or maybe just nothing... ?

But I think settling this is not urgent, as almost nothing is using these yet, so we can revise this at a later date.

Co-authored-by: Codex <codex@openai.com>

### DIFF
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -767,13 +767,21 @@ function latex(io::Context, node::Node, ::MarkdownAST.Strikethrough)
     return
 end
 
+function _warn_raw_html_in_latex(io::Context, kind::AbstractString)
+    source = isempty(io.filename) ? "(unknown source)" : Documenter.locrepr(io.filename)
+    @warn "Raw HTML $kind is not supported in LaTeX output in $(source); stripping tags."
+    return
+end
+
 function latex(io::Context, ::Node, html::MarkdownAST.HTMLBlock)
+    _warn_raw_html_in_latex(io, "block")
     latexesc(io, replace(html.html, r"<[^>]+>" => ""))
     _println(io)
     return
 end
 
 function latex(io::Context, ::Node, html::MarkdownAST.HTMLInline)
+    _warn_raw_html_in_latex(io, "inline")
     latexesc(io, replace(html.html, r"<[^>]+>" => ""))
     return
 end
--- a/test/latexwriter.jl
+++ b/test/latexwriter.jl
@@ -93,10 +93,10 @@ end
     @test _node_to_latex(strikethrough) == "gone"
 
     htmlinline = Documenter.MarkdownAST.Node(Documenter.MarkdownAST.HTMLInline("<span>x</span>"))
-    @test _node_to_latex(htmlinline) == "x"
+    @test_logs (:warn, r"Raw HTML inline is not supported in LaTeX output") _node_to_latex(htmlinline) == "x"
 
     htmlblock = Documenter.MarkdownAST.Node(Documenter.MarkdownAST.HTMLBlock("<div>y</div>"))
-    @test _node_to_latex(htmlblock) == "y\n"
+    @test_logs (:warn, r"Raw HTML block is not supported in LaTeX output") _node_to_latex(htmlblock) == "y\n"
 end
 
 @testset "latex table link fragment PDF regression reproducer" begin


mortenpi:
> 1. the new features require Julia nightly, so it should not (yet!) be in the "Showcase" section in the manual

We could also just add a compat node, that this requires Julia 1.14.

> By the way, there is also one design issue I am ignoring right now: what to do with HTML blocks and inline HTML when rendering to LaTeX?

A similar, but distinct issue is with at-raw blocks. In that case, we just include them just for the appropriate writer (i.e. `@raw html` is in HTML, `@raw latex` is for LaTeX/PDF etc.). Arguably, we could treat the new nodes as `@raw html`, but it's probably not the right way.

A safe option would maybe be to error for now though -- we can then relax it later. If you are "cross-compiling" for HTML and PDF, I think you want to make sure you avoid random HTML popping up in your output (or things being just outright missing).

So.. maybe let's error in LaTeX for now actually?

src/html/HTMLWriter.jl
```diff
@@ -2299,6 +2299,10 @@ domify(dctx::DCtx, node::Node, ::MarkdownAST.BlockQuote) = DOM.Tag(:blockquote)(
 
 domify(dctx::DCtx, node::Node, ::MarkdownAST.Strong) = DOM.Tag(:strong)(domify(dctx, node.children))
 
+domify(dctx::DCtx, node::Node, ::MarkdownAST.Emph) = DOM.Tag(:em)(domify(dctx, node.children))
+
+domify(dctx::DCtx, node::Node, ::MarkdownAST.Strikethrough) = DOM.Tag(:del)(domify(dctx, node.children))
```
mortenpi, line 2304: Arguably, this could be a `<s>`, not a `<del>`, based on the semantics, though it's not clear-cut. I feel that in docs, the strikethrough would be more a stylistic/typesetting choice, rather than indicating an edit.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/del
- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/s

But, for what it's worth, the CommonMark discussions do seem to default to `<del>`, and so does GitHub (e.g. if you inspect this: ~foo~):

- https://talk.commonmark.org/t/feature-request-strikethrough-text/220/7
- https://talk.commonmark.org/t/strikeout-threw-out-strikethrough-strikes-out-throughout/820/22

This is something we can always change later too though if we want.

```suggestion
domify(dctx::DCtx, node::Node, ::MarkdownAST.Strikethrough) = DOM.Tag(:s)(domify(dctx, node.children))
```

fingolfin: In my view, `del` corresponds to `em` and `strong` (those are "semantic"),  while `s` corresponds to `i` and `b` (purely formatting). Since we use `em` and `strong` already, it seemed natural to use `del`. I guess that's also why Codex suggested `del`.

If you don't mind, I'd leave it as it is for now. That said, I don't feel strongly about this. But as you say, we can revisit it in the future if we have reasons to change it.

fingolfin:
> So.. maybe let's error in LaTeX for now actually?

But then we'd break the Julia manual, which already uses `<br>` as I explained (see <https://github.com/JuliaLang/julia/issues/49991>). Of course that is a "bug" there, but surely we should make sure we don't break one of our biggest "customer".

Arguably, the use of `<br>` is a "bug". Alas, one that I don't know how to "fix": Ideally we'd insert a Markdown "hard break", by ending a line in 

So change this (excerpt of the offending table)
```md
| Software Concept   | Julia | C/C++ |
| :---               | :---  | :---  |
| software module    | A Julia "package"        | `.h`/`.hpp` files<br>+compiled `somelib.a`   |

```
to maybe this
```md
| Software Concept   | Julia | C/C++ |
| :---               | :---  | :---  |
| software module    | A Julia "package"        | `.h`/`.hpp` files\
            +compiled `somelib.a`   |
```
Alas this does not work... old table:

| Software Concept   | Julia | C/C++ |
| :---               | :---  | :---  |
| software module    | A Julia "package"        | `.h`/`.hpp` files<br>+compiled `somelib.a`   |

new table:

| Software Concept   | Julia | C/C++ |
| :---               | :---  | :---  |
| software module    | A Julia "package"        | `.h`/`.hpp` files\
            +compiled `somelib.a`   |


It doesn't work in GitHub and it doesn't work in Julia: I think this is one of the major shortcomings of the "table extension" and my guess is that this is one of the main reasons it is not part of the CommonMark standard: it doesn't compose well with "blocks".


I guess we could *special case* this: I'd bet that this usecase (inserting line breaks) is quite common, and replacing `<br>` or `<br/>` by line breaks would be easy enough (though perhaps not Documenter should do it, but rather Markdown.jl, when in "Julia mode"...

mortenpi:
> But then we'd break the Julia manual, which already uses `<br>` as I explained (see [JuliaLang/julia#49991](https://github.com/JuliaLang/julia/issues/49991)). Of course that is a "bug" there, but surely we should make sure we don't break one of our biggest "customer".

That's fair. Maybe let's just warn then? Ideally we should also have an option to opt-in to new errors I think, but I am not sure we have an existing setting for that..?

fingolfin:
> That's fair. Maybe let's just warn then?

Yeah, a warning seems reasonable.

Of course that would mean building the Julia manual PDF would trigger that warning. Alas with https://github.com/JuliaLang/julia/pull/61088 the issue <https://github.com/JuliaLang/julia/issues/49991> would be resolved, so that is taken care of if we merge that PR.

> Ideally we should also have an option to opt-in to new errors I think, but I am not sure we have an existing setting for that..?

I don't think so. If we had, we could have used it instead of adding the `treat_markdown_warnings_as_error` option. I'll open an issue.

PR #2866, opened 2026-01-27
Add top_menu feature for multi-section documentation

lmiq:
Disclaimer: This feature was **heavily** implemented using Claude Opus 4.5. 

It was a feature that I was willing to have for a long time, but had no skills to implement. Seems to be working properly, and might be useful to many people for further organization of the documentation. 

## Add `top_menu` feature for multi-section documentation

Example:

image

image

### Summary

This PR adds a new `top_menu` keyword argument to `makedocs` that enables upper-level organization of documentation with a horizontal navigation bar at the top of the page. This is useful for large documentation projects that need categorization above the sidebar level (e.g., separating "Getting Started", "Guides", "API Reference" into distinct sections).

### Usage Example

```julia
makedocs(
    top_menu = [
        "Getting Started" => [
            "Introduction" => "getting-started/index.md",
            "Installation" => "getting-started/install.md",
        ],
        "User Guide" => [
            "Basics" => "guide/index.md", 
            "Advanced" => "guide/advanced.md",
        ],
    ],
    # pages keyword is omitted when using top_menu
)
```

### Features

- **Section-based navigation**: Each section maintains its own navigation tree in the sidebar
- **Theme-aware styling**: Uses sidebar theme variables for consistent appearance across all themes
- **Responsive design**: Mobile-friendly with proper hamburger menu integration
- **Active section indicator**: Bottom border highlights the current section

### Files Changed

| File | Change |
|------|--------|
| `src/documents.jl` | Added `TopMenuSection` struct |
| `src/builder_pipeline.jl` | Extended `walk_navpages` for section processing |
| `src/html/HTMLWriter.jl` | Added rendering functions |
| `src/makedocs.jl` | Added keyword documentation |
| `assets/html/scss/documenter/layout/_topmenu.scss` | New SCSS styles |
| `docs/src/man/guide.md` | User documentation |
| `test/topmenu.jl` | Unit tests |

### Behavior Notes

- The first section in `top_menu` determines the landing page
- Pages should be unique across sections (a warning is emitted for duplicates)
- Fully backward compatible - existing docs without `top_menu` are unchanged

### Checklist

- [x] Core implementation
- [x] CSS styling with theme support
- [x] Unit tests
- [x] Example build
- [x] Documentation in docstrings
- [x] User guide documentation

PR #2904, opened 2026-03-09
Top menu from pages

lmiq:
This is a variation of #2866 where instead of adding a `top_menu` vector that replaces the `pages` vector, a 

`top_menu=true` 

option is added to `makedocs`, which will interpret the first layer of the documentation menu as the top menu. 

if `top_menu=false` the common (current - no top-menu) behavior is preserved. 

This is a nice alternative for the interface, indeed.

@mortenpi


lmiq:
INFO:

I generates this page correctly (local test here), with and without the `top_menu = true` option:

https://m3g.github.io/ComplexMixtures.jl/stable/

And the PDF output is generated equally with both options.

lmiq:
This is another page using the feature:

https://m3g.github.io/PDBTools.jl/dev/

Where the configuration was just adding a line to `makedocs`:

https://github.com/m3g/PDBTools.jl/blob/7fcdceb7f789e7e092e133e0667839b406a5d57b/docs/make.jl#L12

src/makedocs.jl
```diff
@@ -13,6 +13,7 @@
         sitename = "",
         expandfirst = [],
         draft = false,
+        top_menu = false,
```
mortenpi: I would like this to be an option to `Documenter.HTML`. This way we can also avoid having to pass this through to everything.

src/builder_pipeline.jl
```diff
@@ -127,22 +127,98 @@ function Selectors.runner(::Type{Builder.SetupBuildDirectory}, doc::Documenter.D
...
     userpages = isempty(doc.user.pages) ? sort(mdpages, lt = lt_page) : doc.user.pages
 
-    # Populating the .navtree and .navlist.
-    # We need the for loop because we can't assign to the fields of the immutable
-    # doc.internal.
-    for navnode in walk_navpages(userpages, nothing, doc)
-        push!(doc.internal.navtree, navnode)
-    end
+    # Check if top_menu is being used
+    if !isempty(doc.user.top_menu)
```
mortenpi: Related to the argument, but we really shouldn't have anything here in the builder pipeline. The top menu should be purely a presentation (i.e. a writer) concern.

src/html/HTMLWriter.jl
```diff
@@ -939,6 +939,31 @@ function render(doc::Documenter.Document, settings::HTML = HTML())
...
+                        <title>Redirecting...</title>
+                    </head>
+                    <body>
+                        <p>Redirecting to <a href="$(first_page_path)">$(first_section.title)</a>...</p>
+                    </body>
+                    </html>"""
+                )
+            end
+        end
+    end
```
mortenpi: I'm not sure why we need these redirects? I'd keep the structure of the HTML files unchanged, just change the HTML on each page to include the top bar.

src/html/HTMLWriter.jl
```diff
@@ -995,28 +1020,157 @@ function render_page(ctx, navnode)
...
+    # Fallback to global navtree
+    return ctx.doc.internal.navtree
+end
+
 """
 Renders the main `<html>` tag.
 """
-function render_html(ctx, head, sidebar, navbar, article, footer, extras)
+function render_html(ctx, head, sidebar, navbar, article, footer, extras; top_menu = nothing)
```
mortenpi: The `top_menu` value should be accessed via `ctx`.

mortenpi (commented):
First, sorry for not getting back to this earlier. But I think this would need a bit of iteration. Right now, I couldn't actually get it to work. But I'd be happy to help shepherd this through.

I do also have few concerns around the high-level choices, so I left some comments for those. I did not dig into the front-end part yet, since I'd like to see it build first.

src/makedocs.jl
lmiq: top_menu is now a keyword argument of HTML (e.g. format = HTML(top_menu = true)) and has been removed from makedocs. The documentation and the example in test/examples/make.jl have been updated accordingly.

src/builder_pipeline.jl
lmiq: the builder pipeline now always follows the original code path, with no awareness of top_menu. TopMenuSection was moved to HTMLWriter.jl, and a build_top_menu_sections function derives the per-section navtrees and navlists directly from the already-built global doc.internal.navtree at render time. The navlist keyword argument that had been added to walk_navpages was also removed.

src/html/HTMLWriter.jl
lmiq: The redirect generation has been revised rather than removed entirely. When top_menu = true and no root index.md exists, a root index.html is now generated that redirects to the first index.md found anywhere in the pages tree. The file structure of all other pages remains unchanged — the top bar is still purely a per-page HTML addition.

This was not completely removed for cases like:

```
pages = [ "English" => [ "en/index.md" ], "Spanish" => "es/index.md" ]
```

where it not natural to have a single `index.md` page in the tree. In these cases the first one is pick for redirection.

src/html/HTMLWriter.jl
lmiq: the top_menu keyword argument has been removed from render_html. The function now takes navnode as a positional argument and calls render_top_menu(ctx, navnode) internally, so the top menu is derived from ctx rather than passed in from the call site.

ps: This one is completely vibe-coded. I don't know what ctx is.

lmiq:
Incorporated the changes suggested above.

I'm still testing and fixing bugs.

lmiq:
Seems to be working:

Test pages:

https://m3g.github.io/FundamentosDMC.jl/dev
(does not have a root index.md - picks the first one as landing page, redirecting).

https://m3g.github.io/PDBTools.jl/dev/
(this one has a single index.md at the top level)

https://m3g.github.io/ComplexMixtures.jl/dev/
(this one has a single index.md file - but not at the top level)

When there is no index.md file at the top level, we are redirecting. The issue with avoiding the redirection is that relative paths might break if we move to the top level the index.md file - particularly those within raw html blocks. 

I think that *if* there is no root/index.md file, just creating one that redirects to the first "index.md" file found in the pages tree is a good alternative. If there is a root/index.md file, then no redirection occurs.

fingolfin:
Claude AI and me just resolved the merge conflicts (by running `make themes`), and had a look. Here are the unfiltered review remarks made by Claude:

---

Now the review. I built the example docs locally and poked at a few edge cases; findings below, roughly in order of severity.

### Blockers

**1. The docs describe the wrong API.** `top_menu` is an option of `HTML()`, but `docs/src/man/guide.md` and the CHANGELOG entry both call it a `makedocs` keyword, and the guide's example passes `top_menu = true` straight to `makedocs` — which won't work. The `HTML` docstring gets it right.

Related design question: should it be on `HTML` at all? `top_menu` reinterprets the meaning of `pages`, which is format-agnostic, so the LaTeX writer will silently ignore it and flatten everything. Putting it on `makedocs` would at least make the scope explicit.

**2. The new tests don't run.**

- `test/examples/tests.jl` uses `with_logger`, `ConsoleLogger` and `Logging.Warn`, but that file never imports `Logging` → `UndefVarError` as soon as the testset is reached.
- `@test_nowarn makedocs_with_topmenu(["Empty Section" => []])` doesn't just warn, it throws: `BoundsError` from `first(ctx.doc.internal.navlist)` in `render_sidebar` (HTMLWriter.jl:1496). This one reproduces on `master` without `top_menu` too, so it's a pre-existing bug rather than a regression — but the test asserts the opposite of what happens.
- The "invalid `top_menu` entry (not a Pair)" test is actually exercising the pre-existing `"…" is not an existing page!` error, not any `top_menu`-specific validation.

**3. `prev`/`next` still cross section boundaries.** The guide says "The sidebar's previous/next navigation stays within each section", but `.prev`/`.next` are still populated from the global `doc.internal.navlist` in `builder_pipeline.jl`, untouched by this PR. In the example build, `guide/index.html` (first page of section 2) gets `« Installation`, pointing back into section 1. Either implement per-section chaining or drop the claim.

**4. `hide()` is ignored by the top menu.** `build_top_menu_sections` never looks at `nn.visible`. With

```julia
pages = ["Sec A" => ["index.md"], hide("Sec B" => "other.md", ["hidden.md"])]
```

"Sec B" shows up in the top bar and its dropdown lists the hidden page.

**5. No validation of the `pages` shape.** The docs require `"Title" => [...]` entries, but nothing enforces it. With a flat `pages = ["index.md", "Other" => "other.md"]`, `title = something(top_node.title_override, "")` produces an empty label and you get `<a class="docs-top-menu-link" href><span…>▾</span></a>` in the output — silently. This should be an error.

**6. CI formatting.** `biome check` fails on `assets/html/js/sidebar.js:65` (missing trailing comma after `decodeURIComponent(...)`).

**7. CHANGELOG.** The entry links `[#2866]` (the superseded PR) and there's no matching `[#2866]:` link definition, so the changelog check will fail. Should be `[#2904]`.

### Should fix

**8. The `DOM.jl` change looks like leftover debugging.** `Node(text::AbstractString) = new(TEXT, text, Attributes(), Node[])` adds two allocations to *every text node of every page* in every build. I reverted it locally and the `html-topmenu` example still builds fine — is there a case that actually needs it?

**9. Dead code:** `find_navnode_for_page` (HTMLWriter.jl:1203) is never called.

**10. Per-page linear scans.** `render_top_menu` and `get_section_navtree` each walk every section's `navlist` looking for the current page, on every page render. A `Dict{String,Int}` page → section index built once in `build_top_menu_sections` would replace both.

**11. The redirect `index.html` is a separate feature riding along.** It hand-writes a meta-refresh page and suppresses the "src/index.md missing" warning, but only when `top_menu` is on, and it picks "first path ending in `index.md`" as the target, which is fairly arbitrary. Worth splitting out and discussing on its own.

**12. A section node that has both a page and children loses its own page.** `navtree = isempty(top_node.children) ? [top_node] : top_node.children` drops `top_node` from the sidebar and dropdown, but `first_page` (the section's link target) still points at it — so clicking the section title can land you on a page that isn't in its own sidebar.

**13. Unrelated churn**, worth dropping to keep the diff reviewable: the `(src === nothing) || push!(…)` → `if` rewrite plus a stray blank line in `builder_pipeline.jl`, two blank lines in `documents.jl`, and `docs/Project.toml` gaining `LiveServer` and a reformatted `[sources]` entry.

**14.** The `html-topmenu` example build passes `warnonly = true`, which hides whatever the feature warns about.

### Minor

- Dropdowns are `display: none !important` on touch, so on mobile a section is only reachable via its first page. Reasonable degradation, but not mentioned in the docs.
- No ARIA attributes on the dropdowns, and the caret is a literal `▾` character rather than the icon font used elsewhere in the theme.
- `@test occursin("getting-started", page_content)` duplicates an assertion made a few lines above.
- `mktempdir()` in the edge-case helper is never cleaned up, and `import Markdown` is unused.

Overall the layout and CSS side looks solid and the generated markup is clean — most of what's above is documentation drift, missing validation, and tests that were never executed.

PR #1715, opened 2021-10-17
[review]: sub TOC on the right side of the page

rmNULL:
Prototyping for #1688. 
All of the changes are still in development stage.

Will open for review once i think its ready.

Issue #1688, opened 2021-08-29
Add table of content for a specific page on the right side


logankilpatrick:
I have been reading through: https://code.visualstudio.com/docs and absolutely love that the right side shows a highlight of the page. I feel like it would be pretty straightforward and helpful to pick up on the headers from the page and show them on the top right, especially since the space is not utilized anyway.

mortenpi:
For future reference: there is discussion about this in #1563, including a simple mockup of the TOC.

mortenpi:
Hmm, I was glancing at #1715 and I noticed that this actually duplicates what we already have. The current TOC already shows the `h2` headings of the current page, and the sidebar jumps to that part of the sidebar on page load and should therefore generally be visible to the user. I guess the question is whether that is good enough UX for this already or not.

logankilpatrick:
Yeah, I am not sure, it just seems like wasted valuable space right now

mortenpi:
I don't think we want add more stuff into the left sidebar, so if we're going to have additional information there, then I think it should go on the right. It is also an option to consider removing the `h2` headings from the left sidebar altogether, if we're going to have them on the right. One would need to make sure that the two sidebars also work well on mobile too though.

I think right now the main thing is to clearly decide what exactly should go into the sidebar on the right. Regarding the Base > Filesystem example: we currently distinguish between the headings on a page (at-contents block) and the docstrings on a page (at-index block). Since we already have those blocks though, should we think of the right sidebar to essentially be a out-of-page version of one of those, or even a combination of those?

rmNULL:
> I don't think we want add more stuff into the left sidebar, so if we're going to have additional information there, then I think it should go on the right. It is also an option to consider removing the h2 headings from the left sidebar altogether, if we're going to have them on the right.

> I think right now the main thing is to clearly decide what exactly should go into the sidebar on the right.

Second this.
Having a clear idea about this should also resolve the h2 issue.

i recall this being a problem while implementing the right TOC,
along with other things like same kind of tag showing up multiple times(
they shared the same name but different function signatures).
its an error on my part that i didn't write down these problems while i faced them
and now my memory has completely failed me.
i'll go through the source later and get back in a day with more details(hopefully).
If you already have something in your mind on how things should be organized that would be helpful(
and save some energy on my end :)

> One would need to make sure that the two sidebars also work well on mobile too though.

getting it to work on mobile would require some redesigning but once we have the organization part
[...]

PR #2910, opened 2026-03-17
Add Right TOC

Sisira07:
toc
This PR adds a table of contents on the right side of every page. I have only included the h2 headings but as mentioned in #1688, would be repetitive. 

I have modified the `HTMLWriter.jl` file by adding a function that uses almost the same logic as the function `navitem`, which is used by the sidebar to display the h2 subheadings. 

I kept the UI simple, for demonstration purposes, it can be improved, but might require changes to be made in `HTMLWriter.jl`. 

Referred to #1715 for `CSS` styling. 

Note: I have used AI to refine my code.

### DIFF
--- a/assets/html/scss/documenter/_variables.scss
+++ b/assets/html/scss/documenter/_variables.scss
@@ -80,6 +80,8 @@ $body-size: 16px !default;
 $documenter-max-width: 50rem !default;
 $documenter-sidebar-width: 18rem !default;
 
+$documenter-toc-width:12 rem !default;
+
 $documenter-sidebar-main-gap: 2rem !default;
 $documenter-main-padding-right: 1rem !default;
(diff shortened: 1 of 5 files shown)


fingolfin:
In the current version, this duplicates the TOC on the left and right, that feels odd.

Does the right side TOC scroll out of view long a long page (like the left-side TOC does), or does it stay anchored?

What happens with the right TOC if the window is narrow (or when viewing the page on a phone) ?

I guess having a live preview somewhere would be useful.

fingolfin:
... aaaand of course I should have started with: Great that you are working on this! (Sorry, I tend to think that but then forget to write it down 😢 )

fingolfin:
@Sisira07 still interested in getting this over the finishing line? Then please reply to my questions and we can perhaps find a way forward

PR #2928, opened 2026-04-26
Fix duplicate elseif condition in interpret_repo_and_remotes

### LINKED ISSUE #2868: Duplicate `elseif` condition in `interpret_repo_and_remotes`
Note that both `elseif` have the same condition:

https://github.com/JuliaDocs/Documenter.jl/blob/a56e4d13c879c12be418166d9e9f7cf54a2da7a9/src/documents.jl#L622-L640

AFAICT it has been like that since @mortenpi introduced the code back in 2023?

singhharsh1708:
The second elseif branch had the same condition as the first (startswith(makedocs_root_remoteref.root, makedocs_root_repo)), but its comment described the opposite case: the git repo root being more specific (nested deeper) than the remotes path. Fix it to use startswith(makedocs_root_repo, makedocs_root_remoteref.root). Fixes #2868

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* Fixed a duplicate `elseif` condition in `interpret_repo_and_remotes` that made the branch handling the case where the git repository root is nested inside a `remotes` path unreachable. ([#2868])
 * Fixed headers hidden by the navbar. ([#2905])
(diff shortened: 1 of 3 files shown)


fingolfin (approved):
LGTM

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* Fixed an `Unexpected repository roots` error when `makedocs` runs in a Git repository nested inside a directory configured in `remotes`; the repository's own remote is now used for the main repository, and the `remotes` entry applies to the surrounding paths. ([#2868])
```
fingolfin, line 19: I took the liberty of revising this changelog entry to talk about the effect of the change, not the cause

PR #2929, opened 2026-04-26
Fix LaTeXWriter not stripping $$ delimiters from text/latex MIME output

### LINKED ISSUE #2923: LaTeXWriter does not strip $$ delimiters from text/latex MIME output
## Summary

The `LaTeXWriter` writes `text/latex` MIME output directly without stripping `$$` or `\[`/`\]` delimiters, unlike the `HTMLWriter` which correctly strips them via `_strip_latex_math_delimiters`. This causes `Bad math environment delimiter` errors when packages emit `$$ \begin{equation} ... \end{equation} $$` via their `show(io, ::MIME"text/latex", ...)` methods.

## Reproducer

Any Julia type whose `show(::IO, ::MIME"text/latex", ...)` wraps output in `$$` delimiters around a display math environment will trigger this. A common real-world example is [ModelingToolkit.jl](https://github.com/SciML/ModelingToolkit.jl), which emits:

```latex
$$ \begin{equation}
\frac{dv(t)}{dt} = g
\end{equation}
$$
```

When building PDF docs with `Documenter.LaTeX()` and an `@example` block that displays a ModelingToolkit object, the LaTeXWriter writes this verbatim into the `.tex` file. Since `$$` and `\begin{equation}` both open display math mode, LaTeX fails with:

```
! LaTeX Error: Bad math environment delimiter.

l.5707 $$ \begin{equation}
```

Minimal `make.jl` to reproduce:

```julia
using Documenter, ModelingToolkit

makedocs(;
    format = Documenter.LaTeX(platform = "native"),  # or "tectonic"
    pages = ["index.md"],
    sitename = "Test",
)
```

Where `index.md` contains an `@example` block that displays a ModelingToolkit system.

## Root cause

In `src/latex/LaTeXWriter.jl` around line 496, the `text/latex` MIME branch writes content directly:

```julia
    # If it has a latex MIME, just write it out directly.
    content = d[MIME("text/latex")]
    if startswith(content, "\\begin{tabular}")
        # ...tabular handling...
    else
        _print(io, content)  # <-- no stripping of $$ or \[ \]
    end
```

The `HTMLWriter` handles this correctly in `src/html/HTMLWriter.jl` via `_strip_latex_math_delimiters`, which strips `\[ ... \]`, `$ ... $`, and `$$ ... $$` wrappers before re-wrapping:

```julia
latex = d[MIME"text/latex"()]
# If the show(io, ::MIME"text/latex", x) output is already wrapped in
# \[ ... \] or $$ ... $$, we unwrap it first, since when we output
# Markdown.LaTeX objects we put the correct delimiters around it anyway.
has_math, latex = _strip_latex_math_delimiters(latex)
```

The `LaTeXWriter` should apply the same (or equivalent) stripping logic before writing.

## Related issues

- #1426 — Fixed `$$` stripping in the HTMLWriter (v0.26.0)
- #1709 — Fixed `text/latex` MIME handling in LaTeXWriter (wrote directly instead of round-tripping through Markdown), but did not add delimiter stripping

## Version info

- Documenter v1.8.0 (compat 1.15.0)
- Julia 1.11
- Tested with both `tectonic` (via `tectonic_jll`) and native `latexmk`/`lualatex`

singhharsh1708:
When a type's show(io, ::MIME\"text/latex\", ...) wraps output in \$\$
or \\[...\\] around a display math environment (e.g. ModelingToolkit),
LaTeXWriter was writing the content verbatim. This caused a 'Bad math
environment delimiter' LaTeX error because both \$\$ and \\begin{equation}
try to open math mode.
Add _strip_latex_math_delimiters which strips outer \$\$/\\[...\\] delimiters
only when the inner content already contains a \\begin{...} environment.
Plain math content like \$\$ \\left[...\\right] \$\$ is left unchanged.
Fixes #2923

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Fixed headers hidden by the navbar. ([#2905])
+* Fixed `LaTeXWriter` writing `text/latex` MIME output verbatim when wrapped in `$$` or `\[...\]` around a display math environment (e.g. `\begin{equation}`), which caused a "Bad math environment delimiter" LaTeX error. ([#2923])
 * Tightened `@ref` classification so references are matched more consistently by syntax, with header labels still taking precedence over docstrings and mistaken header/id references less likely to resolve as docstrings. ([#2678])
 
 ## Version [v1.17.0] - 2026-02-20
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -467,6 +467,24 @@ end
 
 # Select the "best" representation for LaTeX output.
 using Base64: base64decode
+
+function _strip_latex_math_delimiters(latex::AbstractString)
+    m_bracket = match(r"\s*\\\[(.*)\\\]\s*"s, latex)
+    if m_bracket !== nothing
+        inner = m_bracket[1]
+        if occursin(r"^\s*\\begin\{"s, inner)
+            return lstrip(inner)
+        end
+    end
+    m_dollars = match(r"^\s*(\${1,2})([^\$]*)\1\s*$"s, latex)
+    if m_dollars !== nothing
+        inner = m_dollars[2]
+        if occursin(r"^\s*\\begin\{"s, inner)
+            return lstrip(inner)
+        end
+    end
+    return latex
+end
 latex(io::Context, node::Node, ::Documenter.MultiOutput) = latex(io, node.children)
 function latex(io::Context, node::Node, moe::Documenter.MultiOutputElement)
     return Base.invokelatest(latex, io, node, moe.element)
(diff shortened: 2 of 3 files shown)


odow:
[I maintain that this is a bug in ModelingToolkit](https://github.com/JuliaDocs/Documenter.jl/issues/2923#issuecomment-4256856004):

```
$$ \begin{equation}
\frac{dv(t)}{dt} = g
\end{equation}
$$
```
That's not valid LaTeX.

singhharsh1708:
You're right that $$\begin{equation}...\end{equation}$$ is invalid LaTeX — the bug belongs in the package generating it.
I've updated the approach: Documenter now emits a @warn when it detects this pattern, explicitly telling users to fix the show(io, MIME"text/latex"(), ...) method of the relevant type. The outer delimiters are still stripped so the PDF build doesn't crash with an opaque "Bad math environment delimiter" error, but the warning makes the real problem visible rather than silently papering over it.
Happy to close this if the preference is to let it fail loudly instead.

odow:
I don't think there needs to be any change in Documenter. We shouldn't try to correct other people's incorrect LaTeX.

odow:
I assume the issue is here:

https://github.com/SciML/ModelingToolkit.jl/blob/8d382c8e10c9cfa77807bcc2716074e5635f9470/lib/ModelingToolkitBase/ext/MTKLatexifyExt.jl#L21-L23

asinghvi17:
It's probably because `latexify` emits `\begin{equation}...\end{equation}` by default - https://github.com/korsbo/Latexify.jl/blob/8b633ca3bbadf5e85b27b57555a76574cc15e786/src/latexequation.jl#L4

Ideally there would be some way to tell it not to - which SciML/MTK could use.  But at a brief glance I could not find one.

singhharsh1708:
Understood, closing this. The fix belongs upstream — either in Latexify.jl to provide a way to suppress the \begin{equation} wrapper, or in ModelingToolkit's MTKLatexifyExt.jl to use that option once available.

PR #2930, opened 2026-04-26
Add test coverage for multi-dash @id header resolution 

### LINKED ISSUE #2732: Header `@id` syntax does not support dashes
```
┌ Warning: Cannot resolve @ref for md"[advanced users guide](@ref advanced-guide)" in src/somefile.md.
│ - No docstring found in doc for binding `Base.-`.
└ @ Documenter ~/.julia/packages/Documenter/iRt2s/src/utilities/utilities.jl:49
```

Maybe it's trying to parse as Julia?  I guess the simple solution is to use a String or something. But I thought it would match first, that's what the docs say.

singhharsh1708:
The root cause (dashed @ref being misclassified as Julia expressions)
was fixed in #2678 via DASHED_XREF_ID_REGEX. This adds an explicit
test for the successful case: a header defined with a multi-dash @id
(multi-word-section-id) and a @ref that correctly resolves to it.
Closes #2732


fingolfin (approved):
Thanks!

PR #2935, opened 2026-05-18
docs: add instructions for modifying IOContext in outputs (Closes #2207)

### LINKED ISSUE #2207: Document how to modify IOContext parameters in outputs
I.e. number of columns of the REPL output etc. -- would be good to have a short how-to guide for this in the docs. Related issues:

* https://github.com/JuliaDocs/Documenter.jl/issues/1450
* https://github.com/JuliaDocs/Documenter.jl/issues/942

@fredrikekre: in reference to https://github.com/JuliaDocs/Documenter.jl/issues/1450, how do you imagine the user using `DisplayAs.(set|with)context` in Documenter? Explicitly call it in the at-example/repl block? I.e. something like this?

````markdown
```@example
f() = collect('a':'z') # hide
f() |> DisplayAs.withcontext(...)
```
````

### PR DESCRIPTION
**Pull request by rainerrodrigues**
Closes #2207

### Description
This PR adds a short section to the `syntax.md` manual explaining how users can use `IOContext` to control the display of large outputs in `@example` and `@repl` blocks. 

Many users struggle with documentation pages becoming bloated when evaluating blocks that return large arrays or dataframes. This addition provides a standard solution by adding:
1. A brief explanation of how to use `Base.show` with a custom `IOContext` at the end of a block.
2. A copy-pasteable example demonstrating the use of `:limit => true` and `:displaysize`.
3. A quick tip on using `ENV["COLUMNS"]` for global doctest width formatting.

### Testing
- [x] Built documentation locally (`julia --project=docs docs/make.jl`).
- [x] Verified that the new section in `syntax.md` renders correctly and the code blocks format properly. 

Let me know if this should be moved to a different section of the manual or if you'd like any wording changes!

### CODE REVIEW

**Comment by rainerrodrigues** [comment]
Hi! I've added the documentation and updated the CHANGELOG.md. It looks like the CI is failing purely due to the flaky test/online/online_githubcheck.jl test (it expected :linkcheck_remotes but evaluated to an empty set, likely hitting a GitHub API rate limit). The PR should be ready for review whenever you have time!

File: CHANGELOG.md [hunk 1]
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Large Markdown tables in LaTeX / PDF output now use `xltabular`, so they break across pages with a repeated header row instead of being silently dropped or aborting the build with `Dimension too large`. ([#2963])
+* Added documentation on how to use `IOContext` to limit the display size of large outputs in `@example` blocks. ([#2207])
```
fingolfin: @mortenpi [elsewhere wrote](https://github.com/JuliaDocs/Documenter.jl/pull/2968)

> FWIW, my policy has been not to add any changelog entries to docs changes, since they are non-functional.

so I am dropping this now
```suggestion
```

docs/src/man/syntax.md
```diff
@@ -432,6 +432,35 @@ As with `@index` if `Pages` is not provided then all pages are included. The def
 
     Documenter will then list the contents of the "Subsection" pages, and they will always appear in the same order as they are in the sidebar.
 
+### Modifying Output Display (`IOContext`)
```
fingolfin, line 435: You made this a subsection of the section `## @contents` block`  which seems wrong. It should be either a section of its own, but still in a more logical place; or a subsection of the section describing `@example`

docs/src/man/syntax.md
```diff
@@ -432,6 +432,35 @@ As with `@index` if `Pages` is not provided then all pages are included. The def
...
+
+When generating documentation, you may want to limit the output of large arrays, dataframes, or deeply nested structures. By default, Documenter evaluates blocks without aggressive truncation, which can occasionally result in excessively long output blocks.
```
fingolfin, line 437: That's factually incorrect, Documenter already sets `:limit => true` everywhere.

This section should be about "how to change the truncation size", not "how to enable truncation".

fingolfin (changes requested):
@rainerrodrigues thanks for your contribution! However, what you did in this PR does not seem to actually do what was asked for in #2207. You explain a hack as a workaround, but it is very brittle (e.g. the `show` must be the last thing in the `@example` block or it'll break). The issue asked for "kwargs" to be defined for `@example` that could control its overall handling of output.

PR #2936, opened 2026-05-22
Add progressmeter extension

langestefan:
This adds a tiny extension for [ProgressMeter.jl](https://github.com/timholy/ProgressMeter.jl) so long doc builds can be tracked over time.

Output will look like this:

```julia
julia> using ProgressMeter

Running doctests: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:01
Expanding pages:  100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:11
Rendering HTML:   100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:03
```


fingolfin:
This sounds useful to me, I'd be in favor. Hopefully @mortenpi as well?

This is marked as a draft, so without having looked deeply at it: what do you consider as "TODO" @langestefan ?

fingolfin:
(BTW I took the liberty of resolving the merge conflict in the changelog so that CI can run again now that we unbroken it on master)

asinghvi17:
This also looks generally reasonable to me.

langestefan:
> This is marked as a draft, so without having looked deeply at it: what do you consider as "TODO" @langestefan ?

I think the core is finished. I will test it again locally to make sure.

I marked this as draft because there is currently no way to pass customization options to `Progress`.

Also we may want to remove the redundant output from:

```julia
Rendering HTML:   100%|██████████████|
```

to

```julia
100%|██████████████|
```

that looks cleaner to me.

ext/DocumenterProgressMeterExt.jl
```diff
@@ -0,0 +1,43 @@
...
+    if result === nothing
+        finish!(p.progress)
+        return nothing
+    end
+    return result
+end
+
+function Documenter.progress_iter(iter::AbstractVector; desc::AbstractString = "")
+    n = length(iter)
+    bar = Progress(n; desc = desc, color = :blue)
```
asinghvi17: Can just pass through kwargs... here no?


```suggestion
function Documenter.progress_iter(iter::AbstractVector; kwargs...)
    n = length(iter)
    keywords = merge((; color = :blue), kwargs)
    bar = Progress(n; color = :blue, keywords...)
```

langestefan: yes, but how do you get the options from `make.jl` to there? `progress_iter` is not a user facing function.

asinghvi17: Would you need to?  Aside from turning it off which we already have a switch for, I'm not sure how much more configuration a user would actually find useful

langestefan: Hard to say what some people might want to use in the future :) I will leave it out for now then!

asinghvi17: Oh.  Actually, I just realized.  This probably doesn't respect `verbose=false`?  We might want to add that in as the only configurability option available

langestefan: sorry what verbose are you referring to?

asinghvi17: Somehow I could have sworn that was a kwarg to makedocs.  But apparently not.  Please ignore me then 😅

asinghvi17:
This looks good to me, but some of the new tests are failing on julia 1.6

langestefan:
yeah some tests need to be conditional on > Julia 1.9 because that's when extensions were added

test/runtests.jl
```diff
@@ -121,4 +121,8 @@ end
     # Running doctest() on our own manual
     @info "doctest() Documenter's manual"
     @quietly include("manual.jl")
+
+    # Last: loading ProgressMeter would draw bars in every subsequent build.
+    @info "Running tests in progress.jl"
+    @quietly include("progress.jl")
```
asinghvi17, line 127:
```suggestion
    if VERSION < v"1.10"
        # Last: loading ProgressMeter would draw bars in every subsequent build.
        @info "Running tests in progress.jl"
        @quietly include("progress.jl")
    end
```

langestefan: should be handled now

fingolfin (changes requested):
Thanks for working on this. As I said, I am in general in favor, but two things need to be addressed from my point of view:

### 1. Activation isn't actually opt-in, and there's no opt-out

The docs and changelog say "Activate by adding `ProgressMeter` to your docs project and writing `using ProgressMeter` in `make.jl`". But that's not quite how package extensions work: the extension loads as soon as ProgressMeter is *anywhere* in the session, including as a transitive dependency of something else in the docs environment. ProgressMeter is a common dep, so plenty of users will get bars they never asked for.

And there's no way to turn them off: `ProgressMeter.Progress` defaults to `enabled = true` and does no TTY detection, so this also fires in CI logs, where `\r` overwriting produces noise rather than a bar.

Minimum: pass `enabled = isinteractive() && !haskey(ENV, "CI")`. Better: a `makedocs` keyword threaded down to `progress_iter`, so it's genuinely opt-in.

### 2. The bar collides with output from the loops it wraps

The wrapped loops are exactly the loops that print. `expand` emits `@warn` in several places and `_doctest` prints failure diffs. Since `Progress` overwrites with `\r`, these interleave. This results in ugliness like this:

```
item 2 counter=1
  40%|████████████▍      |  ETA: 0:00:01[Kitem 3 counter=2
```

Related: `finish!` only runs when iteration completes normally. Any throw from the loop body skips it, so the terminal is left mid-bar and the error message lands on top of it. So this needs some `try`/`finally`.

langestefan:
Good catches! I did think about point 1, but I found it a bit counter-intuitive to have to add an argument to  `makedocs` for something which is only offered as an extension? That feels like I made an error in the design. If you don't mind that, I can implement the suggestion. Otherwise I will investigate other options.

PR #2940, opened 2026-06-02
User manager lexer list for latex output

sandyspiers:
The `LatexWriter.LEXER` list is rather small.

This PR allows user to provide their own list.

This list defaults to the current whitelist, so shouldn't add any breaking changes.

Thanks!

(disclaimer; Claude helped with this one)

### DIFF
--- a/src/latex/LaTeXWriter.jl
+++ b/src/latex/LaTeXWriter.jl
@@ -16,6 +16,15 @@ module LaTeXWriter
 import ...Documenter: Documenter
 using MarkdownAST: MarkdownAST, Node
 
+# Whitelisted lexers.
+const LEXER = Set(
+    [
+        "julia",
+        "jlcon",
+        "text",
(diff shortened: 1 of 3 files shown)


src/latex/LaTeXWriter.jl
```diff
@@ -16,6 +16,15 @@ module LaTeXWriter
...
 using MarkdownAST: MarkdownAST, Node
 
+# Whitelisted lexers.
+const LEXER = Set(
+    [
+        "julia",
+        "jlcon",
+        "text",
+    ]
+)
```
fingolfin, line 26: @mortenpi Why do we even have such a whitelist, as opposed to passing everything to `pygments`; or possibly asking pygments for a list of lexers?

I feel that no matter what we do with this PR, there should be a comment next to this list that provides that justification, if there is one.

PR #2941, opened 2026-06-04
Add workflow_dispatch to yml to allow manual runs

mauro3:
I added  `workflow_dispatch:` to the documentation.yml example. This allows to run this manually as well.

See https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch and https://github.com/orgs/community/discussions/58022

### DIFF
--- a/docs/src/man/hosting.md
+++ b/docs/src/man/hosting.md
@@ -57,7 +57,7 @@ on:
       - master # update to match your development branch (master, main, dev, trunk, ...)
     tags: '*'
   pull_request:
-
+  workflow_dispatch:
 jobs:
   build:
     # These permissions are needed to:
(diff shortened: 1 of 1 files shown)


docs/src/man/hosting.md
```diff
@@ -57,7 +57,7 @@ on:
       - master # update to match your development branch (master, main, dev, trunk, ...)
     tags: '*'
   pull_request:
-
+  workflow_dispatch:
```
fingolfin, line 60:
```suggestion
  workflow_dispatch:

```

fingolfin:
Should we really recommend this generally? Why?

mauro3:
I wanted to do this at the time I made this PR, but I can't remember what for... I don't think most people will notice and probably needs no documentation. Anyway, I have not strong feelings but don't see the harm and see a benefit at times.

This is what it would look like in the action panel:
image

fredrikekre:
One reason could be that if you don't commit the manifest you might want to manually rebuild on the same commit to bring in new releases of doc deps.

fingolfin (approved):

PR #2946, opened 2026-07-02
Revert "Support self-hosted github installations (#2755)"

mortenpi:
I'm inclined to revert #2755 for now. I've been holding off on the release to see if it's enough to make some small changes. But I don't see immediately what the correct solutions are, and I don't want to keep holding off the release any longer.

The two reservations I have with the PR:

- I am not super-happy that we pick up the API URL automagically for _non-Enterprise_ deployments. That feels like a security gotcha you can abuse and get people's `GITHUB_TOKEN`s and compromise their repos if you're clever enough.

  What I would like here is an explicit opt-in to the Enterprise case. And that can then read the environment variables automatically, shifting the responsibility clearly to the users.

  For example, something like `deploy_config = GitHubEnterpise()` or `deploy_config = GitHubActions(:enterprise)`.

- It changed the way we determine the GitHub pages URL for `post_status`. It seems incorrect, and I think regresses it for out-of-repo deployments (i.e. when your're deploying to `gh-pages` of a another repo). I am also not 100% we want that API call in the constructor either.

  Mose also flagged the same or related issue here I think: https://github.com/JuliaDocs/Documenter.jl/pull/2755#discussion_r2934416794

So this needs a bit more thought in my opinion, and unfortunately I am not really comfortable shipping it. But I also don't want to block releases and development until we've had chance to address it.

So I propose we revert this for now. And try to get GitHub Enterprise support in a future release.


mortenpi:
cc @tz-lom @fingolfin

Also, just want to note that the deplodocs code is not the easiest part of the codebase to change / contribute to. It could really use a refactor. But that kinda also forces us to be a bit conservative here unfortunately.

goerz (approved):

tz-lom:
Hi, @mortenpi 

I've missed when you opened this PR, sorry for that.

> That feels like a security gotcha you can abuse and get people's GITHUB_TOKENs and compromise their repos if you're clever enough.
I am not sure, if you can manipulate CI environment you can just patch hardcoded value as well.
However, I have nothing against explicit Enterprise option either

> It changed the way we determine the GitHub pages URL for post_status. It seems incorrect, and I think regresses it for out-of-repo deployments (i.e. when your're deploying to gh-pages of a another repo). I am also not 100% we want that API call in the constructor either.

When I was working on PR I found that test coverage was not covering everything,
I improved coverage, so I would also advise not to roll-back those improvements, some of them test non-enterprise scenarios.

I would like to address this concern with a proper test, and if it would be found that I made a mistake I will fix that do you have a faulty scenario example already?


I am interested to fix implementation rather than throw it away completely, can you share your release cycle plan?

P.S.
Are you attending JuliaCon now? If so, do you want to meet in person?

fingolfin:
@tz-lom @mortenpi and me are at JuliaCon. Not sure if you guys already linked up by other means (e.g. over Slack). Anyway, happy to link up as well if there's interest (though for this PR, I feel @mortenpi is more on top of things so it'd be best if you two were to meet and talk, I am optional ;-) )

mortenpi:
I'll go ahead and merge this now. The regression test failures should be expected at the moment. I'll try to see if we can restore some of the tests in a follow-up PR.

PR #2947, opened 2026-07-07
Support arbitrary `[content](@id name)` anchors (#745)

### LINKED ISSUE #745: Allow adding arbitrary anchors to be linked with `@ref`, e.g. via `[content](@id anchor-name)` syntax
My use case is that I generated thumbnails as previews for plot figures, and I'd like that these thumbnail images be cross-referenced with `@ref` or similar so that when clicked, we can jump to another section in the docs (e.g. to view the full-sized plot or the source code).

Currently, `alt text` inserts a figure and `[Some-Header-Title](@ref)` inserts a parsed cross-reference HTML link to the specified header anchor, and `[text](URL)` is a hyperlink, but I haven't been able to figure out a way to combine inserting the fig + a cross-referenced link. As far as I can tell, `Documenter` only allows `@ref` to headers.

Therefore, a way of specifying arbitrary anchors on any content so that we can `@ref` them elsewhere in the documentation would be nice!

A suggested design could be:
`[content](@anchor anchor-name)` in the markdown, which is then outputted with `<a>` tags with `id`'s in the HTML. Then these anchors can be referenced by using `@ref`.

(thanks to @mortenpi for the suggested design!)

kahliburke:
Closes #745.

## What this does

Generalizes the `@id` mechanism — currently limited to top-level headers — so `[content](@id name)` can attach a named, cross-referenceable anchor to **arbitrary inline content** (text, images). For example, the original use case in this issue: a thumbnail image that jumps to a full-size figure.

```markdown
An inline text anchor: [some phrase](@id my-anchor).

A clickable thumbnail that is itself an anchor target:
[![thumbnail](assets/thumb.png)](@id figure-1)

... elsewhere, on any page ...
Jump to [the phrase](@ref my-anchor) or [Figure 1](@ref figure-1).
```

This follows the direction suggested here by @ettersi (reuse `@id` for consistency with header labels) and @mortenpi (the at-id lookup is currently restricted to headings; generalizing it).

## Behavior

- Resolves through the existing `@ref` pipeline, including forward and cross-page references.
- Shares a single id namespace with header `@id` labels, so a duplicate id (header/header, header/inline, or inline/inline) is caught by the existing uniqueness check.
- Ids are slugified exactly as header `@id`s are, so `[x](@id My Anchor)` yields a valid `My-Anchor` id.
- Written to the `objects.inv` inventory as `std:label` entries, so other projects can reach them via `DocumenterInterLinks` `@extref` — addressing the inventory requests from @moble and @goerz.
- Rendered as `<span id="...">` in HTML and `\hypertarget` in LaTeX (matching the `\hyperlinkref` emitted for a resolved `@ref`). A `<span>` (rather than `<a id>`) keeps the anchored content visually unchanged and avoids nesting an `<a>` inside content that may itself be a link or image.

## Implementation

- New `AnchoredInline` inline element (`documents.jl`).
- `collect_named_anchors!` (`expander_pipeline.jl`) walks each page during template expansion — before the cross-reference stage — and rewrites non-header `@id` links, registering them in the same `AnchorMap` as headers.
- `@contents` population skips non-heading anchors that now share that map.
- HTML and LaTeX writer methods for `AnchoredInline`.

## Tests & docs

- Integration tests in `test/crossreferences.jl` covering HTML output, inventory entries, cross-page references, nesting in admonitions/lists, header-slug collisions, id slugification, empty-content anchors, and duplicate-id failure.
- An inline anchor added to the `examples/` build, with an `objects.inv` assertion via `DocInventories` in `test/examples/tests.jl` (exercises the real HTML + LaTeX integration path).
- A new "Anchors on arbitrary content" section in `man/syntax.md`, plus a CHANGELOG entry.
- Full test suite passes locally.

## Notes for reviewers

- **Naming:** the anchors are stored in `doc.internal.headers` (reusing the header machinery, per the discussion above), which makes that field name a bit of a misnomer now. Happy to rename it to `doc.internal.anchors` throughout if you'd prefer.
- **Scope:** this handles inline `@id` links in page Markdown. Anchors inside docstrings / `@example` output (which live in off-tree ASTs) and block-level anchoring (e.g. labeling a table or equation, as @odow / @moble mentioned) are out of scope here and would be natural follow-ups.

## Demo

*(screenshots of a small demo site — inline text anchor, a thumbnail-gallery that jumps to full-size figures across pages, and the decoded `objects.inv` — to be added.)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Support self-hosted GitHub instances. ([#2755])
+* The `[content](@id name)` syntax can now attach a named, cross-referenceable anchor to arbitrary inline content (such as text or a thumbnail image), not just headers. These anchors resolve via `@ref`, share a single id namespace with header labels, and are written to the `objects.inv` inventory as `std:label` entries. ([#745])
 
 ### Changed
 
@@ -1664,6 +1665,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#697]: https://github.com/JuliaDocs/Documenter.jl/issues/697
 [#706]: https://github.com/JuliaDocs/Documenter.jl/issues/706
 [#744]: https://github.com/JuliaDocs/Documenter.jl/issues/744
+[#745]: https://github.com/JuliaDocs/Documenter.jl/issues/745
 [#756]: https://github.com/JuliaDocs/Documenter.jl/issues/756
 [#764]: https://github.com/JuliaDocs/Documenter.jl/issues/764
 [#774]: https://github.com/JuliaDocs/Documenter.jl/issues/774
--- a/docs/src/man/syntax.md
+++ b/docs/src/man/syntax.md
@@ -294,6 +294,33 @@ document more than once is disallowed.
     Both user-defined and internally generated header reference labels take precedence over
     docstring references, in case there is a conflict.
 
+### Anchors on arbitrary content
+
+The `[content](@id name)` syntax is not restricted to headers: it can attach a named anchor
+to arbitrary inline content, turning it into a cross-reference target. This is useful, for
+example, to link a thumbnail image to the section or figure it previews:
+
+```markdown
+A named anchor on inline text: [some text](@id my-anchor).
+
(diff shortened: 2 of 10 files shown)


kahliburke:
Superseded by #2948 (branch renamed to `kb/inline-id-anchors`; the head-branch rename closed this one).

PR #2944, opened 2026-06-17
Add note about multi-line expressions

BjarkeHautop:
Documents the at least 7 spaces rule.

Fixes #1159

PR #2948, opened 2026-07-07
Support arbitrary `[content](@id name)` anchors (#745)

### LINKED ISSUE #745: Allow adding arbitrary anchors to be linked with `@ref`, e.g. via `[content](@id anchor-name)` syntax

kahliburke:
Closes #745.

## What this does

Generalizes the `@id` mechanism — currently limited to top-level headers — so `[content](@id name)` can attach a named, cross-referenceable anchor to **arbitrary inline content** (text, images). For example, the original use case in this issue: a thumbnail image that jumps to a full-size figure.

```markdown
An inline text anchor: [some phrase](@id my-anchor).

A clickable thumbnail that is itself an anchor target:
[![thumbnail](assets/thumb.png)](@id figure-1)

... elsewhere, on any page ...
Jump to [the phrase](@ref my-anchor) or [Figure 1](@ref figure-1).
```

This follows the direction suggested here by @ettersi (reuse `@id` for consistency with header labels) and @mortenpi (the at-id lookup is currently restricted to headings; generalizing it).

## Behavior

- Resolves through the existing `@ref` pipeline, including forward and cross-page references.
- Shares a single id namespace with header `@id` labels, so a duplicate id (header/header, header/inline, or inline/inline) is caught by the existing uniqueness check.
- Ids are slugified exactly as header `@id`s are, so `[x](@id My Anchor)` yields a valid `My-Anchor` id.
- Written to the `objects.inv` inventory as `std:label` entries, so other projects can reach them via `DocumenterInterLinks` `@extref` — addressing the inventory requests from @moble and @goerz.
- Rendered as `<span id="...">` in HTML and `\hypertarget` in LaTeX (matching the `\hyperlinkref` emitted for a resolved `@ref`). A `<span>` (rather than `<a id>`) keeps the anchored content visually unchanged and avoids nesting an `<a>` inside content that may itself be a link or image.

## Implementation

- New `AnchoredInline` inline element (`documents.jl`).
- `collect_named_anchors!` (`expander_pipeline.jl`) walks each page during template expansion — before the cross-reference stage — and rewrites non-header `@id` links, registering them in the same `AnchorMap` as headers.
- `@contents` population skips non-heading anchors that now share that map.
- HTML and LaTeX writer methods for `AnchoredInline`.

## Tests & docs

- Integration tests in `test/crossreferences.jl` covering HTML output, inventory entries, cross-page references, nesting in admonitions/lists, header-slug collisions, id slugification, empty-content anchors, and duplicate-id failure.
- An inline anchor added to the `examples/` build, with an `objects.inv` assertion via `DocInventories` in `test/examples/tests.jl` (exercises the real HTML + LaTeX integration path).
- A new "Anchors on arbitrary content" section in `man/syntax.md`, plus a CHANGELOG entry.
- Full test suite passes locally.

## Notes for reviewers

- **Naming:** the anchors are stored in `doc.internal.headers` (reusing the header machinery, per the discussion above), which makes that field name a bit of a misnomer now. Happy to rename it to `doc.internal.anchors` throughout if you'd prefer.
- **Scope:** this handles inline `@id` links in page Markdown. Anchors inside docstrings / `@example` output (which live in off-tree ASTs) and block-level anchoring (e.g. labeling a table or equation, as @odow / @moble mentioned) are out of scope here and would be natural follow-ups.

## Demo

image
image
image


kahliburke:
For reviewers: I looked into the failing checks, and the ones that remain red do not appear to be caused by this PR — they also fail on recent `master` runs (e.g. the merges of #2942, #2944, #2945), and this branch does not touch the relevant files.

- **Documentation: html / pdf** and **Runic** — these *were* caused by this PR (a couple of `@ref`s to undocumented types in a new docstring / the manual, plus formatting) and are now fixed and green.
- **Linkcheck: Documenter manual** — fails on `https://codeberg.org` returning HTTP 403 (an existing link in the Forgejo/Codeberg remote docs; the site appears to block the CI client). Intermittent, not introduced here.
- **Linkcheck: online tests** — `test/online/online_githubcheck.jl` is sensitive to the CI `GITHUB_TOKEN`, which has reduced permissions on fork PRs; not touched by this PR.
- **Julia nightly (3 OS)** — nightly-only error-message/parse-format changes in doctest fixtures; the stable, LTS and 1.6 jobs all pass. Also red on `master`.
- **codecov/patch** — soft coverage-threshold on the diff.

Happy to dig into any of these further if it would help — just flagging that they look pre-existing / environmental rather than regressions from this change.

fingolfin:
@kahliburke could you please disclose any AI use, as required by our `AGENTS.md` ?

docs/src/man/syntax.md
```diff
@@ -294,6 +294,33 @@ document more than once is disallowed.
     Both user-defined and internally generated header reference labels take precedence over
     docstring references, in case there is a conflict.
 
+### Anchors on arbitrary content
```
fingolfin: This is right now a subsection under `## Header`. Perhaps a different place would be better? E.g. perhaps in or right after the `## [`@ref` and `@id` links](@id at-ref-at-id-links)` section... which, I just realize, really should be renamed, as it sounds as if `@id` was used to create links, when it is there to make anchors / link targets. Which that section doesn't even properly explain, other by giving an example.

So in that light, it fits even better: you text could be adjusted to explain `@id` in general, and then it is quite natural to state as something that can be applied to lots of things.

docs/src/man/syntax.md
```diff
@@ -294,6 +294,33 @@ document more than once is disallowed.
...
+A clickable thumbnail that is itself an anchor target:
+
+[![thumbnail](assets/thumb.png)](@id figure-1)
+
+... elsewhere ...
+
+Jump to [the anchored text](@ref my-anchor) or [Figure 1](@ref figure-1).
+```
+
+Unlike the header form, the link wrapping the content is *not* removed: the content is
```
fingolfin: I don't understand this, which "link" is not removed, and where? Are you talking about `[thumbnail](@id figure-1)` ?

docs/src/man/syntax.md
```diff
@@ -294,6 +294,33 @@ document more than once is disallowed.
...
+rendered as usual, wrapped in an anchor (`<span id="...">` in HTML, `\hypertarget` in
+``\LaTeX``) that `@ref` links can point to.
+
+These anchors share a single id namespace with header `@id` labels (so a duplicate id is
+reported the same way a duplicate header is), and they are written into the `objects.inv`
```
fingolfin:
```suggestion
These anchors share a single id namespace with header `@id` labels, so a duplicate id is
reported the same way a duplicate header is. They are written into the `objects.inv`
```

src/expander_pipeline.jl
```diff
@@ -1117,18 +1122,48 @@ iscode(x::MarkdownAST.CodeBlock, r::Regex) = occursin(r, x.info)
 iscode(x::MarkdownAST.CodeBlock, lang) = x.info == lang
 iscode(x, lang) = false
 
-const NAMEDHEADER_REGEX = r"^@id (.+)$"
+# Matches the destination of an `@id` link, both for named headers (`# [H](@id name)`,
```
fingolfin, line 1125: Ahh I guess the problem I had above in understanding is that there are (at least) two different ways of interpreting "link" here:
1. the Markdown syntax for writing links is used by Documenter to represent anchors
2. HTML links, e.g. defined by `<a>` elements

I was thinking about 2, but I assume you meant 1. OK, gotcha! Perhaps we should talk about "Markdown link" -- or just call it an "`@id` anchor", and explain the syntax section of the manual that while technically this is a Markdown link element, we used it to define link anchors and hence will refer to it as an anchor...

fingolfin:
I looked over this PR, and it seems solid; except for a few tweaks in the documentation I feel we could merge it soon. Thanks!

I've used the "Update branch" button to merge latest master, so that CI re-runs -- hopefully that'll cure the CI failures.

That said, it *is* marked as draft, so perhaps @kahliburke had something that ought to be changed before we properly review and merge it?

goerz:
It's a nice feature, and the implementation looks pretty solid to me

kahliburke:
@fingolfin The draft status was unintentional, updated that. Thanks.

fingolfin:
I've pushed a couple commits (made in colab with Claude Opus, acknowledged in the commit messages), that fix a couple issue with the code we code, improves the docs, and clears up some naming (e.g. we renamed `doc.internal.headers` to `doc.internal.anchors` for clarity). Details in the commit messages of the four new commits.

fingolfin (approved):
Looks good to me now but I also made major changes, so @kahliburke should say whether those are OK, and perhaps @goerz or @mortenpi want to have another look?

kahliburke:
@fingolfin Thanks for picking this up. All four of your commits look right to me and I've taken them as-is. The rename is an improvement, and pointing `.object` at the live `AnchoredInline` fixes a real slip in my original code.

Two follow-ups pushed on top.

**A `\label` for an `@id` anchor inside a header.** The `in_header` guard in "Clarify `@id` anchor docs and fix LaTeX header case" fixes a genuine problem, but it trades it for a different one: the anchor loses its target entirely, so the `\hyperlinkref` emitted on the referring page points at a destination that is never defined anywhere in the document. `\hyperlinkref` (documenter.sty:91) falls back to `\hyperlink{H}` when `r@H` is undefined, so the link silently goes nowhere.

I built the LaTeX for `## Mixed [thing](@id in-header) here` on one page, with `[the header one](@ref in-header)` on another:

| | emitted for the header | reference resolves | ToC / bookmarks |
|---|---|---|---|
| before your commit | `\chapter{Mixed \hypertarget{H}{thing} here}` | yes | `\hypertarget` rides into the moving argument |
| after your commit | `\chapter{Mixed thing here}` | no, `H` undefined document-wide | clean |
| now | `\chapter{Mixed thing here}` then `\label{H}{}` | yes | clean |

Deferring a `\label` to just after the title keeps your fix and restores the target, which is what `AnchoredHeader` already does for the whole-header case. This only bites on a header with *mixed* content; an `@id` that wraps an entire header takes the `AnchoredHeader` path and was always fine, which is why the suite stayed green through both states. Covered now in `test/latexwriter.jl`.

**Code spans in the docs.** The `declaude` pass turned `` `<span id="...">` `` and `` `\hypertarget` `` into "a span tag" and "a hypertarget command". Those are the exact strings a reader is scanning for, and the manual code-spans tags and TeX commands elsewhere (`<details>` in `showcase.md`, `\DocMainTitle` in `other-formats.md`), so I've restored them. I kept your plain "LaTeX" over ``\LaTeX``: checking the manual, bare "LaTeX" is what it uses in running prose, with the logo form reserved for the LaTeX-syntax page and its heading, so that change was right.

Full test suite green locally (2734 pass, 1 pre-existing broken), Runic clean.

AI disclosure per `AGENTS.md`: this review and both commits were prepared with Claude Opus 5, and the commits carry `Co-authored-by:` trailers.

kahliburke:
The one red check, `Linkcheck: Documenter manual`, looks like an external flake rather than anything in these commits:

```
Error: linkcheck 'https://github.com/JuliaDocs/Documenter.jl/issues/2963' status: 500.
```

GitHub 500ing on one of its own issue pages. That link is `CHANGELOG.md:2302`, added by bcd6802f3 (#2966), and it was already in the tree for the previous run on 6437b5077 where this same job passed. My two commits touch `docs/src/man/syntax.md`, `src/latex/LaTeXWriter.jl` and `test/latexwriter.jl`, and add no URLs.

Rerunning needs admin rights, so could you restart that job when you get a chance? The other 27 checks are green.

fingolfin:
Thanks @kahliburke (also for acknowledging the AI use -- though you still haven't done so for the original submission).

I've made one more change, though: I really don't want us to mention implementation details like `<span>` and `\hypertarget` in the user documentation. We don't do that for any other feature, and I think that's good. So I replace that bit *again*, but this time just dropped it. BTW it was also not quite correct, as with your recent changes an `@id`   anchor inside a header title emits a `\label` instead of `\hypertarget`.

If you are OK with that, I'd say this PR is ready for merging.

mortenpi (approved):

PR #2952, opened 2026-07-24
Make it clearer that `modules` argument is needed for doctests in docstrings

penelopeysm:
When `jldoctest` code block is placed in a docstring it's not run unless `makedocs` is given `modules=[MyModule]`. That's fine, but the only mention of this requirement is under an unassuming heading called 'Module-level metadata'

https://documenter.juliadocs.org/stable/man/doctests/#Module-level-metadata

which makes it pretty hard to find (at least I really struggled to find it).

Besides, I only knew to look for it because I already knew that this sort of thing was possible to do. For a newcomer they might not even realise that it's possible to add a doctest in a docstring.

This PR adds a larger note at the top so that it's just more obvious.


fingolfin (approved):
Thanks!

Issue #2956, opened 2026-08-01
footnote previews cut off code blocks


rokke-git:
another footnotes clipping issue: if there's a code block in the footnote, it correctly preserves whitespace; but since the preview doesn't stay up when hovering over it there's no way to scroll the code block:

<img width="377" height="560" alt="Image" src="https://github.com/user-attachments/assets/77c929f4-24b0-4e94-a8ed-f2f0bcf1be6d" />

fingolfin:
Thanks for the report, definitely something to look into

rokke-git:
in addition to fixing this issue, the other footnote issue also wouldn't be as bad if the footnote stayed up when hovering over the body of the footnote after it popped up. I tried poking around a bit but it looks like the popups are being done with javascript instead of css for whatever reason so I wasn't able to figure out an easy solution

fingolfin:
See PR #2958 for a fix

PR #2958, opened 2026-08-14
Fix footnote rendering in HTML previews and LaTeX output

### LINKED ISSUE #2894: footnote previews in a table get cut off
footnote previews should render above everything else, and not be cut off by the container they live in. this works successfully if the footnote is not in some other container:

Image

but if you put it inside of a table it gets cut off in the most annoying way:

Image

it creates a scrollbar at the bottom of the table; but if you try scrolling, your mouse moves off the footnote and it disappears, causing the scrollbar to no longer be needed, snapping back to the original view. you then instinctively click the footnote to make it sticky, but clicking actually takes you to the bottom of the page instead.

fingolfin:
Fixes #2894 and #2956.

- Previews are now positioned with `position: fixed`, so no scrolling or clipping ancestor (a table, a code block) can cut them off, and they no longer add to a table's scroll width.
- The preview stays open while the pointer is over it, so a long footnote can actually be scrolled.


fingolfin:
Pushed a second commit that adds both cases to the showcase page.

goerz (approved):
Tested this manually. Seems to work fine.

Claude agrees.

LGTM!

PR #2959, opened 2026-08-14
Fix linkcheck CI failures

fingolfin:
Two unrelated linkcheck failures on CI which we just ignore

- `https://www.chiark.greenend.org.uk/~sgtatham/putty/` times out (and I've seen this often in the past)
- `https://codeberg.org` returns 403 from runners (200 from a normal machine).

**`test/online/online_githubcheck.jl`** — the "Failure" testset relied on RegistryInstances being at an untagged version. But that isn't something we can rely on forever. So instead the tests now build a synthetic remote instead, with the desired properties we want to test.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -32,6 +32,10 @@ linkcheck_ignore = [
 if get(ENV, "GITHUB_ACTIONS", nothing) == "true"
     # It seems that CTAN blocks GitHub Actions?
     push!(linkcheck_ignore, "https://ctan.org/pkg/minted")
+    # Times out from Actions runners (curl exit 28).
+    push!(linkcheck_ignore, "https://www.chiark.greenend.org.uk/~sgtatham/putty/")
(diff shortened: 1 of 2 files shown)


docs/make.jl
```diff
@@ -32,6 +32,10 @@ linkcheck_ignore = [
 if get(ENV, "GITHUB_ACTIONS", nothing) == "true"
     # It seems that CTAN blocks GitHub Actions?
     push!(linkcheck_ignore, "https://ctan.org/pkg/minted")
+    # Times out from Actions runners (curl exit 28).
+    push!(linkcheck_ignore, "https://www.chiark.greenend.org.uk/~sgtatham/putty/")
```
mortenpi: I haven't seen the Putty failures myself on CI, so I think it's intermittent. So let's not exclude it just yet.

```suggestion
```

goerz (approved):
LGTM!

mortenpi (approved):

PR #2961, opened 2026-08-14
Let unmatched `@ref` fall through the resolver pipeline

### LINKED ISSUE #2668: `"Header with slug '$slug' is not unique"` error should always be thrown

### LINKED ISSUE #2960: #2678 broke three kinds of working `@ref`
#2678 classifies `@ref` by syntax and makes the header step strict. Three forms that resolved before now error:

1. `[makedocs](@ref)` — un-backticked text naming a docstring; strict ends the pipeline before `Docs` runs.
2. `` [`JULIA_EDITOR`](@ref) `` against `@id JULIA_EDITOR` — backticked text only consults docstrings.
3. Any plugin step, e.g. DocumenterInterLinks' fallback — strict ends the pipeline at `Header`.

Bisect: `9a7549525` green, `16a8cdcb3` red, for Oscar, the Julia manual and DocumenterInterLinks alike.

#2678 was right about one thing. Unmatched slugs used to be parsed as code, so `[Totally Bogus Target](@ref)` resolved to `href="other/#Base.:-"`, silently. Oscar shipped 13 of those (oscar-system/Oscar.jl#6183), Julia 16 (JuliaLang/julia#62718). That must not come back.

Suggested fix: let `Header` decline when it finds no anchor and continue down the pipeline, reporting only if every step declines. `Base.:-` stays rejected because no step claims it.

## Case sensitivity

Raised in oscar-system/Oscar.jl#6183. Nothing changed here: `slugify` is untouched since #2325 and there is no case folding anywhere. On `16a8cdcb3^`, `[Simplicial complexes](@ref)` against `## Simplicial Complexes` resolved to `Base.:-`, never to the header.

Whether `@ref` *should* fold case is open. Slugs are also HTML anchors and inventory keys, so it would reach `@id` and `@extref` too. Cheaper alternative: keep matching exact, and name the case-insensitive near miss in the error.

## Minor

`namedxref` emits "does not exist" and "is not unique" together, and names the page holding the link, which reads as if the header had to live there.

fingolfin:
Fixes #2960.

In PR #2787 (merged October 2025, already released) we made the header step stop the pipeline when it could not resolve a slug. That was meant to fix #2668, where Documenter silently accepted invalid refs. In hindsight that was a mistake (CC @lgoettgens): it violates the principle @goerz expressed, that no resolver can "own" a reference, since plugins may add resolvers for existing syntax.

Then in PR #2678 (merged March 2026, not yet released) we changed @ref to classify targets strictly by syntax.

Together they break three forms that resolved before: ``[makedocs](@ref)``, ``[`JULIA_EDITOR`](@ref)`` naming a header whose title is code, and anything a plugin resolves.

With the changes in this PR, no step stops the pipeline any more. A step that cannot resolve a link records why and leaves it alone; only the pipeline as a whole gives up. This effectively reverts #2787, which also rejected references that resolve correctly (#2843) and kept plugin resolvers from running.

Plain link text is a header reference only. Following @goerz's review, ``[somefunction](@ref)`` is not a docstring reference; that must be written ``[`somefunction`](@ref)``. This keeps #2668 fixed by construction rather than by inspection: plain text is never looked up as code, so ``[Blow-Ups](@ref)`` can no longer parse as `Blow - Ups` and find the docstring of `-`. With the fallback gone, `Docs` runs before `Header`, `Header` absorbs the backticked-header case, and the pipeline is three steps again.

It seems there are no issues with this change in the Julia manual, and in the OSCAR manual just two references were broken (see https://github.com/oscar-system/Oscar.jl/pull/6196)

Note that `Selectors.order` values change: Header 1.0 → 3.0, Docs 3.0 → 1.0. DocumenterInterLinks is unaffected (its fallback resolver is at 10.0), but a plugin that slotted itself between the old values might be in trouble.

Prepared with AI assistance (Claude).


goerz:
> Header now stops only when it must own the reference

Not sure if this is just badly phrased, but what I was trying to say in https://github.com/JuliaDocs/Documenter.jl/issues/2960#issuecomment-5294180799: no resolver can "own" a reference, if we allow plugins to add additional resolvers for existing syntax. Which is what the ["fallback resolution" of `DocumenterInterLinks`](https://juliadocs.org/DocumenterInterLinks.jl/dev/fallback/) does. All a particular resolver knows it "I can't resolve this".

fingolfin:
@goerz @mortenpi this PR should now be "slop-free" (that's not saying it is great), or at least hopefully close enough... but before I try to polish it further, it makes senes to get some fresh feedback from you. You may not even have to look at the code or read the description, just this:

So this PR tries to reduce the impact of PR #2787 (already released) plus #2678 (not yet in a release).

But a major source of complexity in here comes from trying to still support `[somefunction](@ref)` where `somefunction` is the name of a function. By our "new" rules, this should have been written as ``[`somefunction`](@ref)``. 

Do we want to enforce this? In Julia and OSCAR it required fixing a small number of places, but otherwise it was fine. It may annoy more package authors, though.

*If* we say we are willing to require this, I could simplify the code in here, by switching the order in which we evaluate `XRefResolvers.Header` vs. `XRefResolvers.Docs`:  if we evaluate `XRefResolvers.Docs` first, then I can get rid of `XRefResolvers.HeaderFallback` (added in this PR) again.

goerz:
`[somefunction](@ref)` where `somefunction` is the name of a function should not be supported. It’s fine to break existing documentation: that pattern is a bug in that existing documentation (it formats incorrectly). They should be happy that this is now flagged.

Somewhat relatedly:

```
[text](@ref somefunction)
```

would be better as

```
[text](@ref `somefunction`)
```

I think we may already have implemented support for the second form, but I think we should transition to documenting it as the recommended syntax. We’ll have to retain support for the old syntax for backwards compatibility, at least until a 2.0 release

fingolfin:
@goerz agreed! that makes a lot of sense to me. I've updated the PR now, and in particular the documentation.

.github/workflows/regression-tests.yml
```diff
@@ -74,7 +74,6 @@ jobs:
           Pkg.develop(Pkg.PackageSpec(path=pwd()))
           # special hack to make OSCAR tests work *sigh*
           if package == "Oscar"
-            #AbstractAlgebra, Nemo, Hecke, Singular, GAP, Polymake
```
fingolfin, line 77: This is a hack I made to force regression tests to run on this PR. I reveals the expected breakage in OSCAR (until https://github.com/oscar-system/Oscar.jl/pull/6196 is merged there).

goerz: I don't understand. You removed a comment line?

docs/src/man/syntax.md
```diff
@@ -201,19 +201,29 @@ makedocs
 ... [#42](@ref) ...
 ````
 
-The reference target is classified from the link syntax:
+The reference target is initially classified from the link syntax:
```
goerz: I would not add the word "initially". It's just confusing for most people. Most users don't need to worry about the internals of the resolver systems or how the interact with plugins

.github/workflows/regression-tests.yml
fingolfin: yes, I removed a comment line, so that this file is modified; the regression tests are *only* run if either...
1. CI is running on `master` or a `release-*` branch (so not a PR), or
2. if one of `.github/workflows/regression-tests.yml` or `Project.toml` is modified

So this modification is a hack to trigger condition 2 and get the regression tests to run on this PR.

Of course that's not a nice way to handle things, which is why I made PR #2967

docs/src/man/syntax.md
```diff
@@ -201,19 +201,29 @@ makedocs
...
 
 | Reference type | Implicit form | Explicit form |
 | --- | --- | --- |
 | Header by title | `[Header name](@ref)` | `[link](@ref "Header name")` |
-| Header by `@id` | n/a | `[link](@ref header-id)` |
+| Header by `@id` | only if the text is the label itself | `[link](@ref header-id)` |
 | Issue/PR | `[#42](@ref)` | `[link](@ref #42)` |
 | Docstring | ``[`Example.domath`](@ref)`` | ``[link](@ref `Example.domath`)`` |
```
goerz:
```suggestion
| Docstring | ```[`Example.domath`](@ref)``` | ```[link](@ref `Example.domath`)``` |
```

goerz: Two backticks are "math" in Julia markdown

docs/src/man/syntax.md
goerz:
```suggestion
A reference that is a fully qualified name (e.g. ```[`Example.domath`](@ref)```, or ```[`domath`](@ref `Example.domath`)``` will also be resolved in `Main`.
```

docs/src/man/syntax.md
```diff
@@ -229,7 +239,8 @@ current page using the same syntax.
 ### Named `@ref`s
 
 It is also possible to override the destination of an `@ref`-link by adding the appropriate
-label to the link, such as a docstring reference or a page heading.
+label to the link, such as a docstring reference or a page heading. However, the unquoted
+form in the third reference below is best avoided; see the compatibility note below.
```
goerz:
```suggestion
label to the link, such as a docstring reference or a page heading.
```

docs/src/man/syntax.md
```diff
@@ -229,7 +239,8 @@ current page using the same syntax.
...
 
 ```markdown
 All three of the following references point to `g` found in module `Main.Other`:
```
goerz:
```suggestion
The following references point to `g` found in module `Main.Other`:
```

docs/src/man/syntax.md
goerz:
```suggestion
* [`Main.Other.g`](@ref)
* [the `g` function](@ref `Main.Other.g`)
```

docs/src/man/syntax.md
```diff
@@ -261,10 +272,10 @@ Use [`for i = 1:10 ...`](@ref for) to loop over all the numbers from 1 to 10.
...
 !!! note "Compatibility"
 
-    The explicit unquoted docstring form `[text](@ref Example.domath)` remains supported for
-    backward compatibility. However, header references still take precedence, including
-    labels introduced with [`@id`](@ref at-ref-at-id-links), so the explicit backticked form
-    is less ambiguous.
+    Prefer the backticked form ```[text](@ref `Example.domath`)```. The unquoted form
+    `[text](@ref Example.domath)` remains supported for backward compatibility, but it is
+    ambiguous: an unquoted target that matches a header label, including one introduced with
+    [`@id`](@ref at-ref-at-id-links), resolves to that header rather than to the docstring.
```
goerz:
```suggestion
    For backwards compatibility, Documenter also resolves `[text](@ref Example.domath)` without backticks around `Example.domath`. Prefer the backticked form ```[text](@ref `Example.domath`)```. The legacy syntax is ambiguous: an unquoted target that matches a header label, including one introduced with [`@id`](@ref at-ref-at-id-links), resolves to that header rather than to the docstring.
```

src/cross_references.jl
```diff
@@ -152,19 +152,22 @@ module XRefResolvers
...
 
-    - [`XRefResolvers.Header`](@ref) for links like `[Section Header](@ref)`
-    - [`XRefResolvers.Issue`](@ref) for links like `[#11](@ref)`
     - [`XRefResolvers.Docs`](@ref) for links like ```[`Documenter.makedocs`](@ref)```
+    - [`XRefResolvers.Issue`](@ref) for links like `[#11](@ref)`
+    - [`XRefResolvers.Header`](@ref) for links like `[Section Header](@ref)`
+
+    A step only matches the link syntax it serves. Plain text is a header reference and
+    nothing else; backticked text is a docstring reference first and a header second, since
+    a header title may itself be code.
```
goerz, line 161: I had some doubts about documenting backticks in the ```[`Header`](@ref)``` syntax as supported. Ultimately, I'm coming down on this being fine.

However, what is more important (not so much in the documentation, but we should have a test for it): ```[text](@ref `manual-header-id`)``` should work. I'd want to bring this fully in line with ["syntax 3" of `DocumenterInterLink`](https://juliadocs.org/DocumenterInterLinks.jl/stable/syntax/#Syntax):

src/cross_references.jl
```diff
@@ -196,6 +199,10 @@ module XRefResolvers
 
     This runs if the `slug` corresponds to a known local section title, and resolves the
     `node` to link to that section.
+
+    It also makes a second attempt at backticked link text. That syntax ordinarily
+    references a docstring, which is why [`XRefResolvers.Docs`](@ref) gets it first, but a
+    header title may itself be code, as in ```### [`JULIA_EDITOR`](@id JULIA_EDITOR)```.
```
goerz: Using a header with a manual ID is not a good example in this context. This is pretty much a bug in the Julia manual. The way they _should_ have written that has always been

```
### [`JULIA_EDITOR`](@id env-var-julia-editor)

...

See the [`JULIA_EDITOR`](@ref env-var-julia-editor)
```

I'd like to support ```[`JULIA_EDITOR`](@ref `env-var-julia-editor`)``` as well, because otherwise text editors syntax-highlight the text with spell checking, which is not ideal.

What we're presumably trying to support now is

```
### `JULIA_EDITOR`

...

See the [`JULIA_EDITOR`](@ref)
```

Which is neat to work out of the box.

But that's probably the example we should use (and, we should have a test for that exact syntax)

test/docsxref/src/index.md
```diff
@@ -47,6 +52,8 @@ Documenter.hide
 
 ## Two words
 
+## [`ENVVAR`](@id ENVVAR)
```
goerz: See my previous comment.

Just ```## `ENVAR````?

goerz (changes requested):
Look pretty good!

I made some suggestions. In particularly, I'm more strongly discouraging the legacy `[text](@ref function)` without backticks.

docs/src/man/syntax.md
goerz: I might have screwed up the review UI. What I was trying to do: don't advertise the legacy `[domath](@ref Example.domath)` syntax. Leave that to the compatibility callout farther down

src/cross_references.jl
fingolfin: We could look at that, but this is now a feature request, and so I'd really prefer to not mix that into this PR. It could be in a follow-up PR by you, though, and then we can discuss it there.

goerz: Yeah, we shouldn’t add features in this PR. This was in case our implementation already handles this. But even then, I agree that we can add tests or documentation later on.

fingolfin:
Applied all suggestions, except for one which amounts to a behavior change that's beyond the scope of this PR.

fingolfin:
@goerz if you agree with the current version, please approve :-)

goerz (approved):

PR #2964, opened 2026-08-20
Drop trailing slash from documenter/deploy status URL

jkrumbiegel:
I found that if this URL ends with a slash, redirect on privately hosted documentation to the random url that github uses doesn't work. For public ones, it doesn't seem to matter which one it is.

Code was generated with Claude Opus 5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed headers hidden by the navbar. ([#2905])
 * Tightened `@ref` classification so references are matched more consistently by syntax, with header labels still taking precedence over docstrings and mistaken header/id references less likely to resolve as docstrings. ([#2678])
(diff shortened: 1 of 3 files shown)


fingolfin:
> I found that if this URL ends with a slash, redirect on privately hosted documentation to the random url that github uses doesn't work. 

I can't seem to parse this sentence. What redirect? hosted where? I am lost.

The difference in handling between a trailing slash or not seems a question of how the web server is configured where the URL goes to. It is not clear to me whether there is an actual issue here in Documenter, that warrants a fix; or an issue in the configuration of your webserver. Perhaps you could provide a few more details?


> For public ones, it doesn't seem to matter which one it is.

I think that means: "for public GitHub (?) repositories, it doesn't matter whether the trailing slash is there or not", right?

jkrumbiegel:
What I meant was, if you use thr pages feature with private github repositories, the link to the deployed site doesn't work because the servers github redirects to for private pages are not configured to work with such trailing slash urls like the public pages are. As both work with the non-trailing slash version, I thought it would be convenient to change this default.

PR #2966, opened 2026-08-24
Break large LaTeX tables across pages

### LINKED ISSUE #2963: Large Markdown tables are omitted or fail LaTeX PDF builds
Documenter’s LaTeX writer renders every Markdown table as a non-breaking `table` containing `tabulary`:

https://github.com/JuliaDocs/Documenter.jl/blob/master/src/latex/LaTeXWriter.jl#L716-L735

This prevents tables from spanning pages. Sufficiently large tables are either omitted from the generated PDF or, with LaTeX 2026-06-01, cause compilation to fail with `Dimension too large`.

This currently breaks the Julia manual PDF pipeline:

- Julia issue: https://github.com/JuliaLang/julia/issues/62787
- First failing build: https://github.com/JuliaLang/docs.julialang.org/actions/runs/26791202219

## Minimal example

An `index.md` containing:

```markdown
# Large table

​```@eval
using Markdown

entries = Any[Any[
    ["Column"],
    ["Value"],
]]

for i in 1:4000
    push!(entries, Any[
        [string(i)],
        ["x"],
    ])
end

Markdown.MD([Markdown.Table(entries, [:l, :l])])
​```
```

and a standard native LaTeX build:

```julia
using Documenter

makedocs(
    sitename = "Large table reproducer",
    format = Documenter.LaTeX(platform = "native"),
)
```

## Actual behavior

With LaTeX2e 2026-06-01, compilation fails at the end of the generated table:

```text
! Dimension too large.
<to be read again>
+
l. ... \end{tabulary}
```

LaTeX 2026 changed centered tabulars to use a text-mode `\vcenter`. Its implementation calculates the complete table height plus depth in a dimension register, exposing the overflow:

https://github.com/latex3/latex2e/commit/062a0538113f70d131d2740ac8ec105a106ad563

With the older LaTeX implementation, compilation can succeed, but the oversized float is not rendered. For example, Julia 1.12.6’s PDF manual contains the Unicode Input chapter introduction and warning, then immediately starts the next chapter—the entire 3,724-row table is absent.

Pinning an older TeX Live version therefore restores compilation but does not restore the missing table.

## Expected behavior

Large Markdown tables should span multiple PDF pages, ideally with their header row repeated. At minimum, Documenter should not silently omit them.

Potential approaches could include using `longtable`, `xltabular`, or selecting a page-breaking environment once a table exceeds an appropriate threshold.

fingolfin:
Every Markdown table was rendered as a `table` float wrapping `tabulary`, which cannot break across pages. LaTeX silently drops such a float when it overflows a page, and since LaTeX 2026-06-01 the build aborts with "Dimension too large" instead. The 3,724-row Unicode Input table is missing from the Julia 1.12.6 PDF manual for this reason.

Tables with at least 20 rows now use `xltabular` (longtable + tabularx), which breaks across pages and repeats the header row. The `X` column type keeps long cells wrapping, which plain `l`/`c`/`r` would not.

The LaTeX showcase gains a table too long for one page, so the behaviour can be inspected in the built PDF.

Fixes #2963

Prepared with AI assistance (Claude).

Here is the showcase PDF produced with this PR, the table starts on page 8.
[showcaseafter.pdf](https://github.com/user-attachments/files/31365595/showcaseafter.pdf)


mortenpi (approved):
I did not deeply go into this one, but it seems fine to me, and seems to do the right thing. So LGTM.

mortenpi:
I might consider this a "feature" though.

PR #2967, opened 2026-08-24
Allow forcing regression tests on a PR via a label

fingolfin:
The regression tests currently only run on pull requests touching `regression-tests.yml` or `Project.toml`, with no way to opt a PR in.

With this PR, it becomes possible to run the regression test suite on a PR: adding the label `Run Regression Tests` to an open PR starts a run.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -10,12 +10,34 @@ on:
       - '*/*-runci'
     tags: '*'
   pull_request:
-    paths:
-      - '.github/workflows/regression-tests.yml'
-      - 'Project.toml'
+    # 'labeled' so that adding the opt-in label to an open PR starts a new run
+    types: [opened, synchronize, reopened, labeled]
 jobs:
+  # The path filter that used to live on the `pull_request` trigger lives here
+  # instead: a trigger-level filter would keep the workflow from starting at
+  # all, leaving no place to check for the opt-in label.
+  gate:
+    name: 'Gate'
+    runs-on: ubuntu-latest
+    outputs:
+      run-tests: ${{ steps.decide.outputs.run-tests }}
+    steps:
+      - uses: dorny/paths-filter@v3
+        id: filter
+        if: github.event_name == 'pull_request'
+        with:
+          filters: |
+            ci:
+              - '.github/workflows/regression-tests.yml'
+              - 'Project.toml'
+      - id: decide
+        run: |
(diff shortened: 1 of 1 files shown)


fingolfin:
The regression test failures in the Julia manual and in DocumenterInterLinks are real, but not caused by this PR; rather they are the motivation for this PR. They are fixed by https://github.com/JuliaDocs/Documenter.jl/pull/2961

goerz:
I like the `Run Regression Tests` label to opt PRs into regression testing. I wouldn't make a distinction between "cheap" and "heavy" regression tests. Just run all regression tests on `master` and on any PR that applies that label. That's from a perspective of not wasting CI minutes unnecessarily, as well as keeping the setup reasonably simple

fingolfin:
Honestly, if it was up to me, I'd run *all* the regression tests, *always*: because honestly, I am pretty sure we'll forget quite often to make use of this feature. Add in that nobody really checks the status of the regression tests, and (AFAIK) there is no automation in place to signal to us if it fails, and we get the situation we had now: a PR was merged that broken Julia, Oscar and DocumenterInterLinks and we didn't even notice for months.

So I really think we'd be better of not relying on a human thinking "hey perhaps the regression tests should be run here...

That said, perhaps as a compromise, we could enhance the regression test workflow to open an issue as soon as it fails (and close it when it starts working again). That'd probably be a good idea no matter what, as long as we don't run all regression tests on all PRs.... 

In fact, might be a good idea to also run those tests on a schedule (daily or weekly) to also catch regressions caused by external changes (such as: Julia breaking its manual; a GH CI provider for texlive breaking; broken links; Julia changing something about how the docsystem works; etc.).  In my experience it helps a lot to know at least roughly *when* something broke -- i.e. if I know "the Julia manual started failing on day X when it worked on X-1, then it's much easier to pinpoint the cause than when I know "it started to fail at some point in the past 3 months".

mortenpi:
My two cents here: I share the concern that running regression tests on every push is wasteful. Our CI is already pretty massive, even without it.

I think just an opt-in label would be good. And I would  remove the complexity of "cheap" and "expensive" regression tests.

Weekly schedule seems like a good idea too. Problem is notification. Maybe it can open an issue if the weekly check fails? And I guess preferably re-open an existing issue, if that's feasible, so that the automation doesn't start opening too many new issues all the time.

fingolfin:
@goerz @mortenpi changed this as discussed to only check for the labels. Will make a follow-up PR that does the notifications.

goerz (approved):

PR #2968, opened 2026-08-24
Clarify SSH deploy key setup for GitHub Actions

### LINKED ISSUE #2403: Out-of-repo deployment doesn't work for me
I'm trying to move `https://github.com/JuliaGraphics/Luxor.jl` docs to `https://github.com/JuliaGraphics/LuxorManual`. I followed the instructions in the Documenter docs [here](https://documenter.juliadocs.org/stable/man/hosting/#Out-of-repo-deployment).

The two keys were added to the correct locations, as far as I can tell, and the `make.jl` was updated with this:

```julia
repo = "github.com/JuliaGraphics/LuxorManual.git"
withenv("GITHUB_REPOSITORY" => repo) do
    deploydocs(
      repo = repo,
      target = "build",
      push_preview = true,
      forcepush = true,
    )
end
```

The error message [here](https://github.com/JuliaGraphics/Luxor.jl/actions/runs/7543161131/job/20533536894#step:7:147) says:

```
fatal: repository 'https://github.com/github.com/JuliaGraphics/LuxorManual.git.git/' not found
```

so it's possible the `make.jl` code I added is wrong, but not sure how...?

fingolfin:
The deploy key instructions lived in the Travis CI section and quoted a `genkeys` output that no longer matches what DocumenterTools prints. Users following them had to guess where the private key belongs in GitHub's settings, where "Secrets and variables" offers repository, environment and organization secrets plus a separate variables tab. Picking the wrong one yields a 403 from `github-actions[bot]`, because Documenter silently falls back to `GITHUB_TOKEN`.

Move the instructions to the GitHub Actions section, name the exact settings pages for both halves of the key pair, and spell out which of the similar-looking boxes is the right one. Map the 403 error back to its cause. State for out-of-repo deployment which repository receives which half, and that `GITHUB_TOKEN` cannot work there at all.

Resolves #2403, IMHO (but CC @cormullion who might disagree)

@mortenpi should this get a changelog entry?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Change type of `HTML.assets` field to `Vector{HTMLHeadContent}` to allow configuring dynamically in plugins. ([#2925])
 * JuliaMono has been updated from `v0.050` to `v0.63.2`. ([#2953])
+* The SSH deploy key instructions in the manual now cover GitHub Actions directly, naming the settings pages for both halves of the key pair and disambiguating repository secrets from variables and environment secrets. ([#2403])
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


goerz (approved):

cormullion:
I don’t remember much about this - Michael
coaxed to me to a solution at the time but I’ve since forgotten it all - so I don’t disagree with anything. 😅

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Change type of `HTML.assets` field to `Vector{HTMLHeadContent}` to allow configuring dynamically in plugins. ([#2925])
 * JuliaMono has been updated from `v0.050` to `v0.63.2`. ([#2953])
+* The SSH deploy key instructions in the manual now cover GitHub Actions directly, naming the settings pages for both halves of the key pair and disambiguating repository secrets from variables and environment secrets. ([#2403])
```
mortenpi, line 16: > should this get a changelog entry?

FWIW, my policy has been not to add any changelog entries to docs changes, since they are non-functional.

PR #2970, opened 2026-08-24
Add a maintained TeX Live container image

### LINKED ISSUE #2073: Migrate LaTeX Docker image to GCHR

fingolfin:
The image behind `Documenter.LaTeX(platform = "docker")` is built from an [archived repository](https://github.com/JuliaDocs/DocumenterLaTeX.jl/blob/master/docker/Dockerfile) on Ubuntu 18.04 and was last updated seven years ago. It never installed the DejaVu fonts that documenter.sty selects, so the PDF silently falls back to substitutes.

Add a replacement built from `docker/` in this repository and published to ghcr.io/juliadocs/documenter-latex. Keeping it here lets the TeX package list be a single file shared with the native `install-texlive` step in CI, and lets a Dockerfile change be validated by Documenter's own PDF test suite before it is published rather than after.

TeX Live is installed from CTAN rather than from Debian, whose texlive-* packages trail upstream by a year or more. collection-latexextra is included because the predecessor was built on Debian's texlive-latex-extra, and dropping it would silently break custom preambles that work today.

Nothing points at the new image yet, so that it can be published and made public before Documenter starts pulling it.

Resolves #2073.
Closes #2092.

## Action required from a JuliaDocs org admin

The first run of `DockerImage.yml` creates `ghcr.io/juliadocs/documenter-latex` as a **private** org-level package. Someone with admin rights (so not me) must then set it to public (Package settings → Danger Zone → Change package visibility). This is a one-time step; later pushes keep the setting. It is not optional: nothing pulling this image authenticates to GHCR — not the `docs` PDF job here, and not any user running `platform = "docker"` — so while the package is private, every PDF build fails to pull it. If the org restricts package creation or public visibility under Settings → Packages, that needs relaxing first.

**Please do not merge #2971 until this package is published and public**, or its
`Documentation: pdf` job will fail.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -19,9 +19,19 @@ Documenter version.
 | JuliaMono | not used by default; available to custom preambles |
 | Pygments (`pygmentize`) | the `minted` backend for source highlighting |
 
-`texlive-packages.txt` is shared with the `zauguin/install-texlive` step in
-`.github/workflows/CI.yml`, so the containerised and natively installed TeX
-toolchains cannot drift apart.
+[`texlive-packages.txt`](texlive-packages.txt) is shared with the
+`zauguin/install-texlive` step in `.github/workflows/CI.yml`, so the containerised
+and natively installed TeX toolchains cannot drift apart. Keep it a bare list, one
+package per line: that action passes the file to `tlmgr` verbatim, so a `#` comment
+in it becomes a package name and the install silently comes out wrong. Its entries
+are, and why:
+
+| Package | Why |
+| --- | --- |
+| `scheme-medium` | `latexmk`, `lualatex`, `fontspec`, `polyglossia`, `amsmath`, `graphicx`, … |
+| `collection-latexextra` | The predecessor image was built on Debian's `texlive-latex-extra`; dropping it would silently break custom preambles that work today. Also supplies `mwe`, whose `example-image` the cover page test uses. |
+| `hyperref`, `minted`, `newunicodechar`, `tabulary`, `tcolorbox` | Requested explicitly by `assets/latex/documenter.sty` |
+| `transparent`, `upquote` | Pulled in by generated documents rather than by `documenter.sty` |
 
 ## Using it
 
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,70 @@
+# `ghcr.io/juliadocs/documenter-latex`
+
+A TeX toolchain for Documenter's PDF backend, used by
+
+```julia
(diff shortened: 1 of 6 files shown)


docker/README.md
```diff
@@ -0,0 +1,80 @@
...
+* `:latest`.
+
+To build it locally:
+
+```bash
+docker build -t documenter-latex:test docker/
+```
+
+The predecessor, `juliadocs/documenter-latex:0.1` on Docker Hub, is unmaintained;
+its Dockerfile lives in the archived `JuliaDocs/DocumenterLaTeX.jl` repository.
```
fingolfin, line 80: Probably could remove this.
```suggestion
```

goerz (approved):

fingolfin:
It'd be good to also here from @mortenpi

Also, in the end, after merging this, apparently someone with admin rights for the JuliaDocs org will have to click something somewhere (see the "Action required from a JuliaDocs org admin" block in the PR description), so e.g. @mortenpi @ViralBShah @fredrikekre ...

ViralBShah:
@fingolfin This looks good directionally to me -  but this stuff can only be verified once the PDFs come out. Presumably not much risk - since easy to revert or fix issues.

I have also made you owner in JuliaDocs.

fingolfin:
Thanks @ViralBShah 

After some discussion with @goerz and @mortenpi I promised I'll write up properly how I think this should all work before we commit to anything. But it'll be a bit longer till I get to that.

One of the things to ponder is whether to do what this PR does, and regularly (weekly) update the docker image with security updates etc; or whether to aim for a more static image (as we are doing right now, sadly -- 7 years without changes isn't great). The latter is appealing in terms of reproducibility; and security updates may not be so relevant in that context anyway...

But I need to ponder it, and it may be some weeks or even months till I have time to return to this.

ViralBShah:
I feel that if this works, we should at least get it in to have a new image since that would be a strict improvement.

PR #2974, opened 2026-08-25
Report regression test failures on master as an issue

fingolfin:
Follows #2967.

Two gaps today: regression tests only run when something is merged, and a
failure is only visible to whoever thinks to look at the workflow's run
history. Breakage in a downstream package can sit unnoticed for a long time,
and breakage introduced by a downstream package itself is never caught at all,
since nothing on our side triggers a run.

This adds:

- **A weekly scheduled run** (Mondays, 05:17 UTC). Scheduled runs always use
  `master`.
- **A `report-status` job** that opens an issue when the regression tests fail
  on `master`. The issue carries the label
  `Status: Regression Tests Failing`, which is how a later run finds it: a
  repeated failure updates the existing issue instead of opening a second one,
  and a passing run comments and closes it. The label is created on demand, so
  there is no repo setup step.

A failure that is still open gets a reminder comment only from the weekly
scheduled run — so a long-standing breakage nags once a week, rather than once
per merge.

All of this is restricted to `master` (`github.event_name != 'pull_request' &&
github.ref == 'refs/heads/master'`), so a pull request neither opens an issue
nor closes one that is already open.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -10,12 +10,34 @@ on:
       - '*/*-runci'
     tags: '*'
   pull_request:
-    paths:
-      - '.github/workflows/regression-tests.yml'
-      - 'Project.toml'
+    # 'labeled' so that adding the opt-in label to an open PR starts a new run
+    types: [opened, synchronize, reopened, labeled]
 jobs:
+  # The path filter that used to live on the `pull_request` trigger lives here
+  # instead: a trigger-level filter would keep the workflow from starting at
+  # all, leaving no place to check for the opt-in label.
+  gate:
+    name: 'Gate'
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


goerz:
> Worth deciding before merge: whether the "still failing" comment on every
failing run is useful or just noise

If it's once a week, I would consider it useful. Especially because these days, regression failures should be pretty actionable without anyone having to spend much time on them. Your favorite frontier model LLM should be doing very well with resolving such failures.

mortenpi:
> > Worth deciding before merge: whether the "still failing" comment on every
> > failing run is useful or just noise
> 
> If it's once a week, I would consider it useful.

I concur. I think it's good if a bot keep bugging us on a weekly basis. Do we want to make the change? This LGTM as is as well though.

fingolfin:
Updated so that it "nags" once a week if CI is still failing (if not, the issue is closed)

mortenpi (approved):

PR #2976, opened 2026-08-25
fix: updated minisearch version and cdn source

### LINKED ISSUE #2212: Get minisearch.js to cdnjs
I noticed it currently comes from jsdelivr, but it would be nice if we'd have them all on one CDN. A starting point would probably to open a library request on cdnjs.

cc @Hetarth02

### LINKED ISSUE #2973: Update minisearch version
Thanks to @fingolfin, minisearch is now available on `cdnjs`. (Ref: #2212)

But the earliest version available is `6.2.0`, and the we use `6.1.0`.

If we to cdnjs, we also need to upgrade the version.

From what I have checked the upgrade doesn't break anything. (Ref: https://github.com/lucaong/minisearch/blob/master/CHANGELOG.md#v620---2023-10-26)

Hetarth02:
Fixes #2973 
Fixes #2212


fingolfin (approved):
Thanks!

Perhaps we should also update to the latest version, but that should be in a separate PR.

PR #2977, opened 2026-08-25
HTMLWriter: set a no-referrer policy by default

### LINKED ISSUE #2391: Prevent leaking trackable pageview stats to CDNs by default

fingolfin:
The CDNs serving the fonts, stylesheets and scripts see the full URL of the page a reader is on, which makes per-page view counts trivially collectable by a third party.

Emit a document-level `<meta name="referrer">` tag rather than a `referrerpolicy` attribute on the individual `<link>`/`<script>` tags: most of the JS dependencies are pulled in at runtime by requirejs, and those requests inherit the document policy, not the policy of the tag that loaded requirejs.

The policy is configurable via `HTML(referrerpolicy = ...)`, since a document-level policy also strips the referrer from outbound links that the reader clicks; `nothing` omits the tag entirely.

Closes #2391.
Closes #2401.

---

Note: `no-referrer` is document-wide, so readers clicking from the docs to GitHub, Zenodo, etc. arrive with no referrer. In PR #2401, @mortenpi suggested `strict-origin` as an alternative. That would preserve site-level attribution while still hiding the page. I went with `no-referrer` because that's what #2391 asks for, and the keyword makes it a one-line override.


goerz (approved):
LGTM

PR #2978, opened 2026-08-25
Fix Biome tooling for JS formatting

fingolfin:
CI ran `make format-js`, i.e. `biome check --write`, which rewrites files in the runner and exits 0, so formatting violations could never fail the build. Add a `check-js` target that reports without rewriting and propagates Biome's exit code, and use it in CI.

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -11,9 +11,6 @@
     "ignoreUnknown": false,
     "includes": [
       "biome.jsonc",
-      // Hidden directories may hold nested checkouts (e.g. git worktrees) with
-      // their own biome.jsonc, which Biome rejects as a second root config.
-      "!.*",
       "assets/html/**",
       "!!assets/html/themes",
       "!!assets/html/scss/highlightjs"
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -11,6 +11,9 @@
     "ignoreUnknown": false,
     "includes": [
       "biome.jsonc",
+      // Hidden directories may hold nested checkouts (e.g. git worktrees) with
+      // their own biome.jsonc, which Biome rejects as a second root config.
+      "!.*",
       "assets/html/**",
       "!!assets/html/themes",
       "!!assets/html/scss/highlightjs"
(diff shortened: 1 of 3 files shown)


biome.jsonc
```diff
@@ -11,6 +11,9 @@
     "ignoreUnknown": false,
     "includes": [
       "biome.jsonc",
+      // Hidden directories may hold nested checkouts (e.g. git worktrees) with
+      // their own biome.jsonc, which Biome rejects as a second root config.
+      "!.*",
```
mortenpi: This is meant to exclude anything that starts with a `.`, right? If so, it doesn't really make sense to me. Such directories can contain valid content (case in point, `.github`; but also, we may have more such directories in the future, including maybe `.claude`, for repo-specific skills etc). So I don't think this is the right config, unless I misunderstand?

fingolfin: I'd be happy to change it to
```suggestion
      // Ignore nested worktrees
      "!.claude/worktrees",
```
which is my specific use case.

I just thought it might be nice to make it more generic to also cover other agents and tools.

This does *not* prevent us from formatting files in . dirs, though; e.g. if we want to format things in `.github`, we'd just have to add it to this list. In fact since it is *not* on that list right now, we don't format anything in there, regardless of whether or not we add the `!.*` specifier.

fingolfin: @mortenpi just tell me which way you prefer to resolve this and I'll happily do it

fingolfin: Let's drop this from this PR so we can merge the rest, and then I can open a new one for this change (or an alternative one for the same purpose)
```suggestion
```

goerz:
Stupid question, but can't we just [give biome the list](https://biomejs.dev/guides/getting-started/#command-line-interface) of `.js` files we want it to format?

We can use `git ls-files | grep '\.js$'` or something like that in the `Makefile` and CI scripts.

fingolfin:
@goerz Biome also format CSS and other things, and we still need an exclude list for that.  And it wouldn't solve my problem either.

mortenpi (approved):

PR #2981, opened 2026-08-28
docs/src/man/hosting.md: update example GitHub workflows

fingolfin:


mortenpi (approved):

mortenpi:
It would be nice to somehow have dependabot keep this up-to-date too. Not sure that's feasible if we'd have this in a separate file that we'd dynamically include.

PR #2983, opened 2026-08-29
Fail `@contents`/`@index` blocks that do not parse

### LINKED ISSUE #1140: Partial parsing of a broken at-block
The following at-contents block with a syntax error (missing comma)

````markdown
```@contents
Pages = [
    "linked_list.md",
    "mutable_linked_list.md"
    "intset.md",
    "sorted_containers.md",
]
```
````

will print an error message about the syntax into the terminal, but still partially parses and outputs a partial table of contents. It probably should fail completely.

fingolfin:
A syntax error in the block body was reported, but the block was still expanded using the default settings. For `@contents` that meant a table of contents listing the whole document, silently unrelated to what the `Pages` assignment was meant to select.

`parseblock` gains a `parse_error_result` keyword so callers can tell a broken block from an empty one; `buildnode` uses it to return `nothing`, and the two runners then leave the block as the code block it already is.

Fixes #1140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


goerz (approved):
LGTM

PR #2986, opened 2026-08-29
Order `@contents` by the `pages` given to `makedocs`

### LINKED ISSUE #936: `@contents` does not respect the order of `pages` passed to `makedocs`
In https://tkf.github.io/julia-python-snippets/dev/, I noticed that the order of [`pages` specified in `make.jl`](https://github.com/tkf/julia-python-snippets/blob/7df6bcdfb9d7853b62d5b9e8b0e1a7d18cec0bf3/docs/make.jl#L18-L23) is reflected only to the sidebar:

> image

and _not_ to the table of contents:

> image

generated by:

    ```@contents
    ```

Is it possible to fix `@contents` to respect the order of `pages`?

The Travis build is using [Documenter v0.21.1](https://travis-ci.com/tkf/julia-python-snippets/builds/99538708#L550).  I tried using Documenter master 43cb56afcda0d5371c30f3497270894d25c18a82 locally but the result was the same.

fingolfin:
`@contents` entries were sorted by the anchor order assigned during expansion, which walks the pages sorted by filesystem path. A block without an explicit `Pages = [...]` therefore listed the pages alphabetically, disagreeing with the sidebar and the document body.

Fall back to the navigation order when `Pages` is not given. An explicit `Pages` still takes precedence.

The `latex_showcase` references change accordingly: that build declares `pages = ["showcase.md", "docstrings.md"]`, so its table of contents no longer starts with the docstrings chapter.

Fixes #936


goerz (approved):
Strongly approve this feature!

Code seems fine to me, at a glance

PR #2987, opened 2026-08-30
Keep the global `meta` for crossref and rendering

fingolfin:
The `meta` keyword of `makedocs` supplies default values for the `@meta` blocks of every page, but the cross-reference stage started each page by emptying `page.globals.meta`. The defaults were then only restored from `@meta` blocks already in the page — that is, they survived only on pages carrying the very block whose repetition `meta` exists to avoid.

On a page without one, `CurrentModule` was missing while resolving `@ref`, and `CollapsedDocStrings`, `Description`, `EditURL` and `IgnorePage` were missing at render time.

Meant to unblock https://github.com/oscar-system/Oscar.jl/pull/4978 together with PR #2988 (CC @lgoettgens)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,13 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


lgoettgens (approved):
LGTM! Just be aware of conflicts with https://github.com/JuliaDocs/Documenter.jl/pull/2988 in `test/default_meta/tests.jl`

mortenpi (approved):

PR #2988, opened 2026-08-30
Add a `meta` keyword to `doctest`

fingolfin:
`makedocs` gained a `meta` keyword to supply default values for the `@meta` blocks of every page, but the standalone `doctest` builds its own document and did not forward it. Packages that run their doctests through `doctest` rather than `makedocs` therefore had no way to set a global `DocTestSetup` for the doctests on their manual pages; `setdocmeta!` only reaches docstrings.

Meant to unblock https://github.com/oscar-system/Oscar.jl/pull/4978 together with PR #2987 (CC @lgoettgens)


lgoettgens (approved):
LGTM

asinghvi17 (approved):

PR #2991, opened 2026-08-31
Support named anchors on admonitions titles

fredrikekre:
An admonition title of the form `!!! note "[title](at-id name)"` now registers a named anchor, mirroring named headers. The `at-id` link is stripped from the rendered title, the anchor resolves via `at-ref` (also across pages), is written to the `objects.inv` inventory as a `std:label` entry, and its slug becomes the HTML `id` (and permalink target) of the admonition element itself, instead of the content-hash based id that changes whenever the admonition is edited. Admonitions without an `at-id` anchor keep the hash-based id.

Since `MarkdownAST.Admonition.title` is a plain string, the title is matched textually in `collect_named_anchors!` (next to the inline anchors from #2948), and the element is replaced in place with a new `Documenter.AnchoredBlock` that holds the anchor and the original admonition. The anchors share the id namespace and duplicate checks with header and inline `at-id` anchors.

### DIFF
--- a/src/expander_pipeline.jl
+++ b/src/expander_pipeline.jl
@@ -60,7 +60,8 @@ function expand(doc::Documenter.Document)
             Selectors.dispatch(Expanders.ExpanderPipeline, node, page, doc)
             expand_recursively(node, page, doc)
         end
-        # Register arbitrary `[content](@id name)` anchors. This runs after the per-node
+        # Register arbitrary `[content](@id name)` anchors, on inline content as well as in
+        # admonition titles. This runs after the per-node
         # expansion above (so header `@id` links have already been consumed by TrackHeaders)
         # and, crucially, before the CrossReferences pipeline stage, so that `@ref`s to these
         # anchors — including forward references to later pages — can be resolved.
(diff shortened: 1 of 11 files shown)


fingolfin:
Hmm, but in general are those admonition titles pure strings, and not supporting any kind of Markdown at all? It seems really strange to make this one exception. If we do this, then we should support all Markdown there, and have that changed in Julia's stdlib itself?

src/expander_pipeline.jl
```diff
@@ -1127,6 +1128,12 @@ iscode(x, lang) = false
 # collect_named_anchors!).
 const NAMED_ANCHOR_REGEX = r"^@id (.+)$"
 
+# Matches an admonition title of the form `[title](@id name)`, i.e. the whole title wrapped
+# in an `@id` link, mirroring named headers. Since `MarkdownAST.Admonition.title` is a plain
+# string (never parsed as markdown), the pattern is matched textually rather than on a
+# `MarkdownAST.Link` element.
+const ADMONITION_TITLE_ANCHOR_REGEX = r"^\[(.*)\]\(@id (.+)\)$"
+
```
fingolfin: Aha, as I suspected, you have to do that Markdown parsing by foot here.

I am not really thrilled about that :-/

fredrikekre: I believe we do extended parsing like this elsewhere too, but should probably actually use Markdown.parse and not a regex perhaps.

fredrikekre: Updated to Markdown.parse it instead.

mortenpi:
> Hmm, but in general are those admonition titles pure strings, and not supporting any kind of Markdown at all?

Yeah, I agree with this. My philosophy has always been to stick strictly to the semantics of the Markdown AST. As a case in point, the admonitions won't render nicely elsewhere, e.g. in the REPL:

image

So I don't really think we should do this.

PR #2993, opened 2026-08-31
Size PNG and JPG @example output according to its pixel density

jkrumbiegel:
This adds dpi awareness for PNGs and JPGs just like I added to DocumenterVitepress a while ago https://github.com/LuxDL/DocumenterVitepress.jl/pull/224

Claude Opus 5:

---

A PNG saved at a higher pixel density than the CSS default of 96 dpi records that density in a `pHYs` chunk, which is what you get from e.g. a Makie figure saved with `px_per_unit = 2` so it looks crisp on a HiDPI screen. The `img` tag written for an `@example` image output carried no `width`/`height`, so the browser had no reason to consult that chunk and sized the image by its raw pixel dimensions instead. A figure meant to occupy 350x250 CSS pixels is 700x500 pixels on disk, and was displayed at that size, twice as large in each dimension as intended.

The PNG bytes are now read with [SimpleImageMetadata.jl](https://github.com/jkrumbiegel/SimpleImageMetadata.jl), which takes the `IHDR` dimensions and the `pHYs` density from the header, and the resulting size in CSS pixels goes into the tag:

```html
<img src="index-69498fcf.png" alt="Example block output" width="350" height="250"/>
```

Besides fixing the size this gives the browser a high-resolution image to downsample, so the plot stays sharp on HiDPI screens. Both the file and the inlined `data:` URI branch get the attributes.

The parser lives in that package rather than in Documenter, per the discussion below. It was registered in General today, has no dependencies of its own and supports Julia 1.6, so it does not move Documenter's own bounds. The same code was vendored in DocumenterVitepress and QuartoNotebookRunner, which can drop their copies for it too.

PNG and JPEG output are both sized this way. WebP stores no resolution at all, so it is left alone.

## Example

```julia
using CairoMakie, Random
CairoMakie.activate!(px_per_unit = 2)
Random.seed!(1)
lines(cumsum(randn(100)); figure = (; size = (350, 250)))
```

Both screenshots are taken at a device pixel ratio of 2, like on a HiDPI screen.

## Before

figure displayed at twice its intended size

## After

figure displayed at 350x250


src/html/HTMLWriter.jl
```diff
@@ -2719,17 +2719,94 @@ function domify_show_image_binary(dctx::DCtx, filetype::AbstractString, d::Dict{
...
+
+Returns no attributes if the bytes cannot be parsed as a PNG.
+"""
+function png_size_attributes(bytes::Vector{UInt8})
+    metadata = try
+        png_image_metadata(bytes)
+    catch e
+        @debug "Unable to determine the size of a PNG @example output" exception = (e, catch_backtrace())
+        return ()
+    end
```
mortenpi, line 2748: I am quite allergic to unconditional try-catches. Probably better to handle the errors via an explicit return value, or at least a typed error object.

src/html/HTMLWriter.jl
```diff
@@ -2719,17 +2719,94 @@ function domify_show_image_binary(dctx::DCtx, filetype::AbstractString, d::Dict{
...
+    return (:width => string(metadata.width), :height => string(metadata.height))
+end
+
+# Taken from https://github.com/PumasAI/QuartoNotebookRunner.jl/, MIT licensed
+function png_image_metadata(bytes::Vector{UInt8})
```
mortenpi, line 2753: This is quite non-trivial, and re-implemented in at least three places now. I think I would feel better if this was a small utility package, and the maintenance burden shifted away from here.

mortenpi (commented):
So, my gut feel was that this is too much complexity for too little gain. But looking at it -- if we'd maintain the PNG parsing separately, then I think this would be fine.

jkrumbiegel:
How do you mean separately? In terms of complexity, for a parser I think this is pretty low :) it just loops over chunks looking for that one relevant one. That's why I had added it to both DocumenterVitepress and QuartoNotebookRunner like that, felt like too little for a package

goerz:
> felt like too little for a package

Definitely not: it's not very complex, but definitely "not trivial". If we vendor it into Documenter, that puts the maintenance burden on the Documenter maintainers, and it's not really code that any of us understands.

If it's being used in three separate packages now, that definitely warrants making it a separate package. It doesn't have to be a "big" package...

jkrumbiegel:
PngDpiMetadataReader.jl? :) I guess if it's a package it should do a bit more, maybe I can add jpg and webp too which I think have some similar metadata. I'll look into it

mortenpi:
Don't have to overengineer it just for this though. Good thing that, when it's in it's own package, it can grow at its own pace. But, FWIW, another reason to have it as a separate package -- even if nearly trivial now, these things tend to grow in complexity over time :slightly_smiling_face:

jkrumbiegel:
I've edited the description, this PR now uses the newly registered SimpleImageMetadata.jl package which can do the dpi lookup for png and jpg, two of the most common formats. I've also had Claude edit the description correspondingly.

PR #2994, opened 2026-08-31
Keep Biome out of nested checkouts

fingolfin:
Any checkouts of this repo in a nested git worktree under `.claude/worktrees` carry a copy of `biome.jsonc`. Biome treats every such file it walks into as a configuration and aborts with "Found a nested root configuration", so any Biome run from the main checkout fails while a worktree exists.

Negated patterns prune the directory walk, unlike the include patterns, which only select what gets formatted. Excluding the directory therefore stops the discovery without changing the set of formatted files.

---

This is the "controversial" part of PR #2978.

I'd love a user specific fix that I could keep local and not annoy other people with, but I couldn't find one short of patching the config file, or bypassing `Makefile`.

But I think this may affect other people than me anyway, and I can't think of any downsides to this. What's the harm?

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -11,6 +11,10 @@
     "ignoreUnknown": false,
     "includes": [
       "biome.jsonc",
+      // Claude Code puts additional checkouts of this repo here. Biome treats
+      // every biome.jsonc it walks into as a config and rejects the nested
+      // ones; a negated pattern is what stops it descending in the first place.
+      "!.claude/worktrees",
       "assets/html/**",
       "!!assets/html/themes",
       "!!assets/html/scss/highlightjs"


goerz:
I have no objections to this whatsoever, but I’m still not quite clear on why we jump through hoops of trying to get biome to ignore files that certain workflows may or may not create (what folder Claude uses for worktrees is not standardized in any way), as opposed to just telling it which files we do want it to format.

fingolfin:
@goerz this is not about telling Biome which fails to format: we already tell it that precisely.

This is to work around the annoying default behavior of Biome were it searches for *additional* config files in all subdirectories. The only reliable way to stop that is to explicitly exclude directories in `files.include` -- which is super weird because that list otherwise controls which files or directories to *format*, but for reasons I can't fathom, it is also used for this purpose. It is explicitly documented as such, but without an explanation.


And both Codex and Claude Code by default like to put git worktrees into a `.codex` or `.claude` subdirectory. This is not a convention I made up, this is the default one gets when pointing Claude at a git dir and asking it to work in a fresh worktree. It is even [documented](https://code.claude.com/docs/en/worktrees#start-claude-in-a-worktree).

goerz:
Ah, now I get it!

That seems like a “bug” in biome, but we’ll have to deal with it.

I didn’t know that Claude now had a default location for worktrees. Without the command line flag, I’ve seen it put worktrees in various locations depending on the mood of the day.

But with this being documented, that makes a very strong case for us excluding these directories explicitly. So, thanks for the clarification. With that, I would wholeheartedly support this PR.

goerz (approved):

PR #2998, opened 2026-09-05
Match the whole language of a fenced code block

### LINKED ISSUE #2997: New error on code block languages that start with `jldoctest` but have extra trailing characters
Since version 1.18 Documenter errors during doctests when it encounters a code block with a language that `startswith("jldoctest")` but is not equal to `jldoctest`, for example `jldoctests`:
```
┌ Error: Doctesting failed
│   exception =
│    internal error: expected jldoctest code block in experimental/ActionPolyRing/src/Content.jl:1164-1175
```
Previous versions of Documenter silently ignored these blocks.

In our case this did catch wrong doctest blocks because the language was accidentally written as `jldoctests` (with an `s` at the end).

Having a warning for this might be good but I think this is breaking (it did break doctests for released versions of Oscar).
One could argue that it would be valid for someone to come up with a custom `jldoctest_special` language with some extra features. But that would not work anymore.

The error is triggered here:
https://github.com/JuliaDocs/Documenter.jl/blob/4a8287436f861099d5ca375df0358ffd4e6be6e6/src/utilities/utilities.jl#L764-L767

This was introduced in https://github.com/JuliaDocs/Documenter.jl/pull/2912 by @fingolfin.

The caller checked only for `startswith`:
https://github.com/JuliaDocs/Documenter.jl/blob/4a8287436f861099d5ca375df0358ffd4e6be6e6/src/doctests.jl#L108

fingolfin:
A code block whose language merely started with the name of a Documenter block -- ```jldoctests``` for a doctest, ```@examples``` for an example -- was routed to that block's runner, which then rejected the language it had been handed and aborted the build with an `internal error`. Dispatch now compares the whole language, so such a block is passed through as an ordinary code block again, as it was before v1.18.

Silently ignoring them is what makes the typo hard to find, though, so an unrecognized language that starts with the name of a Documenter block is now reported with a plain warning. A deliberate custom language such as `jldoctest_special` is warned about as well; it still renders.

Fixes #2997.


goerz (approved):
Not an in-depth review, but make sense to me, and I'd trust you or your Claude to get the details right.

LGTM

PR #2999, opened 2026-09-10
Change guide for Julia v1.12+

Gregstrq:
Currently, the guide for Julia v1.12+ suggests adding the package to the docs environment as `add PkgName`. As far as I understand, it will tie the docs to the latest registered version of the package and won't even work if the package is not yet registered.
I think we want the docs environment to point to the parent package, so we should `dev ..` instead, as explained in [Pkg.jl docs](https://pkgdocs.julialang.org/v1/creating-packages/#adding-tests-to-packages).

### DIFF
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -37,6 +37,13 @@ $ julia --project=docs
 This will create a `Project.toml` file in the `docs/` subdirectory and add `Documenter` and the top-level package as available packages.
 See also the [Pkg.jl](https://github.com/JuliaLang/Pkg.jl/) documentation on working with [project workspaces](https://pkgdocs.julialang.org/v1/toml-files/#The-[workspace]-section) and [environments](https://pkgdocs.julialang.org/v1/environments/).
 
+For Julia v1.12, one should also manually add
+```
+[sources]
+PkgName = {path = ".."}
+```
+to the `docs/Project.toml` file. This is done automatically in Julia v1.13.
+
 ### Julia v1.11 and earlier
 For Julia versions v1.11 and earlier, creating a separate project in `docs/` is the standard approach.
 To do this, navigate to your package's root folder and do
--- a/docs/src/man/guide.md
+++ b/docs/src/man/guide.md
@@ -30,7 +30,8 @@ Next, open up a Julia REPL in the `docs` subdirectory, enter `pkg>` mode with th
 
 ```
 $ julia --project=docs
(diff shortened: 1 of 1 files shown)


goerz:
Using `dev` instead of `add` doesn't hurt, but that only survives for as long as the `docs/Manifest.toml` is around. Since manifest files usually shouldn't be committed (although opinions on that might differ when it comes to the `docs` environment specifically), I think a much more important recommendation would be to manually add

```
[sources]
PkgName = {path = ".."}
```

to the resulting `docs/Project.toml` file.

That part is currently missing from the Julia 1.12 section, and I'd recommend adding it in this PR.

docs/src/man/guide.md
```diff
@@ -30,7 +30,8 @@ Next, open up a Julia REPL in the `docs` subdirectory, enter `pkg>` mode with th
 
 ```
 $ julia --project=docs
-(PkgName/docs) pkg> add Documenter PkgName
+(PkgName/docs) pkg> add Documenter
+(PkgName/docs) pkg> dev ..
```
fredrikekre:
```suggestion
(PkgName/docs) pkg> dev .
```

Gregstrq:
I have done some experimentation. Bear in mind I have julia v1.13 installed.

 If I try to add an unregistered package, it all breaks down.
If I `dev` the parent folder, than
```
[sources]
PkgName = {path = ".."}
```
Is added automatically to the Project.toml of the workspace directory, even if the there was no Manifest.toml in the parent folder.

goerz:
Oh, nice! Does 1.12 do that as well, or is that a new feature?

In that case, I would rephrase "manually add" to "check that the file contains"

docs/src/man/guide.md
Gregstrq: You are right, my mistake. The guide suggests that you actually run julia in the parent folder.

Gregstrq:
@goerz 
On v1.12.7, `dev` does not add [sources] automatically. It must be a new feature in v1.13.

Gregstrq:
@fredrikekre , @goerz , I have added a line about manual adding the source block on v1.12

goerz (approved):

Gregstrq:
So, is there anything that needs to be done in order for the PR to merge?
I don't think I should update changelog.md...

goerz:
Seems like an obvious improvement to me, so I'm just going to go ahead an merge

PR #3001, opened 2026-09-14
Apply :limit => true to all MIMEs in display_dict

asinghvi17:
Fix infinitely long dataframes in html output.

`display_dict` rendered `text/plain` through `limitstringmime`, which sets `:limit => true` in the `IOContext`.

But every other MIME (`text/html`, `text/latex`, `text/markdown`, images) went through plain `stringmime` with no `:limit`. So an `@example` block returning a 100-row DataFrame emitted a `text/html` table with all 100 rows. The REPL, IJulia and Pluto all show 25 rows because they set `:limit`.

This builds one limited `IOContext` (still layered on top of any caller-supplied `context`) and passes it to `stringmime` for every MIME, including `text/plain`; `limitstringmime` had no other callers and is removed. A regression test is added in `test/utilities.jl`.

Before: `@example` with a 100-row DataFrame -> full 100-row `<table>` (and an `example_size_threshold` warning).
After: 25-row `<table>`, matching the `text/plain` output and the REPL.

https://claude.ai/code/session_014beQkRwLUtB6ipLc2xS8G6


goerz (approved):
Seems like a very good idea, and the implementation doesn't raise any concerns, at a glance

